PR #7, opened 2016-01-29
fix(radar): fix wrong display polarRadiusAxis and fix missing center …

evilucifero:
fix(radar): fix wrong display polarRadiusAxis and fix missing center axis


arcthur:
Test don't pass.

PR #45, opened 2016-02-28
Pie chart label

vanthiyathevan:
When using custom label element, It wasn't getting rendered next to the slice.

- added label formatter for tooltip. (Ex: date formatting)
- pass the pie slice percentage as part of props for custom pie label element


xile611:
Thanks~
Maybe we need to add a props `labelLine` to control the lines between labels and slices.

vanthiyathevan:
Actually `labelRadius` will make sense to position the label from centre.
Ex: `{labelRadius: '50%'}` for positioning in the middle of the slice.
`{labelRadius: 10}` or `{labelRadius: -10}` to position it from the outer radius

PR #53, opened 2016-03-12
Fix tooltip location in case x has margin

orius123:
Top was calculated from viewBox.x instead of y,
when I had a chart with margin right the tooltip was going down out of the box even if it could fit there.
anyways it looks like a typo :smile:


orius123:
thanks!
When do you plan to release? it's really annoying in my app..

xile611:
Thanks! We plan to release a stable version in April. But we will do as soon as possible to support wireless.

PR #60, opened 2016-03-24
stage-0 is need for demo

p2227:
stage-0 is need for demo(when` npm run demo`)


arcthur:
It's my mistake, the stage-0 will be splitted to specific packages. I will fix the babelrc

PR #78, opened 2016-04-27
fix misspelling

doublelduob:


xile611:
Thanks!
But `getTooltipBBox ` is not a misspelling, it means getting bounding box of the tooltip.
If you have used *Raphael.js* before, you may know `getBBox`  which is a method of Raphael element.
I just borrowed the naming from *Raphael.js*.

doublelduob:
My bad, I was reading the recharts awesome source code. Wish me next time
will do something right : )

2016-04-27 19:28 GMT+08:00 xile611 <notifications@github.com>:

xile611:
@doublelduob  Thanks~  Pull request is welcome!

PR #81, opened 2016-04-29
tooltip move smoothly, fix suddenly jump

doublelduob:
when tooltip reach viewBox.x + viewBox.y, it will suddenly jump to another coordination, fix to stay at bottom or top feels no surprise to user :)

### DIFF
--- a/src/component/Tooltip.js
+++ b/src/component/Tooltip.js
@@ -98,6 +98,7 @@ class Tooltip extends Component {
       pointerEvents: 'none',
       display: active ? 'block' : 'none',
       position: 'absolute',
+      transition: '200ms linear',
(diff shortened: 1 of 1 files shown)


xile611:
Well, I don't think your positioning strategy is very appropriate. Because it can cause the problem of tooltip covering the active dots.

doublelduob:
You are right.  I didn't think about that.  This is why the tooltip will jump. Let me try something make the jump smoothly.

xile611:
Yeah, tooltip located at the right of cursor is natural. But when we detect the tooltip will exceed the viewing area, the position of tooltip will be adjusted to the left of the cursor.

PR #83, opened 2016-04-29
feat(tooltip): add tooltip animation

jasonHzq:


src/component/Tooltip.js
```diff
@@ -104,19 +118,32 @@ class Tooltip extends Component {
...
-    outerStyle.left = Math.max(
+    const translateX = Math.max(
       coordinate.x + box.width + offset > (viewBox.x + viewBox.width) ?
       coordinate.x - box.width - offset :
       coordinate.x + offset, viewBox.x);
+
     outerStyle.top = Math.max(
       coordinate.y + box.height + offset > (viewBox.y + viewBox.height) ?
       coordinate.y - box.height - offset :
       coordinate.y + offset, viewBox.y);
```
doublelduob: This implement is good, just ignored the Y axis when chart is horizontal. 

just add a style transition: 400ms linear. It will work.

doublelduob: My bad. horizontal case is good as well.

PR #92, opened 2016-05-10
Feature/stack offset

iuriikozuliak:
Allow to pass different stack offset


arcthur:
the test was failed, please add test case too. thanks

PR #97, opened 2016-05-12
Sync Brush state with upstream data

ciAnd7:
CategoricalChartWrapper 'notify' brush when receiving new data. This lead to reset internal brush's scale state.

CategoricalChartWrapper keep ref to Brush element. May be using 'ref' is not good solution, and there may be other solutions to force Brush to reset...

### DIFF
--- a/src/cartesian/Brush.js
+++ b/src/cartesian/Brush.js
@@ -51,21 +51,13 @@ class Brush extends Component {
     };
 
     if (props.data && props.data.length) {
-      const len = props.data.length;
-      const startIndex = _.isNumber(props.defaultStartIndex) ? props.defaultStartIndex : 0;
-      const endIndex = _.isNumber(props.defaultEndIndex) ? props.defaultEndIndex : len - 1;
-
-      this.scale = scalePoint().domain(_.range(0, len))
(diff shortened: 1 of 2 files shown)


xile611:
Please update the latest code, and fix the build fail problem.

ciAnd7:
@xile611 Ok. 

It seems that problem with refs and tests environment. The only failure I see when running test locally is:
```
Firefox 46.0.0 (Mac OS X 10.11.0) <LineChart /> MouseEnter lineChart should show tooltip, 
active dot, and cursor FAILED
	addComponentAsRefTo(...): Only a ReactOwner can have refs. You might be adding a ref to a 
component that was not created inside a component's `render` method, or you have multiple copies of 
React loaded (details: https://fb.me/react-refs-must-have-owner).

...

@/......./recharts/test/index.js:89873:21 <- webpack:///test/specs/chart/LineChartSpec.js:201:20
```
I will try to fix.

xile611:
@ciAnd7 This bug of Brush has been fixed in version 0.10.7

ciAnd7:
@xile611 Thanks a lot!

PR #126, opened 2016-06-23
Fix issue 125 - wrong tick calculation

george-cz:


xile611:
@george-cz 
The cause of this bug is that the value of `tick` in `Bar` locate in the left of each band, but the value of `tick` in `Line` locate in the middle of each band.

Your change may cause the bug of LineChart.

george-cz:
You're right, I believe my latest commit fixed that. However the code isn't as pretty and straightforward anymore :D

xile611:
Maybe it's better to fix `getTicksOfAxis`, I change this function in the latest version.

xile611:
But, I find `getTicksOfAxis ` is in a mess.

xile611:
@george-cz We think it's better to fix `getTicksOfAxis `, and we have fix it in V0.12.0. Thank you !

PR #127, opened 2016-06-23
Prevent forced axisId for composed bar charts with multiple yAxes

iuriikozuliak:
screen shot 2016-06-23 at 21 58 05

### DIFF
--- a/src/chart/BarChart.js
+++ b/src/chart/BarChart.js
@@ -253,7 +253,7 @@ class BarChart extends Component {
 
     return items.map((child, i) => {
       const { xAxisId, yAxisId } = child.props;
-      const axisId = layout === 'horizontal' ? xAxisId : yAxisId;
+      const axisId = layout === 'horizontal' && yAxisId === 0 ? xAxisId : yAxisId;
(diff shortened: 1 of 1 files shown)


xile611:
@iuriikozuliak  I don't quite understand what's the problem.

iuriikozuliak:
@xile611 Please take a look on it at http://jsfiddle.net/jvhd96pq/ and check error in the console, it happens at this line - https://github.com/recharts/recharts/blob/c326127737687e0e89d75935c799f3b1e70b8e56/src/chart/BarChart.js#L125

`sizeList` is undefined at `this.getBarPosition(bandSize, sizeList[axisId])` since `axisId` equals `xAxisId` for horizontal layout instead of `yAxisId`.

xile611:
@iuriikozuliak Thank you! I have found the reason.
 The [L256](https://github.com/recharts/recharts/blob/c326127737687e0e89d75935c799f3b1e70b8e56/src/chart/BarChart.js#L256) should fixed to the following:

```
const axisId = layout === 'horizontal' ? yAxisId : xAxisId;
```

PR #128, opened 2016-06-24
allows functions to determine fill values on Bar component

sankho:
Was a simple addition I needed for a project I'm using it on. Useful for when you want the bars on a Barchart to be separate colors as determined by their values.

A few thoughts:

- could be better as a separate prop rather than overloading the `fill` prop and allowing 2 different types.
- let me know if the method could be executed in a better place w/n `Bar.js` other than `renderRectangle()`
- when we're ok with the implementation I'd be happy to add this feature to the other components in the `cartesian/` folder.

### DIFF
--- a/src/cartesian/Bar.js
+++ b/src/cartesian/Bar.js
@@ -81,6 +81,10 @@ class Bar extends Component {
   renderRectangle(option, props) {
     let rectangle;
 
+    if (typeof props.fill === 'function') {
+      props.fill = props.fill(props.value)
+    }
+
     if (React.isValidElement(option)) {
       rectangle = React.cloneElement(option, props);
     } else if (_.isFunction(option)) {
--- a/src/cartesian/Bar.js
+++ b/src/cartesian/Bar.js
@@ -28,6 +28,7 @@ class Bar extends Component {
     name: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     dataKey: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
     formatter: PropTypes.func,
+    fillFormatter: PropTypes.func,
 
     shape: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
     label: PropTypes.oneOfType([
@@ -81,10 +82,6 @@ class Bar extends Component {
   renderRectangle(option, props) {
     let rectangle;
 
-    if (typeof props.fill === 'function') {
-      props.fill = props.fill(props.value)
-    }
-
     if (React.isValidElement(option)) {
       rectangle = React.cloneElement(option, props);
     } else if (_.isFunction(option)) {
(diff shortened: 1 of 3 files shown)


xile611:
@sankho  
This approach can lead to presentation attributes of svg elements  is more and more complex, difficult to maintain. Because each components has many presentation attributes, all these attributes have the same problem.

sankho:
@xile611 updated this to not touch `ReactUtils` and overload the presentation attributes.

Instead, setting up a prop `fillFormatter` directly on the Bar itself. I'm hoping this is a more suitable approach, but please advise and I can make further adjustments.

xile611:
@sankho I think it's better to add `style`  which can set all the presentation attributes by a function. But we are still very cautious to add **props**. Because this feature can be implemented in two ways now.
One is add `fill` to the `data` passed to `BarChart`. Another is set the `shape` of `Bar`.

sankho:
@xile611 so I understand this creates another way to implement the feature of having different colored `Bar`'s on `BarCharts` - that's kind ot he point. The other approaches involve creating multiple `Bar` elements for each color type w/ separate sets of data. I was trying to add a convenience method to avoid creating multiple `Bar` elements and iterate over a single set of data.

Understood if you'd rather avoid having multiple ways to do things, but I feel this feature is the most developer friendly, moreso than creating multiple bars. I included a test to help w/ maintenance in the future. If having multiple approaches is less desired than developer friendliness, I can close the commit and use the aforementioned approach.

But I may be misunderstanding, and I'm a bit unclear about how to do this w/ `style` - would you mind elaborating here?

xile611:
@sankho  Thank you very much for your contribution. I really hope that Recharts will become more and more friendly to developers.

But I worry that your implementation will bring more and more props like `strokeFormatter`, `strokeWidthFormatter`, `fillOpactityFormatter`, `strokeOpacityFormatter`, and so on. Too may props is also not developer friendly. Meanwhile our code may become something like this:

```javascript
if (this.props.fillFormatter) {
  const fill = this.props.fillFormatter(entry.value)
}
if (this.props.strokeFormatter) {
  const stroke = this.props.strokeFormatter(entry.value)
}
....
```

So I think what we need is something like `attrFormatter`, which can be a function and can format all the attributes by values. And we may get final  presentation attributes like this:

```javascript
let attrs = getPresentationAttributes(props);

if (_.isFunciton(this.props.attrFormatter)) {
  attrs = { ...attrs, ...this.props.attrFormatter(props.value) };
}
```

Than It seems `attrFormatter` do the similar thing to `shape`.  `attrFormatter` is only simpler than `shape`.
So I advise you can use `shape` to implement your chart.

[Implemented by shape](http://jsfiddle.net/8s13xjth/)
[Implemented by formatting data](http://jsfiddle.net/wg72e1jt/)

sankho:
@xile611, For the record, I think Recharts is absolutely the most developer friendly d3 react library out there, it's why I chose it - regardless of this feature or the approach. I didn't mean anything negative by that comment.

Thanks for detailing the hesitancy of making too many props, I can understand the want to avoid code with endless if statements.

Also, thanks for the fiddle's showing alternative implementations. Personally I'd prefer the library I'm using obfuscate such tasks, so I do like the `attrFormatter` approach as it allows my application code to be less verbose.

So, here's my question: would you want an `attrFormatter` in the code, or would you prefer people use the shape / data formatting approach for this feature? If so I can start implementing, but otherwise I could close this PR. I should have made an issue for this first, so just checking this time around :)

xile611:
@sankho I prefer developers can use the `shape` or  `data` to approach this feature.

sankho:
No problem, will do

PR #178, opened 2016-08-17
fix: fix bug of stacks in barchart are misplaced #177

### LINKED ISSUE #177: Some stacks are shown in unexpected position in Stacked Barchart.
As shown in the attached screenshot, The third stack in a stacked bar chart will be shown in unexpected position.
If a bar chart contains more than three stacks, the position of stacks may be misplaced

Here's a reproduction code.
http://jsfiddle.net/kauzp0n/m7w8c7uq/

And it is not happened if value is negative.
http://jsfiddle.net/kauzp0n/csqxnk43/1/

kazup0n:
This PR fixes #177 .

`series[i][j][1]` is summation of previous values, so there's no need to add it to `positive`.

### DIFF
--- a/src/util/CartesianUtils.js
+++ b/src/util/CartesianUtils.js
@@ -23,7 +23,7 @@ const offsetSign = (series, order) => {
       if (value >= 0) {
         series[i][j][0] = positive;
         series[i][j][1] = positive + value;
-        positive += series[i][j][1];
+        positive = series[i][j][1];
       } else {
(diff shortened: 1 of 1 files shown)


xile611:
@kazup0n Sorry I  didn't see your pr earlier until I released version 0.13.3. And this bug has been fixed in 0.13.3. Thank you for your contribution.

kazup0n:
@xile611 It's OK. thank you!

PR #192, opened 2016-08-24
Fix #136: New RepsonsiveContainer API

### LINKED ISSUE #136: Bar: label formatter doesn't work
Hey,

formatter is declared in Bar's PropTypes but it isn't used anywhere and doesn't work. I would like to pass custom formatting function for labels that sits on top of Bar.

nickpalmer:
* The ResponsiveContainer can make use of optional
  minWidth, minHeight and aspect options to give more
  control over the sizing of the chart in various layouts.

### DIFF
--- a/src/component/ResponsiveContainer.js
+++ b/src/component/ResponsiveContainer.js
@@ -18,14 +18,15 @@ const render = ({ aspect, width, height, minWidth, minHeight, container, childre
        'The aspect(%s) must be greater than zero.',
        aspect);
 
-  let calculatedWidth = container.width;
-  let calculatedHeight = container.height;
+  let calculatedWidth = isPercent(width) ? container.width : width;
+  let calculatedHeight = isPercent(height) ? container.height : height;
 
   if (aspect && aspect > 0) {
     // Preserve the desired aspect ratio
-    if (calculatedWidth / calculatedHeight > aspect) {
+    if (calculatedHeight > 0 &&
+        (calculatedWidth === 0 || calculatedWidth / calculatedHeight > aspect)) {
       calculatedWidth = calculatedHeight * aspect;
-    } else if (calculatedWidth / calculatedHeight < aspect) {
+    } else if (calculatedHeight === 0 || calculatedWidth / calculatedHeight < aspect) {
       calculatedHeight = calculatedWidth / aspect;
     }
   }
--- a/src/component/ResponsiveContainer.js
+++ b/src/component/ResponsiveContainer.js
@@ -7,56 +7,70 @@ import pureRender from '../util/PureRender';
 import { getPercentValue, isPercent } from '../util/DataUtils';
 import { warn } from '../util/LogUtils';
 
-const render = ({ width, height, container, children }) => {
+const render = ({ aspect, width, height, minWidth, minHeight, container, children }) => {
+
   warn(isPercent(width) || isPercent(height),
-    `The width(%s) and height(%s) are both fixed number,
-     maybe you don't need to use ResponsiveContainer.`,
+    `The width(%s) and height(%s) are both fixed numbers,
+     maybe you don't need to use a ResponsiveContainer.`,
     width, height
   );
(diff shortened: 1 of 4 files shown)


src/component/ResponsiveContainer.js
```diff
@@ -7,56 +7,70 @@ import pureRender from '../util/PureRender';
...
+  let calculatedHeight = getPercentValue(height, container.height);
 
-  if (calculatedWidth > 0 && calculatedHeight > 0) {
-    return React.cloneElement(children, {
-      width: calculatedWidth,
-      height: calculatedHeight,
-    });
+  if (aspect && aspect > 0) {
+    // Preserve the desired aspect ratio
+    if (calculatedWidth / calculatedHeight > aspect) {
```
xile611: I found that the value of `calculatedHeight` may be zero. Maybe we should add branches like this:
```js
   if (calculatedHeight <= 0 || calculatedWidth / calculatedHeight < aspect) {
     calculatedHeight = calculatedWidth / aspect;
   } else {
     calculatedWidth = calculatedHeight * aspect;
  }
```

src/component/ResponsiveContainer.js
```diff
@@ -7,56 +7,70 @@ import pureRender from '../util/PureRender';
...
+  warn(!aspect || aspect > 0,
+       'The aspect(%s) must be greater than zero.',
+       aspect);
 
-  warn(calculatedWidth > 0 && calculatedHeight > 0,
-    `The width(%s) and height(%s) of chart should be greater than 0,
-    please check the style of container, or the props width(%s) and height(%s).`,
-     calculatedWidth, calculatedHeight, width, height
-  );
+  let calculatedWidth = getPercentValue(width, container.width);
```
xile611: Because we set the width, height, minHeight, minWidth in [L50](https://github.com/recharts/recharts/pull/192/commits/aec73d638ae9fa14c5e52e543934eb8f1b599d21#diff-64e08d9691c7f3156ef1dfd43b7c94e5R50), `container.width` should be the width we need to render charts.

```
let calculatedWidth = container.width;
let calculatedHeight = container.height;
```

xile611:
@nickpalmer Thank you for your enthusiasm.

src/component/ResponsiveContainer.js
nickpalmer: This is the case with the existing implementation and is exactly the problem I am trying to solve.

That won't work in many cases because BOTH can be zero. The solution is to use minHeight or minWidth to force a render at some minimum size.

src/component/ResponsiveContainer.js
nickpalmer: That remove the ability to override and FORCE a size using height or width.

src/component/ResponsiveContainer.js
nickpalmer: In fact the current test case actually does this. Try checking out master and putting a log of containerWidth and containerHeight. This change set gives you controls to deal with the problem that are backwards compatible.

nickpalmer:
@xile611 No worries. I would like to see this merge so I don't have to check a private branch out of git in npm for the project I am working on.

src/component/ResponsiveContainer.js
xile611: @nickpalmer Not everyone want to to set the value of minHeight and minWidth. So the `calculatedHeight` could be zero. We should avoid the denominator to be zero in a division operation.

src/component/ResponsiveContainer.js
xile611: If I have set the width to be `40%`, then the `calculatedWidth` would be `16%` of the parent node because of the following calculation. 
```
let calculatedWidth = getPercentValue(width, container.width);
```

nickpalmer:
Unfortunately there is no way to test the double percentage because enzyme doesn't actually trigger the ContainerDimensions to calculate. 👅

PR #193, opened 2016-08-24
Add optional scaleToFit to prevent label cutoff.

nickpalmer:
* This adds a new scaleToFit property to Pie which ensures
  that the labels are completely visible. It accomplishes this
  by measuring the actual chart and scaling down if the labels
  do not completely fit in the bounding box on first render.
  While this will shrink the size of the text in the chart
  for many users this will be preferable to clipping.


nickpalmer:
This is probably controversial, but I figured we could discuss it.

nickpalmer:
This doesn't work right on scale up. Closing and looking for a new solution.

PR #202, opened 2016-08-30
Branch get latest recharts

aleib:
Latest Recharts
Updated React 
Local react-smooth


aleib:
Apologies, went to the wrong repo :/

PR #219, opened 2016-09-08
allow label function to return a string

jochenberger:
This allows for easier custom label generation, like in
```jsx
<Pie data={chartData} label={(name, value)=>`${name}: ${value}`} />
```
This addresses only the pie chart, but it should be easy to port the change to the other chart types.


avindra:
Thanks for this @jochenberger  and @xile611 ! Will this be in the next release?

avindra:
I see this was released.

Btw, @jochenberger, your example code above is incorrect, it should be:

```jsx
<Pie data={chartData} label={({name, value})=>`${name}: ${value}`} />
```

Issue #191, opened 2016-08-23
Label in center of PieChart


bduff9:
I see a couple examples where there is text placed into the center of Donut Pie Chart based on active index.  In my case, I need static text placed into the center representing the data (i.e. 30% if that is what the data represents).  I can kind of get by this using custom label, but if I ever need labels outside of the center too, this won't work.

Am I missing something or is this currently accurate in how one might do this?  Thanks!

xile611:
@bduff9  If the width and height is fixed number,  you can place some text in the center of donut pie chart by adding svg elements in `PieChart`.

```jsx
<PieChart width={800} height={400}
   <text x={400} y={200} textAnchor="middle" dominantBaseline="middle">
    Donut
   </text>
   ...
</PieChart>
```

[Here](http://jsfiddle.net/gzzq6jn3/) is a demo.

bduff9:
@xile611 Awesome, thank you so much for the demo.  Unfortunately, the width is not static as I am using responsivecontainer, so I really need cx and cy, like in the custom label function.

xile611:
@bduff9 We'll try to support centered label. But it may take some time. You can try to fix it. Or a workaround is to use a position absolute text.

bduff9:
@xile611  Thank you!  For now, I am fine with my workaround but I appreciate the quick reply

xile611:
New components `Label` and `LabelList` are arrived in version 1.0.0-alpha.0. Centric label can be rendered by the following demo.

```jsx
<Pie>
  <Label value="any text" position="center" />
</Pie>
```

qborreda:
Is there an example of this using the Label component?

sppatel:
What if I want multiple labels inside the pie chart?

For example..

      60%
Completed

essaji:
I tried putting a react component in the `content` property of `Label` but for some reason its not rendering the react component.
e.g. 
```
<Label content={<h4>My label here</h4>} position="center" />
```

but its not working :/ any idea?

nivedhithaVenkatachalam:
@essaji Did you figure out a way to make it work? I am trying to add JSX either to content prop or as children prop (displays as [object Object]) and both does not work.

lecerd:
[...]

Issue #238, opened 2016-09-14
Responsive font size and/or word-wrapping


SachaG:
When resizing the width of the `ResponsiveContainer` chart [here](http://jsfiddle.net/64eoc63w/5/), the XAxis legend adapts by dropping labels. 

What would be a good way to instead make the font size smaller, or maybe wrap words as the overall container width gets narrower?

SachaG:
It looks like word-wrapping in SVG isn't trivial. Here's a way to do it, but it's fairly complex:

https://bl.ocks.org/mbostock/7555321

I'll probably just make my label text shorter…

techniq:
@SachaG I've worked on of PoC for a word-wrapping `<Text>` element to be used for all of rechart's labels.  See https://github.com/recharts/recharts/issues/198 and the PoC http://codepen.io/techniq/pen/rLXwJJ?editors=0011.

I was literally getting ready to start working on this and getting a PR in.  No timeline when I'll have it done, but hopefully "soon".

While I'm working on it, adding a `fit` option would be nice.  Something like http://bl.ocks.org/milkbread/6571949 and http://stackoverflow.com/a/22580176/191902

SachaG:
@techniq oh that's awesome! Would be a huge help for sure.

xile611:
Duplicated with [#198](https://github.com/recharts/recharts/issues/198).

techniq:
@SachaG Might be interested in this incoming PR - https://github.com/recharts/recharts/pull/243

Phonixoop:
.recharts-rectangle.recharts-tooltip-cursor {
  fill: #00000049;
}

PR #243, opened 2016-09-16
New <Text> component.  Closes #198

### LINKED ISSUE #198: Word-wrapping <Text> component
It would be very useful if there was a React `<Text>` component that could be used in place of the direct SVG `<text>` element that supported word wrapping.

This component could then be used for wrapping long axis labels (like [this](https://bl.ocks.org/mbostock/7555321) d3 demo), but also used for custom labels.  For example, I have a similar pie chart to [this example](http://recharts.org/examples#CustomActiveShapePieChart) and would like to word wrap long labels in the center.

I envision you would set a fixed width prop on the component and it would wrap the contents (using `<tspan>s`).  For example:

```jsx
<Text width="200">
  Discovered howto “like” things mentally
</Text>
```
would produce something like (stolen from d3 example):
```html
<text x="0" y="9" dy="0.71em">
  <tspan x="0" y="9" dy="0.71em">Discovered how</tspan>
  <tspan x="0" y="9" dy="1.81em">to “like” things</tspan>
  <tspan x="0" y="9" dy="2.91em">mentally</tspan>
</text>
```

Btw, Victory has a similar concept with their [VictoryLabel](https://formidable.com/open-source/victory/docs/victory-label) component.  I'm not sure if using explicit new lines `\n` is the way to go, or having a `width` prop and having the text wrap to fit.

techniq:
Adds a new `<Text>` component to replace using the built in SVG element `<text>` for better labels, ticks, etc.

Includes the following features:
- Word-wrapping (when `width` prop is defined)
- Scale-to-fit text (`scaleToFit` prop)
- Vertical alignment (`verticalAnchor` prop)
- Rotation (`angle` prop)

The best way to see the features in action is to run `npm run demo` and open the Text demo.

### DIFF
--- a/src/cartesian/Area.js
+++ b/src/cartesian/Area.js
@@ -266,7 +266,7 @@ class Area extends Component {
       labelItem = (
         <Text
           key={props.key}
-          {...getPresentationAttributes(props)}
+          {...props}
           className="recharts-area-label"
         >
           {_.isArray(value) ? value[1] : value}
--- a/src/cartesian/Area.js
+++ b/src/cartesian/Area.js
@@ -6,6 +6,7 @@ import classNames from 'classnames';
 import Curve from '../shape/Curve';
 import Dot from '../shape/Dot';
 import Layer from '../container/Layer';
+import Text from '../component/Text';
 import Animate from 'react-smooth';
 import pureRender from '../util/PureRender';
 import { PRESENTATION_ATTRIBUTES, getPresentationAttributes } from '../util/ReactUtils';
@@ -263,13 +264,13 @@ class Area extends Component {
       labelItem = option(props);
     } else {
       labelItem = (
-        <text
+        <Text
           key={props.key}
           {...getPresentationAttributes(props)}
           className="recharts-area-label"
         >
           {_.isArray(value) ? value[1] : value}
-        </text>
+        </Text>
       );
     }
 
--- a/src/component/Text.js
+++ b/src/component/Text.js
@@ -32,7 +32,11 @@ class Text extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.updateWordsByLines(nextProps, this.props.children !== nextProps.children);
+    const calculateWordWidths = (
+      this.props.children !== nextProps.children ||
+      this.props.style !== nextProps.style
+    );
+    this.updateWordsByLines(nextProps, calculateWordWidths);
   }
 
   updateWordsByLines(props, calculateWordWidths) {
--- a/src/component/Text.js
+++ b/src/component/Text.js
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import reduceCSSCalc from 'reduce-css-calc';
-import { PRESENTATION_ATTRIBUTES } from '../util/ReactUtils';
+import classNames from 'classnames';
+import { PRESENTATION_ATTRIBUTES, getPresentationAttributes } from '../util/ReactUtils';
 import _ from 'lodash';
 
 class Text extends Component {
@@ -110,6 +111,7 @@ class Text extends Component {
       angle,
       lineHeight,
       capHeight,
+      className,
       ...textProps,
     } = this.props;
     const { wordsByLines } = this.state;
@@ -143,7 +145,11 @@ class Text extends Component {
     }
 
     return (
-      <text textAnchor={textAnchor} {...textProps}>
+      <text
+        {...getPresentationAttributes(textProps)}
+        className={classNames('recharts-text', className)}
+        textAnchor={textAnchor}
+      >
       {
         wordsByLines.map((line, index) => (
           <tspan x={x} dy={index === 0 ? startDy : lineHeight} key={index}>
--- a/src/component/Text.js
+++ b/src/component/Text.js
@@ -85,18 +85,15 @@ class Text extends Component {
   calculateWordsByLines(wordsWithComputedWidth, spaceWidth, lineWidth) {
     const { fit } = this.props;
     return wordsWithComputedWidth.reduce((result, { word, width }) => {
-      const currentLine = result[result.length - 1] || { words: [], width: 0 };
+      const currentLine = result[result.length - 1];
 
-      if (currentLine.words.length === 0) {
-        // First word on line
-        const newLine = { words: [word], width };
-        result.push(newLine);
-      } else if (lineWidth == null || fit || (currentLine.width + width + spaceWidth) < lineWidth) {
+      if (currentLine && (lineWidth == null || fit ||
+        (currentLine.width + width + spaceWidth) < lineWidth)) {
         // Word can be added to an existing line
         currentLine.words.push(word);
         currentLine.width += width + spaceWidth;
       } else {
-        // Word is too long to fit on existing line
+        // Add first word to line or word is too long to fit on existing line
         const newLine = { words: [word], width };
         result.push(newLine);
       }
--- a/src/component/Text.js
+++ b/src/component/Text.js
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import reduceCSSCalc from 'reduce-css-calc';
 import { PRESENTATION_ATTRIBUTES } from '../util/ReactUtils';
+import _ from 'lodash';
 
 class Text extends Component {
 
@@ -59,7 +60,7 @@ class Text extends Component {
     // Calculate length of each word to be used to determine number of words per line
     var svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
     var text = document.createElementNS('http://www.w3.org/2000/svg', 'text');
-    Object.assign(text.style, props.style);
+    _.assign(text.style, props.style);
     svg.appendChild(text);
     document.body.appendChild(svg);
 
--- a/src/component/Text.js
+++ b/src/component/Text.js
@@ -1,0 +1,157 @@
+import React, { Component, PropTypes } from 'react';
+import reduceCSSCalc from 'reduce-css-calc';
+import { PRESENTATION_ATTRIBUTES } from '../util/ReactUtils';
+
+class Text extends Component {
+
+  static propTypes = {
+    ...PRESENTATION_ATTRIBUTES,
+    fit: PropTypes.bool,
+    angle: PropTypes.number,
+    textAnchor: PropTypes.oneOf(['start', 'middle', 'end', 'inherit']),
+    verticalAnchor: PropTypes.oneOf(['start', 'middle', 'end']),
+  };
+
+  static defaultProps = {
+    x: 0,
+    y: 0,
+    lineHeight: '1em',
+    capHeight: '0.71em', // Magic number from d3
+    fit: false,
+    textAnchor: 'start',
+    verticalAnchor: 'end', // Maintain compat with existing charts / default SVG behavior
+  };
+
+  state = {
+    wordsByLines: [],
+  };
+
+  componentWillMount() {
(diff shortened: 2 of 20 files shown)


techniq:
@xile611 I'm having trouble getting the CI build to pass due to what I believe is differences in font rendering with Linux/Firefox 31 and my Mac.  Works fine locally in Chrome (53), Firefox (48), and Safari (9.1).  I've tried changing and relaxing my tests to get these to pass, but I still have 1 that fails (surprising change from "2em" to "32px" still fails, even though 2em should be 32px (by default)).

Any idea how to work around this?

---

Also, my hope is to allow passing any of <Text> props defined within a component's `label` prop defined as an object down to `<Text>`.  Currently these are filtered by `ReactUtils.getPresentationAttributes`, and the additional props do not belong there (`verticalAnchor`, `lineHeight`, `fit`, `angle`, `capHeight`)

While the user could import the text label and use it within the label props, like so:

```jsx
import { Bar, Text } from 'recharts';

<Bar dataKey="value" label={<Text verticalAnchor="middle" />} />
```

it would be nice if the following was allowed:

```jsx
<Bar dataKey="value" label={{ verticalAnchor='middle' }} />
```

There are some presentation attributes that `<Text>` uses (primarily `width` to word-wrap), but it would be nice if all props were passed.

Thoughts?

techniq:
@xile611 I know you're busy, but do you have any feedback on this?  Hoping to get it merged soon if possible.

xile611:
@techniq Sorry! Too busy these days! I should take more time to review your code!

You can check [here](https://www.w3.org/TR/SVG/styling.html) for all the styling properties of svg elements.  svg elements don't support `verticalAnchor` and `lineHeight`.

techniq:
@xile611 I know SVG doesn't, but this new `<Text>` component does ;).  See also - https://formidable.com/open-source/victory/docs/victory-label/

xile611:
@techniq It should be implemented by calculating `x` and `y` attributes of `text`.

And I think `Text` should not replace all the `text` in recharts now, because the computation of text width and height would cost much.

techniq:
If you take a look [updateWordsByLines](https://github.com/recharts/recharts/pull/243/files#diff-6bc302cee0e627fcdb36a23c8e081fa6R44), the calculation is only performed if required (`width` or `fit` prop set).  Have you taken a look at the Text demo yet?

xile611:
@techniq  
Sorry. I just have a quick look at your code. I need more time to review every lines of your code.
I feel very tangled about the CSS2 styling  props of Texts, `verticalAnchor `, `lineHeight`. All the props of recharts' components are svg styling props, and the CSS2 styling props can only be added  by the props `style`. I worry that other developers of recharts will be confused.

So I prefer `Text` is a optional component to render labels, not a internal used component.
How about your opinion ?

techniq:
@xile611 I'm fine with not making it an internal component if that's the direction you prefer, I was just trying to simplify using it by not requiring the user to import and building the component themselves.  It seems like a useful feature to use it throughout, and one reason I spent a good bit of time making sure it was performant (for lack of a better word).  It only calculates the width of words if required (when a fixed width (for word wrapping) or when `fit` is set (to size the font dynamically to fit the container)).  It also only does this once at first, and will only recalc if needed (words change, for example).

I'm fine with keeping this as a component you must import and pass in to use if that's the direction you choose.  If you have any feedback on how to fix the failing test(s), I would appreciate it.  It runs run on all the machines / browsers I have, so I think it has to due with the Linux / older Firefox setup used in CI (system font differences, etc).

xile611:
@techniq Thanks! 

image

The test 

```jsx
  it('Wraps long text if styled and not enough room', () => {
    const wrapper = shallow(
      <Text width={200} style={{ fontSize: '32px' }}>This is really long text</Text>
    );

    expect(wrapper.instance().state.wordsByLines.length).to.equal(2);
  });
```

result in 3 lines when the font-family is set to be "Courier". May be you can set the fontFamily of `Text` in the test case to be a very widely used one.

techniq:
@xile611 Explicitly setting the font-family in the tests fixed the CI failures.  Thanks for the tip.

Any chance this could get merged and released soon?  There have been a [few](https://github.com/recharts/recharts/issues/238) [issues](https://github.com/recharts/recharts/issues/259) including my [own](https://github.com/recharts/recharts/issues/198) that are looking forward to some of the features this PR provides.

xile611:
@techniq Thanks! I'll try to finish today!

src/cartesian/Area.js
```diff
@@ -263,13 +264,13 @@ class Area extends Component {
       labelItem = option(props);
     } else {
       labelItem = (
-        <text
+        <Text
           key={props.key}
           {...getPresentationAttributes(props)}
```
xile611: `getPresentationAttributes(props)` can be replaced by `props`, because `Text` is no longer a simple svg element.  `getPresentationAttributes` should be done in `Text`.

src/component/Text.js
```diff
@@ -0,0 +1,162 @@
+import React, { Component, PropTypes } from 'react';
+import reduceCSSCalc from 'reduce-css-calc';
+import { PRESENTATION_ATTRIBUTES } from '../util/ReactUtils';
+import _ from 'lodash';
+
+class Text extends Component {
+
+  static propTypes = {
+    ...PRESENTATION_ATTRIBUTES,
+    fit: PropTypes.bool,
```
xile611: `fit` seems a little hard to understand. Any better name ?

src/component/Text.js
```diff
@@ -0,0 +1,162 @@
...
+      this.setState({ wordsByLines });
+    } else {
+      const words = props.children ? props.children.toString().split(/\s+/) : [];
+      this.setState({ wordsByLines: [{ words }] });
+    }
+  }
+
+  calculateWordWidths(props) {
+    // Calculate length of each word to be used to determine number of words per line
+    var svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
```
xile611: Maybe `var` should be replaced by `const`. Have you run `npm run lint` ?

src/component/Text.js
```diff
@@ -0,0 +1,162 @@
...
+    var text = document.createElementNS('http://www.w3.org/2000/svg', 'text');
+    _.assign(text.style, props.style);
```
xile611: How about:

```js
const style = {...text.style, ...props.style};
text.style = style; 
```
` _.assign` could cause some times of repaint of node.

src/component/Text.js
```diff
@@ -0,0 +1,162 @@
...
+
+    return { wordsWithComputedWidth, spaceWidth };
+  }
+
+  calculateWordsByLines(wordsWithComputedWidth, spaceWidth, lineWidth) {
+    const { fit } = this.props;
+    return wordsWithComputedWidth.reduce((result, { word, width }) => {
+      const currentLine = result[result.length - 1] || { words: [], width: 0 };
+
+      if (currentLine.words.length === 0) {
```
xile611: It seems duplicated here. Maybe these lines can be:

```js
const currentLine = result[result.length - 1];

if (currentLine && (lineWidth == null || fit || (currentLine.width + width + spaceWidth) < lineWidth)) {
   currentLine.words.push(word);
   currentLine.width += width + spaceWidth;
} else {
   const newLine = { words: [word], width };
   result.push(newLine);
}
```

But your implementation is clearer...

src/component/Text.js
```diff
@@ -0,0 +1,162 @@
...
+    }
+    if (angle) {
+      transforms.push(`rotate(${angle}, ${x}, ${y})`);
+    }
+    if (transforms.length) {
+      textProps.transform = transforms.join(' ');
+    }
+
+    return (
+      <text dy={dy} textAnchor={textAnchor} {...textProps}>
```
xile611: `getPresentationAttributes` should be called here to filter some invalidate props.

src/cartesian/Area.js
techniq: Done

src/component/Text.js
techniq: I renamed ot "scaleToFit" to be a little more expressive to what it does

src/component/Text.js
techniq: Fixed.  I tried to get `npm run lint` but it just returns with no output.  Not sure if I have an issue with my eslint but I did fix this issue.  I'm getting lint errors in my editor and when I commit (typically for lines being too long that I break down / refactor)

src/component/Text.js
techniq: I tried this originally but it appears `text.style` does not allow assignment, just it's properties.  I also had `Object.assign` at one point but found out it wasn't polyfilled and thus failed on some older browsers.

src/component/Text.js
techniq: 👍  I like this and implemented it.  I swear I originally tried this but ran into issues and needed the 3 conditions, but works now, and passes tests.

src/component/Text.js
techniq: Oddly, if I set:

```js
<text dy={dy} textAnchor={textAnchor} {...getPresentationAttributes (textProps)}>
```

I'm getting errors on other tests, so I didn't make this change.
screen shot 2016-09-21 at 11 19 05 am

techniq:
@xile611 Thanks for the great feedback.  I've implemented most of them and left comments for each).

Also, I noticed verticalAnchor was not working exactly how I planned and made corrections.  You now have good control to align the text to the 9 positions (top left, bottom right, middle middle, etc) using combinations of `textAnchor` and `verticalAnchor`.  This PR should now also resolve  https://github.com/recharts/recharts/issues/191.  I made some adjustments to the TextDemo which allows you to see the anchor point which should help demonstrate how it works.  It's useful to set the demo to:

- width: `300` (default)
- x: `150` (center)
- y: `100` (center)
- fontSize: `2em` (or greater) to get some word wrapping

Let me know if I can be anymore help on this getting merged in.

src/component/Text.js
xile611:
```jsx
const {
      ...
      className,
      ...textProps,
    } = this.props;
...


return (
      <text
        {...getPresentationAttributes(textProps)}
        className={classNames('recharts-text', className)}
        textAnchor={textAnchor}
      >
      {
        wordsByLines.map((line, index) => (
          <tspan x={x} dy={index === 0 ? startDy : lineHeight} key={index}>
            {line.words.join(' ')}
          </tspan>
        ))
      }
      </text>
    );
```

These test errors can be fixed by parsing the className of `Text`. Because `React` will check the props of native dom and svg elements.

techniq: Done

techniq:
@xile611 👍

James2516:
Does anyone know how to use this to wrap X-axis label?

basicOldCoder:
> 
> 
> Does anyone know how to use this to wrap X-axis label?

I spend time and make this work.  It look like this has be ask in places, and this thread helped me, so I reply now.

https://jsfiddle.net/basicOldCoder/b5ex8j7z/1/

(1) You must use custom tick generator (Must set interval={0} to show all tick):

`<XAxis dataKey="name" height={80} tickMargin={0} interval={0} tick={<CustomizedAxisTick/>}/>`

(2) Define the customized tick generator to use <Text> element.  Be sure make import { Text } from Recharts!:

`return (
    	<g transform={`translate(${x},${y})`}>
        <Text dy={16} fontSize={14} textAnchor="middle" verticalAnchor="start" width={80}>{payload.value}</Text>        
      </g>
    );`

(3) <Text /> has scaleToFit prop, but it work weird, so just set width unless you want some text be big some text be small. Set width and text should be wrapping.

32xnabin:
@basicOldCoder any way to do ellipsis .. in your example

Issue #183, opened 2016-08-19
LineChart with y and x datakeys?


pcmaffey:
Is it possible to create a LineChart (and AreaChart) with numeric data values for both x and y axis? I want a Line calculated with both dataKeys (so that x-axis values are not evenly distributed).

I see how to do this for a ScatterChart (and even connect with linear line), but I want a curved line, and gradients, and options available only with Line/Area charts.

Or is the only option to input connected Null points to simulate that spacing between values?

PS. Great work!

xile611:
@pcmaffey Thank you! I'll try to support Area in ScatterChart. And you can check [this issue](https://github.com/recharts/recharts/issues/154) for the discussion of two numeric axes.

kujon:
Both this issue and https://github.com/recharts/recharts/issues/154 highlight a few of shortcomings:
1) While almost all line/area charts can be drawn with x-axis being categorical, the problem occurs if you'd like to use axis' domain (imagine a line chart, where you want to draw 2 overlapping lines with different x-axes, both of which have different domains - currently impossible). Then, if you start supporting domains for categorical axes, what should happen if the domain is wider than the categories provided?
2) Replicating this scenario using scatter chart is not possible either, since it doesn't support multiple axes (and it wouldn't make much sense for it to do so).
3) Mocking numerical x and y values with scatter chart is also different in terms of behaviour - in scatter chart you interact with data points only, whereas in line/area chart you can be scrolling through their x-axis and the tooltip is always going to show something.

PR #255, opened 2016-09-20
Domain support for categorical axes

kujon:
Please refer to my concerns here: https://github.com/recharts/recharts/issues/183#issuecomment-248284741

Long story short, currently if you were to create e.g. a `LineChart` with 2 different lines and 2 different x-axes, there's no way to make those lines overlap. Think, Google Analytics chart for comparing visits in 2 different periods of time:
GA

With this functionality, you can specify domain for a categorical axis, allowing you to chop off the time period, you are not interested in. Your data could look something like this:
```javascript
const data = [
  {timestamp: '2015-01-01', oldValue: 5, newValue: null},
  {timestamp: '2015-06-01', oldValue: 4, newValue: null},
  {timestamp: '2016-01-01', oldValue: null, newValue: 3},
  {timestamp: '2016-06-01', oldValue: null, newValue: 7}
];
```

And your axis configuration could be:
```javascript
<XAxis dataKey='oldValue' xAxisId='oldValues' domain={[0, 1]} allowDataOverflow />
<XAxis dataKey='newValue' xAxisId='newValues' domain={[2, 3]} allowDataOverflow />
```

Domain configuration is very similar to that of numerical axis, however instead of operating on values, you operate on indices (indices specified are always inclusive - this is different to how `_.range` works, but similar to how you specify domain for the numerical axis), e.g.:
```javascript
const categories = [0, 1, 2];
const domain = [1, 2]; // > visibleCategories === [1, 2];

const categories = [0, 1, 2];
const domain = [-1, 'auto']; // > visibleCategories === [-1, 0, 1, 2];

const categories = ['foo', 'bar', 'baz'];
const domain = ['auto', 3]; // > visibleCategories === ['foo', 'bar', 'baz', 3];
```

~~I probably need to add `BarChart` support as well.~~ done

### DIFF
--- a/src/util/DataUtils.js
+++ b/src/util/DataUtils.js
@@ -46,11 +46,11 @@ export const parseSpecifiedDomain = (specifiedDomain, dataDomain, allowDataOverf
   let domain;
   let domainMin;
   let domainMax;
-  const specifiedDomainMin = specifiedDomain[0];
-  const dataDomainMin = dataDomain[0];
-  const specifiedDomainMax = _.last(specifiedDomain);
-  const dataDomainMax = _.last(dataDomain);
   const isCategorical = dataDomain.length !== 2;
+  const specifiedDomainMin = specifiedDomain[0];
+  const dataDomainMin = isCategorical ? 0 : dataDomain[0];
+  const specifiedDomainMax = specifiedDomain[1];
+  const dataDomainMax = isCategorical ? dataDomain.length - 1 : dataDomain[1];
 
   if (_.isNumber(specifiedDomainMin)) {
     domainMin = allowDataOverflow ?
@@ -75,7 +75,14 @@ export const parseSpecifiedDomain = (specifiedDomain, dataDomain, allowDataOverf
   }
 
   if (isCategorical) {
-    domain = _.range(domainMin, domainMax + 1);
+    domain = [
+      ...(domainMin < dataDomainMin ? _.range(domainMin, dataDomainMin) : []),
+      ...(allowDataOverflow ? dataDomain.slice(
+        Math.max(domainMin, dataDomainMin),
+        Math.min(domainMax, dataDomainMax) + 1
+      ) : dataDomain),
+      ...(dataDomainMax < domainMax ? _.range(dataDomainMax + 1, domainMax + 1) : []),
+    ];
   } else {
     domain = [domainMin, domainMax];
   }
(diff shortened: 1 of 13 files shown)


xile611:
@kujon 




Some bug happens for active dot of LineChart.

src/util/DataUtils.js
```diff
@@ -43,29 +43,41 @@ export const parseSpecifiedDomain = (specifiedDomain, dataDomain, allowDataOverf
...
     return dataDomain;
   }
+  let domain;
+  let domainMin;
+  let domainMax;
+  const specifiedDomainMin = specifiedDomain[0];
+  const dataDomainMin = dataDomain[0];
+  const specifiedDomainMax = _.last(specifiedDomain);
+  const dataDomainMax = _.last(dataDomain);
+  const isCategorical = dataDomain.length !== 2;
```
kujon: I was wondering about that, will fix.

kujon:
@xile611 will address all of those issues, thanks a lot for feedback! On a more general note: what do you think of the idea itself? I find it to be a gimmicky, but useful workaround around some of the current limitations. Ideally IMHO, however, you should simply be able to pass the data in at the Line/Area/Bar level (like you do in the Scatter), rather than at the level of the chart itself. Uber guys have got quite a similar idea here https://github.com/uber/react-vis. This, however, would mean breaking changes. Something worth considering for 0.2.0 imho.

kujon:
I've addressed both of these issues @xile611

kujon:
There seems to be an unrelated test failure. Any ideas? It doesn't happen locally.
```
<Text /> Does not wrap long text if enough width FAILED
timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

xile611:
@kujon Thanks for your advice. 
In the beginning, we want users to pass data in `LineChart` not `Line`, because we often get a table like data from server side. we would use one column to display `XAxis`, some other column to display a line or a bar.  
But in your case, LineChart seems very hard to use. So I check HightChart for some ideas. I found HightChart support two kind of  data([series data](http://www.highcharts.com/demo/line-basic),  [table data](http://www.highcharts.com/demo/line-ajax)). Maybe recharts should support two kind of data too, but the implementation is to be determined.

xile611:
And on this issue, It seems to be gimmicky. May be we can support both XAxis and YAxis to be numerical axis in LineChart, BarChart, and AreaChart.

kujon:
Thanks for feedback. I'll close this PR then and try to come up with a more reasonable solution.

Issue #261, opened 2016-09-21
What should I do if I want to to add custom onClick event to LineChart?


weita0:
I only found mouse event interface in bar chart and pie chart

xile611:
@weita0  Can you describe detail what kind of event do you want to bind to LineChart?

weita0:
For example, I would like to do something like showing details right after the user click the vertical line.

Hectorhammett:
I'm using a Line Chart, and i'm in this trouble as well, i need the user to click a point, and i want to grab the information on that specific point so i can query the database to give the user more detailed information.

A onClick event would be greatly appretiated!. 

Excelent package for react btw.

Hectorhammett:
@weita0 I created a fork of Recharts and added a On Click functionality:
https://github.com/Hectorhammett/recharts/tree/package

I've created it until they add a complete functionality, use it if it serves you.

To use it, in the LineChart element, add a prop called onDotClick (The event fires when you click a dot) and pass a function to it.

The function returns in the first parameter, the label on the bottom axis (x axis) and its value.

@xile611 I don't think my code is perfect, but for the moment i can work with these, you have any plans to add this feature in the future?

weita0:
@Hectorhammett  Excellent job!

Hectorhammett:
@weita0  Thanks! 

I could've made m code better but i was focused in fixing my problem, would you like me to ask for a Pull Request? Again, my code right now is not perfect, but if you like it, i could improve it.

JasperStam:
@Hectorhammett 

Making a PR would be awesome as I have the exact same issue!

xile611:
@Hectorhammett Could you submit a PR? Then I can check the changes of your code much clearer.

jaredramirez:
I found an alternative solution, somewhat hacky though.

To accomplish getting data about the specific point clicked, I first set the `activeDot` prop on the `Line` component to false to removed it entirely. Then I created a custom label component with the same shape/styles of `activeDot` component, and added it to `Line`. Because the label has `payload` as a prop, you now have access to the data behind the point!

LineChart component:

`
import LabelAsPoint from './LabelAsPoint.js';

...

<LineChart {/* your props here */} >
    // other components such as <XAxis />, <YAxis />, < Tooltip />, etc
    <Line
        label={ <LabelAsPoint /> }
        activeDot={false}
        type="monotone"
        dataKey={ key }
        stroke={ colors[index] }
    />
/>`

[...]

PR #289, opened 2016-10-11
add support for onClick on Tooltip

raspo:
Adds an optional `onClick` prop for the `Tooltip` component.

Any chart that implements a `Tooltip` with the `onClick` prop is now clickable. The callback arguments contain the content of the tooltip as well as the index of the clicked element.

Solves the issue of not being able to have a click event on line charts #261 

Let me know if this solution is ok. I originally started working on this because I needed to add a click event on a line chart. I couldn't find any better place to put the `onClick` listener. I tried using the chart `Dot` as click target, but I found it too difficult to click during testing.

### DIFF
--- a/demo/component/BarChart.js
+++ b/demo/component/BarChart.js
@@ -213,6 +213,10 @@ export default React.createClass({
     console.log(`Pv Bar (${index}) Click: `, data);
   },
 
+  handleClickTooltip(data, index, e) {
+    console.log(`Tooltip (${index}) Click: `, data);
+  },
+
   render() {
     const { data, data01, data02 } = this.state;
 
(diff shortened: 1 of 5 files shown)


xile611:
@raspo It's strange to add `onClick` event in `Tooltip`, but  bind event handler in Chart.

raspo:
@xile611 I hear you, do you think there is a better place to put it?

Ideally we would want to use the `Line` or the `Dot` components, but those targets are way too small to be usable.
The `onClick` on `Tooltip` definitely feels strange, but it somehow makes sense as the callback is returning the content of the tooltip, and the `onClick` is triggered only when a tooltip is showing.

xile611:
@raspo How about add a props `triggerEvent` in `Tooltip` ?

raspo:
@xile611 not sure what you mean by that, just renaming `onClick` to `triggerEvent`?

xile611:
@raspo I mean add a props `triggerEventType` which can be 'hover', 'click', and the default value is 'hover'.

yelynn2015:
@xile611 How about Pie chart? If I want to click the area in pie chat, then the tooltip would be presented, how should I do?

PR #307, opened 2016-10-22
Add angle property to PRESENTATION_ATTRIBUTES

joaosoares:
Adding the angle property to the PRESENTATION_ATTRIBUTES array in util/ReactUtils.js allows ticks in the cartesian axes to be positioned at any angle. It was not listed as a presentation attribute before, but I believe it should be, as there is no other way to rotate each tick's text while using tickFormatter, i. e., without creating a custom React class as it's done in the [CustomizedLabelLineChart](http://recharts.org/examples#CustomizedLabelLineChart) in the docs.

### DIFF
--- a/src/util/ReactUtils.js
+++ b/src/util/ReactUtils.js
@@ -3,6 +3,7 @@ import _ from 'lodash';
 
 export const PRESENTATION_ATTRIBUTES = {
   alignmentBaseline: PropTypes.string,
+  angle: PropTypes.number,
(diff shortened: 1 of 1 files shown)


pacarvalho:
Which release contains this change?

arcthur:
I'm sorry.
I already publish 0.15.3. Our work is very busy recently.

pacarvalho:
Thank you!

Issue #287, opened 2016-10-10
Real-time chart


sdoomz:
Hi! 
I'am trying to figure out how to connect rechart with async data without full re-rendering. 
Every second I receive new point for area chart and then call setState() with new data collection. That triggers re-rendering :( 

But my questions are:
1. How to keep previously focused tooltip in opened state after new data was added?
2. How to apply only new data without full re-rendering? 
3. How to move xAxis smoothly? [d3 example](https://bost.ocks.org/mike/path/)

Seems like solving the second issue may solve first automatically, or? 
http://jsfiddle.net/sdoomz/y6dw0fgh/

xile611:
@sdoomz 

For your questions:

1. Tooltip is triggered by mouse event now, may be you can use other method to display the new data, for example an `position: absolute;` div or a customized label of `Area`.

2. Temporarily no idea.

3.D3 use path animation to switch path, I have no idea how to add path transition animation when path update now.

sdoomz:
@xile611 Currently it is possible to solve issue with [dynamic charts](http://jsfiddle.net/sdoomz/ams3jyj2/) using Highcharts.js and [react-highcharts](https://github.com/kirjs/react-highcharts) package. But it would be great to see same functionalities in next versions of Recharts.

xile611:
@sdoomz Thanks for great suggestions! We'll try to implement this functionality, maybe through a wrapper component.

akvanhar:
I would like to add another vote for this feature.
I've been experimenting with it on my own by using AnimateGroup from react-smooth as a wrapper to my recharts components.

SpaceK33z:
I'm also looking to implement realtime data, but there are major frame rate issues at the moment. PR #277 will probably fix some of these issues.

akvanhar:
@xile611 would love to hear an update about this enhancement.

xile611:
@akvanhar Yeah, We're trying to refactor the animation part. But it may take some time.

akvanhar:
Thanks for the update. I'll stay tuned and will try to implement a work-around in the meanwhile.

xile611:
OK! Both PR and suggestions are welcome. @akvanhar

ruaanvds:
Is there any known workaround for this at the moment?

diegocasmo:
[...]

PR #340, opened 2016-11-11
Save chart's state after properties have been updated

sdoomz:
According to this [issue](https://github.com/recharts/recharts/issues/287) it is not possible to save tooltip's position after update.


xile611:
@sdoomz It needs to be discussed whether to update state when data is updated.

PR #353, opened 2016-11-18
Fix #351: Relax propTypes.ticks in CartesianAxis

### LINKED ISSUE #351: PropTypes warning when using XAxis with ticks that are array of strings
Tested with recharts v0.18.0.

Sample code:

```js
      const xAxisTicks = ['Oct 1', 'Oct 2'];

      <LineChart data={chartData} margin={margin} width={width} height={height} >
        <XAxis dataKey={xDataKey} tickLine={false} interval={0} ticks={xAxisTicks} />
        <YAxis
          orientation="right" tickLine={false} width={50}
          tickFormatter={yAxisLabelFormatter}
        />
        <CartesianGrid stroke="#eee" />
        <Tooltip />
        <Line
          type="linear" dataKey={yDataKey} strokeWidth="3" dot={false}
          isAnimationActive={false}
          formatter={lineFormatter}
        />
      </LineChart>
```

PropTypes warning:

```
warning.js:36 Warning: Failed prop type: Invalid prop `ticks` supplied to `CartesianAxis`.
    in CartesianAxis (created by AnimationDecorator(ComposedDataDecorator(LineChart)))
    in AnimationDecorator(ComposedDataDecorator(LineChart)) (created by SpacedGridChart)
    in SpacedGridChart (created by ContainerDimensions)
```

Workaround:

CartesianAxis.propTypes.ticks = React.PropTypes.any;

mwolson:
When interval=0, it's allowed to be an array whose elements are any type.

### DIFF
--- a/src/cartesian/CartesianAxis.js
+++ b/src/cartesian/CartesianAxis.js
@@ -39,13 +39,7 @@ class CartesianAxis extends Component {
     tickLine: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
 
     minLabelGap: PropTypes.number,
-    ticks: PropTypes.oneOfType([
-      PropTypes.arrayOf(PropTypes.shape({
-        value: PropTypes.any,
-        coordinate: PropTypes.value,
-      })),
-      PropTypes.arrayOf(PropTypes.number),
-    ]),
+    ticks: PropTypes.array,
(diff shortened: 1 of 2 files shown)


mwolson:
Note: this new test case won't actually fail on your current master branch because it will cause a non-fatal proptype warning rather than an exception.

PR #389, opened 2016-12-06
set a class on a legend item

briice:
**Why**

I need to customize the legend items. In my case, I've added the ability to toggle elements from a chart and I want to style legend items accordingly.
I could have created a custom `Legend` component for that, but I think it's a nice feature to add to the default one.

**How**

This PR makes it possible by adding a `className` attribute to the `payload` property in the `Legend` component. If set, this string is appended to the other classes on the `li` element.

* Tests are passing
* I have updated the doc on a branch. I'll send a PR if you think this feature is worth it.

### DIFF
--- a/src/component/DefaultLegendContent.js
+++ b/src/component/DefaultLegendContent.js
@@ -23,9 +23,9 @@ class DefaultLegendContent extends Component {
       value: PropTypes.any,
       id: PropTypes.any,
       type: PropTypes.oneOf([
-        'line', 'square', 'rect', 'circle', 'cross', 'diamond', 'square',
-        'star', 'triangle', 'wye',
+        'circle', 'cross', 'diamond', 'line', 'rect', 'square', 'star', 'triangle', 'wye',
       ]),
+      className: PropTypes.string,
     })),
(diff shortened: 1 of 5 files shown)


briice:
Hey, are you just closing it without an explanation? Do you find it useless?

xile611:
@briice It seems strange to add `className` in  `data`.

src/component/DefaultLegendContent.js
```diff
@@ -99,19 +99,26 @@ class DefaultLegendContent extends Component {
...
-      >
-        <Surface width={iconSize} height={iconSize} viewBox={viewBox} style={svgStyle}>
-          {this.renderIcon(entry, iconSize)}
-        </Surface>
-        <span className="recharts-legend-item-text">{entry.value}</span>
-      </li>
-    ));
+    return payload.map((entry, i) => {
+      let className = `recharts-legend-item legend-item-${i}`;
+      if (typeof entry.className === 'string' && entry.className.length > 0) {
```
xile611: Such kind of  if-else seems quite specified.

PR #392, opened 2016-12-10
allows to set a class on a legend item

briice:
**Why**

I need to customize the legend items. In my case, I've added the ability to toggle elements from a chart and I want to style legend items accordingly.
I could have created a custom Legend component for that, but I think it's a nice feature to add to the default one.

**How**

This PR makes it possible by adding a className attribute to the payload property in the Legend component. If set, this string is appended to the other classes on the li element.

- Tests are passing


briice:
Sorry I didn't mean to create a PR here but on a fork so we can use our modifications in the meantime.

PR #399, opened 2016-12-15
Fix #267: Add Cartesian Error Bar Component [enhancement]

### LINKED ISSUE #267: error bars for line charts/bar charts
Would be super sweet if you could somehow pass an error array as a prop to make error bars on line and bar charts (especially bar chart). 

Thanks for making this tool!

McMenemy:
I made a general Errorbar component that can be attached to Scatter, Line, and Bar components as a prop (example code below).

```javascript
const data01 = [ { x: 100, y: 200, z: 200, errorY: [20, 30], errorX: 30 } ...]

<Scatter name="A school" data={data01} fill="#ff7300">
  <ErrorBar dataKey="errorY" width={8} strokeWidth={1} width={0} stroke="blue" direction="y" />
  <ErrorBar dataKey="errorX" width={4} strokeWidth={2} opacity={0.5} stroke="green" direction="x" />
</Scatter>
```

screen shot 2016-12-16 at 6 51 11 pm

screen shot 2016-12-14 at 5 53 38 pm

screen shot 2016-12-16 at 7 04 02 pm


McMenemy:
Also, I know we talked about making this work for candlestick charts, but I actually think they are different conceptually because an error bar is a value you add onto an existing data point (similar to a label) where as a candlestick chart is an actual chart type where each point consists of four values displayed in an histogram like manner. After the error bar PR goes through I can work on the candle chart if you would like.

xile611:
@McMenemy  Thanks!

I have two kinds of thoughts about ErrorBar and candlestick chart now.

1. Add a new component `ErrorBar`, and it should be the child of `Line`, `Area`, `Scatter`, `Bar`, etc.So we can add error bars like the following:

```jsx
  <Line dataKey="pv">
    <ErrorBar dataKey="pvError"  />
  </Line>

  <Scatter data={[...]}>
    <ErrorBar dataKey="pvError"  />
  </Scatter>
```
2.The new component `ErrorBar` should be the child of `LineChart`, `AreaChart`, etc. Something link the error bar of [highcharts](http://www.highcharts.com/demo/error-bar). So we just need add ErrorBar like the following:

```jsx
  <LineChart >
    <Line dataKey="a" />
    <ErrorBar dataKey="aError" />
  </LineChart>

  <ScatterChart>
    <ErrorBar dataKey="pvError"  />
  </ScatterChart>
```

And I also want to know whether the value of an error bar will  be an interval like the demo of highcharts?

src/cartesian/Line.js
```diff
@@ -28,9 +29,10 @@ class Line extends Component {
     ]), PropTypes.func]),
     unit: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     name: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-    dataKey: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
     yAxisId: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     xAxisId: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+    yAxisMap: PropTypes.object,
```
xile611: Why `yAxisMap` and `xAxisMap` ? not `xAxis` and `yAxis` ?

src/cartesian/ErrorBar.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * @fileOverview Render a group of error bar
+*/
+import React, { Component, PropTypes } from 'react';
+import Layer from '../container/Layer';
+
+class ErrorBars extends Component {
```
xile611: The className is different from filename.

xile611: And the value of errorBar will change the domain of `YAxis` in a horizontal LineChart. We should calculate the value of errorBar before calculating the domain of `xAxisMap` and `yAxisMap`.

src/cartesian/Line.js
McMenemy: I was getting the correct xAxis and yAxis by using the respective Ids and Maps. Is there a better way to do this?

McMenemy:
Thanks for the feedback! 

Option 1 of your two ideas looks the best to me.

I'll also make the error an interval instead of a single value. (My use case only needed one value but I can see how people would also need intervals.) 

I'll aim to have these changes done by this evening.

src/cartesian/ErrorBar.js
McMenemy: fixed

McMenemy: Not really sure the best way to change the domain of YAxis before adding error bars. Any thoughts?

McMenemy:
@xile611  Let me know if anything else needs to be done before merge. I think it should be close.

src/cartesian/Line.js
xile611: @McMenemy  I think this mapping can be done in `LineChart`.  Each `Line` has just one x-axis and one y-axis, It's strange to parse `xAxixMap` and `yAxisMap` to `Line`.

src/cartesian/ErrorBar.js
```diff
@@ -73,9 +75,9 @@ class ErrorBar extends Component {
 
       return (
         <Layer className={`recharts-errorBar-${i}`} key={i}>
-          <line {...coordsTop} stroke={fill} strokeWidth={strokeWidth} />;
-          <line {...coordsMid} stroke={fill} strokeWidth={strokeWidth} />;
-          <line {...coordsBot} stroke={fill} strokeWidth={strokeWidth} />;
+          <line {...coordsTop} stroke={fill} strokeWidth={strokeWidth} strokeOpacity={opacity} />;
```
McMenemy: fixed

src/cartesian/Scatter.js
```diff
@@ -162,12 +162,35 @@ class Scatter extends Component {
...
+          dataPointFormatter: dataPointFormatterX,
+        });
+      }
+
+      return React.cloneElement(item, {
+        key: i,
+        data: points,
+        xAxis,
+        yAxis,
+        layout,
```
McMenemy: Fixed. Thanks for pointing out that way to clean up the code.

src/cartesian/Line.js
McMenemy: fixed

McMenemy:
@xile611 I think my most recent updates covered everything you pointed out in the review.

Issue #405, opened 2016-12-20
Tooltip performance


konstantin24121:
Tooltip consume too much CPU load if the cursor is moved very fast. HandleMouseMove method of generateCategoricalChart reacts to mouseMove event. Handler on every tick changed state and it causes a wave of rerender. It doesn't behave the same way if you move cursor over a plot. Otherwise a lot of CPU load is allocated wich is critical fir low end devices. What do you think about adding a throttleDelay between events? Those who dont care about performace can leave throttleDelay empty. According to my tests 200ms delay decreces load from 25% to 8-10%, i attached screenshots below.

I could make a pr.

Before: 
![default](https://cloud.githubusercontent.com/assets/9361325/21359127/fdf120a8-c6fc-11e6-91b4-b1f4c938be0a.PNG)

With throttle

![withthrottle](https://cloud.githubusercontent.com/assets/9361325/21359139/06951b6a-c6fd-11e6-8745-bec85eaadaf5.PNG)

xile611:
@konstantin24121 It's a good suggestion to add throttle delay.  Maybe you can submit a pr ?

And I have a  thought now. The events to show and update tooltip change the state frequently which will also update other unchanged components, like `Legend`, `XAxis`, etc.  If the state about tooltip events could be managed in `eventCenter` of recharts not in `state`, only tooltip will update when mousemove event is triggered.

xile611:
<img width="850" alt="2016-12-21 11 04 20" src="https://cloud.githubusercontent.com/assets/6111424/21376365/4cccda56-c76f-11e6-8da9-1b83326ec53a.png">
<img width="889" alt="2016-12-21 11 04 09" src="https://cloud.githubusercontent.com/assets/6111424/21376366/4ee1d8e6-c76f-11e6-8394-f07c05714914.png">
<img width="1432" alt="2016-12-21 11 03 41" src="https://cloud.githubusercontent.com/assets/6111424/21376371/5b9330b2-c76f-11e6-9dfd-9ff703e1cd32.png">


I try to add `throttleDelay`, but I can't get `event.pageX` every time even I have used `event. persist()`.

[Related issue](https://github.com/facebook/react/issues/7778)

konstantin24121:
@xile611 this (article)[http://blog.revathskumar.com/2016/02/reactjs-using-debounce-in-react-components.html] has explain

PR #406, opened 2016-12-21
Add throttle delay props

konstantin24121:
This props need for performance. If developer not worried about
performance he can leave this prop empty.

I made performance measurement, screens below. All tests were performed on one chart

**No throttle**
nothrottle

**200ms throttle**
throttle200

**400ms throttle**
throttle400


konstantin24121:
See #405

Issue #363, opened 2016-11-23
React Dom Server 15.4.0


earthinferno:
Given that your library has a dependency on react-dom-server and that the good folks at React state 'We don't recommend using React DOM Server on the client in most cases'; are there any plans to remove this dependency or separate it out into a second library?

xile611:
@earthinferno Yeah, of course. But no good solution now.

xiaokekeT:
https://github.com/recharts/recharts/blob/master/src/component/Legend.js#L5

remove this line ?

xile611:
Yeah! It should be removed.

PR #416, opened 2016-12-28
fix: Delete unused react-dom/server #363

xiaokekeT:


xile611:
@xiaokekeT Thanks! It has been finished.

PR #500, opened 2017-02-04
fix: isomorphic/universal rendering failing due to Date.now()

sankalplakhina:
isomorphic/universal rendering failing due to Date.now() usage for id creation in Pie component

- Included a new prop 'clipPathId' for Pie component which can be used for isomorphic apps

### DIFF
--- a/demo/component/Pie.js
+++ b/demo/component/Pie.js
@@ -18,6 +18,7 @@ export default React.createClass({
           endAngle={0}
           cx={250}
           cy={250}
+          clipPathId="pie-demo-id-1"
           outerRadius={200}
           innerRadius={180}
           data={data}
--- a/src/polar/Pie.js
+++ b/src/polar/Pie.js
@@ -71,6 +71,7 @@ class Pie extends Component {
       'spring',
       'linear',
     ]),
+    clipPathId: PropTypes.string,
   };
 
   static defaultProps = {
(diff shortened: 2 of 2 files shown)


sankalplakhina:
@xile611 Request you please take a look at my pull request. Would be helpful if this can be merged in and released with next version. 

I am awaiting for this fix and release to update this library accordingly in personal project.



P.S. Great job done with D3 and react here! I developed few charts with D3+React for my projects as well and I totally understand the effort it takes to get them right :) Really like your project here. Do let me know in case I can help you with anything here.

xile611:
@sankalplakhina Thanks! I think we just need to use lodash.uniqueId to generate the unique id of `Pie`.

sankalplakhina:
@xile611 Yes, that could possibly fix this case.

But there's one more case I have in mind which might fail if we do the above. Since we are assuming lodash will be imported as a singleton on both server and client and therefore would generate the same unique id, but what about the case where ` lodash.uniqueId` is being used at server for any other application specific task before server rendering. That would take the lodash's unique id counter one step ahead of what client would be calculating. 

I suggest we should still keep `clipPathId` as prop and replace the calculation in `defaultProps` with ` lodash.uniqueId`.  That could cover all our cases. 

Please let me know if you agree and I will reopen this PR with necessary changes.

xile611:
@sankalplakhina How about add `uniqueId` in the utils of recharts ? I think the `id` is only a internal variable which should be unique in a single page, keep `clipPathId ` as prop is needless.

sankalplakhina:
@xile611 Completely agree on `clipPathId` as prop is needless. Only suggesting it as it was giving me control from my app to fix this scenario. However, your suggestion of having rechart's own `uniqueId` generator function should solve this 👍

xile611:
@sankalplakhina Thanks! I'll refactor this.

PR #504, opened 2017-02-08
lodash method per package

OlivierCo:
- direct load method as a package from lodash definition
- instead of loading the entire lodash library just load what needed.


xile611:
@OlivierCo We have use `lodash-webpack-plugin` and `babel-plugin-lodash`. I think recharts only require the necessary files of `lodash`. You can check the `lib` of `recharts`.

PR #516, opened 2017-02-13
Update Legend.js

tgyou:
prevent re-render change dimension with legend for server side rendering


tgyou:
I am so sorry. I thought it was PR from a forked repo.

PR #526, opened 2017-02-20
Update Pie.js

davide-giardini:
Now you can pass an explicit `id` to the Pie component in order to better support server side rendering.


xile611:
@davide-giardini Can you give an example why we need to set the `id` of `Pie` in SSR ?

xile611:
@davide-giardini  The PR hasn's been updated for a long time, you can reopen it if any updates is submitted.

zation:
This is because the implementation of `uniqueId` in DataUtils.js. In server side, DataUtils.js will be imported once but run several times across several requests. So in client side, the id will be much smaller than server side. That causes checksum issue.

Issue #446, opened 2017-01-11
Display data in Brush


ZephD:
Is it possible to display graphable data in the brush?

For example, the dataKey is the x-axis key, but can we also have show one (or more) of our y-axis data in the brush?

[Like this](https://i.stack.imgur.com/JYBwS.png)

GregPhyn:
Has this been implemented?
Is usage like 
<Brush dataKey="the_dataKey" /> ?

xile611:
@GregoryPotdevin Not yet.

Eric24:
+1

evisong:
Hi, I've created a PR #586 for this ticket, could you please help review it when you have some time? Thanks.

xile611:
Fixed in version 0.22.0

gknapp:
This capability is available from 0.22.0? In a later issue (#792) @xile611 said it's not implemented yet?

https://github.com/recharts/recharts/issues/792

I'm confused, can this be achieved with recharts as per #586 ?

Issue #205, opened 2016-09-01
Legends within parent SVG


derrickpelletier:
Curious if there is a particular reason that the legend is not part of the main chart `<svg>`?

Use-case is that users customize charts within our app, and when done are able to download/export them as either a standalone `.svg` or `.png`. In either case, the legend being separate complicates this.

derrickpelletier:
Maybe even a separate legend component designed to be composed into the main svg, keeping the functionality of both... like an `<SvgLegend>` or something. Just thinking out loud...

xile611:
@derrickpelletier We use dom nodes to render legend because the layout of dom nodes is much simpler than svg nodes. For svg nodes, we must calculate the width of each text.

SachaG:
I'm working on making charts exportable as images (to let people share them as Twitter cards) so I also think this would be useful.

derrickpelletier:
@SachaG I had started a decorator here based on that need: [react-download-svg](https://github.com/derrickpelletier/react-download-svg) for (ideally) use with any svg-based component. It's just a first working pass, open to ideas on a better api for it 👌

SachaG:
This looks really cool! Does it convert the SVG to PNG right in the browser? I didn't even know you could do that!

That being said for Twitter Cards I need to be able to point my meta tag at an image file, so I think I would have to do the conversion server-side first and host the file somewhere.

ammmze:
Anybody try [html2canvas](https://html2canvas.hertzen.com) which I guess can export to PNG? Or maybe this one https://github.com/tsayen/dom-to-image/blob/master/README.md

I'm gonna need this functionality soon, so will be trying it out soon-ish

marcramser:
Any update? We would also need it to export the chart (with the legend) to png. Or does anyone else has a solution?

cristianocca:
> 
> 
> Any update? We would also need it to export the chart (with the legend) to png. Or does anyone else has a solution?
> 
> Update:
> I used https://github.com/tsayen/dom-to-image/blob/master/README.md and it worked perfectly fine.
> 
> ```
>   exportChart = () => {
>     if (this.currentChart && this.currentChart.container) {
>       domtoimage.toPng(this.currentChart.container, { bgcolor: '#ffffff' })
>         .then((dataUrl) => {
>           const download = document.createElement('a');
>           download.href = dataUrl;
>           download.download = `chart.png`;
>           download.dispatchEvent(new MouseEvent('click'));
>         })
[...]

PR #586, opened 2017-03-21
Support embedded chart as panoram in Brush

evisong:
We're leveraging Recharts in our project (thanks for the great lib!), and #446 became a key dependency of ours recently.

This PR is trying to add to Brush an optional embedded chart as a "panoram". This chart: 

- Shares the same data with the outer one, but ignores startIndex and endIndex;
- Shows in "compact" mode, which only renders the chart content (Area, Bar, Line);
- Keeps the same size with containing brush, lays in front of brush background and behind brush slide.

```
  <Brush>
    <LineChart>
      <Line dataKey="uv" stroke="#ff7300" dot={false} />
    </LineChart>
  </Brush>
```


caiolopes:
Great!! I hope it will be merged soon...

evisong:
Here is a screenshot of panoramic brush (similar to Highstock): 

spark_ui

src/chart/generateCategoricalChart.js
```diff
@@ -44,6 +44,7 @@ const generateCategoricalChart = (ChartComponent, GraphicalChild) => {
     static propTypes = {
       ...ChartComponent.propTypes,
       syncId: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+      compact: PropTypes.bool,
```
xile611: @evisong About this name of this props, do you have any idea when user want to use this props to generate svg elements (something describe in [this issue](https://github.com/recharts/recharts/issues/205)) ?

evisong: @xile611 For this `compact` prop, my main consideration was to exclude:

- HTML DOMs including <div>, Legend and Tooltip
- X-axises and y-axises
- Brush

All those exclusions are specifically made for the "panoramic brush" use case. Do you think this is way too specific? When looking back into this, I found myself may still need grid inside the panorama. Maybe it's better to extend this prop to an array (of chart parts to be excluded, like `['grid', 'chart', 'children']`... a little tricky...)?

Anyway for this MR, this prop is necessary, since when cloning element in `Brush`:  https://github.com/recharts/recharts/pull/586/files/6c29a98d8e445ffe456d968cd9b0a3080e47b5bb#diff-44d40453cfbdcbd566bcc8d82acd5c5bR306 , I don't have any chance to exclude parts inside the wrapping `Surface`.

Looking forward to your further feedbacks.

----

About https://github.com/recharts/recharts/issues/205 you mentioned, since certain parts (HTML, Tooltip, etc.) would be excluded by setting `compact`, it would naturally become one single exportable svg. But I don't think this meets the requirement in #205, as Legend is excluded as well. 

Just like you mentioned in that ticket, (HTML) DOM nodes handles flow layout better than SVG nodes. My experience also tells that embedding HTML DOM node in SVG node is much harder than embedding SVG node in HTML DOM ones, due to poor compatibility of svg foreignObject: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/foreignObject#Browser_compatibility. So any component that contains, or possibly contains HTML, is better to be rendered at outside of chart's main SVG, which is exactly the current solution in Recharts (i.e. Legend and Tooltip).

As #205 requested, exporting to all-in-one SVG, requires either: 

- Implement Legend with pure SVG, need to calculate text's width and handle potential line wrapping, or:
- Keep using HTML, but convert it to SVG after rendered, unfortunately this is not doable in browser: http://stackoverflow.com/a/26963688/245345 , so this must be done together with a backend.

src/cartesian/Brush.js
```diff
@@ -280,6 +294,26 @@ class Brush extends Component {
     );
   }
 
+  renderPanoram() {
```
evisong: My bad, the noun form of panoramic is panorama... Will push a fix soon.

mooncakelmn:
+1 to have this merged soon...

src/cartesian/Brush.js
evisong: Fixed typo in 985fc69

src/chart/generateCategoricalChart.js
evisong: > All those exclusions are specifically made for the "panoramic brush" use case. Do you think this is way too specific? When looking back into this, I found myself may still need grid inside the panorama. Maybe it's better to extend this prop to an array (of chart parts to be excluded, like ['grid', 'chart', 'children']... a little tricky...)?
>
> Anyway for this MR, this prop is necessary, since when cloning element in Brush: https://github.com/recharts/recharts/pull/586/files/6c29a98d8e445ffe456d968cd9b0a3080e47b5bb#diff-44d40453cfbdcbd566bcc8d82acd5c5bR306 , I don't have any chance to exclude parts inside the wrapping Surface.

As what I've done in e597e3b, `compact=true` now would render complete svg contents within the wrapping `<Surface />`. 

If a user would like to render a panoramic brush, he needs to specify a certain Chart as child of `<Brush />`. Brush will ensure that to `compact=true` is set to its containing Chart, but user needs to specify what the contents are inside the Chart:

```jsx
<Brush>
  <AreaChart>
    <CartesianGrid />
    <YAxis hide domain={['auto', 'auto']} />
    <Area dataKey="price" stroke="#ff7300" fill="#ff7300" dot={false} />
  </AreaChart>
</Brush>
```

Please refer to a complete demo in 7470e8e.

Anyway, Tooltip, Legend and event handlers will be ignored when `compact=true` is set to the Chart.

evisong: > As #205 requested, exporting to all-in-one SVG, requires either:
> 
> - Implement Legend with pure SVG, need to calculate text's width and handle potential line wrapping, or:

As of e597e3b, Chart's `compact=true` may also be used as a foundation of "all-in-one SVG", once pure SVG Legend is supported, #205 could be resolved.

evisong:
@xile611 I've made several adjustments, please refer to new comments above and review the changes when you've some time. Thanks.

Issue #581, opened 2017-03-18
Changing the number of Lines in LineChart always redraws the last Line


eanders-MS:
Thanks for this awesome library. I think I might have found a bug. I have a LineChart with three Lines. I pass the Lines to LineChart via an array: `{lines}`. When populating the lines array, if I skip any of the lines in the array, when I later re-include the skipped Line, the last Line in the array will reanimate back in.

Repro on JsFiddle: http://jsfiddle.net/dimetrodon/hhg4a89q/

skvale:
@eanders-MS It looks like LineChart doesn't use the keys you are passing in on the `<Line />` component.  Instead it is computing keys on the `<LineChart />` children by index.  What initially seems off here is that the newly added `<Line />` is becoming the first index.

So the newly added one gets a key of `line-0` and the last one goes from `key=line-1` to `key=line-2`, causing the re-render.

As a quick work-around, see JsFiddle http://jsfiddle.net/4z48g1bL/, which clears up the poorly computed keys by reversing the lines before the render. (:eyes: Line 19)

eanders-MS:
Thanks so much @skvale !

PR #615, opened 2017-04-07
fix: re-rendered element bug when adding new elements

skvale:
Fixes: #581 

[index as a key is an anti-pattern](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318)

If you add new children to a chart it will compute the key based on the index of that element.

A problem comes up when you add a new child to the beginning of the children array.  This will bump the other indexes forward and cause a re-render if the key has changed.

The fix I propose is to use children keys if present so a user can specify their own.

Perhaps a more robust solution would be to use something like `child.props.dataKey` as the default.
`key: child.key || child.props.dataKey`

### DIFF
--- a/demo/component/LineChart.js
+++ b/demo/component/LineChart.js
@@ -419,6 +419,7 @@ export default React.createClass({
         </div>
 
         <p>LineChart with two y-axes</p>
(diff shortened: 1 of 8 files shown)


xile611:
@skvale The `dataKey` can be a function now, so we can only use `child.key`.

PR #620, opened 2017-04-10
Add `none` legend type

### LINKED ISSUE #368: Omit specified line/bar/area/etc. from legend and tooltip
Is there a way to omit a specific line/bar/area/etc. from the legend and tooltip, other than using a custom content for these? It would be desirable to be able to control what data series appear in the legend and tooltip at the data series component level.

IanVS:
Fixes #368 

This does two main things:

- Consolidates the legend type PropType array into one place, in `ReactUtils`, to ensure they stay consistent.
- Adds a `"none"` item to valid legend types, which if used will cause the legend item to not be included.  This is useful in some cases for example when you are plotting comparison data between two periods, but you don't want to have two legend items.

I've also added a spec and demo as well.

### DIFF
--- a/demo/component/Legend.js
+++ b/demo/component/Legend.js
@@ -23,15 +23,15 @@ const data3 = [
 ];
 
 export default React.createClass({
-  render () {
+  render() {
     return (
       <div>
         <div style={{ position: 'relative', height: 200 }}>
           <Legend width={500} height={30} payload={data} />
         </div>
 
-        <div style={{ position: 'relative', height: 200 }}>
-          <Legend layout='vertical' width={200} height={100} payload={data2} />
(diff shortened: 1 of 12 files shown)


IanVS:
I've also submitted https://github.com/recharts/recharts.org/pull/31 for the doc updates.

Eric24:
@IanVS : Thank you for this solution; it works great for the legend. I'm also needing a way to omit an item from the tooltip. Currently, even setting the name to " " (which was a suggestion elsewhere) still renders the value in the tooltip. Do you have any thoughts on an approach to this?

IanVS:
It's been a while since I took a look at this, but my best guess is that you'll want to render a custom tooltip component, so that you'll have full control of what is and isn't shown.  Maybe there's a better way, but that's what we do.

PR #622, opened 2017-04-12
use prop-types instead of React.PropTypes

jochenberger:
As of React 15.5, `React.PropTypes` is deprecated. https://facebook.github.io/react/blog/

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "d3-scale": "1.0.4",
     "d3-shape": "1.0.4",
     "lodash": "~4.17.4",
+    "prop-types": "^15.5.7",
     "react-resize-detector": "0.3.3",
     "react-smooth": "0.1.20",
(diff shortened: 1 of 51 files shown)


cmdrmcdonald:
The dependency on react-resize-detector means that this error will still occur when using recharts.  It looks like the author of react-resize-detector has updated their package; if the version number for this dependency can be bumped to 0.4.1 it should remove the error.

jochenberger:
@cmdrmcdonald then I suggest, you create a PR to address that.

nrkin:
Hello
One observation. Because of the above change, I need to explicitly add the prop-types library to our dependencies although I haven't yet migrated to 15.5+ (actually I am migrating to React 15.4.2 still).
I wanted to understand whether Recharts 0.22.4 will work fine with 15.4.2?
Thanks!

PR #645, opened 2017-04-21
Dot custom classname bug fix

### LINKED ISSUE #644: custom classname to Dot from Line
Bug

When i try to set custom class for dots from, it's not working
```javascript
<Line type="monotone" dataKey="uv" dot={{r: 6, className: 'dot-red'}} />
```

The Dots circle won't have `dot-red` className cause

kwolfy:
#644

### DIFF
--- a/src/cartesian/Line.js
+++ b/src/cartesian/Line.js
@@ -268,7 +268,8 @@ class Line extends Component {
     } else if (_.isFunction(option)) {
       dotItem = option(props);
     } else {
-      dotItem = <Dot {...props} className="recharts-line-dot" />;
+      const className = option.className || 'recharts-line-dot';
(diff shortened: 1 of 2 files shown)


src/cartesian/Line.js
```diff
@@ -268,7 +268,8 @@ class Line extends Component {
     } else if (_.isFunction(option)) {
       dotItem = option(props);
     } else {
-      dotItem = <Dot {...props} className="recharts-line-dot" />;
+      const className = option.className || 'recharts-line-dot';
```
xile611: Don't use `||`, use  

```jsx

const className = classNames('recharts-line-dot', option.className);
```

kwolfy:
Done

PR #656, opened 2017-04-26
Change react-smooth version

egor20041:
Needs for remove warning about React.Props


jochenberger:
I think #657 is more complete.

egor20041:
I agree

PR #689, opened 2017-05-15
Allow viewing a single bar in a bar chart

JonathanUsername:
Currently it's not possible for me to view a single bar in my barchart, since the loop that decides the bandSize can't handle when the ticks array has only one element. BandSize is returned as 0, and that means the bar has 0 width.

My fix for this was to take the axis width if there is only 1 tick in the array. I'm not certain if this is the best way to fix this, but I didn't see any other issues about trying to view a single bar in a bar chart, so I assumed this was an edge case.


xile611:
@JonathanUsername Please resolve conflicts.

PR #701, opened 2017-05-22
Improve performance when checking tick text width

alyssaq:
While debugging the chart performance, I noticed that ~1sec is spent calculating the tick width before the charts are rendered. 
screen shot 2017-05-22 at 3 06 53 pm

Digger a little deeper, you'll see that ~700 ms is spent setting the CSS style and text content of the tick text.
screen shot 2017-05-22 at 3 23 12 pm

According to this [SO](http://stackoverflow.com/questions/17199958/why-does-setting-textcontent-cause-layout-thrashing), textcontent causes layout thrashing.
By creating a text node and updating its text and setting the custom css style only if a user provides one, this is reduced to ~4ms.
screen shot 2017-05-22 at 4 07 11 pm
screen shot 2017-05-22 at 5 02 28 pm

### DIFF
--- a/src/util/DOMUtils.js
+++ b/src/util/DOMUtils.js
@@ -61,14 +61,19 @@ export const getStringSize = (text, style = {}) => {
 
   try {
     let measurementSpan = document.getElementById(MEASUREMENT_SPAN_ID);
+    const textNode = document.createTextNode('');
     if (!measurementSpan) {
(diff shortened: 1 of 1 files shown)


alyssaq:
Ah damn this didnt work. Let me take out the styles checking into a separate PR

PR #712, opened 2017-05-29
Fix `iteratee is not a function` error

kcjpop:
I got an error `iteratee is not a function` when trying the sample snippet on Recharts homepage.

```jsx
      <LineChart width={500} height={300} data={data}>
        <XAxis dataKey="name" />
        <YAxis />
        <CartesianGrid stroke="#eee" strokeDasharray="5 5" />
        <Line type="monotone" dataKey="uv" stroke="#8884d8" />
        <Line type="monotone" dataKey="pv" stroke="#82ca9d" />
      </LineChart>
```

screenshot from 2017-05-29 13-57-47

I did check [Lodash's docs](https://lodash.com/docs/4.17.4#sortBy) but apparently passing a single function didn't work in this case, but an array of iteratees did.

### DIFF
--- a/src/chart/generateCategoricalChart.js
+++ b/src/chart/generateCategoricalChart.js
@@ -1067,7 +1067,7 @@ const generateCategoricalChart = ({
 
       return {
         tooltipTicks,
-        orderedTooltipTicks: _.sortBy(tooltipTicks, o => o.coordinate),
+        orderedTooltipTicks: _.sortBy(tooltipTicks, [o => o.coordinate]),
         tooltipAxis: axis,
         tooltipAxisBandSize: getBandSizeOfAxis(axis),
       };


alyssaq:
Came across this same issue and this PR fixes it. Please merge it.

xile611:
@alyssaq  What about the version of lodash. I try the latest lodash, it works when only path a single function.

image

kcjpop:
@xile611 Just tested with recharts@1.0.0.alpha-4 and the problem gone. It's weird because as I said in the first comment, I checked with Lodash's docs and apparently it's possible to pass a single function into `sortBy()`. Lodash's version at that time was 4.17.4 and it seems there is no newer version since then.

alyssaq:
Seems to be fine now

PR #719, opened 2017-06-02
Add number of ticks to tick props

Heedster:
This is helpful when trying to render each tick based on the total available width and the number of ticks visible.


src/cartesian/CartesianAxis.js
```diff
@@ -378,6 +378,7 @@ class CartesianAxis extends Component {
         ...customTickProps,
         ...tickCoord,
         index: i, payload: entry,
+        totalTicks: finalTicks.length,
```
xile611: @Heedster I prefer props like `visibleTicksCount`.

PR #724, opened 2017-06-06
Mixed components

txgruppi:
Updates on how data is selected and axis _points_ are calculated.

#### Changed the order in which `getDisplayedData` looks for the data.

The data used to calculated the Axis _points_ are returned in this order:

1. Data from the graphical items;
2. Data from the item properties;
3. data from the chart props.

This allows us to use a general data set for a complement like `Line` while using a specific data set for another component like Scatter.

The example below has the chart's data property defined as the general data used for the `Line` components while the `Scatter` has its own data property.

Without this code changes the `Scatter` would never render since it would use the data from the chart.

The `linesData` and `scatterData` are completely different and they don't use the same `points` on any axis. They are just displayed together.

```jsx
// linesData = [ { label: string, a: number, b: number }, ... ];
// scatterData = [ { x: number, y: number }, ... ];

<ComposedChart width={600} height={400} data={linesData}>
  <CartesianGrid />

  <XAxis xAxisId="a1" dataKey="x" type="number" />
  <YAxis yAxisId="a2" dataKey="y" type="number" />

  <XAxis xAxisId="a3" dataKey="label" type="category" orientation="top" />
  <YAxis yAxisId="a4" type="number" orientation="right" />

  <Scatter xAxisId="a1" yAxisId="a2" data={scatterData} fill="#ff0000" line />

  <Line xAxisId="a3" yAxisId="a4" dataKey="a" type="monotone" stroke="#00ff00" />
  <Line xAxisId="a3" yAxisId="a4" dataKey="b" type="monotone" stroke="#ff00ff" />
</ComposedChart>
```

#### Changed how an axis _points_ are calculated.

Updated the `getAxisMapByAxes` send to `getDisplayedData` `graphicalItems` which are connected with the given axis.

This prevents the axis from considering data it is not supposed to display.

The example code is the same used in the previous item.

Without this change:

screen shot 2017-06-06 at 16 13 14

With this change:

screen shot 2017-06-06 at 16 13 25

### DIFF
--- a/src/chart/generateCategoricalChart.js
+++ b/src/chart/generateCategoricalChart.js
@@ -263,7 +263,9 @@ const generateCategoricalChart = ({
       stackGroups, dataStartIndex, dataEndIndex }) {
       const { layout, children, stackOffset } = props;
       const displayedData = this.constructor.getDisplayedData(props, {
-        graphicalItems, dataStartIndex, dataEndIndex,
+        graphicalItems: graphicalItems.filter(item => item.props[axisIdKey] === axisId),
+        dataStartIndex,
+        dataEndIndex,
       });
       const len = displayedData.length;
       const isCategorial = isCategorialAxis(layout, axisType);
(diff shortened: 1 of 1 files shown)


morinted:
This is behavior that I thought the library would have by default.

I tried this PR and can confirm it solves my use case. I've reviewed the code it looks fine to me.

PR #747, opened 2017-06-23
fix faulty logic check in inRange function

dsifford:
This little bugger had me scratching my head today for much longer than I'd care to admit! 😄


dsifford:
Thanks for merging @xile611 

Any chance you can push this to npm?

Issue #860, opened 2017-09-01
Cartesian grid vertical fill color


saravanareddy:
I have a requirement to alternate the fill color in the grid. I am able to set the fill color, but is there a way to customize it so that I can alternate between two colors. Diagram attached. 

It would be great if the fill took an array as input and cycle through the input colors. Something like:

`
<CartesianGrid vertical horizontal={false} verticalFill={['#555555', '#444444']} fillOpacity={0.2} />
`
Thank you. 

![chart](https://user-images.githubusercontent.com/1454752/29993250-51208336-8fcf-11e7-9723-a392f307a98d.jpg)

saravanareddy:
Is there a way to specify style to the vertical grid lines? I want to make my lines white.

saravanareddy:
@xile611 thank you for adding this to features list. 

For the time being I could set the background to one color and use ReferenceArea to create a rectangular block of the second color with a white stroke to meet my requirements. 

My question is how do I get the starting co-ordinates of the ticks markers along the x-axis? If I know that, I think I will be good. Thank you.

costagolub:
#867 pushed request about this

saravanareddy:
@costagolub anyway for me to specify the grid line color? Would be nice to have a prop under CartesianGrid to specify color for the vertical and horizontal grid lines.

costagolub:
@saravanareddy so what exactly do you want to achieve. Do specify the colors for the grid lines in the props of CartesianGrid? I added this into the pull request where possible to specify `verticalFill ` and `horizontalFill` that accepts arrays with colors that will be altered as well as controlled opacity like `fillOpacityVertical`. Does that make sense?

saravanareddy:
@costagolub thank you for your commits. `verticalFill` and `horizontalFill` along with the opacity option give me the ability to shade the rectangular regions in whichever way I want. That's great!

But, if you look at the reference image in my original post. You will notice that the grid has white lines at the ticks, followed by the rectangular region(using verticalFill and fillOpacityVertical). Right now, by default the grid lines are drawn in `gray` color and I would like to customize that to any color of my choice. Hope my requirement makes sense. Thank you.

costagolub:
@saravanareddy I see now what you mean. Thank you for clarification. 
If you're using `XAxis` and `YAxis` for adding lines with ticks you can just disable them with prop `axisLine={false}` like `<XAxis axisLine={false} dataKey="name" stroke="#666" /> `

[...]

PR #867, opened 2017-09-04
Add horizontal / vertical grid lines filled with colors 

costagolub:
As a feature tagged there [https://github.com/recharts/recharts/issues/860](https://github.com/recharts/recharts/issues/860)


src/cartesian/CartesianGrid.js
```diff
@@ -174,7 +254,9 @@ class CartesianGrid extends Component {
 
     return (
       <g className="recharts-cartesian-grid">
-        {this.renderBackground()}
```
xile611: @costagolub We can use `renderBackground` to render a background rectangle. If you remove this function, we can only achieve this by rendering several rectangles which will bring some strange gaps.

src/cartesian/CartesianGrid.js
```diff
@@ -131,6 +136,81 @@ class CartesianGrid extends Component {
...
+  /**
+   * Draw vertical grid stripes filled by colors
+   * @param {Array} verticalPoints either passed in as props or generated from function
+   * @return {Group} Vertical stripes
+   */
+  renderVerticalStripes(verticalPoints) {
+    const { verticalFill } = this.props;
+    if (!verticalFill || !verticalFill.length) { return null; }
+
+    const { fillOpacityVertical, x, y, width, height } = this.props;
```
xile611: The props `fillOpacityVertical` seems to be very strange.

src/cartesian/CartesianGrid.js
```diff
@@ -131,6 +136,81 @@ class CartesianGrid extends Component {
...
+    let verticalPointsUpdated = verticalPoints;
+    if (x === 0) {
```
xile611: I can't understand this line. Why not unshift x when x !== verticalPoints[0]?

src/cartesian/CartesianGrid.js
```diff
@@ -131,6 +136,81 @@ class CartesianGrid extends Component {
...
+      verticalPointsUpdated.unshift(0);
+    }
+
+    const items = verticalPointsUpdated.map((entry, i) => {
+      let lineWidth = Math.max(0, (i === 0) ? verticalPointsUpdated[i+1]-x : verticalPointsUpdated[i+1] !== undefined ? verticalPointsUpdated[i+1] - entry : width - entry);
```
xile611: You need lint for this line.
And you should simplify this line.
```js
const lineWidth = !_.isNil(verticalPointsUpdated[i + 1] ? verticalPointsUpdated[i + 1] - entry :  (x + width - entry);

if (lineWidth <= 0) { return null; } 
```

src/cartesian/CartesianGrid.js
costagolub: What do mean by strange? You think is it better to use the default FillOpacity property both for horizontal and vertical lines?

src/cartesian/CartesianGrid.js
costagolub: Ok, I need to check your suggestion how it works. Thanks!

src/cartesian/CartesianGrid.js
costagolub: @xile611 yes, I actually returned it back in the amended commit.

src/cartesian/CartesianGrid.js
costagolub: ok, I will do, thanks.

src/cartesian/CartesianGrid.js
xile611: @costagolub How about use the default fillOpacity?

costagolub:
#895

norman0612:
will this go out with the next release?

src/cartesian/CartesianGrid.js
```diff
@@ -131,6 +136,81 @@ class CartesianGrid extends Component {
...
+   * @return {Group} Vertical stripes
+   */
+  renderVerticalStripes(verticalPoints) {
+    const { verticalFill } = this.props;
+    if (!verticalFill || !verticalFill.length) { return null; }
+
+    const { fillOpacity, x, y, width, height } = this.props;
+    const verticalPointsUpdated = verticalPoints;
+    if (x !== verticalPoints[0]) {
+      verticalPointsUpdated.unshift(0);
```
xile611: @costagolub You should unshift `x`.

src/cartesian/CartesianGrid.js
```diff
@@ -131,6 +136,81 @@ class CartesianGrid extends Component {
...
+   * @return {Group} Horizontal stripes
+   */
+  renderHorizontalStripes(horizontalPoints) {
+    const { horizontalFill } = this.props;
+    if (!horizontalFill || !horizontalFill.length) { return null; }
+
+    const { fillOpacity, x, y, width, height } = this.props;
+    const horizontalPointsUpdated = horizontalPoints;
+    if (y !== horizontalPoints[0]) {
+      horizontalPointsUpdated.unshift(0);
```
xile611: @costagolub It's same here. You should unshift `x`.

xile611:
@costagolub How about create another PR for Tooltip?
It's OK for the feature`horizontal / vertical grid lines filled with colors`. But I'm not sure about the `Tooltip` feature now.

xile611:
@costagolub We has a component `DefaultTooltipContent` now, may be we can add some  TooltipContent to support tooltip has `corner`.

costagolub:
hey @xile611, sure it's not a problem to shift the tooltip component to another PR. 
But do you propose to extend the `defaultTooltipContent `component with the corner?

src/cartesian/CartesianGrid.js
costagolub: @xile611 will fix it, thanks

xile611:
@costagolub By now, I prefer to create another `TooltipContent`.

PR #896, opened 2017-09-18
add props id to Pie

Tennen:
it seems when using SSR, the id of Pie would be different between client side and server side, 
so would it be better if I can set id by myself?


xile611:
@Tennen You should only change the files in src ....

Tennen:
@xile611 I'm sorry... can you fix this problem or should I do another pull requests?

xile611:
@Tennen Can you submit another PR ?

Issue #927, opened 2017-09-29
Specify arbitrary domain for the Cartesian X and Y axes


klandell:
## What is the current behavior?
Currently the domain accepts numbers, 'auto', or addition/subtraction from dataMin and dataMax.

## What is the expected behavior?
I would like to be able to pass functions into the domain config to calculate the domain based on the dataMin and dataMax values.

klandell:
Please see the linked pull requests for the implementation of this feature request.  Thanks!

PR #928, opened 2017-09-29
Allow an arbirary domain for cartesian X and Y axes

klandell:
This pull request is the implementation for feature request #927

### DIFF
--- a/src/util/ChartUtils.js
+++ b/src/util/ChartUtils.js
@@ -873,7 +873,7 @@ export const parseSpecifiedDomain = (specifiedDomain, dataDomain, allowDataOverf
     const value = +MIN_VALUE_REG.exec(specifiedDomain[0])[1];
 
     domain[0] = dataDomain[0] - value;
-  } else if (toString.call(specifiedDomain[0]) === '[object Function]') {
+  } else if (_.isFunction(specifiedDomain[0])) {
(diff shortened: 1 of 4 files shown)


src/util/ChartUtils.js
```diff
@@ -884,6 +886,8 @@ export const parseSpecifiedDomain = (specifiedDomain, dataDomain, allowDataOverf
     const value = +MAX_VALUE_REG.exec(specifiedDomain[1])[1];
 
     domain[1] = dataDomain[1] + value;
+  } else if (toString.call(specifiedDomain[1]) === '[object Function]') {
```
xile611: How about use _.isFunction to detect whether `specifiedDomain[1]` is a function?

src/util/ChartUtils.js
```diff
@@ -873,6 +873,8 @@ export const parseSpecifiedDomain = (specifiedDomain, dataDomain, allowDataOverf
     const value = +MIN_VALUE_REG.exec(specifiedDomain[0])[1];
 
     domain[0] = dataDomain[0] - value;
+  } else if (toString.call(specifiedDomain[0]) === '[object Function]') {
```
xile611: How about use _.isFunction to detect whether `specifiedDomain[0]` is a function?

klandell:
Good call, the PR has been updated.

Issue #847, opened 2017-08-23
Reference Area does not cover entire area specified


aarohmankad:
## Do you want to request a feature or report a bug?

Report a Bug

## What is the current behavior?

Even though I am specifying the boundaries of the Reference Area, there are parts of the graph left uncolored.

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/qbmg1567/).

http://jsfiddle.net/y7zb5h12/

![image](https://user-images.githubusercontent.com/3311321/29630781-2dfeb81c-87f2-11e7-8b87-f8f50b4e5078.png)

## What is the expected behavior?

When I specify the boundaries of the Reference Area as the full domain/range of the graph, I expect that entire area to be covered.

messerc:
I'm still unclear on what your issue is - everything seems to be working as intended within the fiddle. You're defining `x1, x2, y1, y2` explicitly within the `<ReferenceArea />` component and it's correct.

aarohmankad:
I wanted the spaces that aren't covered to be covered. Currently expanding any of  `x1, x2, y1, y2` makes the reference area disappear

remojansen:
@aarohmankad I'm not sure but maybe this is what you are looking for https://github.com/recharts/recharts/pull/931 ?

faeron:
I think I have the same issue as @aarohmankad had, so I. In my case, I want the reference area to represent a threshold range, to see if the data exceeds it at some point.

The reference area is only drawn if both y-values are in range of the y-Axis domain.
Now I would expect, that even if I set the y2 value outside of that domain, that the reference area will adapt and still render, but with the upper/lower y-value set to the min/max values of the y-Axis domain.

Here a sample I borrowed and updated from first post:
https://codesandbox.io/s/m72oxy7oqy

#931 is part of the solution, because now I can omit x values, as it should span the full x range.

If that is the desired behavior for ReferenceArea, I think the fix would be to check in https://github.com/recharts/recharts/blob/master/src/cartesian/ReferenceArea.js#L69-L91 first if the values are in their respective axis domain, before scaling them.

PR #931, opened 2017-10-03
Allow ReferenceArea to cover available space

remojansen:
I needed the `ReferenceArea` element to be able to cover all the available space given a point. For example, cover the entire X axis given one of the Y positions.

Before it was required to specify at least the two components of a pair:

- `x1` and `y2`
- `x1` and `y2`

Now it is possible to declare only one of the components of the area:

- `x1`
- `x2`
- `y1`
- `y2`

For example the following:

screenshot from 2017-10-03 17-19-46

Can be achieved using:

```ts
<ReferenceArea
    stroke="red"
    fill="red"
    y2={threshold}
    strokeOpacity={0.2}
    fillOpacity={0.1}
/>
```

```ts
<ReferenceArea
    stroke="red"
    fill="red"
    y1={threshold}
    strokeOpacity={0.2}
    fillOpacity={0.1}
/>
```

This feature is useful if you want to highlight when a value goes above or below a threshold.

I'm not sure but I think the issue https://github.com/recharts/recharts/issues/847 is related.

### DIFF
--- a/src/cartesian/ReferenceArea.js
+++ b/src/cartesian/ReferenceArea.js
@@ -55,7 +55,7 @@ class ReferenceArea extends Component {
     strokeWidth: 1,
   };
 
-  getRect(hasX, hasY) {
+  getRect(hasX1, hasX2, hasY1, hasY2, hasX, hasY) {
     const { x1: xValue1, x2: xValue2, y1: yValue1, y2: yValue2, xAxis,
       yAxis } = this.props;
     const xScale = xAxis.scale;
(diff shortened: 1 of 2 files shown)


remojansen:
Do you guys have an idea about when will this be merged and released? It is blocking a feature in one of my projects at work. Thanks a lot for the support in advanced.

src/cartesian/ReferenceArea.js
```diff
@@ -55,7 +55,7 @@ class ReferenceArea extends Component {
     strokeWidth: 1,
   };
 
-  getRect(hasX, hasY) {
+  getRect(hasX1, hasX2, hasY1, hasY2, hasX, hasY) {
```
xile611: @remojansen It seems you can remove `hasX` and `hasY`.

remojansen:
@xile611 thanks for the code review. I have removed the unused arguments.

remojansen:
Hi @xile611 can you please release `v1.0.0-alpha.7`. We are building an application with recharts and we need this feature for one of our releases. I would be very thankful if you could release it. We are planning to contribute more in the future.

remojansen:
Thanks a million for the release!

devison:
Yes, thanks indeed!

PR #937, opened 2017-10-05
Add className prop to Label Component

akram-rameez:
I wanted to add a className on Label, for easy static content modification via css classes rather than using css styles.

### DIFF
--- a/src/component/Label.js
+++ b/src/component/Label.js
@@ -1,6 +1,7 @@
 import React, { cloneElement, isValidElement } from 'react';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
+import classNames from 'classnames';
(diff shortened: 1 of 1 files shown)


akram-rameez:
Referencing issue #938

src/component/Label.js
```diff
@@ -297,7 +298,7 @@ function Label(props) {
 
   return (
     <Text
-      className={"recharts-label" + (className ? ' ' + className : '')}
+      className={`recharts-label ${className}`}
```
xile611: @akram-rameez Use the util function of `classnames`, please.
```jsx
<Text className={classNames('recharts-label', className);}>
```

akram-rameez: alright 👍

PR #951, opened 2017-10-12
 when maxBarSize is small, each bar should only be divided by realBar…

BOZ88:


src/util/ChartUtils.js
```diff
@@ -245,16 +245,22 @@ export const getBarPosition = ({ barGap, barCategoryGap, bandSize, sizeList = []
       originalSize >>= 0;
     }
     const size = (maxBarSize === +maxBarSize) ? Math.min(originalSize, maxBarSize) : originalSize;
+    // when maxBarSize is small, each bar should only be divided by realBarGap, not wide empty space
+    let sum = sizeList.reduce((res, entry) => (res + size), 0);
```
xile611: It seems no `size` in `sizeList`.

BOZ88: yes, I made a mistake. should change to 
    
const sum = len * size + (len - 1) * barGap;

BOZ88:
fix sum calculation.

src/util/ChartUtils.js
```diff
@@ -245,16 +245,21 @@ export const getBarPosition = ({ barGap, barCategoryGap, bandSize, sizeList = []
       originalSize >>= 0;
     }
     const size = (maxBarSize === +maxBarSize) ? Math.min(originalSize, maxBarSize) : originalSize;
+    // when maxBarSize is small, each bar should only be divided by realBarGap, not wide empty space
+    const sum = len * size + (len - 1) * barGap;
```
xile611: `bargap` can be a percent ...

PR #959, opened 2017-10-20
add id prop to Pie Component

Tennen:
when using SSR, the id of Pie component would be different between server side and client side,
add a prop to set a unique id could solve this problem.


src/polar/Pie.js
```diff
@@ -74,6 +74,7 @@ class Pie extends Component {
       'spring',
       'linear',
     ]),
+    id: PropTypes.string,
```
xile611: @Tennen You should add a default value of props `id`, and remove `id = uniqueId('recharts-pie-');`.

Tennen: @xile611 done~

Issue #955, opened 2017-10-18
RadialBar props not working


juliussohn:
Changing most of the RadialBar props has no effect, including: minAngle, maxAngle, startAngle, endAngle, clockWise.

I tested it in the  [example in JSFiddle](http://jsfiddle.net/4qo7ba5L/?utm_source=website&utm_medium=embed&utm_campaign=4qo7ba5L)

tongyy:
It appears that it removed the related props at this [diff ](https://github.com/recharts/recharts/commit/8dfd83dc5e2f6e32958cec16f3009ddc14600cc0#diff-232275603e425a80513a3b40847f82b8)

thomax:
Nice to know it's not my fault that I can't get RadialBar to behave :)

thomax:
I use `PieChart` as a temporary workaround.

xile611:
Because only one group of bar is supported in Recharts 0.x. We want to support multi groups of bar in 1.0+ just like <BarChart />, so we move the props `startAngle`, `endAngle` to RadialBarChart.

![image](https://user-images.githubusercontent.com/6111424/32935928-306bc4e2-cbad-11e7-9310-a00313d35bd3.png)

And If you have specified the props `starAngle`, `endAngle`, `clockWise` is no longer needed, so this two props are removed.

The props `minAngle` and `maxAngle` are used to specify  the domain of `PolarAngleAxis`, so they are removed too. Try to set the `domain` of `PolarAngleAxis` after I resolve some bug.

xile611:
Try to set the range of `PolarAngleAxis` in recharts 1.0.0-beta.3

PR #1014, opened 2017-11-15
:bug: #955 Fix minAngle and maxAngle for the polar RadialBar

huei-tan-sociomantic:
This pull request fixes only the prop `minAngle` `maxAngle` of `Polar/RadialBar`, 

I have two questions

1. Isn't it better if this prop being set in `RadialBarChart` since it affects all the cells of the radial bar inside the RadialBarchart? Unless there is a requirement to set two `RadialBar`s within one RadialBarChart.
1. should we use the prop name `minPointSize` `maxBarSize` instead of `minAngle` `maxAngle`, I think the latter is more proper with the given word `Angle`

_Example code and screenshot :_

```es6
const data = [
  { uv: 100, fill: '#8884d8' },
  { uv: 0, fill: '#83a6ed' },
  { uv: 0, fill: '#8dd1e1' },
  { uv: 0, fill: '#82ca9d' },
  { uv: 0, fill: '#a4de6c' },
  { uv: 0, fill: '#d0ed57' },
  { uv: 0, fill: '#ffc658' },
];

<RadialBarChart
    width={500}
    height={300}
    cx={150}
    cy={150}
    data={data}
    innerRadius={20}
    outerRadius={140}
    startAngle={0}
    endAngle={180}>

    <RadialBar minAngle={22.5} maxAngle={90} background dataKey="uv" >
    {
        data.map((entry, index) => (
            <Cell key={`cell-${index}`} fill={colors[index]}/>
        ));
    }
    </RadialBar>

</RadialBarChart>
```

screen shot 2017-11-15 at 12 49 31

### DIFF
--- a/demo/component/RadialBarChart.js
+++ b/demo/component/RadialBarChart.js
@@ -61,7 +61,7 @@ export default class Demo extends Component {
             barSize={10}
             data={data}
           >
-            <RadialBar minPointSize={15} background dataKey="uv" >
+            <RadialBar minAngle={15} background dataKey="uv" >
               {
                 data.map((entry, index) => (
                   <Cell key={`cell-${index}`} fill={colors[index]}/>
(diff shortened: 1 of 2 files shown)


xile611:
@huei-tan-sociomantic  The props `minPointSize` is used to set a visible angle when the value is very small and no bar is visible.  
`minAngle` and `maxAngle` were used to set the `domain` of `angleAxis`. I plan to remove  `minAngle` and `maxAngle`, and developers can set these props by the `domain` of `PolarAngleAxis` which is  same to `XAxis`.

huei-tan-sociomantic:
Thanks, looking forward to your bug resolved 😃

Issue #935, opened 2017-10-04
Error: React.Children.only expected to receive a single React element child.


jgentes:
## Do you want to request a feature or report a bug?
bug

## What is the current behavior?
When using an iterator (scattercharts.map(s => <Scatter.../>) to generate Scatter components from an array, an error occurs during render.

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/qbmg1567/).
I would! The current fiddles are broken due to an npmcdn.com issue with React 16 :(

## What is the expected behavior?
The parent <ScatterChart/> container should allow multiple children.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Alpha 5 of recharts, latest Chrome. Not sure if this is new or a regression.

bcornett:
I'm doing the same thing with a `LineChart` and get this error.
```js
{Array.from({length: lineCount}).map((x, i) => (
  <Line key={i} dataKey={`line${i}value`} />
))}
```

sorodrigo:
Same here :(

eldyvoon:
This happens in RadarChart too, other charts is fine.

ryanand26:
I got the same issue with a composed chart. I worked around it by storing the created array of lines/areas, not recreating it on each render.

sorodrigo:
Another fix is spreading the resulting array.
```jsx
...[/* full with magic */].map(line => (<Line stuff={stuff} />))
```

jgentes:
@sorodrigo I wasn't able to use the spread operator to fix the issue. And I'm not sure that works:

`...Array.map` - wouldn't it need to be `[...Array].map`?

rafaelc0sta:
Also experiencing this :(

ryanand26:
@rafaelc0sta @jgentes Can you create a demo on jsfiddle? The code I had with the issue was really convoluted so not suited to a demo.

xile611:
Any idea to reproduce this case ? I can't reproduce this case in my local project.

jgentes:
I've been trying, and as a result of trying I somehow got it working, although I don't know what I've done differently. But I'm happy now, so I'll go ahead and close this. Thanks for the help @ryanand26 and @xile611

rafaelc0sta:
[...]

PR #1031, opened 2017-11-21
Possible fix: "Error: React.Children.only expected to receive a single React element child."

ryanand26:
Relates to the issue logged at https://github.com/recharts/recharts/issues/935
* Added test case for updating a line chart from a container component, this catches the exception
* Removed the use of `Children.only` from the isSingleChildEqual call. Appears to resolve the issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
+## 1.0.0-beta.1(Nov 06, 2017)
+
+### feat
+
+- Add index to line props in Pie
+- Update ReferenceDot.js
+
+### chore
+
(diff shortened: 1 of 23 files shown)


xile611:
Thanks~ @ryanand26 
But I still can't reproduce this problem ...

ryanand26:
Cheers for merging. It was happening pretty consistently for me, Windows 7 & 10 machines, Chrome & Firefox.
Any plan on when v1 might make it out of beta? I'm using recharts through TypeScript and when it drops I'll probably go in and correct some of the definitions.

PR #1033, opened 2017-11-22
Add Animation Decorators

huei-tan-sociomantic:
There are too many Animations repeatedly implementation across many different components, since we have the PureRender Decorator, using another `AnimationDecorator` is the best solution to avoid repeating.

However, this pull request update doesn't include `chart/Treemap` `shape/Rectangle`, those are fully customized for the functional work and don't follow the same rule with others.

for a better reviewing, please use this link with the append param `w=1`
https://github.com/recharts/recharts/pull/1033/files?w=1


huei-tan-sociomantic:
Thanks for the CI test integration, it helps me identify the potential bugs! 🐛

xile611:
@huei-tan-sociomantic Thanks for your contribution. But I think the AnimationDecorator seems to be very strange now which has some inelegant codes like the following. 
We need to find a best point between code duplication and readability. So I think such kind of `AnimationDecorator` is not so elegant now.
```jsx
+function componentWillReceiveProps(nextProps) {
 +  const {
 +    isAnimationActive, animationId, points, data, baseLine, sectors,
 +  } = this.props;
 +
 +  if (nextProps.isAnimationActive !== isAnimationActive) {
 +    if (sectors) this.setState({ prevSectors: [] }); // Pie
 +  } else if (nextProps.animationId !== animationId) {
 +    if (points) this.setState({ prevPoints: points }); // Line Area Scatter Radar
 +    if (data) this.setState({ prevData: data }); // Bar RadialBar
 +    if (baseLine) this.setState({ prevBaseLine: baseLine }); // Area
 +    if (sectors) this.setState({ prevSectors: sectors }); // Pie
 +  }
 +}
```

PR #1050, opened 2017-11-27
Fix sankey diagram not re-rendering when updating data prop

pewniak747:
When `Sankey` component `data` prop changed, the chart did not update
right away, instead it was deferred until next props update.

This was because `componentWillReceiveProps` function used `this.props` instead of `nextProps` to calculate the next state.

Demo:
http://jsfiddle.net/s2tz0ey7/1/

Affected version:
1.0.0-beta.5, possibly earlier versions

PS. Thank you for maintaining recharts, I've been using it for months and find it very useful 😄  ❤️

### DIFF
--- a/src/chart/Sankey.js
+++ b/src/chart/Sankey.js
@@ -375,7 +375,7 @@ class Sankey extends Component {
       nextProps.height !== height || !shallowEqual(nextProps.margin, margin) ||
       nextProps.iterations !== iterations || nextProps.nodeWidth !== nodeWidth ||
       nextProps.nodePadding !== nodePadding || nextProps.nameKey !== nameKey) {
-      this.setState(this.createDefaultState(this.props));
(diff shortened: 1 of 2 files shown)


pewniak747:
@xile611 ping :)

BTW, what's the status of `Sankey` component? I'm not seeing it in the docs, perhaps it's not so stable yet?
I'm asking because I'd be interested in making more improvements to `Sankey`.

xile611:
@pewniak747 The `Sankey` is not stable now. So we have not added it into the docs. Maybe we'll change the components of `Sankey` completely.

pewniak747:
@xile611 thanks :)

I might have another bugfix to `Sankey` component (in certain cases, when the nodes are wide and not far apart, the links aren't drawn accurately).
Would you be interested in such bugfix or is it better to wait until the `Sankey` refactor?

PR #1059, opened 2017-12-01
Create property on tooltip for gpu rendering

richbai90:
#1060

replaces translate with translate3d on the tooltip wrapper for smoother animation at the cost of some memory. 

```js
    <BarChart>
          <Tooltip gpuRendering />
          ...
      </BarChart>
```

### DIFF
--- a/src/component/Tooltip.js
+++ b/src/component/Tooltip.js
@@ -58,7 +58,7 @@ const propTypes = {
   ]),
   itemSorter: PropTypes.func,
   filterNull: PropTypes.bool,
-  gpuRender: PropTypes.bool,
+  useTranslate3d: PropTypes.bool,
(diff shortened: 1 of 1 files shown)


xile611:
@richbai90 Thanks! I have a little doubt about the props name `gpuRender` which  is `gpuRendering` in your demo. Maybe `isTranslate3d` or `useTranslate3d` is better.

khakulov:
why not always use `translate3d` istead of `translate`?

richbai90:
I think `translate3d` or `useTranslate3d` is perhaps more clear about what is actually changing behind the scenes sure.

@khakulov Some may not wish to use `translate3d` if memory is a consideration. Just a thought.

Issue #1063, opened 2017-12-03
reduce-css-calc really useful?


hlehmann:
reduce-css-calc is mostly use on build process.

It's only use there:
https://github.com/recharts/recharts/blob/master/src/component/Text.js#L141 (3 times)

Does it really worth to load reduce-css-calc and its dependencies in order to precalculate something the browser easily calculate ?

xile611:
@hlehmann PR is welcome, `reduce-css-calc` is not necessary.

PR #1085, opened 2017-12-13
 Remove reduce-css-calc dependency

hlehmann:
Remove unnecessary reduce-css-calc dependency

https://github.com/recharts/recharts/issues/1063

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,8 +58,7 @@
     "prop-types": "^15.6.0",
     "react-resize-detector": "1.1.0",
     "react-smooth": "1.0.0",
-    "recharts-scale": "0.3.2",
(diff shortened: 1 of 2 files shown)


hlehmann:
Well in fact it appear that `dy` attribute of `tspan` doesn't allow `calc`.

`reduce-css-calc` avoided this by removing the `calc` but only if `lineHeight` and `capHeight` have the same unit, and let the `calc` otherwise.

At this end we should force `lineHeight` and `capHeight` to have the same unit and calculate `startDy` without using `reduce-css-calc`

xile611:
@hlehmann OK! So we should maintain `reduce-css-calc` temporarily.

Issue #1039, opened 2017-11-23
BarChart: Tooltip is broken since beta.3


jochenberger:
See:
http://jsfiddle.net/ad071owf/
Same Fiddle with beta.2:
http://jsfiddle.net/ad071owf/1/

If you hover over one of the bars, there is a gray rectangle shown at the left side.
I tried a `git bisect` to find the commit that introduced the issue, but I cannot reproduce it in the demo.

xile611:
@jochenberger I can't reproduce it in my local project too. jsfiddle seems very strange now. I can't select texts correctly too.

jochenberger:
Seems to work again with beta.4: http://jsfiddle.net/ad071owf/2/

jochenberger:
No, I must have missed something. It's still wrong in beta.4.

jochenberger:
Beta 5 too: http://jsfiddle.net/ad071owf/3/

mkreim:
Hi,

I am observing the same problem for the bar charts. A good example can be seen here:
expected behavior: http://recharts.org/#/en-US/examples/SimpleBarChart
buggy behavior: http://jsfiddle.net/v8gxrbyu/
For me it looks like the coordinates of the tooltip are always 0,0. So the tooltip is not moving around and the cursor shows a gray box at the left side of the chart.

Also this issue seems to be a duplicate of https://github.com/recharts/recharts/issues/1079

Thanks an best regards

Michael

Golodhros:
This seems to be the case also with the line chart and its tooltip:
![fullscreen_12_30_17__10_59_am](https://user-images.githubusercontent.com/190833/34456655-cf516a20-ed50-11e7-8529-23ff2b088a7d.png)

It's funny how the version of the docs works fine.

mpodlasin:
I use beta.10 version and I am experiencing this exact issue as well.

anthony2025:
I am using 1.1.0 and still having this issue. It is only affecting BarCharts on my project though,  LineCharts work beautifully.

JohnStarich:
I ran into this on the latest version as well (1.6.2). I've opened a PR that fixes my issue: https://github.com/recharts/recharts/pull/1811
I've reversed the order of the rendered elements, placing the tooltip toward the back. Not the greatest solution, but it seemed to do the trick for me.

PR #1095, opened 2017-12-15
replace lodash _.get with simple Array.prototype.find

khakulov:
In out Project we used Webpack Plugin `LodashModuleReplacementPlugin` which strippes out path feature of lodash. `_.get` uses path feature. Which caused strange bugs like #1039

This is much simple form for `_.get(arr, '[0]')` which uses ES5 `Array.prototype.find` feature


dmr:
Is there something I can do to get this merged into master?

`find` should be supported by all browsers IMO (https://caniuse.com/#feat=es5)

dmr:
Thank you :)

Issue #791, opened 2017-07-19
Tooltip not appearing on pie charts when cells re-render


alexandersoto:
I'm trying highlight the selected sector of a pie chart, and the following code works. The issue is that the tooltip won't appear. If I remove the `onMouseEnter` and `onMouseLeave` events, then the tooltips show up fine. But of course, I can't render the selected sector differently in that case. I think there's an issue with tooltips not appearing when the Cells re-render. 

```
<ResponsiveContainer
  width="100%"
  height={this.props.height}
>
  <PieChart
    width={diameter}
    height={diameter}
  >
    <Tooltip/>
    <Pie
      data={this.props.data}
      dataKey="value"
      onMouseEnter={(e, activeIndex) => this.setState({ activeIndex }) }
      onMouseLeave={() => this.setState({ activeIndex: -1 }) }
    >
      {
        _.map(this.props.data, (entry, i) => (
          <Cell
            key={`cell-${i}`}
            fill={COLORS.CONTRASTING_COLORS[i]}
            fillOpacity={this.state.activeIndex === i ? 1 : 0.25}
          />
        ))
      }
    </Pie>
  </PieChart>
</ResponsiveContainer>
```

I tried to fix this issue by "actively" controlling the tooltip, but that didn't work either: https://github.com/recharts/recharts/issues/790

zinroc:
Any progress with this? I just ran into the same issue

zinroc:
Figured it out, this works for me: 

<Tooltip content={<CustomTooltip />} wrapperStyle={{visibility: this.props.activePieSlice ? 'visible' : 'hidden'}} />

trabek:
anything new?

bluetidepro:
I have this same bug in `1.0.0-beta.10`, did you ever get a way to fix this, @alexandersoto?

alexandersoto:
Unfortunately I never did.

hugobarragon:
> Hi, I faced this problem.
> After using `<CustomTooltip />`, it works and can show tooltip with re-render pie-chart.
> 
> ```
>         <Tooltip
>             active={true}
>             wrapperStyle={{
>               visibility: 'visible',
>             }}
>             content={<CustomTooltip />}
>           />
> ```
> but this not worked
> 
> ```
>         <Tooltip
>             active={true}
>             content={<CustomTooltip />}
>           />
> ```

Could you provide the custom tool tip you are using?

xile611:
Duplicated with #791

[...]

PR #1113, opened 2017-12-26
bugfix/791 Fix combineEventHandlers

yaroslav-perec:
Relates to the issue logged at #791

After call method `childHandler`, method `componentWillReceiveProps` inside `generateCategoricalChart` is called and it call `setState()`, defaultHandler also call `setState()`, 
`setState()` - works asynchronous and changes overwritten, so I think we need to call them synchronously.


xile611:
@yaroslav-perec Thanks! But this can't fix the problem! I'll fix the bug of `componentWillReceiveProps` in `generateCategoricalChart`  which shouldn't update when all the props are not changed.

PR #1117, opened 2017-12-27
fix: ensure Line has data before calling SVGPathElement.getTotalLength() for IE

dadamssg:
In the Line component if all point values are null or undefined, the SVGPathElement. getTotalLength() function is causing the following error in IE:

> Unexpected call to method or property access

[Fiddle showing the error(Open in IE).](http://jsfiddle.net/j64rfyju/)

### DIFF
--- a/src/cartesian/Line.js
+++ b/src/cartesian/Line.js
@@ -154,6 +154,13 @@ class Line extends Component {
   }
 
   getTotalLength() {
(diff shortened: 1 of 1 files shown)


xile611:
@dadamssg Thanks! It's not quite safe to check the points. I'll add `try...catch...` in the code to fix the bug in IE.

Issue #735, opened 2017-06-17
CSP problems on string length calculation


maxvektor:
I want to report a bug:
`setAttribute('style'...)` breaks a CSP when using `style-src: self` policy .
https://github.com/recharts/recharts/blob/7cb33048cc795810b4cba2836def37d86074af36/src/util/DOMUtils.js#L70
This place causes multiple errors in console when script is used with CSP policy that forbids inline styling.
Unfortunately there is no way to show it on jsfiddle. As I understand they don't ban inline styling and neither does your demo page.
I think there is a simple way to fix it:
https://stackoverflow.com/questions/24713440/banned-inline-style-csp-and-dynamic-positioning-of-html-elements/29089970#29089970
The only problem I see is to deal with caching.

jordanch:
Any movement on this? We need to adhere to the same CSP policy and have adopted recharts.

whphhg:
I'm running into the same `inline-src` issue while using CSP.

adrianriddle:
Any movement on this issue?

PR #1145, opened 2018-01-18
 Update to allow CSP compliance on setting styles

adrianriddle:
Need to change the way that the style is being set so that CSP compliance works.
The problem can be found in this issue:
#735


adrianriddle:
@xile611 could you please have a look at this pull request, would be remove a huge blocker on a project if I can get this change into the recharts library?

Issue #1146, opened 2018-01-19
Recharts is slow with large data


ofk:
## Do you want to request a feature or report a bug?

a feature

## What is the current behavior?

I use Recharts in my application. My application often uses large data (> 3 MB). But the drawing is very slow. For example, [ofk/recharts-large-data-test](https://github.com/ofk/recharts-large-data-test) drawing takes 4 seconds in my environment.

## What is the expected behavior?

Rechart should fast draw large data.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

I tried 1.0.0-beta.9 in Mac Chrome.

christiancho:
It could be that you're doing too much data transformation right before rendering. If you can offload some of this data transformation to a microservice, I'd suggest that. Also, are you using react v 16? React 16 uses the browser's `requestIdleCallback`. As a result, computationally heavy tasks will keep getting delayed until there is enough bandwidth for the browser to take those on.

ofk:
Thank you very much.
However, I'm very happy that recharts will be speed up. [This example](https://github.com/ofk/recharts-large-data-test) I published to github is the small one. In the future, I will draw larger data. So recharts is also required speedup.
Specifically, #1147 is the only problem.

And I'm using react v16.

xile611:
Duplicated with #1356

PR #1147, opened 2018-01-19
Replace reduce to _.flatmap

ofk:
I found the bottleneck.
My large data drawing is 6x faster (4.2s → 0.7s) with this change.

Other reduce's flatmap was not rewritten because I unconfirmed.

#1146


khakulov:
this pull request contains bug. _.isArray(value) is not checked. It broke my chart

Issue #324, opened 2016-11-05
[Legend] 'unknown props' warning when passing <div> as content property


petermikitsh:
There is a bit of a gotcha with the `Legend` component. If you pass a React component to the Legend, you'll end up with `unknown props` console warnings.

```jsx
<Legend
  verticalAlign="top"
  height={75}
  content={
    <div>
      <h4 style={{
        fontWeight: 400,
        color: '#000'
      }}>
        {this.props.title}
      </h4>
    </div>
  }
/>
```

In the right panel of the React debugger, you can see the unexpected props being passed to the component:

<img width="970" alt="screen shot 2016-11-05 at 4 53 05 pm" src="https://cloud.githubusercontent.com/assets/1571918/20034358/71fb1894-a378-11e6-8e86-627566ee5d97.png">

And here are the console errors:

<img width="1103" alt="screen shot 2016-11-05 at 4 55 34 pm" src="https://cloud.githubusercontent.com/assets/1571918/20034367/b783aec6-a378-11e6-993b-0170428cd031.png">

A workaround is to pass a function that returns a react component to the Legend.

```jsx
<Legend
  verticalAlign="top"
  height={75}
  content={() => {
    return (<div>
      <h4 style={{
        fontWeight: 400,
        color: '#000'
      }}>
        {this.props.title}
      </h4>
    </div>);
  }}
/>
```

PR #1173, opened 2018-02-03
Fix unknown props warning on Legend

riwu:
Fixes https://github.com/recharts/recharts/issues/324
When the `content` prop is a React element, eg. ```<Legend content={<SomeComponent />} />```, users should not be expecting the props to be passed to `SomeComponent` as that intention is not clear from the declaration.

To pass in the props, it should be declared as 
```<Legend content={props => <SomeComponent {...props} />} />``` 
or simply
```<Legend content={SomeComponent} />```.


xile611:
@riwu We use `cloneElement` to pass some props calculation in `Legend`. So If you use `content` props of `Legend`, you should allow the passed props.

Issue #1195, opened 2018-02-12
`utcTime` is actually called `utc`


Philipp91:
If I understand correctly, the correct value for `<XAxis scale="..."/>` is `utc` and not `utcTime`, as listed in the [documentation](http://recharts.org/#/en-US/api/XAxis) and in the [source code](https://github.com/recharts/recharts/blob/b48b6d6c912e354a918d2b6307f5c8c6ada210df/src/util/ReactUtils.js#L136). I tested both values inside a `<LineChart/>` and only `utc` works, with `utcTime` I don't get any axis ticks or labels at all. The correct value is documented at d3 here. Apparently, when they [switched](https://github.com/d3/d3-scale/commit/2cc073c99a682f290bb7f689c21baeefa625177b) to naming things `scaleXyz`, they changed `utcTime` to `scaleUtc`.

Interestingly, the [PropTypes validation](https://github.com/recharts/recharts/blob/b48b6d6c912e354a918d2b6307f5c8c6ada210df/src/cartesian/XAxis.js#L52) does _not_ fail for `utc`, though, so maybe I'm overlooking something.

Philipp91:
Sorry, the PropTypes validation actually _does_ fail. But when I apply the fix (see my commit 1b523c1), it works fine.

edaleeta:
Thanks for calling this out! I was wondering why `utcTime` didn't seem to work as expected.
Switching my scale to `utc` was the fix!
I do now encounter the same PropTypes validation error, but thanks for sharing your fix. :)

hlehmann:
any updates on this ?

PR #1196, opened 2018-02-12
#1195 Replace axis scale value `utcTime` with `utc`

Philipp91:
The d3-scale dependency renamed this export here: https://github.com/d3/d3-scale/commit/2cc073c99a682f290bb7f689c21baeefa625177b


Philipp91:
Disclaimer: I did not run `npm run test` or `lint`.

Issue #895, opened 2017-09-17
Adding arrow to tooltip ?


abhayjeetgupta:
## Do you want to request a feature or report a bug?
Feature Request
## What is the current behavior?
I can create a custom tooltip. 

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/qbmg1567/).
No
## What is the expected behavior?
I want to add triangle to tooltip so it should point to the related bar or pie.

costagolub:
@abhayjeetgupta I will prepare pull request about this feature if it's okay.

abhayjeetgupta:
@costagolub yes please.

nathmack:
I have a similar request https://github.com/recharts/recharts/issues/911

jabsatz:
@costagolub where you able to get some work done on this?

xile611:
Sorry, not supported yet. It's recommended to use customized `content` of `Tooltip` to cover default tooltip.

PR #1261, opened 2018-04-18
Added tooltip arrow

jabsatz:
Added a tooltip arrow to show where exactly the data is coming from
2rkqxwyac0

Issue for this: https://github.com/recharts/recharts/issues/895
This is my first PR here so feedback is absolutely welcome!

### DIFF
--- a/src/component/Tooltip.js
+++ b/src/component/Tooltip.js
@@ -41,13 +41,11 @@ const propTypes = {
   }),
 
   label: PropTypes.any,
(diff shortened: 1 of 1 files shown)


xile611:
@jabsatz Thanks for you suggestion! I'll add a new className to the wrapper element of <Tooltip /> which can  allow you to know the relatively position of Tooltip. And It's recommended to use css to render the arrow or other icon.

Issue #929, opened 2017-10-02
Pie Chart - No Label Shown


quangas:
Trying to get a label shown on a simple pie chart. No lines or anything are shown.

```
var COLORS = ['#3DCC91', '#FFB366', '#FF7373', '#FFCC00', '#3B22FF'];

        <PieChart width={300} height={200}>
          <Pie data={data} cx="50%" cy="50%" dataKey="value" label={true} labelLine={true}>
            {
              data.map((entry, index) => (
                <Cell key={index} fill={COLORS[index % COLORS.length]}  />
              ))
            }
          </Pie>
          <Tooltip />
          <Legend />
        </PieChart>
```

Why do none of the examples work on jsFiddle or on the main site? It just renders blank page.

UPDATE
So, I dropped back to version 0.22.4 and it shows the label. The latest version doesn't show it.

Danielh21:
I have the same issue, but I would like not to roll back!

davidsonsns:
I had the same problem. I came back to the version as the @quangas said and it worked. in the end i went to another lib

stevelacy:
@davidsonsns which lib did you switch to?

fgiarritiello:
Having the same problem. Switching back to 0.22.4 works but then I get other problems.

davidsonsns:
I'm using this @stevelacy [react-chartjs-2](https://github.com/jerairrest/react-chartjs-2)

pebanfield:
In my case setting isAnimationActive={false} resolves the issue. Of course this means doing without the animation.

The line of code below seems to be the problem - 

https://github.com/recharts/recharts/blob/master/src/polar/Pie.js#L346

The intent is to not render labels until animation completes which makes sense. However, there appears to be a race condition that means labels are never rendered.

xile611:
http://jsfiddle.net/8wLgvst6/

The labels are visible when isAnimationActive is false or true.

pebanfield:
Hmm - well I am definitely reproducing the issue but I have more than one Pie shape in my PieChart. 

This issue also appears to be present in the examples. I am away from my computer but I will try produce a fiddle to reproduce the issue when I get back. Thanks!

pebanfield:
@xile611 - I added an additional Pie shape to your fiddle and it works fine. So that's not the problem. I also tried changing a few other properties to try to isolate the issue I am seeing in my build and did not reproduce the issue in your fiddle. It seems to be pretty stable.

[...]

PR #1275, opened 2018-04-24
Issue #929

nickppa:
Fix the label will not shown.


nickppa:
To fix the issue #929

xile611:
We can't setState in a render function.

PR #1308, opened 2018-05-18
Add explicit prop to activate tooltip

vincentljn:
Addresses an issue when we have to re-render the chart as keeping the tooltip displayed.
We allow the developer to explicitly set the tooltip activation status with a `isTooltipActive` prop. It will be used to set the default value of the homonym state property


xile611:
@vincentljn Thanks for your advice. The new props means to show the tooltip by default or not. So I think it's better to rename the  props to `defaultShowTooltip`!

Issue #798, opened 2017-07-24
Tooltip wrapperStyle applies to wrapper AND it's recharts-default-tooltip child


sospirited:
## Do you want to request a feature or report a bug?
report a bug

## What is the current behavior?
If you add the following to your tooltip component:
```
<Tooltip wrapperStyle={{ background: red, top: '200px' }} />
```

Then I would expect only the `recharts-tooltip-wrapper` to get the above styles.
However, if you inspect the demo whilst hovering (using chrome-dev-tools shortcut), you would see that not only does the above wrapper element get these inline styles, but so does it's child element with class `recharts-default-tooltip`.

## If the current behavior is a bug, please provide the s`teps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/qbmg1567/).
See above, it's quite easy to reproduce, you can even inspect the above classes with selectors `$$(".recharts-tooltip-wrapper")` and `$$(".recharts-default-tooltip")` respectively, even without hovering.

## What is the expected behavior?
The `wrapperStyle` property should only apply to the wrapper element.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
1.0.0-alpha.2

PR #1322, opened 2018-05-30
Remove wrapperStyle on <DefaultTooltipContent>

cjies:
Does not apply `wrapperStyle` on `<DefaultTooltipContent>` component, since this prop already applied on [tooltip wrapper](https://github.com/recharts/recharts/blob/master/src/component/Tooltip.js#L144).

**Affected version**: 1.0.0-beta.10
**Related issue**: #798 
**Issue playground**: https://jsfiddle.net/n0nhzk1s/
**Screenshot**:
screen shot 2018-05-30 at 10 07 01 am

### DIFF
--- a/src/component/DefaultTooltipContent.js
+++ b/src/component/DefaultTooltipContent.js
@@ -21,7 +21,6 @@ class DefaultTooltipContent extends Component {
   static propTypes = {
     separator: PropTypes.string,
     formatter: PropTypes.func,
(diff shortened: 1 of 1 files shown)


kenips:
Hey guys this is removing the ability to style the wrapper and it's breaking for us. If we don't want to have any styles associated there as it's styled by the tooltip wrapper, we should remove all styling there. Currently it's forcing its own background.

PR #1374, opened 2018-07-06
Allow reference areas which extend past the canvas bounds

paulmelnikow:
I’m working on a chart with several reference areas, some of which extend past the canvas bounds. Though I want to draw the part of the reference area that lies on the canvas, it doesn’t make sense to draw the part that spills out.

It's possible to grow the canvas to accommodate the reference area by specifying `alwaysShow`. Though after thinking about this for a while, it seems that _both_ these things should be possible: to show reference items which are partly off the canvas, and to extend the chart domain to accommodate reference items. This deprecates the `alwaysShow` prop, and adds an `ifOverflow` prop which supports four values:

- `ifOverflow="discard"` (default) – If the reference item falls partly outside the canvas, don't display it. This is the old default.
- `ifOverflow="extendDomain"` – If the reference item falls partly outside the canvas, extend the domain so it fits. This is the old `alwaysShow` behavior.
- `ifOverflow="clip"` – If the reference item falls partly outside the canvas, clip it to the canvas.
- `ifOverflow="keep"` – If the reference item falls partly outside the canvas, draw it anyway.

### DIFF
--- a/src/cartesian/ReferenceArea.js
+++ b/src/cartesian/ReferenceArea.js
@@ -8,10 +8,13 @@ import classNames from 'classnames';
 import pureRender from '../util/PureRender';
 import Layer from '../container/Layer';
 import Label from '../component/Label';
-import { PRESENTATION_ATTRIBUTES } from '../util/ReactUtils';
-import { isNumOrStr } from '../util/DataUtils';
 import { LabeledScaleHelper, rectWithPoints } from '../util/CartesianUtils';
+import { ifOverflowMatches } from '../util/ChartUtils';
+import { isNumOrStr } from '../util/DataUtils';
+import { warn } from '../util/LogUtils';
+import { PRESENTATION_ATTRIBUTES } from '../util/ReactUtils';
 import Rectangle from '../shape/Rectangle';
+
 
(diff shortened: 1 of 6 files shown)


xile611:
@paulmelnikow 

Do you think the props name can be consistent with the attribute `overflow`? The the values can also be part of consistent:

* `overflow="discard"` (default) – If the reference item falls partly outside the canvas, don't display it. This is the old default.
* `overflow="extendDomain"` – If the reference item falls partly outside the canvas, extend the domain so it fits. This is the old alwaysShow behavior.
* `overflow="hidden"` – If the reference item falls partly outside the canvas, clip it to the canvas.
* `overflow="visible"` – If the reference item falls partly outside the canvas, draw it anyway.

xile611:
@paulmelnikow OK! It's best to remain the `overflow` property in `PRESENTATION_ATTRIBUTES`.

xile611:
@paulmelnikow Any idea to update the value of `ifOverflow`:  `keep` to `visible` and `clip` to `hidden`?

paulmelnikow:
Updated!

PR #1376, opened 2018-07-07
Lint cleanup

paulmelnikow:
- Turn off no-multi-assign
- Clean all other lint
   - Except one rule which needs separate attention: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md
- Enforce all rules as errors
    - When rules are set as warnings, there's too much output to sift through, and it's too easy for legitimate programmer errors to be missed.
- Run lint in CI
- Run lint and test on pre-push (not pre-commit)
- Modernize eslint config

Let me know if you have any questions about this!

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,35 +7,35 @@
   },
   "parser": "babel-eslint",
   "rules": {
-    "valid-jsdoc": 2,
-    
-    "react/sort-comp": 0,
-    "react/jsx-uses-react": 2,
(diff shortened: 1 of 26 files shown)


paulmelnikow:
I really need to get these merge into my copy because it will make me much more productive while working on the library. However once I do that it will be hard to contribute my changes due to conflicts. It would be great to get this merged soon. @xile611 any thoughts?

PR #1380, opened 2018-07-09
Reference line segment by specifying a pair of endpoints

paulmelnikow:
This is useful for drawing a regression line limited to the range of the data and at an arbitrary slope/intercept.

Depends on #1374. That commit is included here too.

### DIFF
--- a/test/specs/cartesian/ReferenceLineSpec.js
+++ b/test/specs/cartesian/ReferenceLineSpec.js
@@ -45,6 +45,19 @@ describe('<ReferenceLine />', () => {
     expect(wrapper.find('.recharts-reference-line-line').length).to.equal(2);
     expect(wrapper.find('.recharts-label').length).to.equal(2);
   });
+  it('Renders line segment in ReferenceLine when segment is set', () => {
+    const wrapper = render(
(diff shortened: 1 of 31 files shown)


pengcheng95:
Hello, so I am trying to draw reference lines on specific x axis locations using the segment prop, however it is unable to draw anything on the last x axis location. Is there anyway to customize the ReferenceLine props to display a reference line on the last x axis location?

test/specs/cartesian/ReferenceLineSpec.js
```diff
@@ -45,6 +45,19 @@ describe('<ReferenceLine />', () => {
...
     expect(wrapper.find('.recharts-label').length).to.equal(2);
   });
+  it('Renders line segment in ReferenceLine when segment is set', () => {
+    const wrapper = render(
+      <BarChart width={1100} height={250} barGap={2} barSize={6} data={data} margin={{ top: 20, right: 60, bottom: 0, left: 20 }}>
+        <XAxis dataKey="name" orientation="top" />
+        <YAxis tickCount={7} orientation="right" />
+        <Bar dataKey="uv" />
+        <ReferenceLine x="201106" stroke="#666" label="201106" />
+        <ReferenceLine y={0} stroke="#666" label="0" labelPosition="start" />
```
gaurav5430, line 55: This test case is not using the `segment` prop

PR #1385, opened 2018-07-12
Changed current behavior touch events

alreadyExisted:
Old behavior touch event did not work.
- Changed handlers touch events


src/chart/generateCategoricalChart.js
```diff
@@ -1079,19 +1084,33 @@ const generateCategoricalChart = ({
 
     handleTouchMove = (e) => {
       if (e.changedTouches != null && e.changedTouches.length > 0) {
-        this.handleMouseMove(e.changedTouches[0]);
+        const { onTouchMove } = this.props;
```
xile611: @alreadyExisted The function handleMouseMove isn't called here. Can the tooltip show when we touch the chart ?

.eslintrc
```diff
@@ -54,8 +54,10 @@
     "no-multi-assign": "off",
     "react/jsx-closing-tag-location": "error",
     "react/no-unused-state": "error",
-    "react/prefer-stateless-function": "error"
+    "react/prefer-stateless-function": "error",
 
+    // Errors Expected linebreaks to be 'LF' but found 'CRLF'
+    "linebreak-style": ["error", "windows"]
```
xile611: @alreadyExisted This'll will cause a lot of errors in continuous-integration test.

PR #1388, opened 2018-07-13
Allow to add more classes in tooltips

hackedbellini:
Since #798 was fixed, there was no way to change the tooltip wrapper
style other than defining another class in css for
"recharts-default-tooltip" and mark everything as "important.

The problem with that approach is that we would be changing all the
tooltips, but what if we want to change it for only one chart?

This will allow to define extra classes  for both the wrapper and the
label, allowing us to customize a single chart easily.

### DIFF
--- a/src/component/DefaultTooltipContent.js
+++ b/src/component/DefaultTooltipContent.js
@@ -4,6 +4,7 @@
 import _ from 'lodash';
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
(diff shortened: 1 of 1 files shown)


hackedbellini:
Hi @paulmelnikow 

Thanks for approving my PR. I noticed that you didn't merge it yet. Do I need to do anything else here?

paulmelnikow:
I don't have push access; I'm just contributing my +1 😄

hackedbellini:
hahaha I see. Thanks :)

Issue #1305, opened 2018-05-17
Using custom tooltip with onMouseMove in Line/Area/ComposedChart


jzaczek:
## Do you want to request a feature or report a bug?
Report a bug.

## What is the current behavior?
I've stumbled upon a problem described in: https://github.com/recharts/recharts/issues/991. Updating state in onMouseMove interferes with rendering custom tooltip component. JSFiddle below.

https://jsfiddle.net/nxr4gwkz/

Note that I've added both custom tooltip (simple div that displays payload) and onMouseMove. onMouseMove prevents tooltip from being rendered. 

## What is the expected behavior?

Tooltip is rendered and setState works properly when onMouseMove is set on Line/Area/ComposedChart

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

I've tested this only on Recharts version 1.0.0-beta.10.

llamamoray:
@jzaczek I've just solved this by using the onMouseOver prop instead.

llamamoray:
When using onMouseMove if you put a delay on the `setState` using `setTimeout` you momentarily see the custom tooltip so it's definitely the act of calling setState and the resulting render that is casuing the custom tooltip to not display. I've spent a few hours looking through and debugging the code but can't see the problem. My only suspicion is in the logic here:

https://github.com/recharts/recharts/blob/db0f3f254f12d59583052e77a43822163c476c9f/src/component/Tooltip.js#L83-L91

Do any of the maintainers have any ideas? I actually can't use my workaround of using onMouseOver as it only captures mouse overs of the Bars in my BarChart rather that the entire vertical space.

jzaczek:
Thanks for the PR, but I'm still experiencing the same issue in my app. Which is weird, because the jsfiddle I've attached earlier that looks more-or-less like my app works fine:

https://jsfiddle.net/nxr4gwkz/5/

Do you have any suggestions? What I'm trying to achieve is to show both tooltip and activeLabel on the XAxis. JSFiddle works ok, while in my app I see only the activeLabel - tooltip is not showing.

simonhessner:
I am still experiencing the problem in 1.8.5. Did you find a solution?

jesperlandmer:
I am also experiencing this issue. Have you found a solution?

Issue #1220, opened 2018-03-12
Tooltip disappears after Component.setState()


EthanJStark:
With gratitude to the maintainers and contributors to this fine library!  If this is an error in my code, do let me know! I took a look at the source code to see if I could make this a PR instead of an issue, but couldn't determine the cause of this issue.

## Do you want to request a feature or report a bug?
Bug

## What is the current behavior?

As soon as `Component.setState()` is called by another component (not the recharts graph), the tooltip disappears. 

## Reproduction steps

Bug is present in this [deployed app](https://inbox-ai.herokuapp.com/detail/2).

Before adding the following code in the code for `<Sidebar />`, tooltip was working perfectly. `Sidebar` is a component unrelated to the recharts graph.

```js
  updateSidebar = event => {
    if (event && event.activeTooltipIndex) {
      const activePoint = this.state.graph.data[0].points[
        event.activeTooltipIndex
      ]

      this.setState(prevState => ({
        ...prevState,
        graph: Object.assign({}, this.state.graph, { activePoint })
      }))
    }
  }
```

Changing the contents of the setState function doesn't make a difference, as long as `setState()` gets called, `<Tooltip />`'s props get set to `active: false` and the tooltip disappears.

May be related to #1179, though I don't see the chart redraw animation happening each time setState is called.

I can recreate in a JSFiddle if someone has a working template (current templates are broken)

## What is the expected behavior?

Tooltip can persist even as `state` is updated.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Chrome 64.0.3282.186
Mac OS 10.12.6
Recharts 1.0.0-beta.10

Thanks for any insights!

khennicb:
It looks like this issue is there since a long time. More than a year
[https://github.com/recharts/recharts/issues/553](https://github.com/recharts/recharts/issues/553)

Here is a up-to-date JSFiddle to show the issue : 
[https://codesandbox.io/s/6x979p8rr3](https://codesandbox.io/s/6x979p8rr3)

I wonder if there is still some people working on this library, or if a misconception is preventing any fix on that.

OlivierNguyen:
I've made an MR to fix that: https://github.com/recharts/recharts/pull/1406

andynguy3n:
Sadly this issue is still there
https://codesandbox.io/s/a-moving-chart-from-rechart-x7juw

andynguy3n:
[...]

Issue #553, opened 2017-03-07
Tooltip Behavior


Eric24:
Do you want to request a feature or report a bug?
Feature (unless there is a way to handle this now)

What is the current behavior?
If the user moves the mouse inside the chart area and triggers the tooptip, it will disappear as soon as the chart is updated (i.e. receives new data). That may be fine for static charts, but for charts that are receiving new data in real-time, it's very annoying. As a possible workaround, we had considered "pausing" the data updates when the tooltip is shown (although this isn't the best solution), but there is no event that lets us know if the tooltip is visible or not. But in any case, the better behavior would be for the tooltip to remain visible even if the chart is updated.

If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/9y45ut13/).

What is the expected behavior?
Ideally, the tooltip would remain visible until the user moves the mouse outside the chart area.

Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Latest ReCharts; any browser.

gvaldambrini:
Same issue here. I made a fix (not extensively tested to be honest) for the 0.22.4 version here:
https://github.com/develersrl/recharts/commit/b5fa5c2cc6690d34b90693e0f5c2e15d90d9defd

dmr:
Another behavior that we noticed is that the tooltip disappears after a short while if the user doesn't move the cursor --> Is this configurable somehow?

WeeJeWel:
+1

James-E-Adams:
Same issue here. This is how it manifests for me:

![tooltip](https://user-images.githubusercontent.com/9067274/36246998-61d3fbd8-1286-11e8-9b33-34b7c2eb3b64.gif)

simonhessner:
I have the same issue in the newest version (1.8.5). Why was this closed?

Related:
https://github.com/recharts/recharts/issues/1220
https://github.com/recharts/recharts/issues/2143

simonhessner:
I made a pull request that fixes the problem: https://github.com/recharts/recharts/pull/2150

PR #1406, opened 2018-07-30
fix Tooltip disappears when using setState()

OlivierNguyen:
Bug 
---
- Tooltip disappears when using setState on `onClick` or `onMouseMove` on the chart. It is because the defaultState is always set when the component updated. 

Issues
---
https://github.com/recharts/recharts/issues/1305
https://github.com/recharts/recharts/issues/1220
https://github.com/recharts/recharts/issues/553

### DIFF
--- a/src/chart/generateCategoricalChart.js
+++ b/src/chart/generateCategoricalChart.js
@@ -189,21 +189,15 @@ const generateCategoricalChart = ({
         // update configuration in chilren
         const hasGlobalData = !_.isNil(nextProps.data);
         const newUpdateId = hasGlobalData ? updateId : updateId + 1;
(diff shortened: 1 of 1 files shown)


simonhessner:
It seems like this change is not present in version 1.8.5 anymore. Was this reverted on purpose? I am experiencing a disappearing tooltip in the current version

dankgen-tobias:
> It seems like this change is not present in version 1.8.5 anymore. Was this reverted on purpose? I am experiencing a disappearing tooltip in the current version

Same for me. Tooltip is disappering in LineChart when setting the state with onMouseMove

simonhessner:
I made a fix: https://github.com/recharts/recharts/pull/2150
The PR has not been merged yet, but maybe you could still try if it helps.

PR #1414, opened 2018-08-01
Replace webpack with rollup for umd bundling

TrySound:
Provded multi config to bundle both umd and min version at one step.
Added size snapshot plugin to track different sizes. For esm bundles
it's able to show treeshakability information.

Recharts.js 1.86M
Recharts.min.js 512.9K

Recharts.js 1.56M
Recharts.min.js 421.3K


paulmelnikow:
How does the build time compare?

TrySound:
Before 11s, after 67s. Do you think it's relevant? There's a different in output. New build provides more: development bundle and production minified bundle, information about circular dependencies, information about bundle size.

paulmelnikow:
When I’m pulling in recharts as a dependency with yarn link, and hacking on the library, I have to run the es6 build after every change. It feels like slow iteration as is; I feel a minute is too long.

TrySound:
It's just `build-umd` command. `build-es6` is not changed.

paulmelnikow:
Ah, gotcha.

I don't think the umd build is going to be anyone's primary development environment, especially if the build is going to take a minute. Wouldn't the development bundle and circular dependencies be better identified earlier in the process?

Also, doesn't the karma config depend on webpack? Whatever the gain here, is it worth the complexity of having two bundlers?

TrySound:
Webpack is still used for demo. I think it's okay to have two bundlers. We choose what is the best for users. About circular dependencies: I actually gonna bundle esm and cjs in the next PR. I will add a condition to bundle esm fast. Only full rebuild will take a time. But you will see even more info about your bundles. Size snapshot is able to produce treeshakability info for esm format.

TrySound:
Webpack still bigger. Rollup produces the most readable and compact code.

TrySound:
Okay, will go to write simple and fast charts to finally remove this dependency bloat.

Ronnyrr:
Recharts is still the biggest package in my vendor. The last version got even bigger. Any plans on making it smaller?

TrySound:
@Ronnyrr I guess there are no plans. I would help, but the owner of this library is not interested.

arshbhatti8:
@TrySound  @rnnyrk  What about  https://github.com/recharts/babel-plugin-recharts ?

Does this not help decreasing the bundle size? Honest question.

TrySound:
It's a hack which requires to cherry pick internals. Bad option.

Anyway I migrated to pure d3+react and cut 50 lines of code and 0.5mb in my bundle.

Issue #1419, opened 2018-08-02
ReferenceLine generates warning about alwaysShow


DavidHenri008:
There is a bug when using a ReferenceLine or ReferenceDot or ReferenceArea in a chart.

When a one of the reference is used in a chart a warning saying "The alwaysShow property is deprecated" is displayed when the mouse hover the chart. Obviously the alwaysShow property is not set on the reference.
It seems to be related with the modification of July 6th commit 51c55a6ea4c93034b756700bc350d5d538b22425, by @paulmelnikow. The warning should have been: 
```
 warn(alwaysShow === undefined,
      'The alwaysShow prop is deprecated. Please use ifOverflow="extendDomain" instead.');
```
Where the condition is "===" instead of "!==" since the warn function condition is what you are expecting to avoid the warning and not what cause the warning to be displayed. 

Obviously the workaround is to set the alwaysShow property to something different than undefined.

paulmelnikow:
Oops! Sorry about that. I’ll open a PR with a fix.

DavidHenri008:
Great thank you!

matsilva:
Can close this issue now?

DavidHenri008:
Let's close it.

dkulakowska:
I still get that warning in the console, don't really know why.

olelivalife:
How to fix this?

crickford:
`ifOverflow` is a prop on the `ReferenceLine`, `ReferenceDot`, and `ReferenceArea` components which should be used rather than the `alwaysShow` prop referenced in the docs. It is used to configure how reference components which fall outside of the domain of the chart are rendered, here are the options it accepts (from @types/recharts 1.1.22):
`'hidden' | 'visible' | 'discard' | 'extendDomain'`

crickford:
I've created a PR for the docs here: https://github.com/recharts/recharts.org/pull/101
Here's an example of how to use the `ifOverflow` prop: https://jsfiddle.net/fqt7xw0m/

PR #1430, opened 2018-08-11
Fix alwaysShow warn() condition in ReferenceLine

Philipp91:
The condition is the wrong way round because warn() works like assert().


paulmelnikow:
Thanks for opening this! This fixes #1419.

Philipp91:
Is there anything that prevents this from being merged?

paulmelnikow:
@xile611 is the one who can do it.

nathan-haines:
@xile611 is this something that may get merged in soon?

damilaufer:
I'd also love to see this patch merged... too many warnings!

dangh:
@Philipp91 can you fix it for ReferenceArea and ReferenceDot also?

PR #1434, opened 2018-08-17
Composed rectangle cursor

IanVS:
This restores the wide `rectangle` tooltip cursor when using `ComposedChart`, instead of the `1px` wide cursor that has started to be used in recent versions of recharts.  We also need to allow pointer events, so that we can use the `zoom-in` cursor when hovering a clickable report section.

Finally, because this is a fork and we want to use it within nutshell, we have to have a good way to build the project.  I found https://github.com/exogen/postinstall-build which seems to work great, preventing us from having to check in build artifacts as we've done with other forks in the past (`react-joyride`).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,10 +20,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.44",
-        "jsesc": "2.5.1",
-        "lodash": "4.17.10",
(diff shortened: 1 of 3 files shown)


IanVS:
Doh, I thought I was making a PR onto our own internal fork.  Sorry for the noob mistake.

PR #1438, opened 2018-08-20
Added animateNewValues property to Line

ymoran00:
I need the ability to add items dynamically to a line chart in a smooth way. Currently a "magic number" is calculated when items are added, which creates a large animation, but it's too much of a show. 
I've added attribute animateNewValues to Line. It's default is true, and then the Line behaves as usual. When it's false, new points are not animated.

### DIFF
--- a/src/cartesian/Line.js
+++ b/src/cartesian/Line.js
@@ -62,6 +62,7 @@ class Line extends Component {
     onAnimationEnd: PropTypes.func,
 
     isAnimationActive: PropTypes.bool,
+    animateNewValues: PropTypes.bool,
     animationBegin: PropTypes.number,
     animationDuration: PropTypes.number,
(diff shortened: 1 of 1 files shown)


src/cartesian/Line.js
```diff
@@ -336,9 +339,12 @@ class Line extends Component {
                 }
 
                 // magic number of faking previous x and y location
-                const interpolatorX = interpolateNumber(width * 2, entry.x);
-                const interpolatorY = interpolateNumber(height / 2, entry.y);
-                return { ...entry, x: interpolatorX(t), y: interpolatorY(t) };
+                if (that.animateNewValues) {
```
xile611: There is no need to read `animateNewValues` from `that`, just read this from `this.props`.

BionBit:
Please update the documentation as well.

PR #1481, opened 2018-09-20
Fix #1458

### LINKED ISSUE #1458: Fails to render chart for exponential values
Hi,
As a rechart user I often encounter values which are large. Sometimes of the order of e^22. In cases like these, rechart fails to render Bar, Bar Horizontal, Area and Line. It renders only Pie and Donut chart. Any fixes or workarounds?

ofk:
I fixed #1458.

I know the cause is in [recharts-scale](https://github.com/recharts/recharts-scale).
I copied recharts/recharts-scale#7 as `src/util/ScaleUtils.js`, simplified the code and deleted the dependency of recharts-scale.

We will have higher performance recharts available.
Thank you.


ofk:
I downgraded npm v5 and I reduced package-lock.json's changes.

By the way, this repository has yarn.lock. Would you use npm or yarn for this project?
Unnecessary lockfile should be deleted.

xile611:
@ofk Thanks for your awesome work!  [recharts/recharts-scale#7](https://github.com/recharts/recharts-scale/pull/7) has been merged. And recharts-scale@0.4.0 has been deployed. Maybe you can use recharts-scale@0.4.0 to update this mr!

PR #1509, opened 2018-10-09
add: find yAxisWithNiceTicks and choose it over getAnyElementOfObject

puemos:
When rendering the grid the method is grabbing the first YAxis component
The new feature is that it will prefer a YAxis with ticks and will fallback to the first one if none of them got "niceTicks"

oct-09-2018 18-18-15


xit4:
This is amazing, we have been waiting for this fix!

Issue #1529, opened 2018-10-18
AreaChart height not consistent


BionBit:
The height of the AreaChart differs with different data sets - even though a fixed height is specified.

For example,
Correct behavior: https://jsfiddle.net/r1tf29ap/
(The height of the chart is '100', like specified)

Incorrect behavior: https://jsfiddle.net/8415b7yx/
(The height of the chart is '84.17' - it should stretch to '100')

Last working version was 0.7.x

xile611:
@BionBit 
Thanks for your opinion ~

It's supposed to specify the domain of `YAxis` to be `['dataMin', 'dataMax']`. If you want to fill the fixed height of AreaChart.

```jsx
const {AreaChart, Area, YAxis} = Recharts;
const data = [
      {name: 'Page A', uv: 101, pv: 2400, amt: 2400},
      {name: 'Page B', uv: 10, pv: 1398, amt: 2210}
];
const TinyAreaChart = React.createClass({
	render () {
  	return (
    	<AreaChart width={200} height={100} data={data}
            margin={{top: 0, right: 0, left: 0, bottom: 0}}>
        <YAxis domain={['dataMin', 'dataMax']} hide />
        <Area dataKey='uv' stroke='#8884d8' fill='#8884d8' />
      </AreaChart>
    );
  }
})

ReactDOM.render(
  <TinyAreaChart />,
  document.getElementById('container')
);

```

BionBit:
@xile611 
Thank you for the workaround.

I've tested all the demo cases in various conditions (different data, with and without X and Y axes) and couldn't find any erratic behavior.

Can you point me to some areas you think this commit can be dangerous, so I can refactor the underlying design if possible?

PR #1531, opened 2018-10-20
[BUG-1529] fix inconsistent chart height due to faulty domain update

BionBit:
Fixing unwanted behavior, demos included in the issue.
https://github.com/recharts/recharts/issues/1529

The problem,
If we update the domain even when we don't show an axis (YAxis), then the 'y' offset for the max value of the data is different for each set of data.
(since the max tick value may differ from the max value of the data supplied)

This means that if an app displays several simple charts (tiny*chart) in a row with different sets of data, the charts will have different heights, resulting in inconsistent UX.

### DIFF
--- a/src/util/ChartUtils.js
+++ b/src/util/ChartUtils.js
@@ -756,7 +756,7 @@ export const calculateDomainOfTicks = (ticks, type) => {
  * @return {Object}      null
  */
 export const getTicksOfScale = (scale, opts) => {
-  const { realScaleType, type, tickCount, originalDomain, allowDecimals, hide:axisHidden } = opts;
+  const { realScaleType, type, tickCount, originalDomain, allowDecimals, hide : axisHidden } = opts;
   const scaleType = realScaleType || opts.scale;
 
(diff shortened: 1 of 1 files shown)


xile611:
@BionBit It's supposed to specify the `domain` of `YAxis`. 
```jsx
const {AreaChart, Area, YAxis} = Recharts;
const data = [
      {name: 'Page A', uv: 101, pv: 2400, amt: 2400},
      {name: 'Page B', uv: 10, pv: 1398, amt: 2210}
];
const TinyAreaChart = React.createClass({
	render () {
  	return (
    	<AreaChart width={200} height={100} data={data}
            margin={{top: 0, right: 0, left: 0, bottom: 0}}>
        <YAxis domain={['dataMin', 'dataMax']} hide />
        <Area dataKey='uv' stroke='#8884d8' fill='#8884d8' />
      </AreaChart>
    );
  }
})

ReactDOM.render(
  <TinyAreaChart />,
  document.getElementById('container')
);

```
Your commit will change the behavior of many cases. So I'll close this.

PR #1539, opened 2018-10-25
Scattercharts have been broken since 1.0.0-alpha.6

jgentes:
Fixing this typo fixes them. See https://github.com/recharts/recharts/commit/fe6452f0fbd75d62881a289d94a29e5f10b0b4ed#r25878004


jgentes:
Ok, I may have gotten ahead of myself here. Looking into the test failure..

jgentes:
I've narrowed this down a bit - the problem surfaces with multiple scatters rather than all data plotted on a single scatter.

PR #1567, opened 2018-11-16
Update package.json

Glitches:
Updating this package prevents triggering resizeObserver error

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "d3-shape": "~1.2.0",
     "lodash": "~4.17.4",
     "prop-types": "~15.6.0",
-    "react-resize-detector": "~2.3.0",
+    "react-resize-detector": "~3.2.0",
     "react-smooth": "~1.0.0",
     "recharts-scale": "^0.4.2",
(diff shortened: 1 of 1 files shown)


Glitches:
sorry for my late reply.
The error is reported in thi issue: https://github.com/maslianok/react-resize-detector/issues/45
I have a list of data displayed with a modified pie chart and as soon as I resize the window, the error pops up. In production wouldn't have been a problem if Sentry didn't catch it, but we had to shut that error because it was catching a lot of them. 
So, the point is that the error is generate from that package and may affect other users, but is not really blocking. Just annoying

dmr:
Does recharts still support react@15?

I would like to use recharts inside storybook and react-resize-detector has an issue which was fixed in 3.1.0 (https://github.com/maslianok/react-resize-detector/issues/34) --> What can I do to help this pull request getting merged?

Glitches:
@dmr You could fork and pack recharts yourself, if you think is reasonable enough having a forked package.

xile611:
We'll release a major version to only support react@16, then we'll update react-resize-detector ，thanks!

PR #1630, opened 2019-01-29
fix onMouseEnter and Tooltip for Pie on FireFox

### LINKED ISSUE #1587: Pie chart onMouseEnter events not firing in Firefox
Bug
The `onMonuseEnter` event never fires for the PieChart.
Note - possibly related to this recent PR - https://github.com/recharts/recharts/pull/1582
In Firefox - use the standard demo of the two-level pie chart here - https://jsfiddle.net/alidingling/hqnrgxpj/

In Google chrome - observe the `onMouseEnter` behavior here - https://jsfiddle.net/alidingling/hqnrgxpj/
Firefox is the only version effected I've noticed so far.

### LINKED ISSUE #1602: PieChart not working on Firefox
Bug

PieChart mouse events are not working on the latest version of Firefox (v 64)

Here (https://codesandbox.io/s/kxyv8xn58o) is a simple example of a pie chart that show an alert message when clicking on a pie. Using Chrome or and older Firefox version it works correctly, showing the pointer cursor when hovering the chart and, when clicked, the alert is shown.
When using Firefox 64, the cursor doesn't change and the click has not effects.
(I tested some other charts and it looks like it happens only on PieChart, the other charts are working fine)

The same problem occurs with the Recharts Docs example (http://recharts.org/en-US/examples/CustomActiveShapePieChart). The active shape should change when hovered, but it is not working on Firefox 64.

The click event should behave correctly on all browsers.

Recharts version: (1.4.2)
Browser: Firefox 64
OS: Tried on Linux (Ubuntu 16.04) and Windows 10

### LINKED ISSUE #1613: Tooltips are not appearing the Firefox v64 for the Pie Charts
Report a bug
Tooltips for PieChart are appearing in Google Chrome but not appearing in the Firefox
Open the Piecharts demo which has tooltips in Firefox
Tooltips should appear even in Firefox
Using recharts v1.4.1 , Firefox v64 , Mac OS - Mojave

Fonger:
The latest Firefox fails to trigger events for svg elements if wrapped with a non-existing `clip-path`.
`clip-path` is not necessary for `Pie` so removing the tag just works

fixes #1613, fixes #1602, fixes #1587

### DIFF
--- a/src/polar/Pie.js
+++ b/src/polar/Pie.js
@@ -493,7 +493,7 @@ class Pie extends Component {
 
   render() {
     const { hide, sectors, className, label, cx, cy, innerRadius,
-      outerRadius, isAnimationActive, prevSectors, id } = this.props;
+      outerRadius, isAnimationActive, prevSectors } = this.props;
 
     if (hide || !sectors || !sectors.length || !isNumber(cx) ||
       !isNumber(cy) || !isNumber(innerRadius) ||
@@ -505,9 +505,7 @@ class Pie extends Component {
 
     return (
       <Layer className={layerClass}>
-        <g clipPath={`url(#${_.isNil(id) ? this.id : id})`}>
-          {this.renderSectors()}
-        </g>
+        {this.renderSectors()}
         {label && this.renderLabels(sectors)}
         {Label.renderCallByParent(this.props, null, false)}
         {(!isAnimationActive || (prevSectors && _.isEqual(prevSectors, sectors))) &&


berzniz:
Thanks @Fonger ! -> I would love to see this merged

KarolBartyzel:
It blocks us, is it possible to merge it shortly ?

RParvathi:
When can we have the latest version with this fix available ?

hugobarragon:
i have done a fork with the fix with lib updated so you can temporarily npm install my fork until the fix its official merged
https://github.com/hugobarragon/recharts

Issue #1083, opened 2017-12-13
Animation rendering fails into Pie Chart.


lcustodio:
## Do you want to request a feature or report a bug?

I would like to report a bug.
Animation rendering fails into Pie Chart, there is space for improvement on the fallback.

## What is the current behavior?

There is an intermittent failure on the chart rendering. As exposed on the gif below:

1. The first rendering works nicely as expected.
2. The second rendering (page reload) is intercepted somehow and a default rendering takes place removing the animation and the legend information.

![out2](https://user-images.githubusercontent.com/585672/33879490-d8458844-df2e-11e7-93c8-dcf245849fa7.gif)

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/ndLhnegs/).

http://jsfiddle.net/lcustodio/x5x1de34/

## What is the expected behavior?

The animation to always work, and if some error happens fallback like it's the case currently, but the `label` and `labelLine` function should be called as well.
In addition, some error logging in order to understand and try to solve the problem.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

All the tests were made with Chrome 62 + macOS 10.13.1

ptdede:
I also facing the same issue. Exactly after facing data from internet using 'redux-saga', the animation stop immediately. I think it's because react triggered re-render after props changed. But, when using 'LineChart' the animation works perfectly.

@lcustodio did you find how to handle this already?

squierts:
For me the Line Chart has the same problem and is resolved by turning off animations.

AlexMiroshnikov:
Regarding the original (PieChart) problem - the reason is that `src/polar/Pie.js`::`handleAnimationEnd()` is not being called in case there is more than one pie chart on the page, meaning this component never gets state `isAnimationFinished:true`, which leads the labels not to appear.

lcustodio:
@ptdede we removed the animation.
Did you manage to try @AlexMiroshnikov's suggestion?

KalvinHom:
Ran into the same issue, and had to disable animation to get labels to show on initial render.  I am also using two piecharts in one page, which sounds like is the cause of the bug.

xile611:
@KalvinHom Any demo to reproduce this problem ?

eglenfield:
@xile611 

[...]

PR #1635, opened 2019-02-04
Disable caching if pie sector state is the same

eglenfield:
Add condition if new sectors match prev sectors don't cache. This should resolve the issue seen with this issue: https://github.com/recharts/recharts/issues/1083

*Edit:* Updated to include Bar chart update as well.

### DIFF
--- a/src/cartesian/Bar.js
+++ b/src/cartesian/Bar.js
@@ -181,9 +181,12 @@ class Bar extends Component {
 
   id = uniqueId('recharts-bar-');
 
-  cachePrevData = (data) => {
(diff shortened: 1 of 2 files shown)


xile611:
@eglenfield We check much more props in [generateCategoricalChart#L180](https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.js#L180) to check whether we need to update the pie or not. 
We can't just check the data of Pie.
If you have any idea to reproduce the problem in #1083, please give me an online demo. We'll try to fix the problem.

Issue #919, opened 2017-09-27
Use of peerDependencies or dependencies?


dmr:
In our project we use "d3-scale" 1.0.6.
If recharts would define d3-scale >= 1.0.4 as a peerDependency then we would not need to include both bundles into our build.

We have the same problem/topic with prop-types, core-js and lodash.

What do you think, are you interested in a pull request to change these dependencies to peerDependencies?

RickWong:
Terrible that recharts includes both `lodash` and `lodash-es` in dependencies. That needlessly increases the bundle size.

RickWong:
I could be mistaken. This is my `yarn.lock` file:

```

recharts@1.1.0:
  version "1.1.0"
  resolved "https://registry.yarnpkg.com/recharts/-/recharts-1.1.0.tgz#2ceea5f7438ba32eff54126b1136d5295f1dfbf8"
  dependencies:
    classnames "2.2.5"
    core-js "2.5.1"
    d3-interpolate "^1.1.5"
    d3-scale "1.0.6"
    d3-shape "1.2.0"
    lodash "~4.17.4"
    lodash-es "~4.17.4"
    prop-types "^15.6.0"
    react-resize-detector "1.1.0"
    react-smooth "1.0.0"
    recharts-scale "0.3.2"
    reduce-css-calc "1.3.0"
```

TrySound:
@RickWong `yarn.lock` is not responsible to what is bundled. How look your `recharts` imports?

RickWong:
I suspect `react-smooth` to be the culprit, it requires `lodash`, while the `recharts` selects `lodash-es`.

msholty-fd:
This type of thing seems like a no-brainer, and a very simple PR to open. I intend to open a PR to update this later today.

If you go into your node_modules in a project that has recharts, and delete the lodash folder in `recharts/node_modules`, and also add `lodash` to `peerDependencies` rather than `dependencies`, it removes lodash from `recharts` part of the bundle and makes it use the installed lodash instead.

Side-by-side comparison:
With lodash listed in `dependencies`:
![screen shot 2019-02-11 at 10 32 01 am](https://user-images.githubusercontent.com/13840573/52573532-5dbf6e00-2de8-11e9-9ca7-d79473c17857.png)

With lodash listed in `peerDependencies` and not shipped in the `recharts` dep:
[...]

PR #1641, opened 2019-02-11
Add lodash as a peer dependency instead of a dependency

msholty-fd:
See issue #919 for more details

Right now lodash is packaged in together with recharts, even if the project using recharts has lodash installed. I imagine most people are shipping two separate versions of lodash to production, even if they are the same version. Having lodash packaged in together with recharts as a `dependency` makes it so it cannot be "deduped" correctly with a build tool such as webpack.

Comparisons:

screen shot 2019-02-11 at 10 32 01 am

screen shot 2019-02-11 at 10 32 12 am

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -977,6 +977,15 @@
         "to-fast-properties": "^2.0.0"
       }
     },
+    "@sinonjs/commons": {
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.3.0.tgz",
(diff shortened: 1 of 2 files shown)


wojtekmaj:
I don't see a problem with lodash being deduped, as long as both you and recharts require versions ranges that overlap in at least one point.

Moving it to peerDependency is not a good idea; if you have non-compatible version of lodash, you will get a warning each time and recharts might cease to work.

The way to fix it is to be a little more liberal towards dependency versions. Like, `prop-types` is required to be `~15.6.0`, but nothing would break if it used `15.7.0` or later. So, correct version definition here would be simply `^15.6.0`. Same would apply to `lodash`, I suppose.

PS. Conflict in package-lock.

msholty-fd:
@wojtekmaj yeah, I think you're right about this PR. It may have been premature to assume moving to peerDependencies actually fixed it. I think your PR solves the problem, and in a better way. Closing this one.

PR #1644, opened 2019-02-13
Use more liberal approach to dependency versions

wojtekmaj:
This allows e.g. prop-types 15.7.0 to be used without issues. Without this change:
* many users will have libraries duplicated in their code because no versions are matching both e.g. ^15.7.0 and ~15.6.0 at the same time, or
* users will be forced not to update the libraries which happen to be also recharts dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "recharts",
-  "version": "1.4.2",
+  "version": "1.4.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -5365,7 +5365,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


dmr:
I think it would be even better to move most dependencies to peerDependencies.

I.e. we use core-js and other things in our projects in different versions for various reasons. With recharts forcing me to have a specific version my bundle blows up :(

msholty-fd:
@dmr I also thought this, but I think this can be handled without using peerDependencies. If the version that `recharts` uses is within a safe range, it allows any project to use `recharts` without shipping another version of lodash or core-js, and they can ship "their" version of those dependencies. Most build tools will dedupe the extraneous dependency.

If we move `core-js` and `lodash` to a peer dependency, you have to install both to make the project work period, lest you get a noisy complaint in your terminal. It's a bit awkward for a project to install a dependency just to make another dependency work. `core-js` and `lodash` are "ubiquitous" in the javascript ecosystem, but even then requiring you to list those out in your package.json is unnecessary imo.

dmr:
From my experience having these in dependencies can lead to weird problems and bigger bundles because people just assume that things work without thinking about depencencies of libraries and after hours of debugging somebody dives into the `node_modules/recharts/node_modules` folder to find a second lodash version which causes a problem.

It happened a few times to me in the last years but it is ok to say primary goal is that `yarn add recharts` gives a full installation.

Bigger ranges would be great 👍

PR #1655, opened 2019-02-27
use loop to find [min,max] to avoid RangeError

willr3:
Math.min.apply and Math.max.apply are recursive and will throw a RangeError: Maximum call stack size exceeded when the input array is too long. Looping over the elements and collecting the min and max in a single loop prevents the RangeError.


willr3:
I am using this branch for my work on larger data sets and wanted to check if this is something that could be merged into recharts. Is anyone opposed to this change?

xile611:
@willr3 Thanks! But I think it's better to use `lodash.max` and `lodash.min` to fix this.

willr3:
@xile611 is there a branch / commit I could test to see if it works with the my dataset?

Issue #1649, opened 2019-02-18
How to use ReferenceLine as separator for BarCategories?


EugeneDraitsev:
## Do you want to request a feature or report a bug?
feature proposal / approach discussion

## What is the current behavior?
Before version 1.5.0 (in 1.4.4) ReferenceLine worked incorrect ([1643](https://github.com/recharts/recharts/issues/1643), [1521](https://github.com/recharts/recharts/issues/1521))

And now I need to explain my problem. I'm trying to separate BarCategories with ReferenceLine like this:
![image](https://user-images.githubusercontent.com/16540432/52941014-2e6cac00-3368-11e9-8531-19cee49ea269.png)

It obviously worked fine before 1.5.0 just with this simple code:
```jsx
{map(xValues.slice(1), xValue =>
  <ReferenceLine key={xValue} x={xValue} stroke="#DBDBDB" />,
)}
```
But in 1.5.0, when ReferenceLine started work correctly, my lines shifted to correct positions — center of BarCategories:

![image](https://user-images.githubusercontent.com/16540432/52941220-90c5ac80-3368-11e9-8e69-a595ab16a37c.png)

So, my question is: How I can separate BarCategories with correct behaviour of ReferenceLine?
Maybe make sense to add some `position` prop for ReferenceLine with values `['start', 'middle', 'end']`, that allows you to select position of your ReferenceLine inside range of x/y. Or maybe you have any ideas how to implement this behavior with current API? 

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/ndLhnegs/).

That's not a bug, but example available [here](https://codesandbox.io/s/l71j57y3nl)
You can click on 'Toggle ugly fix' button to see my expected result

## What is the expected behavior?

I described it above

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Recharts: 1.4.3/1.5.0
OS: any os

shlangus:
In according to [docs](http://recharts.org/en-US/api/ReferenceLine#isFront) there is an `isFront` prop that should do what you need. Unfortunately, it seems broken in 1.5.0 as well as 1.4.2.

EugeneDraitsev:
@shlangus I think that `isFront` means "will be rendered over bars (like with higher z-index)", but maybe I'm wrong)

shlangus:
[...]

PR #1685, opened 2019-03-27
Added position props for ReferenceLine to allow to control offset of it.

EugeneDraitsev:
This Pull Request is connected to [1649](https://github.com/recharts/recharts/issues/1649)


EugeneDraitsev:
Horizontal and vertical demos are [available here](https://repl.it/repls/InternationalGrandioseCrash)

PR #1695, opened 2019-04-05
Add prop to restart animation after dataset change

delold:
As sometimes the animation of data change might be a buggy, this MR adds a prop `restartAnimationOnChange`, which will enforce an animation restart after every change of data, instead of animating intermediate values.

### DIFF
--- a/src/cartesian/Area.js
+++ b/src/cartesian/Area.js
@@ -68,6 +68,7 @@ class Area extends Component {
     onAnimationEnd: PropTypes.func,
     animationId: PropTypes.number,
     isAnimationActive: PropTypes.bool,
+    restartAnimationOnChange: PropTypes.bool,
     animationBegin: PropTypes.number,
     animationDuration: PropTypes.number,
     animationEasing: PropTypes.oneOf(['ease', 'ease-in', 'ease-out', 'ease-in-out', 'linear']),
@@ -90,6 +91,7 @@ class Area extends Component {
     hide: false,
 
     isAnimationActive: !isSsr(),
+    restartAnimationOnChange: false,
     animationBegin: 0,
     animationDuration: 1500,
     animationEasing: 'ease',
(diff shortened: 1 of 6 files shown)


slorenzo:
Hope this solves the animation issues!  🙏

xile611:
@delold The default value of `restartAnimationOnChange ` maybe should be `true`.

Mikiemm:
@xile611 Wouldn't this be a breaking change? This would impact apps which do not have this prop set and would suddenly see different animation.

delold:
@xile611 Wouldn't changing the default prop value to `true` cause a (unintended?) breaking change? After upgrade the animation will suddenly change, if the prop is not specified.

src/cartesian/Area.js
```diff
@@ -198,9 +200,9 @@ class Area extends Component {
   state = { isAnimationFinished: true };
 
   componentWillReceiveProps(nextProps) {
-    const { animationId, points, baseLine } = this.props;
+    const { restartAnimationOnChange, animationId, points, baseLine } = this.props;
 
-    if (nextProps.animationId !== animationId) {
+    if (!restartAnimationOnChange && nextProps.animationId !== animationId) {
```
xile611: @delold  the default value of `restartAnimationOnChange`  is set to be `false` now. 

Then if `nextProps.animationId !== animationId`, `prevPoints` will be updated. Then if `!_.isEqual(prevPoints, points)` , animation will be triggered.

if the value of `restartAnimationOnChange` is set to be `true`,   `prevPoints` will always be `null`, then every time the component will be rendered with animation.

is this behavior what you want?

delold: I believe this is not the case, as the condition basically prevents calling `cachePrevData` with previous data. 

When the `restartAnimationOnChange` is set to `false`, the `cachePrevData` should either be a no-op or be called with an empty array. 

PS: forgive my deleted comments, it is quite late here and I'm not thinking that straight 😅

delold: It is true though, that this can be considered confusing, it took me a while to understand it. If you want, I can redo the condition.

elieteyssedou:
Any update on this ? :)

slorenzo:
Hi @elieteyssedou, I experienced this issue. I just wanted to leave some comments to avoid it, a possible workaround. I realized that I had unnecessary renders when the states were updated. `React.Memo` helps you manage it before going to the recharts components. The animations work perfectly if the components are called once for status updates.

PR #1713, opened 2019-04-22
Adds forceCornerRadius prop to RadialBar

ycanales:
This adds the `forceCornerRadius` prop, which allows you to have a bar with rounded corners (a circle actually) when a small value in the `RadialBar` results in a bar shorter than the specified `cornerRadius`, in which case a normal bar with straight edges is drawn.

Hope it's useful!


ycanales:
Before:
Screen Shot 2019-04-22 at 17 34 26

After, with `forceCornerRadius` (and `cornerRadius`):

after

PR #1714, opened 2019-04-23
Use height for aspect

cruth544:
New to open source so bear with me.
I wanted to use ResponsiveChart with aspect in relation to height and not width. Currently aspect only works with width, but I feel it should also work with height as well. 
Updated test spec as well to accommodate new feature.


cruth544:
Force push to change user account

PR #1717, opened 2019-04-24
improve brush drag

dreamniker:


xile611:
@dreamniker Please fix the test errors.

PR #1736, opened 2019-05-15
Customized component.

dawiwt:
customized svg elements by rechart instance props and state.

eg:

```jsx
<Customized component={({ yAxisMap })=>{
    const yAxis = yAxisMap[0]
    const { x, y, width, height } = yAxis
    return <rect x={x + width - 4} y={y} width={4} height={height} fill="red" stroke="none" />
}}/>
```


xile611:
@dawiwt Thanks for your work! Everything is ok, except the component name `Customer`. How about `Customized` ?

dawiwt:
@xile611 Good idea!

PR #1757, opened 2019-05-30
Brush drag fix

peteruithoven:
When while dragging a brush the mouse is moved outside of the brush the brush will stick to the mouse, even when the mouse button was released while outside. This PR adds a document wide mouseup listener fixing this issue. 
When moving the mouse outside the brush area while dragging the brush, the dragging would pause. This PR also enables dragging the brush while outside the brush area.

This could be seen as a successor to #1717.

### DIFF
--- a/src/cartesian/Brush.js
+++ b/src/cartesian/Brush.js
@@ -448,8 +448,6 @@ class Brush extends Component {
         className={layerClass}
         onMouseMove={this.handleDrag}
         onMouseLeave={this.handleLeaveWrapper}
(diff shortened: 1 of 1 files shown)


Rusya44:
Ran into this issue right now, good PR 😄

xile611:
We'll try to fix this bug. It's strange to add / remove event in every update.

peteruithoven:
I don't understand why you think the code add / removes events on "every update", that's not what's happening...

lifeisaloha:
@xile611 any news? It's really weird behaviour.

PR #1811, opened 2019-07-19
Fix BarChart's hover rectangle covering up data

JohnStarich:
Might close https://github.com/recharts/recharts/issues/1039 (it seems the issue may have taken different forms over the past year)

I am sure this is *not* the best solution, but it does work and is remarkably simple. If you know of a better way to fix this bug, I'll happily update the PR. 😄 

One thing we could do instead of `.reverse()` at the end, is using `Array.unshift` instead of `Array.push` and also unfurling the `results` array after reversing them first.

Before:
Screen Shot 2019-07-18 at 11 05 10 PM

After:
Screen Shot 2019-07-18 at 11 02 48 PM

### DIFF
--- a/src/util/ReactUtils.js
+++ b/src/util/ReactUtils.js
@@ -486,7 +486,7 @@ export const renderByOrder = (children, renderMap) => {
     }
   });
 
-  return elements;
+  return elements.reverse();
 };
 
 export const getReactEventByType = (e) => {


JohnStarich:
Oof. That's embarrassing. After putting up this PR, I started experimenting with the order of the children inside my `<BarChart>` component. I discovered that placing the tooltip *after* the bars is what caused my issue.

Perhaps we should change this PR to enforce the tooltip's position at the beginning of the children or document this behavior somewhere?

xile611:
@JohnStarich It's recommended to change the order of component to change the `zIndex` of related elements.
Maybe you can help to enhance our document.

PR #1844, opened 2019-08-16
Add overflow to tooltip styles

jinwood:
Currently tooltips will overflow to the edge of their container if the text content is large, as you can see here:

before_cropped

With fix:

after_crop


xile611:
@jinwood It's supported to use props `contentStyle` of `Tooltip` to override the default style.

PR #1897, opened 2019-10-16
Fixed Text Component crash.

### LINKED ISSUE #1893: Text component crashes in v1.8.0 when scaleToFit is enabled

Bug report

When rendering a Text component with the `scaleToFit` prop set to true, the component crashes due to `wordsByLine` being empty but trying to access the first item https://github.com/recharts/recharts/blob/463f3c2c50fbfb5cb2de3675ccfc9853f8227bdc/src/component/Text.js#L160
`this.updateWordsByLines` was called in `componentWillMount` and `componentWillReceiveProps`, but moving to `componentDidMount` and `componentDidUpdate` results in a render before this. This means that `wordsByLine` has a length of 0 on first render.

I discovered the bug in a chart that is based on this example.

v1.7.1: https://codesandbox.io/embed/recharts-horizontal-bar-with-dual-y-axis-25v91
v1.8.0: https://codesandbox.io/embed/recharts-horizontal-bar-with-dual-y-axis-8orez

Text to render without crashing when `scaleToFit` is true.

Works in 1.7.1, breaks in 1.8.0.

### LINKED ISSUE #1894: v1.8.0 is causing various crashes
Recharts started crashing a lot after upgrading to `1.8.0`. I'm working on sandbox demo to reproduce the crash.

But this is the error I'm getting:

```
Error: TypeError: Cannot read property 'left' of undefined

in BarChart (at Purchases.js:369) in div (created by ResponsiveContainer) in ResponsiveContainer (at Purchases.js:368) 
```

The same error happens for `PieChart` and other charts. I'm using functional components.

sainthkh:
Fixes #1893 
Fixes #1894

### DIFF
--- a/src/component/Text.js
+++ b/src/component/Text.js
@@ -98,11 +98,6 @@ class Text extends Component {
     return this.getWordsWithoutCalculate(props);
   }
 
-  updateWordsWithoutCalculate(props) {
-    const words = !_.isNil(props.children) ? props.children.toString().split(BREAKING_SPACES) : [];
-    this.setState({ wordsByLines: [{ words }] });
-  }
-
   getWordsWithoutCalculate = (props) => {
     const words = !_.isNil(props.children) ?
       props.children.toString().split(BREAKING_SPACES) :
--- a/src/component/Text.js
+++ b/src/component/Text.js
@@ -49,7 +49,7 @@ class Text extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      wordsByLines: []
+      wordsByLines: this.getWordsByLines(props, true),
     };
   }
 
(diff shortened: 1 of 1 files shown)


src/component/Text.js
```diff
@@ -79,28 +85,31 @@ class Text extends Component {
...
       );
-      this.setState({ wordsByLines });
-    } else {
-      this.updateWordsWithoutCalculate(props);
     }
+    return this.getWordsWithoutCalculate(props);
   }
 
   updateWordsWithoutCalculate(props) {
     const words = !_.isNil(props.children) ? props.children.toString().split(BREAKING_SPACES) : [];
```
xile611: this line can be replaced by `this.getWordsWithoutCalculate(props)`

xile611:
@sainthkh Thanks for you commit.  Would you want to be one of the maintainers?  If so, maybe send me an e-mail and contact me. Thanks!

src/component/Text.js
sainthkh: That'll cause eslint consistent-return error. Because undefined is returned `(props.width || props.scaleToFit) && !isSsr()` is false.

sainthkh:
@xile611 Sent you an email to dingling112@gmail.com. I hope the address is correct.

PR #1905, opened 2019-10-17
Used UNSAFE_componentWillReceiveProps.

sainthkh:
Currently, recharts 1.18.x is really unstable. It happened because all `componentWillReceiveProps` are changed to `componentDidUpdate`.

Actually, it's a wrong choice for recharts. Because `componentWillReceiveProps` are called before `render` and `componentDidUpdate` is called after `render`. 

Currently, `componentWillReceiveProps` lifecycle methods are used to memoize values related to rendering things. 

So, they should be fixed with [`memoize-one`](https://github.com/alexreardon/memoize-one) or rewritten as functional components with hooks.

### DIFF
--- a/src/cartesian/Area.js
+++ b/src/cartesian/Area.js
@@ -197,10 +197,11 @@ class Area extends PureComponent {
 
   state = { isAnimationFinished: true };
 
-  componentDidUpdate(prevProps) {
-    const { animationId, points, baseLine } = prevProps;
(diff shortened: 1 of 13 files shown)


avindra:
Nice find. `componentWillReceiveProps` hides a lot of bad / legacy behavior. I've found it often requires slowly changing methods to avoid subtle bugs. Function components with hooks is a great way to get away from it.

PR #1908, opened 2019-10-17
Allow tooltips to explicitly escape the viewBox

dimitropoulos:
This PR allows for an option wherein the tooltip can move outside of the viewBox (in either direction) if the user wants it.  In case you're thinking: "I don't wan the tooltip to be outside of the viewBox", consider that recharts already allows this in two cases:
1. [[demo](https://jsfiddle.net/dimitropoulos/7r9g4c3x/7/)] when the tooltip is larger (in any one dimension) than the viewBox
1. [[demo](https://jsfiddle.net/dimitropoulos/epaLqm8y/8/)] when the tooltip is positioned explicitly.

This PR simply gives the user explicit control of _when_ the tooltip goes outside of the viewBox (for each direction/axis independently, `x`, and `y`).

The use case is for when you just want the tooltip to follow the mouse (just like it does with the standard behavior, before this PR) but you don't care that the tooltip goes outside of the viewBox (i.e. the change this PR introduces).  For example: check out pretty much any tooltip on the public Grafana for GitLab:
https://dashboards.gitlab.com/d/RZmbBr7mk/gitlab-triage?refresh=30s&orgId=1

Screenshot_20191017_141759

### DIFF
--- a/src/component/Tooltip.js
+++ b/src/component/Tooltip.js
@@ -28,6 +28,10 @@ const getUniqPaylod = (option, payload) => {
 };
 
 const propTypes = {
+  allowEscapeViewBox: PropTypes.shape({
+    x: PropTypes.boolean,
+    y: PropTypes.boolean,
+  }),
   content: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
   viewBox: PropTypes.shape({
     x: PropTypes.number,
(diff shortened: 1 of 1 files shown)


dimitropoulos:
thanks @xile611! please take a peek at the update to [the docs](https://github.com/recharts/recharts.org/pull/98) as well as [the types](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39230) when you get a chance.

PR #1912, opened 2019-10-18
Fix demo.

sainthkh:
Demo doesn't work on 2.0. So, it is fixed.


sainthkh:
Closed in favor of #1918.

PR #1913, opened 2019-10-19
TypeScriptified shape

sainthkh:
## Note
* Removed prop-types.
* `PRESENTATION_ATTRIBUTES` `EVENT_ATTRIBUTES` became unnecessary because we have better ones defined in @types/react.
* As for `Dot` and `Curve`, they have different event handler argument list, so @types/react version of `DOMAttributes` had to be modified and created new `adaptEventHandlers`. 
* Now, we don't have to manually filter prop-types. Everything is done statically by TypeScript. So, `getPresentationAttributes` and `filterEventAttributes` are gone.


sainthkh:
Closed in favor of #1918.

PR #1916, opened 2019-10-21
Adding Composed chart to rescaled charts

### LINKED ISSUE #1887: ComposedChart margin not work inside ResponsiveContainer
I have a problem when I using ComposedChart inside ResponsiveContainer
I set ComposedChart margin at zero, but not work.

This is my example:
https://codesandbox.io/s/responsivecontainer-seu9b

Where is the problem?

antoninklopp:
This should fix #1887 
This is the result I get using the same code from the sandbox after the fix. 

responsive-container

### DIFF
--- a/src/util/ChartUtils.js
+++ b/src/util/ChartUtils.js
@@ -557,7 +557,7 @@ export const parseScale = (axis, chartType) => {
     }
 
     if (type === 'category' && chartType && (chartType.indexOf('LineChart') >= 0 ||
-      chartType.indexOf('AreaChart') >= 0)) {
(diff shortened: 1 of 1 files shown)


xile611:
@antoninklopp This PR has been reverted, because of feedback.

It's recommended to add <XAxis scale="point"/> if you want to change the default behavior.

antoninklopp:
@xile611 Very sorry to hear this, I thought that I was just fixing the bug. What does it break?

xile611:
@antoninklopp It's break when <ComposedChart /> has <Bar />

PR #1919, opened 2019-10-22
2.0: rewrite polar/ by ts

xile611:


src/polar/Pie.tsx
```diff
@@ -1,99 +1,110 @@
 /**
  * @fileOverview Render sectors of a pie
  */
-import React, { PureComponent } from 'react';
-import PropTypes from 'prop-types';
+import React, { PureComponent, FunctionComponent, Component, ReactNode, ReactElement } from 'react';
```
sainthkh:
```suggestion
import React, { PureComponent, ReactElement } from 'react';
```

src/util/types.ts
```diff
@@ -27,9 +31,31 @@ export type PresentationAttributesWithProps<P, T> = AriaAttributes &
...
+  TIME = 'time', 
+  BAND = 'band', 
+  POINT = 'point',
+  ORDINAL = 'ordinal', 
+  QUANTILE = 'quantile', 
+  QUANTIZE = 'quantize', 
+  UTC = 'utc', 
+  SEQUENTIAL = 'sequential', 
+  THRESHOLD = 'threshold',
+};
```
sainthkh:
```ts
export type ScaleType = 'auto' | 'linear' | 'pow' | ...
```
Isn't above is shorter and better than enum type? Will we use it like `scaleType == ScaleType.AUTO`?

xile611: Good suggestion! I'll change this.

Philipp91:
The type `ReactElement<SVGElement> | ((props: any) => SVGElement)` occurs several times in this PR. It doesn't work for me in practice. I'm trying to feed a function component to `Pie.activeShape`, which TypeScript refuses saying that `Type '(props: ...) => JSX.Element' is not assignable to type '(props: any) => SVGElement'.`

Should it be `ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>)` instead? I.e. "either you have rendered the element already or you supply a function that will".

PR #1921, opened 2019-10-24
TypeScriptify util.

sainthkh:
Coming soon...


src/polar/types.ts
```diff
@@ -0,0 +1,10 @@
+import { Props as PolarAngleAxisProps } from './PolarAngleAxis';
+import { Props as PolarRadiusAxisProps } from './PolarRadiusAxis';
+
+// TODO: 1. what is the type of scale? 2. Should we include scale to each Props?
+export type AngleAxisProps = PolarAngleAxisProps & { scale: any };
```
xile611: There is a cycle-import here. `PolarAngleAxis` import  some attributes from `types.ts`.

src/util/ChartUtils.ts
```diff
@@ -1,19 +1,26 @@
...
 } from 'd3-shape';
 import { isNumOrStr, uniqueId, isNumber, getPercentValue, mathSign, findEntryInArray } from './DataUtils';
+// @ts-ignore
 import ErrorBar from '../cartesian/ErrorBar';
 import Legend from '../component/Legend';
+// @ts-ignore
 import { findAllByType, findChildByType, getDisplayName } from './ReactUtils';
+import { ReactElement, ReactNode } from 'react';
+import { RadiusAxisProps, AngleAxisProps, TickItem } from '../polar/types';
+import RadialBar from '../polar/RadialBar';
```
xile611: cycle-import here.

PR #1927, opened 2019-10-25
2.0: rewrite cartesian/ and numberAxis/ by ts

xile611:


sainthkh:
Build failed.

sainthkh:
I'll review them tomorrow.

xile611:
@sainthkh 
I think It's better to keep the default behavior of `filterProps` consistent with `getPresentationAttributes` which will make it easier to migrate from `getPresentationAttributes` to `filterProps`. 

And add a optional parameter `includeEvents` to `filterProps` to allow filter all the presentation attributes and events.

This is the main change in the latest commit.

sainthkh:
@xile611 Then, we need to make a plan for removing `PRESENTATION_ATTRIBUTES`.

src/cartesian/Area.tsx
```diff
@@ -1,79 +1,90 @@
...
+interface AreaPointItem extends CurvePoint {
+  value?: number | number[];
+  payload?: any;
+}
 
 
-class Area extends PureComponent {
+interface InternalAreaProps {
+  xAxis?: Omit<XAxisProps, 'scale'> & { scale: D3Scale<string | number> };
+  yAxis?: Omit<YAxisProps, 'scale'> & { scale: D3Scale<string | number> };
```
sainthkh: I found this pattern a lot in your code. How about creating `InternalXAxisProps` and `InternalYAxisProps`?

src/cartesian/Area.tsx
```diff
@@ -95,10 +106,10 @@ class Area extends PureComponent {
     animationEasing: 'ease',
   };
 
-  static getBaseValue = (props, xAxis, yAxis) => {
+  static getBaseValue = (props: Props, xAxis: Props['xAxis'], yAxis: Props['yAxis']): number => {
```
sainthkh: If you agree to create `InternalXAxisProps` and `InternalYAxisProps`, we can change `Props[`xAxis`]` and `Props[`yAxis`]` to them.

src/cartesian/Area.tsx
```diff
@@ -120,7 +131,20 @@ class Area extends PureComponent {
...
   static getComposedData = ({ props, xAxis, yAxis, xAxisTicks, yAxisTicks, bandSize,
-    dataKey, stackedData, dataStartIndex, displayedData, offset }) => {
+    dataKey, stackedData, dataStartIndex, displayedData, offset }: {
+      props: Props;
+      item: Area;
+      bandSize: number;
+      xAxis: InternalAreaProps['xAxis'];
+      yAxis: InternalAreaProps['yAxis'];
+      xAxisTicks: XAxisProps['ticks'];
+      yAxisTicks: YAxisProps['ticks'];
```
sainthkh: I also see `XAxisProps['ticks']` pattern a lot. What do you think is better, `XAxisProps['ticks']` or `TickItem[]`?

src/cartesian/Area.tsx
xile611: @sainthkh If we add `InternalXAxisProps` in `utils/types.ts`, `utils/types.ts` need to import `XAxis`, but `XAxis` need to import `utils/types.ts`. It leads to a cycle.

sainthkh: What I was talking about is putting them under `/cartesian/types.ts`. Because those `props` are only used in `/cartesian`.

sainthkh: Good. Let's leave them as-is for now. We need to think about them later.

Issue #1849, opened 2019-08-21
Bug: Y-Axis data point extends beyond chart (and "auto" or "dataMin|Max" has no effect)


Eric24:
## Do you want to request a feature or report a bug?
Bug

## What is the current behavior?
See the screenshot below. Notice how the 3.20 value is rendered off the bottom of the chart. The behavior is the same for Y-Axis domain of `["auto", "auto"`] and `["dataMin", "dataMax"]`. The only way to "fix" it is to set a fixed domain value, like `[0, "auto"]`.
![image](https://user-images.githubusercontent.com/685316/63444503-56fb5e80-c3fc-11e9-8de2-ec07f6a13e91.png)

## What is the expected behavior?
The Y-Axis should auto-adjust to the full range of values in "auto" or "dataMin|Max" domain modes.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Currently using ReCharts 1.6.2
I can't be sure if it worked in a previous version or not. It's not something I've seen before, but it may be data-related. The only "unusual" thing about this data is that some of the values are nulls, but I wouldn't think that should cause a problem.

adoel90:
After i face the same problem with dataMin, dataMax in Recharts,  i think instead use domain=["dataMin", "dataMax"].

We can simply to solve the bug dataMin dataMax with domain=[Math.min(...valueTemperatureY), Math.min(...valueTemperatureX)]

xile611:
Sorry for not handling your issue in time, please try the latest version. If the problem persists, please open a new issue according to the issue guide.

mattboan:
Could be that your data is an array of strings? That was my problem.

joshuadschoep:
> Could be that your data is an array of strings? That was my problem.

Anyone coming in after, check for this. This ended up being my silly problem as well.

PR #1936, opened 2019-11-08
Avoid using lodash.max & lodash.min which doesn't accept number string

phanect:
The max value of the Y-axis is not properly calculated currently.

Here's the reproduction code:

```jsx
import React from 'react';
import ReactDOM from 'react-dom';
import { LineChart, Line, Legend, CartesianGrid, XAxis, YAxis, Tooltip } from "recharts";

ReactDOM.render((<div>
  <LineChart width={1400} height={360} data={[ { x: 0, y: "1651.00" }, { x: 1, y: "6929.00" }, { x: 2, y: "16989.00" }]}>
    <Line type="monotone" dataKey="y" />

    <Tooltip />
    <CartesianGrid />
    <XAxis dataKey="x" />
    <YAxis dataKey="y" />
    <Legend />
  </LineChart>
</div>), document.getElementById('root'));
```

This results following graph:

Screenshot: Max value is not calculated properly

Although the max value of the Y-axis should be 16909.00, actually the max value is 8000 and the graph is only partially displayed.

In this case, `y` values are given as a string, not a float. The max value of Y-axis is correctly calculated and it works as expected if you give the `y` values as a float.
However, it is sometimes easier to develop if you don't need to convert string to float. For example, when you load data from CSV, some CSV parsing libraries give number data as strings.

This problem is caused by lodash's `_.max()`. ([here](https://github.com/recharts/recharts/blob/master/src/util/ChartUtils.js#L36))
If you give number string, `_.max()` doesn't work properly in some cases.
For example, `_.max([ "1651.00", "6929.00", "16989.00" ])` returns `"6929.00"`.

In this pull request, I reverted to use `Math.min()` and `Math.max()` which works properly when you give the data as string values.
According to the changelog, the core developers seem to have switched from `Math.min()` and `Math.max()` to `_.min()` and `_.max()` to avoid `.apply`, but you can use [destructuring syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) instead.

### DIFF
--- a/src/cartesian/Area.js
+++ b/src/cartesian/Area.js
@@ -273,12 +273,12 @@ class Area extends PureComponent {
     const startX = points[0].x;
     const endX = points[points.length - 1].x;
     const width = alpha * Math.abs(startX - endX);
-    let maxY = _.max(points.map(entry => (entry.y || 0)));
+    let maxY = Math.max(...points.map(entry => (entry.y || 0)));
 
     if (isNumber(baseLine)) {
       maxY = Math.max(baseLine, maxY);
     } else if (baseLine && _.isArray(baseLine) && baseLine.length) {
-      maxY = Math.max(_.max(baseLine.map(entry => (entry.y || 0))), maxY);
+      maxY = Math.max(Math.max(...baseLine.map(entry => (entry.y || 0))), maxY);
     }
 
     if (isNumber(maxY)) {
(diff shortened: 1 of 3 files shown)


phanect:
This issue might solve #1849.

xile611:
@phanect Why not convert the string-number to number before? We use Math.min and Math.max before, but refactor to use _.min and _.max in recharts@1.6.0, because Math.min and Math.max may throw error in some cases.

PR #1979, opened 2019-12-27
Rebase on upstream 1.x branch

IanVS:
This keeps the changes we made in our fork, but pulls us up to the latest 1.x commit.  2.0 seems to indeed have some breaking changes, so we're not quite ready for that.

This also removes the `postinstall-build` package, opting instead for using the `prepare` step that is baked into npm and yarn.


IanVS:
I meant to submit this into my own fork, sorry for the noise.

Issue #1600, opened 2018-12-28
How overide Traveller in Brush component?


wnetking:
Want add my custom traveller but i can find API for this. 
Have you any plans for this?

Thanks for this package :+1: 
https://github.com/recharts/recharts/blob/485b377dcc961d97255f3b542b52d353ff55f582/src/cartesian/Brush.js#L329

PR #1981, opened 2020-01-04
feat: Add styling customization options to Brush

Tiamanti:
fix: Fix some TSC errors in Brush

Recently started to use Recharts and found that Brush lacks some customization options.

All changes are backward compatible and new props will only be used when explicitly specified.
This PR adds the following optional props to Brush:
- backgroundFill?: string;
- backgroundStroke?: string;
- slideFill?: string;
- slideStroke?: string;
- slideOpacity?: number;
- slideStrokeWidth?: number;
- travellerContent?: FC\<TravellerProps>;
- travellerFill?: string;
- travellerStroke?: string;

I additionally fixed some TSC errors in Brush, added script to check types and added an example in LineChart demo.

This also addresses #1600

### DIFF
--- a/src/cartesian/Brush.tsx
+++ b/src/cartesian/Brush.tsx
@@ -1,7 +1,7 @@
 /**
  * @fileOverview Brush
  */
-import React, { PureComponent, Children, ReactText, MouseEvent, ReactElement, TouchEvent } from 'react';
+import React, { PureComponent, Children, ReactText, ReactElement, TouchEvent, FC } from 'react';
 import classNames from 'classnames';
 import { scalePoint, ScalePoint } from 'd3-scale';
 import _ from 'lodash';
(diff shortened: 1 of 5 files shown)


Tiamanti:
@xile611 Could you check out this PR?

src/cartesian/Brush.tsx
```diff
@@ -370,36 +389,44 @@ class Brush extends PureComponent<Props, State> {
...
-        />
-        <line
-          x1={x + 1}
-          y1={lineY + 2}
-          x2={x + travellerWidth - 1}
-          y2={lineY + 2}
-          fill="none"
-          stroke="#fff"
-        />
+        { TravellerContent
```
xile611: TravellerContent should support component and function.

bsplosion:
It doesn't look like this was merged before closing - any particular reason why?  I'm running into issues with the lack of style options on the brush traveller as well and it could really use an API.

PR #1991, opened 2020-01-16
Added the angle as key which need to be used in the Label align.

hardikbeladiya:


Markus-ipse:
@xile611 Any chance this will get any attention in the near future?
Seems like the Recharts repo has been at a standstill since December, are you in need of additional maintainers?

PR #2006, opened 2020-01-29
Allow axis line customization through axisLine prop

donysukardi:
Currently `axisLine` prop on `<XAxis />` is not respected. This change fixes that.


dmfilipenko:
@xile611 any chance to take a look at this?

PR #2013, opened 2020-02-03
fix usage of hooks in Tooltip, Label, Legend and Customized

bertho-zero:
Fix 
image

### DIFF
--- a/src/component/Customized.tsx
+++ b/src/component/Customized.tsx
@@ -1,7 +1,7 @@
 /**
  * @fileOverview Customized
  */
-import React, { isValidElement, cloneElement, Component, FunctionComponent } from 'react';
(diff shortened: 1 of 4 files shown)


noisae:
This change introduced a new problem, now for customized components, we are re-creating every time the component inside the Layer component. With the render prop react was handling better it and rerendering the components but now it recreates.

Could you @bertho-zero explain better why we need this change? I tried some hooks on tooltips and it seems to work with render prop... tks! :) https://stackblitz.com/edit/react-vd6sjt?file=src%2FApp.js

bertho-zero:
It would take 2 separate props:

render: a function that returns a component
component: a component function

Like here: https://github.com/final-form/react-final-form/blob/master/src/renderComponent.js

PR #2025, opened 2020-02-10
Invert cartesian label position based on negative values

### LINKED ISSUE #1058: Unexpected label positioning for vertical bar chart with negative values
Labels (LabelList) are not positioned as expected with vertical bar charts with negative values.
When setting negative bar LabelList position as `insideLeft` the label is positioned incorrectly.
actual

Expected layout is illustrated in this screenshot.
expected

Sample code:
http://jsfiddle.net/keopyn58/

kellytruter:
### Summary
I have been running into an issue where my Bar Chart labels do not render as expected when the layout is set to vertical. eg. where the label position is set to "right" - all bars that are positive, display with the label on the outside right-aligned to the bar, but all negative bars display the label on the inside left of the bar.

I would expect the label on the negative bars to display on outside of the bar and to the left of the bar, as it would be the inverse of the positive label position.

After some digging, I found that @tmikoss had a fix for the same issue but only for horizontal bar charts:
see https://github.com/recharts/recharts/pull/1346.

The solution I applied here extends its solution to apply to all [position options](http://recharts.org/en-US/api/Label#position).

**Note**: I have also added two bar charts to the demo/component/BarChart.js to demonstrate the usage of the labels (but I am fine with removing them).

Resolves #1058

image

### DIFF
--- a/demo/component/BarChart.js
+++ b/demo/component/BarChart.js
@@ -10,10 +10,10 @@ import { changeNumberOfData } from './utils';
 const colors = scaleOrdinal(schemeCategory10).range();
 
 const data = [
-  { name: 'food', uv: 2000, pv: 2013, amt: 4500, time: 1, uvError: [100, 50], pvError: [110, 20] },
-  { name: 'cosmetic', uv: 3300, pv: 2000, amt: 6500, time: 2, uvError: 120, pvError: 50 },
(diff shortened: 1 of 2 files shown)


kellytruter:
@xile611,
Any chance you or anyone else could review these changes?

kellytruter:
Thanks @xile611

tusharagarwal49:
Hi,
After resolving this, I'm facing another issue. Labels for negative bars with big values are overlapping the y-axis.

image

has anyone resolved this?

PR #2062, opened 2020-03-19
Get legend wrapper boundingRect to correctly compute legend offset

liborvachal:
xAxis is overflowing legend for first render, because `getBBox()` called on instance returns zero height. `wrapperNode` have already correct size so we should use `wrapperNode.getBoundingClientRect()` instead.


xile611:
@liborvachal  I think it's better to update the `getBBox` of `Legend`.

PR #2124, opened 2020-05-12
Pass tickFormatter as a prop

donysukardi:
Currently custom tick has no access to the formatted value or even supplied `tickFormatter`. This change increases the composability of different charts that are able to share custom tick styling


donysukardi:
I'm wondering if it'd be useful to pass in already formatted tick value too.

PR #2173, opened 2020-06-18
fix(AreaDot Type): add option to use a function that returns a react element

MatanAvneri:


MatanAvneri:
@xile611 can you please take look?

PR #2275, opened 2020-09-10
Adding SVG passthrough props to axis prop types

ranneyd:
In `CartisianAxis.tsx:361` it's passing props that on line 392 include the props given to the axis. Thus, if you give `fontSize` and `fontFamily` to an `XAxis`, it will properly apply those attributes to the underlying `<text>` elements of the axis labels. However, the type def for `BaseAxisProps` doesn't include any SVGElement props.

These are the two that I need. Instead of trying to add an exhaustive list I figure people can request them or make their own PRs.

### DIFF
--- a/src/component/Label.tsx
+++ b/src/component/Label.tsx
@@ -45,6 +45,7 @@ interface LabelProps {
   children?: ReactNode;
   className?: string;
   content?: ContentType;
+  angle?: number;
 }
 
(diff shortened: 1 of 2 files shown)


HunterEl:
Can you show an example of what this looks like after running with your changes. @ranneyd ?

ranneyd:
Update: I also added `angle` as a valid prop to `Label`, since it is valid.

PR #2291, opened 2020-09-29
fix: defer when syncing to other charts

forivall:
With 8 charts syncing, i noticed some serious jank with the `Line` and `Tooltip`. Simple fix by yielding the event loop. Debouncing could also be effective.

Also includes some typescript typing improvements.

If desired, I'm also open to contributing more typing improvements.

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1199,47 +1199,12 @@ const generateCategoricalChart = ({
     };
 
     handleReceiveSyncEvent = (cId: number | string, chartId: string, data: CategoricalChartState) => {
-      const { syncId, layout } = this.props;
-      const { updateId } = this.state;
+      const { syncId } = this.props;
 
       if (syncId === cId && chartId !== this.uniqueChartId) {
-        const { dataStartIndex, dataEndIndex } = data;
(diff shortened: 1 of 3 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -1198,48 +1198,13 @@ const generateCategoricalChart = ({
...
-                x: layout === 'horizontal' ? tooltipTicks[activeTooltipIndex].coordinate : validateChartX,
-                y: layout === 'horizontal' ? validateChartY : tooltipTicks[activeTooltipIndex].coordinate,
-              }
-            : originCoordinate;
-
-          this.setState({ ...data, activeLabel, activeCoordinate, activePayload });
-        } else {
-          this.setState(data);
-        }
+        (window.requestAnimationFrame || window.setImmediate || window.setTimeout)(
```
xile611, line 1205: this line may  throw error in node?

forivall: i'm guessing that would be in SSR? and so, the environment for SSR doesn't provide a `window` global?

xile611:
@forivall  could you fix the conflicts and reopen this PR ?

Issue #1667, opened 2019-03-11
Stacked area charts produce unexpected graphs with negative/zero values


7hibault:
## Do you want to request a feature or report a bug?

Bug

## What is the current behavior?

Stacked area charts produce unexpected graphs with negative/zero values

![Capture d’écran de 2019-03-11 22-58-05](https://user-images.githubusercontent.com/8068131/54161061-5922bf80-4451-11e9-9236-f24e4f719a29.png)

https://jsfiddle.net/x8nscgk0/

## What is the expected behavior?

The stroke should be at zero.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Recharts : 1.5.0
OS : Ubuntu 18.04.2 LTS
Browser : Firefox Quantuim 65.0.1 (32-bit)

PR #2292, opened 2020-09-29
fix: readd sign support for stacks

sakulstra:
Currently area charts which are stacked on v2 are not able to be plotted diverging.
When looking trough the code i found "sign" which seems to be similar, but doesn't work in version 2.
https://codesandbox.io/s/wandering-feather-meu6s?file=/src/Chart.js

Not sure i it would make sense to drop the custom "sign" code in favor of "diverging"(that's how it's called in d3 as i get it).

With this patch:
image

Examples from the docs like: http://recharts.org/en-US/examples/BarChartStackedBySign are completely broken on vs atm

### DIFF
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -4,6 +4,7 @@ import * as d3Scales from 'd3-scale';
 import {
   stack as shapeStack,
   stackOrderNone,
+  stackOffsetDiverging,
   stackOffsetExpand,
   stackOffsetNone,
   stackOffsetSilhouette,
@@ -739,6 +740,7 @@ export const offsetSign = (series: any) => {
 
 const STACK_OFFSET_MAP: Record<string, any> = {
   sign: offsetSign,
+  diverging: stackOffsetDiverging,
   expand: stackOffsetExpand,
   none: stackOffsetNone,
   silhouette: stackOffsetSilhouette,
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -4,7 +4,6 @@ import * as d3Scales from 'd3-scale';
 import {
   stack as shapeStack,
   stackOrderNone,
-  stackOffsetDiverging,
   stackOffsetExpand,
   stackOffsetNone,
   stackOffsetSilhouette,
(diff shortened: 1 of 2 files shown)


src/util/ChartUtils.ts
```diff
@@ -743,13 +742,13 @@ const STACK_OFFSET_MAP: Record<string, any> = {
...
-};
+} as const;
 
-export const getStackedData = (data: any, stackItems: any, offsetType: string) => {
+export const getStackedData = (data: any, stackItems: any, offsetType: keyof typeof STACK_OFFSET_MAP) => {
   const dataKeys = stackItems.map((item: any) => item.props.dataKey);
   const stack = shapeStack()
     .keys(dataKeys)
-    .value((d, key) => Math.max(+getValueByDataKey(d, key, 0), 0))
+    .value((d, key) => getValueByDataKey(d, key, 0))
```
sakulstra, line 751: while this indeed fixes my issues it reverts what was done in https://github.com/recharts/recharts/commit/e458cb59043f79626ff60d1ecf033c0562876a24#diff-3943890095cad7d9dc4dd87a63a1c157 to fix another issue. Will have to check why that is.

Perhaps @xile611 can enlighten me :sweat_smile:
I saw this change was implemented to fix https://github.com/recharts/recharts/issues/1667 but I don't actually understand the bug :thinking: Why shouldn't negative values be negative? How would you recommend me to move forward with this? I guess I could special case this to `offsetType === 'sign' ? getValueByDataKey(d, key, 0)) : Math.max(+getValueByDataKey(d, key, 0), 0)` - would this still solve he issue you were trying to solve?

sakulstra:
As this was a blocker for me I published a fork on `@sakulstra/recharts` containing this fix.

src/util/ChartUtils.ts
dbertella: I was thinking exactly the same to add this offsetType check before to do the Math.max, should we do that? I'm not an expert in d3 what does diverging do will it be the same?

xile611: Thanks, I'll add a new kind of `offsetType = 'positive'` to resolve #1667 . And revert [e458cb5](https://github.com/recharts/recharts/commit/e458cb59043f79626ff60d1ecf033c0562876a24#diff-3943890095cad7d9dc4dd87a63a1c157)

dbertella: Thanks a million, just tested on beta.8 and it works as expected.

PR #2310, opened 2020-10-29
allow react 17 as peer dependency

ranneyd:
React 17 doesn't have breaking changes, at least not for a project as new as v2 of this lib. Can't see why this would be a problem


avindra (approved):
This looks good. Should merge to unblock yarn users

Issue #2262, opened 2020-08-28
Remove core-js from dependencies


daveisfera:
- [X] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[Current package.json](https://github.com/recharts/recharts/blob/9d4ec579d290fe26e061c18b48e7176d21789f21/package.json#L58)

### Steps to reproduce
yarn install recharts

### What is expected?
`core-js` is not required by `recharts`

### What is actually happening?
`core-js` is installed when `recharts` is installed

| Environment | Info |
|---|---|
| Recharts | v1.8.5 |
| React | 16.13 |
| System | macOS 10.15.6 |
| Browser | Safari 13.1.2 |

---
`core-js` should be selected by the user of `recharts` when they use/bundle it and a specific version shouldn't be listed in `dependencies` of `recharts`

avindra:
#2321 open for removing **core-js**, as well as the vestigial polyfill module.

avindra:
ICYMI: #2321 is merged and released in beta 8:

https://github.com/recharts/recharts/releases/tag/v2.0.0-beta.8

The raw bundle size drops by 15KB after that:

https://bundlephobia.com/result?p=recharts@2.0.0-beta.8

Issue #2195, opened 2020-07-15
Upgrade to core-js@3


SpawnAtis:
Hi. Could u ,pls,  make new release with this changes [Migrate to core-js@3](https://github.com/recharts/recharts/commit/33350eae56532d7a5e28b7393977bef9f9fc8fcb#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) ? I can not upgrade my project to core-js@3 without it.

avindra:
You can pin recharts to the beta version that got released https://github.com/recharts/recharts/releases/tag/v2.0.0-beta.7

#2262 is tracking removing `core-js` as a dependency.

PR #2321, opened 2020-11-09
Remove core-js

avindra:
This fixes Issue #2262 and #2195 as it increases portability.

It addresses usage of the `number` import in `core-js` as a polyfill. The first (`math`) has already been fixed  by @xile611  in 2017.

Thus, the entire `polyfill` module is not needed, and it is removed.

## core-js/es/math

`Math` polyfill is removed because `Math.sign` was already ported in [560dc6f81dfe94d13be34db5fab557174655f3bd](https://github.com/recharts/recharts/commit/560dc6f81dfe94d13be34db5fab557174655f3bd) 

https://github.com/recharts/recharts/blob/d08e0cfad24140bd47340fd07724226dc7c4b45d/src/util/DataUtils.ts#L3-L12

## core-js/es/number

`Number.isFinite` is still in use, so ~I added the reference polyfill from MDN~. EDIT: updated to just fix  1 line to use [`isFinite`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite#Browser_compatibility) instead of qualified [`Number.isFinite`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite#Browser_compatibility) reference, which does not work in IE.

https://github.com/recharts/recharts/blob/d08e0cfad24140bd47340fd07724226dc7c4b45d/src/chart/generateCategoricalChart.tsx#L1599

**Note**: The package lockfile needs to be regenerated to reflect the removal of _core-js_. I can do this, but only using the new lock format in npm 7+.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
-    "core-js": "^3.4.2",
     "d3-interpolate": "^1.3.0",
     "d3-scale": "^3.1.0",
     "d3-shape": "^1.3.5",
@@ -79,7 +78,6 @@
     "@babel/plugin-proposal-function-bind": "^7.0.0",
     "@babel/plugin-proposal-object-rest-spread": "^7.6.2",
     "@babel/plugin-transform-runtime": "^7.6.2",
-    "@babel/polyfill": "^7.6.0",
     "@babel/preset-env": "^7.6.3",
     "@babel/preset-react": "^7.6.3",
     "@babel/preset-typescript": "^7.6.0",
--- a/package.json
+++ b/package.json
@@ -7,8 +7,7 @@
   "jsnext:main": "es6/index",
   "types": "types/index.d.ts",
   "sideEffects": [
-    "./*/index.js",
-    "./*/polyfill.js"
+    "./*/index.js"
   ],
   "files": [
     "*.md",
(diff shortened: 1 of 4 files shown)


src/polyfill.ts
```diff
@@ -1,16 +0,0 @@
-import 'core-js/es/math';
-import 'core-js/es/number';
-/* eslint no-proto: 0 */
-const testObject = {} as any;
-
-if (!(Object.setPrototypeOf || testObject.__proto__)) {
-  const nativeGetPrototypeOf = Object.getPrototypeOf;
-
-  Object.getPrototypeOf = object => {
```
avindra, line 9: I grep for `getPrototypeOf` and seems it is unused (as of latest `master` code).

So I just removed this code too.

package.json
```diff
@@ -7,8 +7,7 @@
   "jsnext:main": "es6/index",
   "types": "types/index.d.ts",
   "sideEffects": [
-    "./*/index.js",
-    "./*/polyfill.js"
+    "./*/index.js"
```
avindra, line 10: Can I remove `index.js` as well ? Not sure if it was added as having side effects because of the polyfill.

avindra (commented):
Questions for the reviewers.

package.json
xile611: Yes, I think so.

Issue #2330, opened 2020-11-27
"Glitches" appear when pie (donut) chart is animating, while rounded corners are set


bibixx:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://jsfiddle.net/akzc1wpu/](https://jsfiddle.net/akzc1wpu/)

### Steps to reproduce
1. Create a simple donut chart
2. Set the `cornerRadius` on the chart'
3. Create some data points that are relatively small
4. Run the chart so that it animates on start

### What is expected?
There should be no glitching

### What is actually happening?
There is some glitching visible (triangles going to the center of the chart)

| Environment | Info |
|---|---|
| Recharts | v1.8.5 |
| React | React 15.6.2 |
| System | MacOS 11.0.1 (Big Sur) |
| Browser | Google Chrome/Firefox/Brave Browser |

bibixx:
Fixed by #2331

PR #2331, opened 2020-11-30
Fixing Pie Charts rendering SVG Paths without inner arc commands

dillonreedy:
When the render functionality for the Pie component is generating its' sector SVG paths, if a developer has specified "corner radius" in the props of the Pie component, the SVG path won't include the inner arc commands in certain situations.

You can see this issue demonstrated [here](https://github.com/recharts/recharts/issues/2330), by setting the AnimationDuration prop on the pie component to a large value like "10000".

This commit will fix the jumping animation effects we see when rendering a Pie Chart with round corners and an inner radius defined.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "recharts",
-  "version": "2.0.0-beta.7",
+  "version": "2.0.0-beta.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


bibixx:
I'm have issued the issue mentioned in the PR and wanted to confirm that the fix seems to fix the problem ❤️ Thanks @dillonreedy!

bibixx:
@arcthur when could we expect this PR to be released?

PR #2347, opened 2020-12-21
WIP add correct typings for activeShape prop

fes300:


jmfrancois:
no yarn lock in this project please use npm

avindra:
@fes300  can you remove the checked-in `yarn.lock`?

Adding `yarn.lock` to `.gitignore` is probably a good idea.

PR #2358, opened 2020-12-23
Pass props to custom legend icon

juajang:
#### Pass props to custom legend icon
Example
```javascript
const data5 = [
  { value: 'Apple', color: '#ff7300', legendIcon: <CustomLegendIcon /> },
  { value: 'Samsung', color: '#bb7300', legendIcon: <CustomLegendIcon /> },
  { value: 'Huawei', color: '#887300', legendIcon: <CustomLegendIcon /> },
  { value: 'Sony', color: '#667300', legendIcon: <CustomLegendIcon /> },
];
```
#### Add custom legend icon demo
image


src/component/DefaultLegendContent.tsx
```diff
@@ -104,7 +106,8 @@ class DefaultLegendContent<TValue, TID> extends PureComponent<Props<TValue, TID>
       );
     }
     if (React.isValidElement(data.legendIcon)) {
-      return React.cloneElement(data.legendIcon);
+      const iconProps: any = _.omit(data, 'legendIcon');
+      return React.cloneElement(data.legendIcon, iconProps);
```
jmfrancois, line 110: please do not use lodash omit for that.
* it is slow
* it has been removed from lodash 5.x

consider this approach

```
      const iconProps: any = { ...data };
      delete iconProps.legendIcon;
```

juajang: Thank you for your review! I modified that.

Issue #2360, opened 2020-12-29
Failed to compile : "export 'Props' (reexported as 'AreaProps') was not found in './cartesian/Area'


bisnard29:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Steps to reproduce
1) update package.json from "recharts": "^2.0.0-beta.8" to "recharts": "^2.0.0".
2) npm build

### What is expected?
Build successfull 

### What is actually happening?
Creating an optimized production build...
Failed to compile.

./node_modules/recharts/es6/index.js
"export 'Props' (reexported as 'AreaProps') was not found in './cartesian/Area'

| Environment | Info |
|---|---|
| Recharts | v2.0.0 |
| React | 16.14.0 |
| System | Ubuntu 20.04 |
| Browser | Chrome |

---
I had the issue both on Linux and Windows. Re-updating to  "recharts": "^2.0.0-beta.8" worked on my windows instance (not on linux one)

Lappintade:
Also has this issue on OSX.

goelkinshuk:
I also got this and i just installed recharts in my project . what's the fix ?

studiojms:
I am having the same issue with recharts 2.0.0 and react 17.0.1

mehrdadrafiee:
having the same issue here. ```react 17.0.1``` and ```recharts 2.0.0-beta.5```. I also tested on ```recharts 2.0.0``` and got the same error.

tactycadmin:
Same issue here  with `recharts 2.0.0-beta.6 `and `recharts 2.0.0-beta.8` as well.

avindra:
What version of Typescript are you guys using? I pulled `master` and the project still builds. At the moment recharts is using:

```sh
❯ tsc --version
Version 3.9.7
```

studiojms:
Not using typescript here, the project is only js

Skitelman:
This appears to be the problem commit https://github.com/recharts/recharts/commit/62636887336f1267017e06a3d23b89ec69066a0a#diff-1496ed5301e625b1e310643256ed6b2af380dd919af777f25d65beecac8f7b22

gruvin:
Same issue -- not using TypeScript. MacOS 10.15, React 16.13.1. Thanks ;-)

barbalex:
I just updated to v2.0.0 and am getting a lot more of these warnings in dev:

```
[...]

PR #2361, opened 2020-12-29
[1/3] Bump devel, migrate CI

avindra:
Refreshing devel:

 * This moves the CI integration to github actions including xvfb based browser test
   - See builds here for now: https://github.com/avindra/recharts/actions
   - Increase visibility of build status on PRs and commit
   - Note :information_source:  builds will not be passing until the other PRS that were split are merged.
 * Update development and testing dependencies
 * eslint
   - update to eslint 7 and prettier 2
   - `eslint --fix` run to re-establish "Gold" eslint baseline
   - harden module rules, including the banning of default exports
 * remove babel decorator plugin
   - only used [in removed 2016 code](https://github.com/recharts/recharts/commit/c6e17106af8c3ce0df759d4ed9db525c9801d421)
 * webpack 4 -> 5
   - recently unblocked by [`karma-webpack v5.0.0-alpha.5`](https://github.com/ryanclark/karma-webpack/releases/tag/v5.0.0-alpha.5)

Note: development now requires at least Node 14 (which is [latest LTS](https://nodejs.org/en/)).

-------

See also the following PRs that were split:

 * PR #2366 for lint fixes where there is no autofixer
 * PR #2367 for the recharts 2.0 release issue
   - also contains naming of all exports


src/component/Legend.tsx
```diff
@@ -50,7 +50,7 @@ export type Props<TValue, TID> = DefaultProps<TValue, TID> & {
     right?: number;
   };
   payloadUniqBy?: UniqueOption<TValue, TID>;
-  onBBoxUpdate?: (box: ClientRect | DOMRect | null) => void;
+  onBBoxUpdate?: (box: DOMRect | null) => void;
```
avindra, line 53: FYI: `ClientRect` seems to be a pre-standard version of the canonical `DOMRect`:

https://bugs.chromium.org/p/chromium/issues/detail?id=719246

.eslintrc
```diff
@@ -85,9 +85,15 @@
     "react/jsx-props-no-spreading": "off",
     "react/state-in-constructor": "off",
     "react/static-property-placement": "off",
+    "arrow-parens": ["error", "always"],
+
+    "no-use-before-define": "off",
+    "@typescript-eslint/no-use-before-define": ["error"],
```
avindra, line 91: See https://stackoverflow.com/a/64024916/270302

.eslintrc
```diff
@@ -85,9 +85,15 @@
     "react/jsx-props-no-spreading": "off",
     "react/state-in-constructor": "off",
     "react/static-property-placement": "off",
+    "arrow-parens": ["error", "always"],
```
avindra, line 88: This makes the linter compatible with the new Prettier 2.0 default

.eslintrc
```diff
@@ -85,9 +85,15 @@
...
+
+    "no-use-before-define": "off",
+    "@typescript-eslint/no-use-before-define": ["error"],
 
     "@typescript-eslint/explicit-function-return-type": "off",
-    "@typescript-eslint/no-explicit-any": "off"
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/explicit-module-boundary-types": "off",
+    "@typescript-eslint/ban-types": "off",
```
avindra, line 96: `@typescript-eslint/ban-types` is complaining about all existing types that are just `Function`.

.eslintrc
```diff
@@ -85,9 +85,15 @@
...
     "react/static-property-placement": "off",
+    "arrow-parens": ["error", "always"],
+
+    "no-use-before-define": "off",
+    "@typescript-eslint/no-use-before-define": ["error"],
 
     "@typescript-eslint/explicit-function-return-type": "off",
-    "@typescript-eslint/no-explicit-any": "off"
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/explicit-module-boundary-types": "off",
```
avindra, line 95: This seems like a good rule to enable, but the codebase has `390` instances it it currently. I Would recommend someone try to fix these in a separate PR.

src/cartesian/ZAxis.tsx
```diff
@@ -20,9 +21,7 @@ export interface Props {
   domain?: AxisDomain;
 }
 
-function ZAxis(props: Props): any {
```
avindra, line 23: In general, the types should be safer now. The `FunctionComponent` return type is not defined as `any`, but rather all the types that are valid as a React component.

src/polar/PolarRadiusAxis.tsx
```diff
@@ -9,11 +9,6 @@ import Layer from '../container/Layer';
 import { polarToCartesian } from '../util/PolarUtils';
 import { PresentationAttributes, filterProps, BaseAxisProps, TickItem, adaptEventsOfChild } from '../util/types';
 
-interface TickIem {
```
avindra, line 12: Unused interface. The correct version `TickItem` is already being imported, so can remove this one.

src/util/ReactUtils.ts
```diff
@@ -19,8 +19,6 @@ const REACT_BROWSER_EVENT_MAP: any = {
   touchstart: 'onTouchStart',
 };
 
-type eventFunc = (data: any, index: number, e: any) => {};
```
avindra, line 22: Unused type

avindra (commented):
Breadcrumbs for reviewer ( @xile611 )

src/util/ChartUtils.ts
```diff
@@ -759,7 +758,6 @@ export const offsetPositive = (series: any) => {
       } else {
         series[i][j][0] = 0;
         series[i][j][1] = 0;
-        negative = series[i][j][1];
```
avindra, line 762: Also not used. Probably copied from a previous version like `offsetNegative` or somewhere else.

.eslintrc
```diff
@@ -27,7 +27,8 @@
         "singleQuote": true,
         "tabWidth": 2,
         "useTabs": false,
-        "trailingComma": "all"
+        "trailingComma": "all",
+        "arrowParens": "avoid",
```
avindra, line 31: Prettier 2.0 changed this rule: https://prettier.io/docs/en/options.html#arrow-function-parentheses

This puts it back to the 1.0 setting so we don't have to change any of the parens right now.

avindra:
@xile611 added a fix for #2360 . Lmk if you want to see any of these commits on a separate PR, I can split if needed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Install dependencies
+      run: npm install
+
+    - name: Test
+      run: |
+        tsc --noEmit
```
avindra, line 29: Added typescript check to CI, which should help catch extra problems not detected in linter or unit testing.

avindra:
@xile611 / @arcthur  I moved the commits where I had to manually fix the linter problems to #2366

webpack.config.js
```diff
@@ -32,9 +32,6 @@ const config = {
         ],
         use: {
           loader: 'babel-loader',
-          query: {
```
avindra, line 35: `query` does not exist in webpack 5. Furthermore, the lodash plugin is already configured in `.babelrc` so we do not need to re-iterate it in the webpack config.

Issue #2363, opened 2020-12-30
Remove  console.log


QoobIY:
https://github.com/recharts/recharts/blob/98ddefb28e057a5ca8c01504ae11330e72eca91f/src/component/Text.tsx#L88

Remove  console.log, please :)

fengpei0114:
recharts/src/component/Text.tsx: Line 172 
`console.log(textProps.x, textProps.y, this.props.children, wordsByLines, this.props);`


Please remove this console.log, too :)

avindra:
Included this on #2361

avindra:
Fixed in 984c7d867ee6eb1335804cd0e6110392324c41ae

(Release pending)

PR #2366, opened 2020-12-31
[2/3] Fix/squelch problems reported by linter

avindra:
This PR was split from #2361 to help make reviewing a little easier. It addresses unused variables, `console.log`, an unused type and a legacy type.

 * Manually address lint warnings without autofixers
 * remove console.log (#2363)
 * Remove `ClientRect` (causes errors after doing upgrades in #2361)
 * Remove a few unused variables and types

After these changes, eslint reports 0 errors and 0 warnings. (Previously 12 warnings).

----

 :information_source: reviewer note: This should be merged _before_ the third and final PR #2367 which converges on the module definition.

### DIFF
--- a/src/cartesian/XAxis.tsx
+++ b/src/cartesian/XAxis.tsx
@@ -1,6 +1,7 @@
 /**
  * @fileOverview X Axis
  */
+import type { FunctionComponent } from 'react';
 import { BaseAxisProps, AxisInterval, PresentationAttributes } from '../util/types';
 
 /** Define of XAxis props */
@@ -30,9 +31,7 @@ interface XAxisProps extends BaseAxisProps {
 
 export type Props = PresentationAttributes<SVGElement> & XAxisProps;
 
-function XAxis(props: Props): any {
-  return null;
-}
+const XAxis: FunctionComponent<Props> = () => null;
 
 XAxis.displayName = 'XAxis';
 XAxis.defaultProps = {
--- a/src/component/Legend.tsx
+++ b/src/component/Legend.tsx
@@ -50,7 +50,7 @@ export type Props<TValue, TID> = DefaultProps<TValue, TID> & {
     right?: number;
   };
   payloadUniqBy?: UniqueOption<TValue, TID>;
-  onBBoxUpdate?: (box: ClientRect | DOMRect | null) => void;
+  onBBoxUpdate?: (box: DOMRect | null) => void;
 };
 
 interface State {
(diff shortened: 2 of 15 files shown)


src/cartesian/XAxis.tsx
```diff
@@ -30,9 +31,7 @@ interface XAxisProps extends BaseAxisProps {
 
 export type Props = PresentationAttributes<SVGElement> & XAxisProps;
 
-function XAxis(props: Props): any {
-  return null;
-}
+const XAxis: FunctionComponent<Props> = () => null;
```
avindra, line 34: `function Component(props: Props): any`  is unsafe, and also introduces an unused variable, `props` which emits linter warnings.

The change to use the native `FunctionComponent` type is safer, since its return type is not defined as `any`, but rather all the types that are valid as a React component.

src/component/Legend.tsx
```diff
@@ -50,7 +50,7 @@ export type Props<TValue, TID> = DefaultProps<TValue, TID> & {
     right?: number;
   };
   payloadUniqBy?: UniqueOption<TValue, TID>;
-  onBBoxUpdate?: (box: ClientRect | DOMRect | null) => void;
+  onBBoxUpdate?: (box: DOMRect | null) => void;
```
avindra, line 53: FYI: [`ClientRect` seems to be a pre-standard version of the canonical `DOMRect`](https://stackoverflow.com/a/65501184/270302).

Users who are referring to `ClientRect` probably need to change to `DOMRect` as well.

src/polar/PolarRadiusAxis.tsx
```diff
@@ -9,11 +9,6 @@ import Layer from '../container/Layer';
 import { polarToCartesian } from '../util/PolarUtils';
 import { PresentationAttributes, filterProps, BaseAxisProps, TickItem, adaptEventsOfChild } from '../util/types';
 
-interface TickIem {
```
avindra, line 12: Unused. The correct `TickItem` is already being imported, so we can just delete this one.

src/util/ChartUtils.ts
```diff
@@ -759,7 +758,6 @@ export const offsetPositive = (series: any) => {
       } else {
         series[i][j][0] = 0;
         series[i][j][1] = 0;
-        negative = series[i][j][1];
```
avindra, line 762: `negative` is unused and is probably leftover code copied from other functions.

src/util/ReactUtils.ts
```diff
@@ -19,8 +19,6 @@ const REACT_BROWSER_EVENT_MAP: any = {
   touchstart: 'onTouchStart',
 };
 
-type eventFunc = (data: any, index: number, e: any) => {};
```
avindra, line 22: Not used.

avindra (commented):
Moved review pointers from #2361

Issue #2368, opened 2021-01-01
React 17


cujarrett:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Is there a roadmap to update the peerDependency to react 17 (released in October 2020).

### What does the proposed API look like?
N/A

rafipiccolo:
could u change this in the package.json to allow react v17 ?

i can confirm that it works for me with react 17, if i force install, but it would be better to just install without force.
```
"peerDependencies": {
    "react": "^16.0.0",
    "react-dom": "^16.0.0"
},
```
it could be

```
"react": "^16.0.0 || ^17.0.0",
"react-dom": "^16.0.0 || ^17.0.0"
```

rafipiccolo:
pull request done. could you merge it please ?

PR #2369, opened 2021-01-01
Update package.json

rafipiccolo:
allow react 17
see #2368

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,8 +49,8 @@
   },
   "homepage": "https://github.com/recharts/recharts",
   "peerDependencies": {
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "react": "^16.0.0 || ^17.0.0",
+    "react-dom": "^16.0.0 || ^17.0.0"
   },
   "dependencies": {
     "classnames": "^2.2.5",


package.json
```diff
@@ -49,8 +49,8 @@
   },
   "homepage": "https://github.com/recharts/recharts",
   "peerDependencies": {
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "react": "^16.0.0 || ^17.0.0",
+    "react-dom": "^16.0.0 || ^17.0.0"
```
cujarrett, line 53: Should it be this?

```js
    "react": "^17.0.0",
    "react-dom": "^17.0.0"
    ```

Comment (cujarrett) [reply package.json:53]: Should the package version also be bumped?

```js
  "name": "recharts",
  "version": "2.0.1",
  ...
```

cujarrett (commented):
Should it just mention `^17.0.0`?

package.json
rafipiccolo: As you wish.
- Allowing 16 & 17 means that the package will accept both instead of forcing only one. React 16 and 17 are compatible en theory so I feel you shouldn't bother playing with maintaining 2 versions of recharts
- the version needs a bump but major ? Minor ?

cujarrett: I think you’re right, ignore my 16/17 comment. Thanks!

rafipiccolo: Worth noting that some dependencies of recharts need the same annoying patch for the installation to work without warnings.

avindra (approved):
:+1:

PR #2376, opened 2021-01-07
Add aria-hidden to measurementSpan

arranfw:
This span should not be available to screen readers as it provides irrelevant information as well as potentially causes massive layout shifts.


avindra (approved):
lgtm

PR #2377, opened 2021-01-09
fix(package.json): disable side effects explicitly

Fonger:
Somehow sideEffects config is removed. I guess this is when we remove the polyfill array .
The default is `true` or `true for array of file paths` so we need to explicitly specify `false` to make tree shaking work

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "module": "es6/index",
   "jsnext:main": "es6/index",
   "types": "types/index.d.ts",
+  "sideEffects": false,
   "files": [
     "*.md",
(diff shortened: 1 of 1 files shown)


avindra (approved):

avindra:
Yes, this is the way:

https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

However, after upgrading to Webpack 5 (#2361), it is not necessary to specify it manually:

https://webpack.js.org/blog/2020-10-10-webpack-5-release/#side-effect-analysis

Issue #961, opened 2017-10-22
Word-wrap overflown X-axis labels


James2516:
https://github.com/recharts/recharts/issues/198

This is supposedly fixed by https://github.com/recharts/recharts/pull/243 but it doesn't seem to work.

How do I use it? `<XAxis dataKey="name" interval={0} width={30} label={<Text width={30} />} />`?
@techniq

techniq:
I haven't used recharts in a while, and have [extracted](https://github.com/techniq/react-svg-text) out the `<Text />` component to use in other projects (like [vx](https://github.com/hshoff/vx)) but all that is required to have `<Text />` wrap is an explicit width like you have.  I don't know how best to integrate Text with an axis but maybe you are supposed to pass props directly to it, such as:

```js
<XAxis dataKey="name" interval={0} width={30} label={props => <Text width={30}>{props.text}</Text>} />
```

but I'm just guessing.  I took a quick peak at the source of [XAxis](https://github.com/recharts/recharts/blob/master/src/cartesian/XAxis.js) but didn't see how it uses the label prop.

James2516:
Doesn't work unfortunately...

havietduc91:
> Doesn't work unfortunately...

Hi James2516, I just only use `width` param for XAxis or YAxis and it's working fine, maybe it helps you!
![screenshot 3](https://user-images.githubusercontent.com/4966192/46184312-454e2b80-c2ff-11e8-8267-52628999acf6.png)

Please click the bellow link to know more!
https://jsfiddle.net/vuh6cLyz/3/

lisamartin00:
I was finally able to figure it out based on their Line Chart Example fiddle, which uses a custom component passed into the `<XAxis> tick` property.
https://jsfiddle.net/alidingling/5br7g9d6/

You can do something similar to what they've done and return the `<Text>` element
https://jsfiddle.net/lisamartin00/exasjofp/
Now you can set a fixed width and the text will wrap.  You may still need to write some custom media queries to adjust the `font-size` for smaller screen widths, e.g.
` @media screen and (max-width: 480px){
.recharts-cartesian-axis-ticks {
      font-size: 10.5px;
    }
`

deadlyfingers:
If you need word wrap on long string or word which has no spaces I have a fiddle which shows how to chunk long text strings and splits it up into multiple `tspan` elements. https://jsfiddle.net/deadlyfingers/jngsxqyo/77/

bertho-zero:
[...]

PR #2380, opened 2021-01-11
[wip] karma -> jest

avindra:
:construction:  WIP (don't merge yet). Opening PR now for visibility + builds.

This PR, for the sake of portability is:

 * moving the tests from Karma into jest
 * removing `enzyme` for testing React

----
## Progress

If you'd like to help, name a [folder from the set of test cases](https://github.com/recharts/recharts/tree/v2.0.3/test/specs) you want to port so everyone else knows to avoid that one.

- [ ] cartesian
- [ ] chart
- [ ] component
- [x] numberAxis
- [ ] polar
- [ ] shape
- [x] util

------
## Toolchain changes

Note to other contributors: continue test using the existing toolchain. Any of the assertion statements are trivial to convert to Jest, and Jest supports the exact BDD pattern that we were using with karma. Coverting `enzyme` and `sinon` uses is less trivial.

For testing without enzyme, see:

 * https://reactjs.org/docs/testing-recipes.html
 * https://reactjs.org/docs/test-utils.html

For sinon replacements, see:

 * https://jestjs.io/docs/en/manual-mocks
 * https://jestjs.io/docs/en/jest-object#jestspyonobject-methodname

PR #2387, opened 2021-01-12
add maxLines prop to Text component

Morantron:
Taking https://github.com/recharts/recharts/issues/961#issuecomment-618265830 as inspiration, I've added a `maxLines` props to the `<Text />` component

image

# TODO

- [ ] Update docs

### DIFF
--- a/src/component/Text.tsx
+++ b/src/component/Text.tsx
@@ -76,6 +76,7 @@ const calculateWordsByLines = (
   scaleToFit?: boolean,
 ): Array<Words> => {
   const shouldLimitLines = isNumber(props.maxLines);
+  const text = props.children as string;
 
   const calculate = (words: Array<WordWithComputedWidth> = []) =>
     words.reduce((result, { word, width }) => {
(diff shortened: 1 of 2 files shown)


src/component/Text.tsx
```diff
@@ -67,26 +68,56 @@ interface State {
...
+  if (!shouldLimitLines) {
     return result;
-  }, []);
+  }
+
+  let tempText = props.children as string;
+  const suffix = '…';
+
+  while (result.length > props.maxLines || findLongestLine(result).width > lineWidth) {
+    tempText = tempText.slice(0, -1);
```
xile611, line 110: How about use binary search algorithm to find the right result?

Morantron: good idea, i'll give a try :+1:

Morantron:
Added binsearch approach. Might be cool in the future to extract all these helper functions to a TextUtils.ts file and test them independently. Maybe after https://github.com/recharts/recharts/pull/2380 is merged?

avindra:
#2380 may take long (1-3mo if I am working on it alone). If you want to write the test now, I'll convert it later on. If you can test without enzyme that'd be :ok_hand: but if not, nbd

PR #2397, opened 2021-01-15
Add `payload` to `Payload` interface.

Semigradsky:
See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0ebb554617399eadb906c4262823bf9f3fcbfa6f/types/recharts/index.d.ts#L925


Semigradsky:
@xile611 Shouldn't the types have been re-generated?
This change was added to `src/...` but not to `types/...`
The same with #2408

Semigradsky:
Related to #2410 I think

PR #2406, opened 2021-01-20
fix(ChartUtils): prevent rerender errors in ResponsiveContainer

MarcusNotheis:
default ticks array of `calculateActiveTickIndex` to an empty array to prevent `Cannot read property length of undefined` error.

I'm doubt that this really fixes the root cause, but as from my tests it prevented the error and in the next render cycle the chart looked like expected.

Fixes #2385 
Fixes #2402

PR #2409, opened 2021-01-24
Bugfix for undefined ticks

AnotherCodeArtist:
Hi!

I ran into a problem using a bar chart within a `<ResponsiveContainer  ... />` that was in turn inside a tab. It turned out that `ticks` was undefined, which caused an exception in `calculateActiveTickIndex`. The changes provided by this PR fixed this problem for me. Would be great if you would pull it ASAP and publish a new version.


ckifer:
#2406 already fixes this issue though it can't hurt to double check

PR #2413, opened 2021-01-26
Update Examples section in the README, and typing state and setState

derekvelzy:
### README ###
Added more clarity to the Examples section of the README.

### /component ###
Included typing of the state and setState properties of ResponsiveContainer and Legend


xile611:
@derekvelzy please fix the error and reopen an PR? And I can't understand why we need a setState statement? Is that supported by react Component.

PR #2422, opened 2021-02-05
Add activeChild to tooltip info to identify the specific item being hovered

daveisfera:
There might be a better way to get this info, but the point is to have the information to identify the specific item that's hovered over, so information about just that piece can be shown in the tooltip rather than the information about the entire bar.

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -239,8 +239,8 @@ const getTooltipData = (state: CategoricalChartState, chartData: any[], layout: 
     const activePayload = getTooltipContent(state, chartData, activeIndex, activeLabel);
     const activeCoordinate = getActiveCoordinate(layout, ticks, activeIndex, rangeData);
 
-    const { x: mouseX, y: mouseY } = rangeObj;
-    const getActiveEntry = (o) => _.get(o, ['props', 'data', activeIndex]);
+    const { x: mouseX, y: mouseY } = rangeData;
+    const getActiveEntry = (o: object) => _.get(o, ['props', 'data', activeIndex]);
     let activeChild = _.find(formatedGraphicalItems, (entry) => {
         const activeEntry = getActiveEntry(entry);
         if (activeEntry) {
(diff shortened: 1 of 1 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -239,11 +239,31 @@ const getTooltipData = (state: CategoricalChartState, chartData: any[], layout:
     const activePayload = getTooltipContent(state, chartData, activeIndex, activeLabel);
     const activeCoordinate = getActiveCoordinate(layout, ticks, activeIndex, rangeData);
 
+    const { x: mouseX, y: mouseY } = rangeData;
+    const getActiveEntry = (o: object) => _.get(o, ['props', 'data', activeIndex]);
+    let activeChild = _.find(formatedGraphicalItems, entry => {
+      const activeEntry = getActiveEntry(entry);
+      if (activeEntry) {
+        const { x, width, y, height } = activeEntry;
```
xile611, line 247: this will only works for rectangle coordinate charts, such as line, bar.
How about pie-chart? radial-bar-chart?

daveisfera: We currently don't need this data for pie-chart and radial-bar-chart, because there's not a child/sub-element that we need to know about for the tooltip, so that's why it's not an issue for our use case, but if there's a better way to handle this, then I'd definitely be open to using it instead.

KlonD90 (approved):

src/chart/generateCategoricalChart.tsx
xile611: Can you show more detail about your case?

daveisfera: The current tooltip on a stacked bar chart shows all of the values and we'd like to show just the value being hovered over. Currently, there's not a way to tell which one is being hovered over and the intent of this change is to provide that information. If there's a better way to accomplish that, then we're definitely open to that, but this was one approach to doing that.

xile611: OK! Got it! I will think about this feature.

daveisfera:
Is there an example of how to use the new feature that was added? Also, is there ETA on when it will be released?

PR #2426, opened 2021-02-08
Fix lint err and warnings

avindra:
Address build failure on caught lint issues.

### DIFF
--- a/src/component/Legend.tsx
+++ b/src/component/Legend.tsx
@@ -2,7 +2,7 @@
  * @fileOverview Legend
  */
 import React, { PureComponent, CSSProperties } from 'react';
-import _ from 'lodash';
+import { isFunction, uniqBy } from 'lodash';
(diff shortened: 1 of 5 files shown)


src/component/Legend.tsx
```diff
@@ -27,13 +27,14 @@ function renderContent(content: ContentType, props: Props) {
...
     return React.cloneElement(content, props);
   }
-  if (_.isFunction(content)) {
+  if (isFunction(content)) {
     return React.createElement(content as any, props);
   }
 
-  const { ref, ...otherProps } = props;
+  const propsWithoutRef = { ...props };
+  delete propsWithoutRef.ref;
```
avindra, line 35: The other way exposes a `ref` variable that is unused. This approach does the same thing without creating a dangling reference.

avindra:
```
Error:   39:36  error  Replace `propsWithoutRef·as·PropsMinusRef` with `(propsWithoutRef·as·PropsMinusRef)`  prettier/prettier
```

WTF. I give up.

PR #2436, opened 2021-02-13
fix: filter out nil elements

### LINKED ISSUE #300: Chart keeps redrawing
Not sure if this is a bug or perhaps something in my code, but I I'm using the ResponsiveContainer and AreaChart and if I update the state of the parent component or even click on a menu above the chart, it redraws the chart.  The data is static, so it's not changing anything.  Any thoughts?

franciscofabian:
Filter out nil elements in `renderByOrder`. Fixes #300 happening again since v2.0.0-beta.8

### DIFF
--- a/src/util/ReactUtils.ts
+++ b/src/util/ReactUtils.ts
@@ -337,7 +337,7 @@ export const renderByOrder = (children: React.ReactElement[], renderMap: any) =>
     }
   });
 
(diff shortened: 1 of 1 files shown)


xile611:
@franciscofabian  Why use `_.flattenDeep`？ can you show the case ?

franciscofabian:
@xile611 you're right, that is not really necessary. I don't have any use case where that is needed. I have reverted that change.

Issue #2403, opened 2021-01-18
Bar component goes out of char lines when used inside ComposedChart


ovidijusr:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-90xxt?file=/package.json)

### Steps to reproduce
- Create a Bar component and wrap around with ComposedChart
- Use 2.0.3 version of recharts
- bars overlaps the chart
- Change version to 1.8.5
- bars does not overlap

### What is expected?
The most left and right bars should not go out of chart, but stay inside.
(like in 1.8.5 or wrapping in BarChart)

### What is actually happening?
The most left and right bars goes out of chart.

| Environment | Info |
|---|---|
| Recharts | v2.0.3 |
| React | 16.13.1 |
| System | Mac os big sur |
| Browser | Google chrome 87 |

datnguyen14:
Same issue with me.
![image](https://user-images.githubusercontent.com/67571368/104978360-46f20f00-5a34-11eb-8f2c-174cc0d42d59.png)

ckifer:
@ovidijusr @datnguyen14 
You can add padding to either side of the XAxis for now, but that will also move your lines. Agreed that this needs addressed.
`<XAxis dataKey="name" padding={{ right: 60, left: 60 }} />`

ovidijusr:
Yeah, but the issue is when you have different data on same graph, then the margins needs to be different

badbod99:
Same issue here, but with Composed chart with vertical bar.

![image](https://user-images.githubusercontent.com/7858558/107933057-2c994b80-6f76-11eb-89c8-a4e51adc4bc8.png)

KlonD90:
I research this bug. It related to parseScale function and ComposedChart type. What's the right way to fix it? @xile611

KlonD90:
Possibile temporarily decision add to XAxis: scale="band"

PR #2448, opened 2021-02-17
Fix for ComposedChart + Bar overflowing width

KlonD90:
Use 'band' scale for composed chart.

Related to: https://github.com/recharts/recharts/issues/2403


xile611:
@KlonD90 Thanks for your PR. Maybe we should return 'band' for ComposedChart when the chart has <Bar />, if not so, return 'point'.

KlonD90:
Ok i'll try it today

PR #2449, opened 2021-02-17
Chore: use npm ci in gha

KlonD90:
Use `npm ci` in GHA for increase test immutability.


avindra (approved):
Huh TIL. Lgtm

https://docs.npmjs.com/cli/v7/commands/npm-ci

Issue #2091, opened 2020-04-08
Compute padding for bar chart on a continuous (number) x-axis


chilversc:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
I have a combination bar/line chart. The line is plotted using a continuous series (as it may not have data at discrete intervals). There is also categorical data that is being displayed using bars.

The component is hosted in a ResponsiveContainer, and the bar's width is auto-calculated. As such I cannot find a simple way to calculate the required padding for the x-axis to avoid the bars overlapping either end of the chart.

There should be an option to calculate the correct padding to ensure the first/last bar on the chart does not overlap the Y axis, or the end of the chart. The exact amount of padding needed would vary depending on the total width of the bars, and how far from the edge the first/last are (it is assumed the bars will lie inside the domain).

It would be nice to include two options for this padding:
* Aligns the edges of the bars with the edges of the graph (no gap).
* Add the inter-bar gap to the padding to allow some space between the edge of the bar and the edge of the graph.

### What does the proposed API look like?
```
<XAxis padding="auto"/>
<XAxis padding="auto-no-gap"/>
```

### Example of the issue

https://codesandbox.io/s/rechart-bar-padding-37o7u?file=/src/App.js

![image](https://user-images.githubusercontent.com/145440/78815185-ccd22b80-79c7-11ea-900d-e291986fddd8.png)

harveyconnor:
Great idea! Have you had any luck with this?

monmush:
I have the same problem. Is there anyone currently working on this feature?

PR #2457, opened 2021-02-22
Feature Request: Allow automated axis padding for "gap" and "no-gap" for barcharts with continuous axis

anselmpaul:
Trying to solve #2091
Here's what it looks like:
image

### DIFF
--- a/src/util/CartesianUtils.ts
+++ b/src/util/CartesianUtils.ts
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import { getTicksOfScale, parseScale, checkDomainOfScale, getBandSizeOfAxis } from './ChartUtils';
 import { findChildByType } from './ReactUtils';
 import { Coordinate, AxisType } from './types';
+import { getPercentValue } from './DataUtils';
 
 /**
  * Calculate the scale function, position, width, height of axes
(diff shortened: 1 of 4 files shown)


xile611:
@anselmpaul  Thanks for your awesome work. Would you please add some test cases for this feature?

src/util/CartesianUtils.ts
```diff
@@ -32,15 +32,43 @@ export const formatAxisMap = (props: any, axisMap: any, offset: any, axisType: A
     const { orientation, domain, padding = {}, mirror, reversed } = axis;
     const offsetKey = `${orientation}${mirror ? 'Mirror' : ''}`;
 
-    let range, x, y, needSpace;
+    let calculatedPadding, range, x, y, needSpace;
+
+    if (axis.padding === 'gap' || axis.padding === 'no-gap') {
+      const diff = domain[1] - domain[0];
+      let smallestDistanceBetweenValues = Infinity;
+      const sortedValues = axis.categoricalDomain.sort();
```
xile611, line 40: There is an assumption here, the type of `axis` must be `number` when we `axis.padding === 'gap' || axis.padding === 'no-gap'`?

src/util/CartesianUtils.ts
```diff
@@ -32,15 +32,43 @@ export const formatAxisMap = (props: any, axisMap: any, offset: any, axisType: A
...
+      sortedValues.forEach((value: number, index: number) => {
+        smallestDistanceBetweenValues = Math.min(
```
xile611, line 42: calculate `smallestDistanceBetweenValues` only when `index > 0` ?

src/util/CartesianUtils.ts
```diff
@@ -32,15 +32,43 @@ export const formatAxisMap = (props: any, axisMap: any, offset: any, axisType: A
...
+      });
+      const smallestDistanceInPercent = smallestDistanceBetweenValues / diff;
+      const rangeWidth = axis.layout === 'vertical' ? offset.top + offset.height : offset.width + offset.left;
+
+      if (axis.padding === 'gap') {
+        calculatedPadding = (smallestDistanceInPercent * rangeWidth) / 2;
+      }
+
+      if (axis.padding === 'no-gap') {
+        calculatedPadding = smallestDistanceInPercent * rangeWidth * 0.35;
```
xile611, line 55: why is `0.35` ?

src/util/CartesianUtils.ts
anselmpaul: fixed

src/util/CartesianUtils.ts
anselmpaul: fixed

src/util/CartesianUtils.ts
anselmpaul: refactored to be calculated correctly (0.35 was just a guess)

anselmpaul:
> @anselmpaul Thanks for your awesome work. Would you please add some test cases for this feature?

I've looked at the test suite (I've never written enzyme tests before) and it looks like you can only test _what_ is rendered (e.g. count of rendered elements) and not _how_ it is rendered (width of element). Since these changes only affect the width, how would you test this?

xile611:
> > @anselmpaul Thanks for your awesome work. Would you please add some test cases for this feature?
> 
> I've looked at the test suite (I've never written enzyme tests before) and it looks like you can only test _what_ is rendered (e.g. count of rendered elements) and not _how_ it is rendered (width of element). Since these changes only affect the width, how would you test this?

You can check the coordinate of bar in test cases.

src/util/CartesianUtils.ts
```diff
@@ -32,15 +33,47 @@ export const formatAxisMap = (props: any, axisMap: any, offset: any, axisType: A
     const { orientation, domain, padding = {}, mirror, reversed } = axis;
     const offsetKey = `${orientation}${mirror ? 'Mirror' : ''}`;
 
-    let range, x, y, needSpace;
+    let calculatedPadding, range, x, y, needSpace;
+
+    if (axis.type === 'number' && (axis.padding === 'gap' || axis.padding === 'no-gap')) {
+      const diff = domain[1] - domain[0];
+      let smallestDistanceBetweenValues = Infinity;
```
xile611, line 40: `axis.categoricalDomain.length === 1` can be a special case here.

src/util/CartesianUtils.ts
```diff
@@ -32,15 +33,47 @@ export const formatAxisMap = (props: any, axisMap: any, offset: any, axisType: A
...
+          smallestDistanceBetweenValues = Math.min(
+            (value || 0) - (sortedValues[index - 1] || 0),
+            smallestDistanceBetweenValues,
+          );
+        }
+      });
+      const smallestDistanceInPercent = smallestDistanceBetweenValues / diff;
+      const rangeWidth = axis.layout === 'vertical' ? offset.height : offset.width;
+
+      if (axis.padding === 'gap') {
```
xile611, line 53: `smallestDistanceBetweenValues` should not be `Infinity`

src/util/CartesianUtils.ts
```diff
@@ -32,15 +33,47 @@ export const formatAxisMap = (props: any, axisMap: any, offset: any, axisType: A
...
+        calculatedPadding = (smallestDistanceInPercent * rangeWidth) / 2;
+      }
+
+      if (axis.padding === 'no-gap') {
+        const gap = getPercentValue(props.barCategoryGap, smallestDistanceInPercent * rangeWidth);
```
xile611, line 58: `smallestDistanceBetweenValues` should not be `Infinity`

src/util/CartesianUtils.ts
anselmpaul: This is the same behaviour as for ticks, how would you avoid this?

src/util/CartesianUtils.ts
anselmpaul: This is the same behaviour as for ticks, how would you avoid this?

anselmpaul:
Sorry for the delay. I've finally added some tests.

src/util/CartesianUtils.ts
anselmpaul: Can you elaborate? In case of `axis.categoricalDomain.length === 1` the `smallestDistanceBetweenValues` will be the value of the single element

src/util/CartesianUtils.ts
xile611: just let `calculatedPadding`  to be 0 when `smallestDistanceBetweenValues  === Infinity `

PR #2459, opened 2021-02-22
TypeScript fix for allowing AreaChart to be valid child of ResponsiveContainer

tylertyssedal:
AreaChart was not a valid child of ResponsiveContainer and would fail to compile in a TypeScript project. Changing the Props interface from `children: ReactElement;` to `children: ReactElement | ReactElement[];` fixes this.


tylertyssedal:
You know what, scratch this. It turns out it that it was somehow trying to process the commented code as a child... so I removed the comments and it builds.

PR #2466, opened 2021-02-24
Fix parameters passed to syncMethod callback function

almudena-vila-cognito:
If syncMethod is a function, the callback function should get access to tooltipTicks and data.


gcourtemanche (approved):
CI errors should be fixed by this https://github.com/recharts/recharts/pull/2465

almudena-vila-cognito:
Closed and open to trigger the checks.
Checks now passing

Issue #2304, opened 2020-10-23
Dots are clipped at the sides for allowDataOverflow YAxis


Retribution98:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/nostalgic-buck-7ph3s?file=/src/App.js)

### Steps to reproduce
1. Use LineChart or AreaChart with dots
2. Use YAxis with allowDataOverflow

### What is expected?
Points are only clipped vertically or not clipped 

### What is actually happening?
Dots are clipped both vertically and horizontally

| Environment | Info |
|---|---|
| Recharts | v1.8.5 |
| React | react 16.9.0 |
| System | Windows 10 |
| Browser | Google Chrome Version 80.0.3987.87 |

---
I use graphs to display real-time sensor data and use a fixation for Y Axis. But I don't want the dots on the graph to be clipped for X Axis

xile611:
This seems to be be right behavior.

ckifer:
merged in #3602

PR #1571, opened 2018-11-19
refactor clipPathId of Area

KayBeSee:
This mimics the behavior of [this PR](https://github.com/recharts/recharts/commit/c1423726fb7b7e5c22e076d9bf556c8644b4a5a3#diff-ab5884655462c4807891c3e26ca065ef) but for Area charts.

It correctly cuts off dots on the Y-Axis for that chart type. Screenshots below show the changes.

screen shot 2018-11-19 at 4 33 18 pm
screen shot 2018-11-19 at 4 33 43 pm

PR #2488, opened 2021-03-09
feat: extend clipping behavior for the allowDataOverflow property

tylerben:
### Overview

This PR covers two main updates related to the clipping behavior associated with the `allowDataOverflow` prop that is available on the `XAxis` and `YAxis` components.
- ensuring that the clip path generated when setting the `allowDataOverflow` prop to true on an axis only clips the axis it was called on (calling this the "Independent Axis Clipping")
- adding the ability to specify whether dots should be clipped by the clip path when the `allowDataOverflow` prop is set to true (calling this "Dot Clipping Configuration")

**Related issues and PRs**
- #2304 
- https://github.com/recharts/recharts/pull/1237
- https://github.com/recharts/recharts/pull/1571

### Independent Axis Clipping
We noticed that the clip path does not take into consideration which axis the `allowDataOverflow` prop is attached to so if the prop is set to true on either axis, the clip path impacts both axes. This PR expands the existing functionality to ensure that the clip path only impacts the axis where the `allowDataOverflow` is set to true. With the updates, if you want the clip path to apply to both a y and x axis, you would need to set `allowDataOverflow` to true on both components.

This update was made to the following cartesian components:
- Area
- Bar
- Line
- Scatter

**Screenshots**

Previous behavior where `allowDataOverflow` prop does not respect the axis it is set on. `allowDataOverflow` is set to true on the `YAxis` and false on the `XAxis` component but the chart is clipped on both axes anyways.

image

Updated behavior where only the `yAxis` clipping is in play like it should be for the above configuration.

image

### Dot Clipping Configuration

Currently, if `allowDataOverflow` is set to true, dots are clipped at the axis bounds. We have a use case (similar to #2304) where we wanted the dots to to be visible and not cut off by the clip path. We solved this by adding support for a `clipDot` property that can be used to easily clip or not clip the dot (defaults to true aka clip the dot). If the user has specified that they would not like to clip the dot, a second clip path is used that offsets and extends the clip path using the chart width/height and radius/stroke width.

Here is an example

```js
<Area dataKey="count" fill="#15CC09" dot={{ r: 3, strokeWidth: 2, clipDot: false }} />
```

This update was made to the following cartesian components:
- Area
- Line

#### Screenshots

Existing behavior with `allowDataOverflow` set to true.

image

Updated behavior with the ability to opt out of dot clipping.
image


semoal (approved):

abartolo:
Any update if this will be merged soon?

abartolo:
disregard my comment above - my clipping was resolved by adding `margin` to the <LineChart />


```
      <LineChart
        data={data}
        margin={{
          top: 30,
          right: 30,
        }}
      >

```

jufemaiz:
Given the PR is a year old and has an approval from a member of the recharts team it's still worthwhile asking:

1. Will this be merged soon?
2. If not, what is the blocker for merging?

Fullchee:
I am willing to contribute as well but I'm not sure what there is left to do

tylerben:
As the author of the PR would love to see this get merged in as well. Are there any blockers for merging this @semoal?

jufemaiz:
> Given the PR is a year old and has an approval from a member of the recharts team it's still worthwhile asking:
> 
> 1. Will this be merged soon?
> 2. If not, what is the blocker for merging?

Would some of the last few recent committers please provide some guidance?

* @jmfrancois 
* @dillonreedy
* @arcthur
* @xile611 

Thanks!

src/cartesian/Area.tsx
```diff
@@ -554,20 +554,35 @@ export class Area extends PureComponent<Props, State> {
     const { isAnimationFinished } = this.state;
     const hasSinglePoint = points.length === 1;
     const layerClass = classNames('recharts-area', className);
-    const needClip = (xAxis && xAxis.allowDataOverflow) || (yAxis && yAxis.allowDataOverflow);
+    const needClipX = xAxis && xAxis.allowDataOverflow;
+    const needClipY = yAxis && yAxis.allowDataOverflow;
+    const needClip = needClipX || needClipY;
     const clipPathId = _.isNil(id) ? this.id : id;
+    const { r, strokeWidth } = filterProps(dot) || { r: 3, strokeWidth: 2 };
+    const { clipDot = true } = dot as DotProps;
```
ckifer, line 562: any way we can avoid casting this? Can we create a typeguard?

add to `DotProps` a 
```
...
propsType: 'dot';
...
```

then

`const isDotProps = (dot: AreaDot): dot is DotProps => 'propsType' in dot && dot.propsType === 'dot'`?

ckifer (changes requested):
This project doesn't use yarn, please do your install with npm.

If we can avoid the `DotProps` cast and remove the `yarn.lock` I think we're good.

ckifer:
Made a new PR for this since it hasn't been active or responded to. See https://github.com/recharts/recharts/pull/3602

ckifer:
Going to close this as stale in favor of linked PR above - this is still linked to that issue for the details that are provided here

ckifer:
merged in #3602

ckifer:
released in 2.7

Issue #2127, opened 2020-05-14
Bar isn't drawn if data has only one datapoint


jukkahuuskonen:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

Actually it's a duplicate, but earlier issue was closed without fixing it (maybe due to not using this issue-helper), so I'm re-reporting it properly now.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-barchart-nloyr?file=/index.js)

### Steps to reproduce
Chart 1 has 3 data points and works fine, chart 2 has 1 data point and doesn't render bar.
button swaps the data for charts and once again the chart with only 1 data point doesn't render. However, Tooltip does show correctly for both charts.

### What is expected?
Chart with  only 1 data point should render that bar.

### What is actually happening?
Bar isn't rendered.

| Environment | Info |
|---|---|
| Recharts | v2.0.0-beta.6 |
| React | 16.9.0 |
| System | Windows 10 |
| Browser | Firefox 76.0.1 (64-bit)/chrome Version 81.0.4044.138 (Official Build) (64-bit) |

---
This seems to be problem that affects all versions of recharts starting from v1.0.0 (swap data button works only from v1.6.1 upwards). 
There has been a report about this earlier, but it has been closed without fixing it.
There was mention that it affects chart only if axis type is 'number', I haven't tested that myself.
This is a major bug that prevents usage of recharts in many cases. For example, when I'm showing weekly data (for current week), when there is data only for first day of the week, the bar doesn't render and it forces me to switch into axis-type 'category' for these cases. I prefer to use 'number' as then I can easily zoom using domain.

FlavienRemy:
Hi, do we have any news about this problem?

PR #2512, opened 2021-03-30
Fix barchart for a single data point

andy128k:
Attempt to fix https://github.com/recharts/recharts/issues/2127


semoal (approved):

marnixhoh:
It appears as if this issue has resurfaced on 2.0.10... see #2601

pabrodez:
Attempt to fix: #2603

Issue #2506, opened 2021-03-24
ReferenceLine with segment not drawing correctly when using BarChart 


lhkraus:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/musing-ptolemy-5i9vs?file=/src/App.js)

### Steps to reproduce
In a BarChart using strings as the XAxis value, create a ReferenceLine using the segment prop.

### What is expected?
To the reference line draws at the entire XAxis.

### What is actually happening?
The reference line stops drawing at the beggining of the last XAxis point configured on the segment prop.

| Environment | Info |
|---|---|
| Recharts | v2.0.8 |
| React | 17.0.2 |
| System | any |
| Browser | any |

ckotyan:
I too stumbled upon this one and was going crazy trying to get it working. Is there a work around for this?

ckotyan:
> I too stumbled upon this one and was going crazy trying to get it working. Is there a work around for this?
> using position={'end'} on the ReferenceLine does not work either.

I found the the reason why this is happening. There is a line of code in CartesianUtils.js that is not passing the position attribute while calculating the endpoints for the line. I can fork a PR for this change if that is preferred.

MLazarevic36:
The issue still persists, does anyone have solution?

lukask-proxora:
@MLazarevic36 Does the issue still persist with the current version(2.4.3)?
I just forked the original code box and updated the version to 2.4.3 and it seems to work as expected now: https://codesandbox.io/s/modest-wind-56086u

PR #2515, opened 2021-03-31
Forward position attribute while calculating Reference endpoints

ckotyan:
Issue link here:  https://github.com/recharts/recharts/issues/2506


ckotyan:
Discarding to start afresh

PR #2599, opened 2021-07-21
Add chart type to tooltip payload

pabrodez:
This is a small PR that adds the type of chart to the tooltip item generated by the `getTooltipItem()`. 
This allows a custom tooltip content component to have access to the type of chart the entry comes from, and render content conditionally. For example, in a custom content, we may want to:

```javascript
{entry.chartType === 'line' ? <SvgLine /> : <SvgRect />}
<span >{entry.name}</span>
<span >{props.separator}</span>
```


semoal (approved):

pabrodez:
Thank you, @semoal  Does the workflow execution need approval?

Issue #2601, opened 2021-07-22
Bar is not showing on a BarChart when there is only one data point after upgrading to 2.0.10 (from 2.0.9)


marnixhoh:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-undefined-data-vz7e4?file=/src/index.js)

### Steps to reproduce
Pass only one data point into the `data` prop on the `BarChart` component. When multiple are passed everything works as expected.

Note that this did not happen on Recharts 2.0.9. This broke after upgrading to 2.0.10

### What is expected?
Bars to show even if there is only one data point.

### What is actually happening?
No bars are shown when there is only one data point.

| Environment | Info |
|---|---|
| Recharts | v2.0.8 |
| React | 16.14.0 |
| System | MacOS |
| Browser | Chrome |

pabrodez:
Related #2512

andy128k:
That's interesting. I observe totally opposite. 2.0.9 did not work (showed no bars) and with #2512 works properly.

OTOH, I have a pinned version of recharts `=2.0.9` with a patch on top of it (done with [patch-package](https://www.npmjs.com/package/patch-package)). So, I cannot say anything about 2.0.10.

staevs:
2.0.10 works fine for me with one data point

pabrodez:
Using latest version, did a bit of digging, and found that `Rectangle.tsx` is being passed `width= 0` and therefore returning null and not rendering at the `path`.  If I overwrite the height and other variables in the code with any values, it will actually render
![image](https://user-images.githubusercontent.com/18014700/126811277-9d547ba8-48b2-4c51-bb15-a2b9f3c7c205.png)

andy128k:
@pabrodez this is what I saw 3 months ago with 2.0.9. Function `getBandSizeOfAxis` returns `0` when there is only one data point. And this leads to a zero width bar (assuming they are vertically oriented).

andy128k:
Ok, here is my brain dump:

When there is a single data point, function `getBandSizeOfAxis` returns `0` (which is a bug itself).
There are scenarios (like mine) when `maxBarSize` is given, and my patch solves the problem.
[...]

PR #2603, opened 2021-07-24
Return default bandSize and change Xaxis default domain

pabrodez:
Attempts to solve #2601 (no drawing for single data points)

In the reproducible example of the above issue, changing the XAxis type to category, or removing the type prop seems to work. By passing number to type, and because the default number of ticks is 5, it produces 5 ticks, being the value of dataKey the last of those 5. By passing to XAxis domain={['dataMin', 'dataMax']} it only calculates one tick and centers it properly, however the issue of bars not being drawn persists.
As an user, when there's only one single data point in a linear xaxis scale, I would prefer to se an XAxis with one single tick and centered.

`getBandSizeOfAxis()` could just return something sensible like axis.width instead of 0. 

 The XAxis default props include:
```
  tickCount: 5,
  type: 'category',
  domain: [0, 'auto'],
```
In a numeric type of axis, this forces, unless overriden, a single data point array to have 5 ticks starting from 0. If I change domain to `['dataMin', 'dataMax']` + the change to getBandSizeOfAxis() I mention above, now I get a nicely centered single data point tick with both bars drawn.

I think the sensible thing to do is setting ['dataMin', 'dataMax'] as default and allow the user to set [0, 'dataMax'] to have a numeric/linear axis starting from 0 when type is numeric.

### DIFF
--- a/src/cartesian/XAxis.tsx
+++ b/src/cartesian/XAxis.tsx
@@ -44,7 +44,7 @@ XAxis.defaultProps = {
   xAxisId: 0,
   tickCount: 5,
   type: 'category',
-  domain: [0, 'auto'],
+  domain: ['dataMin', 'dataMax'],
   padding: { left: 0, right: 0 },
   allowDataOverflow: false,
   scale: 'auto',
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -1112,7 +1112,7 @@ export const getBandSizeOfAxis = (axis: any, ticks?: Array<TickItem>, isBar?: bo
     return bandSize === Infinity ? 0 : bandSize;
   }
 
-  return 0;
+  return axis?.width ?? 0;
 };
 /**
  * parse the domain of a category axis when a domain is specified


src/util/ChartUtils.ts
```diff
@@ -1112,7 +1112,7 @@ export const getBandSizeOfAxis = (axis: any, ticks?: Array<TickItem>, isBar?: bo
     return bandSize === Infinity ? 0 : bandSize;
   }
 
-  return 0;
+  return axis?.width ?? 0;
```
andy128k, line 1115: Hard to say if this fixes an issue in all scenarios. There is a `return` statement two lines above which, also may return 0.

src/cartesian/XAxis.tsx
```diff
@@ -44,7 +44,7 @@ XAxis.defaultProps = {
   xAxisId: 0,
   tickCount: 5,
   type: 'category',
-  domain: [0, 'auto'],
+  domain: ['dataMin', 'dataMax'],
```
andy128k, line 47: This looks like a change in an interface and may require a major version bump.

andy128k:
Let me put my 2 cents here.

IMHO, calculations related to bar-charts require a deep redesign. And patches like this (and #2512) do not solve the root cause, but rather address different use cases.

andy128k (approved):

pabrodez:
> Let me put my 2 cents here.
> 
> IMHO, calculations related to bar-charts require a deep redesign. And patches like this one (and #2512) do not solve the root cause, but rather address different use cases.

💯 Changing the creation of ticks when there's only a single value to return a single tick by default, assigning it a coordinate that is the middle of the axis width.

PR #2614, opened 2021-07-29
Refactor lodash imports to reduce bundle size by 34kb

pabrodez:
This is a refactor of `lodash` imports to reduce the size of the bundle by 34kb.
- `lodash `overall size down to 34kb from 68kb according to `npm run analyse`
- fix `analyse` script in `package.json`


mohamedmansour:
This is good! Try to fix the build and include @arcthur on your pr

PR #2634, opened 2021-08-26
Replace classnames with clsx

wojtekmaj:
clsx is fully compatible with classnames API, is quickly gaining traction and is about 1/3rd smaller than classnames.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "recharts",
-  "version": "2.0.9",
+  "version": "2.1.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
-      "version": "2.0.9",
+      "version": "2.1.2",
       "license": "MIT",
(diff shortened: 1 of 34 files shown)


wojtekmaj:
The chances of classnames being in the bundle is slightly higher indeed. But clsx is slowly closing the gap: https://npm-stat.com/charts.html?package=classnames&package=clsx

By migrating popular packages to the latter my hope was to speed up this trend.

HHongSeungWoo:
@ckifer What do you think about this issue? It's faster, smaller, and doesn't require installing @types.

ckifer:
Looks fine to me - though not having to install types isn't a huge benefit. 

It's a drop-in replacement so no concern from me

wojtekmaj:
It's also now more popular than classnames: https://npmtrends.com/classnames-vs-clsx

ckifer:
Merged!

PR #2660, opened 2021-09-19
Fixes #2636 - ResponsiveContainer makes legend overlapping with chart when re-rendering

### LINKED ISSUE #2636: ResponsiveContainer makes legend overlapping with chart when re-rendering
Report a bug

I'm using the `<AreaChart />` component for a project and I added a custom content component into the `<Legend />` component itself. I have a button that is meant to switch between a Table view and a Chart view, however, It seems like it interferes with the `<ResponsiveContainer />`. Each time the chart is re-rendered, it takes the whole height of the responsive container without considering the `<Legend/>`, which leads to overlapped components (see video attached)

For now, the `<Legend />` component is in `position: absolute`, which matched the desired design, but even if I switch the component to a `relative` position, the chart is still taking into account the `<Legend />` within its height. 

https://user-images.githubusercontent.com/29457898/130966017-228cc4e4-76df-45da-9a44-2df5df6ea3d7.mp4

`<AreaChart />` should always have the defined height when re-rendering, no matter If there is a Legend or not. 

recharts@2.1.2 - Chrome, Firefox, Safari

neilsonz:
Fix demonstrated on `ResponsiveContainer` demo.  Prior to this change, the chart would overlap the legend item (`pv`) and only after resizing the window the `ResponsiveContainer` would re-render and fix the overlapping issue.

## Before (Legend overlap):
image

## After (Legend no longer overlapping):
image

Note: I think due to way that `size` was set after mounting previously (as a class component, from version `2.0.10`), the previous size calculation took into account its children correctly as `renderChart()` was called subsequently after mounting.  When this was converted to a function component in `2.1.0`, this calculation happened sooner and didn't take into account its rendered children sizes, and never gets recalculated until after the component sized has changed (e.g. after a window resize).  This PR is effectively introduces a hack to re-render later, while doing so in a way that causes the current tests to still pass.

### DIFF
--- a/demo/component/ResponsiveContainer.tsx
+++ b/demo/component/ResponsiveContainer.tsx
@@ -206,6 +206,7 @@ export default class Demo extends Component {
                 animationBegin={1300}
                 dot
               />
+              <Legend />
             </AreaChart>
           </ResponsiveContainer>
         </div>
--- a/src/component/ResponsiveContainer.tsx
+++ b/src/component/ResponsiveContainer.tsx
@@ -49,7 +49,7 @@ export const ResponsiveContainer = forwardRef(
     const containerRef = useRef<HTMLDivElement>(null);
     useImperativeHandle(ref, () => containerRef, [containerRef]);
 
-    const mounted = useRef<boolean>(false);
+    const [mounted, setMounted] = useState<boolean>(false);
 
     const getContainerSize = () => {
       if (!containerRef.current) {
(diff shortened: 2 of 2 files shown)


ritwikvd:
This is still not working :(

The initial render makes the legend overlap with the ticks; a subsequent re-render fixes the alignment.

tinhbeng:
I still have the same problem with version 2.1.6

khalitsyn:
Setting debounce={1} helped me with the similar problem which occured after update to 2.1.6

i3130002:
### One thing that gives similar result and might be helpful is [this code ](https://codesandbox.io/s/zen-moore-9jonx?file=/src/App.js)
### Result:
image


### Expected:
image

#### Using debounce={1} fixes things but please fix the issue

Issue #2658, opened 2021-09-18
d3-shape => Error [ERR_REQUIRE_ESM]: Must use import to load ES Module


alithecodeguy:
I used recharts regularly before but since this morning , i got this error when i try to build the nextjs app : 

> Build error occurred
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/**/**/node_modules/d3-shape/src/index.js
require() of ES modules is not supported.
require() of /Users/**/**/node_modules/d3-shape/src/index.js from /Users/**/**-

I updated node and nextjs but there is no luck.

node version : 14.17.6
nextjs version : 11.1.2
recharts version : 2.1.3

update : downgrade to 2.1.2 resolved the issue

apancutt:
See thread [here](https://github.com/recharts/recharts/commit/3115b4d869042f7daf94a267117f75adc5049e23#r56692229)

arcthur:
https://github.com/d3/d3-format/issues/114 you can see this

brc-dd:
Literally none of things mentioned in that thread or the linked issue are working on Next.js `v11.1.3-canary.24`. Here is all that I have tried (and in almost every sensible combination):

* Update `browserslist` to `["> 0.5%", "not dead"]` (earlier had `[">0.3%", "not ie 11", "not dead", "not op_mini all"]`, don't know how switching to `0.5` would have even worked).

* Update webpack config to [this](https://github.com/d3/d3-format/issues/114#issuecomment-826338584) (although it needs to be modified for webpack-5).

* Used `next-transpile-modules` for both `recharts` and `d3-shape`.

* Set `experimental.esmExternals` to `'loose'`.

The only thing that's left is using `next/dynamic`. But for importing `Bar, BarChart, CartesianGrid, Legend, Tooltip, XAxis, YAxis`, I'll have to write 7 dynamic imports. 😕 Downgrading to `2.1.2` appears as the only feasible option.

arcthur:
I'm revert d3 version, and try again.

apancutt:
Fixed now. Thanks.

apancutt:
PR #2707 reintroduces this issue.

apancutt:
Fixed in 2.1.8

rohasplit:
`2.1.13` re-introduces this. Reverting to `2.1.12` fixes the issue.

MauricioHernanCabrera:
[...]

PR #2707, opened 2021-11-05
Update d3-interpolate, d3-scale and d3-shape

Philipp91:
They all have the same remark in their changelog (and no other declared breaking changes otherwise):

> This package now requires Node.js 12 or higher. For more, please read [Sindre Sorhus’s FAQ](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).

I'm not sure if recharts already requires Node.js 12 or higher. If not, this should be considered a breaking change.

Also, we could consider adding an `"engines"` section to `package.json` like [here](https://github.com/d3/d3-shape/blob/005b0e4a12adf6a1039df2bef84bdc8541e34c88/package.json#L52-L54).


apancutt:
This change has reintroduced #2658 - intentional?

arcthur:
miss the problem, and I fixed on 2.1.8

Semigradsky:
Should it be reopened?

Issue #2709, opened 2021-11-09
PieProps ref using incorrect type definitions


HannesArni:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/pie-chart-of-two-levels-forked-zn6wq?file=/src/App.tsx)

### Steps to reproduce
Define the props for a Pie component in an object, and set the type of the object to PieProps. Then pass those properties to a Pie component, and you'll get a type error

### What is expected?
The PieProps type should match the Pie component props

### What is actually happening?
PieProps does not match Pie components props

| Environment | Info |
|---|---|
| Recharts | v2.1.6 |
| React | 16.14.0 |
| System | macOS Big Sur 11.6 |
| Browser | Chrome 95.0.4638.69 |

binhpv:
I second this issue. Probably related to https://github.com/recharts/recharts/issues/2665
The ref that the component exposed is the default class component ref, pointing to the instance of that component and not the DOM element.

Sigkar:
Ran into this issue creating a library component wrapper for a line chart.

A quick fix for this i've been using is to `Omit` the type value for `ref` and add a generic/typed ref in its place, like, and just throwing some basic idea out:

```ts
Omit<PieProps, 'ref'> & Record<'ref', React.Ref<any>>;
```

Also affects `LineProps` for those wondering. Might be an issue in more than those two places.

PR #2765, opened 2022-01-14
Update ref to use the implicit ref type when consuming LineProps

Sigkar:
Kind of a working idea. I made these changes and cross-tested them in a separate repository as well, TypeError went away and `ref` could be used properly. Can't really create a proper test for this in Karma because I only saw `.js` files?

Error:
```
No overload matches this call.
  Overload 1 of 2, '(props: Props | Readonly<Props>): Line', gave the following error.
    Type '{ string?: string | number | undefined; type: CurveType; className?: string | undefined; color?: string | undefined; height?: number | undefined; id?: string | undefined; ... 486 more ...; yAxis?: (Omit<...> & { ...; }) | undefined; }' is not assignable to type 'IntrinsicClassAttributes<Line>'.
      Types of property 'ref' are incompatible.
        Type 'LegacyRef<SVGPathElement> | undefined' is not assignable to type 'LegacyRef<Line> | undefined'.
          Type '(instance: SVGPathElement | null) => void' is not assignable to type 'LegacyRef<Line> | undefined'.
            Type '(instance: SVGPathElement | null) => void' is not assignable to type '(instance: Line | null) => void'.
              Types of parameters 'instance' and 'instance' are incompatible.
                Type 'Line | null' is not assignable to type 'SVGPathElement | null'.
                  Type 'Line' is missing the following properties from type 'SVGPathElement': addEventListener, removeEventListener, pathLength, getPointAtLength, and 263 more.
  Overload 2 of 2, '(props: Props, context: any): Line', gave the following error.
    Type '{ string?: string | number | undefined; type: CurveType; className?: string | undefined; color?: string | undefined; height?: number | undefined; id?: string | undefined; ... 486 more ...; yAxis?: (Omit<...> & { ...; }) | undefined; }' is not assignable to type 'IntrinsicClassAttributes<Line>'.
      Types of property 'ref' are incompatible.
        Type 'LegacyRef<SVGPathElement> | undefined' is not assignable to type 'LegacyRef<Line> | undefined'.
```

Reproduction

```ts
import { LineProps } from 'recharts';

interface SomeLibraryComponent {
  line: LineProps
}

const SomeComponent: SomeLibraryComponent = ({line}) => {
  return (
    <Line // Error location of above
      {...line}
    />
  )
}
```

Find it kind of strange for the `ref` object to be the actual class.. would an `any` work better in this case?

### DIFF
--- a/src/cartesian/Line.tsx
+++ b/src/cartesian/Line.tsx
@@ -75,7 +75,7 @@ interface LineProps extends InternalLineProps {
   id?: string;
 }
 
-export type Props = Omit<CurveProps, 'points' | 'pathRef'> & LineProps;
+export type Props = Omit<CurveProps, 'points' | 'pathRef' | 'ref'> & LineProps & Record<'ref', React.Ref<Line>>;
 
 interface State {
   isAnimationFinished?: boolean;


Sigkar:
Looks like this affects more than just LineProps.

#2709 looks like its running into the same issue.

I also noticed it is affecting the cartesian grid. Maybe can create an arged type for the ref to apply back when exporting the interfaces?

ckifer:
@Sigkar can you confirm this is still an issue? Thanks :)

ckifer:
Going to close this as stale, we'll come back to these when we fix refs for the next major release

PR #2770, opened 2022-01-27
feat: allow axis domain to accept a callback

tommilligan:
We would like to generate the domain of an axis dynamically, based on the minimum and maximum data values.

This is not possible via the current API, where the minimum and maximum values can be calculated via callback independently. This PR adds a route for the entire domain (both min and max) to be calculated in one go.

The callback has access to the context relevant for `recharts` to calculate concrete domain values currently.

---

For a real world example, our sample use looks like:

```typescript
import { XAxis } from "recharts";
import { getNiceTickValues } from "recharts-scale";

const TICK_COUNT = 5;

function customDomain([dataMin, dataMax]: [number, number], _allowDataOverflow: boolean): [number, number] {
  const dataAbsMax = Math.max(Math.abs(dataMin), Math.abs(dataMax));
  const tickValues = getNiceTickValues([-dataAbsMax, dataAbsMax], TICK_COUNT, true);
  return [tickValues[0], tickValues[TICK_COUNT - 1]];
};

...

<XAxis type="number" domain={customDomain} />
```


tommilligan:
Matching documentation PR opened here: https://github.com/recharts/recharts.org/pull/178

PR #2784, opened 2022-02-15
Fixed ResponsiveContainer ref

mrskiro:
Until now, if user wanted to access a ref object, they had to use `ref.current.current`.

This was unfriendly and not clear, so I fixed it.

### DIFF
--- a/src/component/ResponsiveContainer.tsx
+++ b/src/component/ResponsiveContainer.tsx
@@ -47,7 +47,7 @@ export const ResponsiveContainer = forwardRef(
       containerHeight: -1,
     });
     const containerRef = useRef<HTMLDivElement>(null);
-    useImperativeHandle(ref, () => containerRef, [containerRef]);
+    useImperativeHandle(ref, () => containerRef.current, [containerRef]);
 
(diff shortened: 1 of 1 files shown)


KittyGiraudel:
This would be a breaking change though, so I guess it would have to wait for a major version — which might never happen given the library appears to be relatively unmaintained.

ckifer:
Added some labels, most likely fine to add this as a breaking change in 3 if we ever get there

ckifer:
Not sure what you mean. Those that use those functionality have this referenced a certain way. If we change it in a minor or patch release it will break them.

purp1eeeee:
@ckifer 
Do you plan to merge these?
I recently had plans to UPGRADE my recharts and was reminded of this PR.

ckifer:
Yeah but it's a breaking change, so not until 3.x. Which is still timeframe TBD as we gather other breaking changes

purp1eeeee:
I understand.
Thanks for all the maintenance！

ckifer:
Going to close this PR in favor of re-opening the issue https://github.com/recharts/recharts/issues/3718. Harder to keep a PR up to date while waiting for a major version

Issue #2742, opened 2021-12-21
Axis scale="band" no longer works as of Recharts 2.x.x


danatemple:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/histogram-recharts-2-1-8-b31h8)

### Steps to reproduce
Run the code sandbox, which is closely based on the recharts bars example. It also fails on 2.0.x.

### What is expected?
I expect that the X axis ticks are placed "around" the bars in the same way that scale="band" used to do in Recharts 1.8.5.
See this 1.8.5 example: https://codesandbox.io/s/histogram-recharts-1-8-5-wvvqu

### What is actually happening?
scale="band" has no effect. You can delete it from the <Xaxis> without the graph changing at all.

| Environment | Info |
|---|---|
| Recharts | v2.1.6 |
| React | 17.0.1 |
| System | Codesandbox on Chrome, Windows 10 |
| Browser | Chrome 96 |

---
I use scale="band" to display histograms, where the axis ticks need to be at the points between the bars, because they are showing the start and end of the histogram bin.

There is still an issue with the last tick, but I work around this by adding a second axis with just that tick.

semoal:
I think it changed internally from the d3-scale library of how it works.

You could give it a try to this and see if has effect:

```ts
import { scaleBand } from "d3-scale";
const scale = scaleBand().domain(["a"]).range([0,100]);
...
      <XAxis dataKey="name" scale={scale} padding={{ left: 10, right: 10 }} />
```

danatemple:
@semoal thanks, but I just tried in CodeSandbox, didn't seem to have any effect.

semoal:
Cool, looks like a regression we'll look into it as soon as possible. I want before the end of this year to decide a roadmap and start giving some fresh air to the library.

danatemple:
@semoal just wondering what the expectation is here for a fix?

It is kind of blocking us, as we can't upgrade React beyond 16.4 (we hit a very obscure crash with Recharts 1.x) until we upgrade Recharts to 2.x.

Would be happy to pay for someone to work on this, as I appreciate there are many demands on time...

bsplosion:
[...]

PR #2794, opened 2022-03-02
Fix for #2742 Axis scale=band no longer works as of Recharts 2.x.x

danatemple:
This is the fix to https://github.com/recharts/recharts/issues/2742


danatemple:
@arcthur just wanted to check if you'd seen my comment to #2742 regarding the remaining 2 px offset?

https://github.com/recharts/recharts/issues/2742#issuecomment-1060693679

Issue #2801, opened 2022-03-15
Improving accessibility support of Recharts - Contribution offer


cscrosati:
- [x ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Developers using Recharts today don't have a comprehensive set of features and guidance on creating accessible charts for their end customers.

Developers are currently asking for this via the [Issues](https://github.com/recharts/recharts/issues?q=is%3Aissue accessibility) and [Discussions], and a couple of [Discussions](https://github.com/recharts/recharts/discussions) threads. With legal requirements to meet accessibility standards continually expanding, developers need to be able to create accessible charts with Recharts, or they will use a different library.

The end customers benefits from improved accessibility support in many aspects, including but not limited to:
- Improved contrast between data sets, which improved both low vision users. It also greatly improves visibility of data when used with low contrast projectors.
- Improved keyboard only navigation of the data shown in the chart. This benefits users with reduced motor skills, along with users who prefers keyboard only usage.
- Improved screen reader user experience in understanding the graphically displayed data with properly labeled roles and components.

We have assessed the Recharts library and identified areas of improvement and would like to collaborate with the Recharts on contributing a good chunk of accessibility features. Some are APIs additions, and some are documentation additions with guidance for developers. This feature request is a starting point for a discussion on how we can help achieve this with the Recharts maintainers.

Our ultimate goal would be to enable developers using Recharts to meet [WCAG 2.2](https://www.w3.org/TR/WCAG22/) "AA" level compliance requirements.

### What does the proposed API look like?
There is no finite API for this feature request because it encompasses a lot of details related to different aspects of accessibility. We recognize that not all of this will be API fixable, and will be instead guidance within the Recharts documentation.

Here are a few identified things that could be done through API, or through documentation:
- Expose a way to consume WAI-ARIA attributes on the charts.
- Document how to add proper roles and labels to component of a chart.
- Associate good default roles for major components of Rechart.
- Include a default ARIA role for all charts (currently none, proposal could be FIGURE or GROUP or IMG, depending on certain criterias)
- Improve Tooltip to behave in accordance to the "[Content on Hover](https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html)" principles
- Offer a default or optional theme that meets color contrast requirements for non-text elements, preserve current contrast for text elements are these meet the requirements today.

This list is not complete.

[...]

PR #2814, opened 2022-04-04
Add ARIA1.2 attributes to the SvgElementPropKeys filter array

cscrosati:
*This PR is an incremental addition to help with #2801 . It is just a first step to the larger work needed for accessibility in Recharts.*

This PR addresses how one can add ARIA attributes to their Recharts instances. As of now, the ARIA attributes are filtered out by the internal filterProps function using the `SVGElementPropKeys` filtering array. This PR adds the ARIA attibutes to the filtering array to allow them in. For reference, the [SVG2 specification includes the WAI-ARIA attributes](https://www.w3.org/TR/SVG2/struct.html#WAIARIAAttributes).

The following code snippet (extracted from the LineCharts demo page) doesn't render the label in the DOM before:

`<LineChart width={400} height={400} data={data02} syncId="test" aria-label="A simple LineChart with fixed domain y-axis">`

After this change, it now works as expected, with the following DOM render:

image

I didn't include a test suite since it doesn't appear `SVGElementPropKeys` has tests associated with it. Let me know what requirements you would like on that aspect.


fyoung-cpi:
Thanks for taking this first step! Looks good to me.

PR #2843, opened 2022-04-25
chore: move type deps into devDependencies

jmfrancois:
Types dependencies are usefull only to build the package and check types.

Move @types/d3* into devDependencies


lmaillet (approved):

jmfrancois:
cc @xile611 please review !

PR #2852, opened 2022-05-12
Handling of undefined values and type checks in DefaultTooltipContent and associated DataUtil functions

dhassouni:
There were several errors at time of compilation relating to the fact that operations were being performed that cannot be done on undefined values. Also fixed the `isNumOrStr` type check function to accept as an input any parameter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "dh-recharts",
+  "name": "recharts",
   "version": "2.1.9",
   "description": "React charts",
   "main": "lib/index",
@@ -42,13 +42,13 @@
   ],
   "repository": {
     "type": "git",
-    "url": "https://github.com/dhassouni/recharts.git"
+    "url": "https://github.com/recharts/recharts.git"
   },
-  "author": "recharts group, modfied by DHassouni",
+  "author": "recharts group",
   "bugs": {
-    "url": "https://github.com/dhassouni/recharts/issues"
+    "url": "https://github.com/recharts/recharts/issues"
   },
-  "homepage": "https://github.com/dhassouni/recharts",
+  "homepage": "https://github.com/recharts/recharts",
   "peerDependencies": {
     "react": "^16.0.0 || ^17.0.0",
     "react-dom": "^16.0.0 || ^17.0.0"
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "recharts",
+  "name": "dh-recharts",
   "version": "2.1.9",
   "description": "React charts",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "recharts",
+  "name": "dh-recharts",
```
jkklapp, line 2: In order to merge to upstream, I think these changes need to be reverted 🤔

src/component/DefaultTooltipContent.tsx
```diff
@@ -127,7 +127,7 @@ export class DefaultTooltipContent<TValue extends ValueType, TName extends NameT
     const wrapperCN = classNames('recharts-default-tooltip', wrapperClassName);
     const labelCN = classNames('recharts-tooltip-label', labelClassName);
 
-    if (hasLabel && labelFormatter) {
+    if (hasLabel && labelFormatter && !_.isNil(payload)) {
```
jkklapp, line 130: I think it would be better to just compare like

```javascript
payload === null
```

Instead of using loadsh

jkklapp (changes requested):
At least the changes in package.json need to be reverted

package.json
sea-dhassouni: Whoops! Good catch. I was playing around locally. This wasn't supposed to make the PR. Good catch. I will fix it now.

src/component/DefaultTooltipContent.tsx
sea-dhassouni: Fair enough. Lodash was already being used, but I have no problem with removing this dependency on another package.

package.json
dhassouni: My mistake! this wasn't supposed to be in the PR. I'll remove the changes to package. json. Thanks for catching this.

src/component/DefaultTooltipContent.tsx
dhassouni: Fair enough. The package was already using lodash, but I have no problem reducing reliance on other package for this.

dhassouni: For what it is worth, this line of code was just 5 lines above it:
`const hasLabel = !_.isNil(label);`
I'm not going to reach out of scope and removing lodash elsewhere, but just so that we are on the same page about the usage of lodash in this package.

package.json
dhassouni: @jkklapp Thanks for looking at my code. I've made the suggested changes if you want to have a look again. Thanks!

jkklapp (approved):

PR #2898, opened 2022-07-11
Take letter-spacing and font-size into consideration while rendering ticks

saghan:
This is an attempt to improve accessibility in recharts. According to [Web Content Accessibility Guideline](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html) there should not be loss of content when letter-spacing is increased upto 0.12em. CartesianAxis component wasn't taking into consideration letter-spacing while calculating size of ticks. I am trying to fix it. (NOTE: All tests pass)
LineChart before fixing (with letter-spacing 0.5em):
Screen Shot 2022-07-10 at 9 09 22 PM
LineChart after fixing:
Screen Shot 2022-07-10 at 9 10 57 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
+## 2.1.11 (Jun 24, 2022)
+
+# feat
+
+-  Adds react `^18.0.0`  as valid peerDependency (#2820)
+
 ## 2.1.10 (May 19, 2022)
 
 # feat
 
 - Add ARIA1.2 attributes to the SvgElementPropKeys filter array
 - Added Storybook Badge (#2840)
-- Handling of undefined values and type checks in DefaultTooltipContent…
+- Handling of undefined values and type checks in DefaultTooltipContent
 
 # fix
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "recharts",
-  "version": "2.1.10",
+  "version": "2.1.11",
   "description": "React charts",
   "main": "lib/index",
   "module": "es6/index",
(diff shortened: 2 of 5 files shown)


cscrosati (changes requested):
Can you add tests to the PR to ensure the functionality doesn't regress?

saghan:
> Can you add tests to the PR to ensure the functionality doesn't regress?

Done.

cscrosati:
I have another test suggestion:

FontSize/TextSpacing is X1/Y1, and the ticks are {a/b/c}
then when changing FontSize/TextSpacing to X2/Y2, ticks are now {a/c}

Something that confirms that applying a new computed style values does indeed discard or add the ticks as needed. Also independently change FontSize and TextSpacing to prove they independently are impacting the result.

saghan:
> I have another test suggestion:
> 
> FontSize/TextSpacing is X1/Y1, and the ticks are {a/b/c} then when changing FontSize/TextSpacing to X2/Y2, ticks are now {a/c}
> 
> Something that confirms that applying a new computed style values does indeed discard or add the ticks as needed. Also independently change FontSize and TextSpacing to prove they independently are impacting the result.

getComputedStyle does not work  as expected with Karma. That's why the change works fine in a sample but not in test browser. So I have added a test to ensure that at least the states are picked correctly when getComputedStyle is stubbed.

cscrosati:
@arcthur @xile611 This change was made to improve the accessibility of the axis overlapping ticks. This resolves concern with https://www.w3.org/WAI/WCAG22/Understanding/text-spacing.html requirements. Let us know if you have additional concerns with this proposed solution.

cscrosati:

jdom:
LGTM. Thanks for contributing this. We are running into some of these accessibility issues in our planning phase on whether to use the library for our project too.

nmoinvaz:
It seems the calls to `getElementsByClassName` introduced in this PR broke IE 11 support. I didn't see any babel polyfill that would provide this function.

PR #2900, opened 2022-07-12
Add inactiveShape prop to Pie component

rockcs1992:
Hi,

Our team has the need to change the appearance of a piechart when one of the sectors is hovered/selected. For the hovered sectors we could achieve that with the existing api `activeShape`. However (correct me if I'm wrong) the existing api doesn't provide the ability to customize the look of other sectors. Thus I'm proposing to add an API for that.

I'm calling the new API `inactiveShape`, similar to `activeShape`. The API, if provided, allows the users to customize how inactive sectors look like, when a valid `activeIndex` prop is provided.

Please let me know how you think of this. Also raised a PR to update the docs. 
https://github.com/recharts/recharts.org/pull/196
I'll also update the demo and add a codesandbox link once ethis PR is merged.

Thanks

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -87,6 +87,7 @@ interface PieProps extends PieDef {
   data?: any[];
   sectors?: PieSectorDataItem[];
   activeShape?: PieActiveShape;
+  inactiveShape?: PieActiveShape;
   labelLine?: PieLabelLine;
   label?: PieLabel;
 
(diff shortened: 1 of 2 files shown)


src/polar/Pie.tsx
```diff
@@ -87,6 +87,7 @@ interface PieProps extends PieDef {
   data?: any[];
   sectors?: PieSectorDataItem[];
   activeShape?: PieActiveShape;
+  inactiveShape?: PieActiveShape;
```
cscrosati, line 90: nitty: Should we consider renaming the interface PieActiveShape to PieShape? Disregard if this would cause a breaking change though.

cscrosati (approved):

cscrosati:
@arcthur and @xile611 , this PR proposal is a contribution that isn't accessibility driven directly, but allows to developer to interact and redraw the non active shapes of a pie, such that their accessibility characteristics are improved in relation to the active one.

Let us know if you have concerns with the naming of the API, or any other topics.

cscrosati:

rockcs1992:
@arcthur @xile611 Please let us know if you have any opinions/ideas/concerns about this PR. Thanks

PR #2914, opened 2022-07-25
Allow tooltip to be dismissed using Escape key

saghan:
This change is trying to enable users to dismiss tooltip using Escape key. According to [Web Content Accessibility Guideline](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html), content that appears on hover should be able to be dismissed without moving the mouse. Hopefully, this change will help Recharts users interact with the chart more easily through keyboard. P.S. The change in the test file also contains many style changes that have been automatically applied on save due to eslint rules in Recharts

### DIFF
--- a/src/component/Tooltip.tsx
+++ b/src/component/Tooltip.tsx
@@ -115,6 +115,8 @@ export class Tooltip<TValue extends ValueType, TName extends NameType> extends P
   state = {
     boxWidth: -1,
     boxHeight: -1,
+    dismissed: false,
+    dismissedAtCoordinate: { x: 0, y: 0 },
   };
 
   private wrapperNode: HTMLDivElement;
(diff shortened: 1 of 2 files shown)


cscrosati (approved):
Looks good, @xile611 and @arcthur , what do you think? It resolves the "Dismissible" accessibility concern of a "Content on hover" requirement.

cscrosati:

colinmorelli:
I noticed in this PR that the tooltip is only closed if esc is pressed while the tooltip has focus. The result is if the tooltip is open, and (without moving the mouse) you click on the chart, and then tap escape, the tooltip doesn't close.

However, we're also running into another unintended consequence of having focus captured on hover. In our app, we have popovers that close on blur, so if you accidentally move your mouse too far outside of a popover and it lands on a chart, the tooltip will open and focus, which will then close the popover.

I think we can still keep the same accessibility behavior here by just binding an event listener to document.body when the tooltip is opened, and not capturing focus in the tooltip element itself. Would that work/would you accept a PR for this?

PR #2916, opened 2022-07-27
Add formatter function type to tooltip props

cravend:
This updates the prop validation of a tooltip `formatter` prop to not allow any arbitrary function.


saghan:
@cravend @arcthur This change causes build error shown in screenshot below:
Screen Shot 2022-08-01 at 1 15 52 PM

PR #2924, opened 2022-08-01
Fix typing of default tooltip formatter

cravend:
Tied to #2916, this PR updates the typing to the default formatter to prevent it from failing the TSC build. 

Closes #2925

### DIFF
--- a/src/component/DefaultTooltipContent.tsx
+++ b/src/component/DefaultTooltipContent.tsx
@@ -6,8 +6,8 @@ import React, { PureComponent, CSSProperties, ReactNode } from 'react';
 import classNames from 'classnames';
 import { isNumOrStr } from '../util/DataUtils';
 
-function defaultFormatter<T>(value: T) {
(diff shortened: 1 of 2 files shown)


saghan:
@cravend I am looking into it.

saghan:
@cravend Could you try disabling the linter with these comments?
return (
      // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
      <div
        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
        tabIndex={0}
        role="dialog"
        onKeyDown={event => {
          if (event.key === 'Escape') {

---
We need to disable the linter here because the div needs to be able to have focus in order to listen to keypress

saghan:
I have also created a draft here:
https://github.com/recharts/recharts/pull/2925
I think it might be better to have your PR contain the changes. This way all errors will be resolved in a single PR and we don't have to force merge any PR with build errors.

PR #2925, opened 2022-08-01
add eslint exception to fix failing build

saghan:
Adding an eslint exception because we need to listen to 'Escape' key to dismiss Tooltip although it is non-interactive element. (The build failure is due to unrelated changes that are merged)


cravend:
Would it be better to change this to an interactive element and remove all styles?

saghan:
> 

I looked into what interactive element might be suitable as a Tooltip but didn't find any. The only interaction that is expected from users is pressing Escape key. I am open to any suggestions that you might have.

PR #2955, opened 2022-08-30
Add a sample to show how to fill a chart with patterns

saghan:
Hi,
We are planning to use ReCharts in Microsoft Azure. This PR adds a sample which shows how to fill charts with SVG patterns instead of colours. This fulfills the [WCAG](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html) requirement which states that if any information is conveyed through colour difference, then we we should have a visual way of conveying the same information without the difference in colour, for example by using different patterns. It helps people who have trouble distinguishing different colours.

I have not added any test since this PR does not add any feature. This is how the sample looks:
Screen Shot 2022-08-30 at 2 31 20 PM

### DIFF
--- a/demo/component/PieChart.tsx
+++ b/demo/component/PieChart.tsx
@@ -1,10 +1,9 @@
 import React, { Component, ReactElement, ReactSVGElement } from 'react';
-import { PieChart, Pie, Legend, Cell, Tooltip, ResponsiveContainer, Sector,
-  Label, LabelList } from 'recharts';
+import { PieChart, Pie, Legend, Cell, Tooltip, ResponsiveContainer, Sector, Label, LabelList } from 'recharts';
 import { scaleOrdinal } from 'd3-scale';
 import { schemeCategory10 } from 'd3-scale-chromatic';
-import { changeNumberOfData } from './utils';
 import * as _ from 'lodash';
+import { changeNumberOfData } from './utils';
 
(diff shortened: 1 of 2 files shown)


nikolasrieble:
Where is the best place to add examples? Is it in the website repo? Or in this demo? Further, in the demo there are two places of examples - (1) for the api and (2) on the examples page.

saghan:
> Where is the best place to add examples? Is it in the website repo? Or in this demo? Further, in the demo there are two places of examples - (1) for the api and (2) on the examples page.
> 
> I would search for an example in the website repo under the examples page. https://github.com/recharts/recharts.org
> 
> Alternatively, it might make sense to update and aggregate all examples in https://github.com/dillonreedy/recharts-storybook

Thank you for your feedback. I have created a PR that shows how to fill pattern in API example. https://github.com/recharts/recharts.org/pull/205

saghan:
@arcthur Could you please take a look at this change?

nikolasrieble:
Since I wrote my first comment, my perspective has changed. The repo and the website are not active in development. Please make the example in the storybook instead. 

There is a group of people who want to fork, thus not all is lost, but I expect examples to be added more frequently and by more people to the storybook.

saghan:
Right now recharts.org is the best place to put the sample. I might add it to the storybook later.

PR #2960, opened 2022-09-05
Add roles to chart components

saghan:
This PR adds roles to various chart components to meet accessibility guidelines of [WCAG](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11)

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -307,6 +307,7 @@ export class Bar extends PureComponent<Props, State> {
             className="recharts-bar-rectangle"
             {...adaptEventsOfChild(this.props, entry, i)}
             key={`rectangle-${i}`} // eslint-disable-line react/no-array-index-key
+            role="img"
           >
             {Bar.renderRectangle(shape, props)}
           </Layer>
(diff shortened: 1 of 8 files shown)


src/cartesian/Bar.tsx
```diff
@@ -307,6 +307,7 @@ export class Bar extends PureComponent<Props, State> {
             className="recharts-bar-rectangle"
             {...adaptEventsOfChild(this.props, entry, i)}
             key={`rectangle-${i}`} // eslint-disable-line react/no-array-index-key
+            role="img"
```
mgifford, line 310: I think this is the only place that role="img" should be used. If there are references for implementing it elsewhere, please let me know.

src/chart/Sankey.tsx
```diff
@@ -667,6 +669,7 @@ export class Sankey extends PureComponent<Props, State> {
       <div
         className={classNames('recharts-wrapper', className)}
         style={{ ...style, position: 'relative', cursor: 'default', width, height }}
+        role="region"
```
mgifford, line 672: Why apply this role here? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role

src/cartesian/Bar.tsx
mgifford: Sorry, just looked it up and my assertion doesn't seem to be correct, see: 

https://tink.uk/accessible-svg-line-graphs/

Still we should be inserting a <title> so it can be more valuable.

PR #3001, opened 2022-10-07
DefaultTooltipContent.tsx Solving type error for entry.value and entry.name

gabgagnon:
**Error for entry.value**

let value: TValue | undefined
Argument of type 'TValue | undefined' is not assignable to parameter of type 'TValue'.
  'TValue' could be instantiated with an arbitrary type which could be unrelated to 'TValue | undefined'.
  
  
**Error for entry.name**

let name: TName | undefined
Argument of type 'TName | undefined' is not assignable to parameter of type 'TName'.
  'TName' could be instantiated with an arbitrary type which could be unrelated to 'TName | undefined'.

### DIFF
--- a/src/component/DefaultTooltipContent.tsx
+++ b/src/component/DefaultTooltipContent.tsx
@@ -78,8 +78,8 @@ export class DefaultTooltipContent<TValue extends ValueType, TName extends NameT
           ...itemStyle,
         };
         const finalFormatter = entry.formatter || formatter || defaultFormatter;
-        let { name, value } = entry;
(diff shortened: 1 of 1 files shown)


jeremy0dell:
Now when value is 0 (which can happen in a huge variety of circumstances), the formatter function will not run. This is definitely not the intended behavior.

gabgagnon:
> Now when value is 0 (which can happen in a huge variety of circumstances), the formatter function will not run. This is definitely not the intended behavior.

you are right, sorry for this big mistake

Issue #2644, opened 2021-09-02
AreaChart - area is not filled to the bottom when domain dataMin is negative


ryancrunchi:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/simple-area-chart-forked-vu0ls)

### Steps to reproduce
- Create an AreaChart

- Add YAxis with negative domain dataMin

- Add Area to fill bellow curve

### What is expected?
Area is filled to the bottom of the chart (until XAxis) and should take into account dataMin domain

### What is actually happening?
Area is filled only to 0. If dataMin < 0, the area between 0 and dataMin is not filled.

| Environment | Info |
|---|---|
| Recharts | v2.0.8 |
| React | 17.0.1 |
| System | macOS 11.3 |
| Browser | Chrome 93.0.4577.63 |

akamfoad:
You can set `baseValue` prop on `AreaChart` to desired base value, and it'll work. Look at this Fork of your [CodeSandbox
](https://codesandbox.io/s/simple-area-chart-forked-tds0l)

Found out by digging into the source code
https://github.com/recharts/recharts/blob/1a452a291b0f5b95ff0e11bed3f0143210e4495f/src/cartesian/Area.tsx#L121

PR #3013, opened 2022-10-12
Fix `baseValue` prop of Area

### LINKED ISSUE #2518: baseValue area chart is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes
The issue which is not created via https://recharts.github.io/recharts-issue-helper will be closed immediately.

---
I am having an AreaCharr problem. Basevalue has not been accepted as props by Componenent. With typescript. Please help me to solve this problem

注意：不是用 https://recharts.github.io/recharts-issue-helper 创建的 issue 会被立即关闭。

kotokrad:
Currently, the only way to set `baseValue` is passing it to `AreaChart` as a prop but this doesn't typecheck. This fix moves `baseValue` prop to `Area` according to types, so it's possible to use it, as well as to assign different values to different areas.

Fixes #2518 and #2644

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -118,8 +118,9 @@ export class Area extends PureComponent<Props, State> {
     animationEasing: 'ease',
   };
 
-  static getBaseValue = (props: Props, xAxis: Props['xAxis'], yAxis: Props['yAxis']): number => {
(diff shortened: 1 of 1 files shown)


jeremyodelldv:
I would really like to see this get merged in and fixed

PR #3026, opened 2022-10-19
fix: let formatter function run even when value is falsy

jeremy0dell:
Changed the logical AND operators (&&) added in c6cf8dc to nullish coalescing operators (??) to allow formatter function to run even when value is falsy. 0 is a falsy value that still ought to let the formatter function run

### DIFF
--- a/src/component/DefaultTooltipContent.tsx
+++ b/src/component/DefaultTooltipContent.tsx
@@ -79,7 +79,7 @@ export class DefaultTooltipContent<TValue extends ValueType, TName extends NameT
         };
         const finalFormatter = entry.formatter || formatter || defaultFormatter;
         let { value, name } = entry;
-        if (finalFormatter ?? value ?? name) {
+        if (finalFormatter && value != null && name != null) {
           const formatted = finalFormatter(value, name, entry, i, payload);
           if (Array.isArray(formatted)) {
             [value, name] = formatted as [TValue, TName];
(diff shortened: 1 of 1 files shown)


src/component/DefaultTooltipContent.tsx
```diff
@@ -79,7 +79,7 @@ export class DefaultTooltipContent<TValue extends ValueType, TName extends NameT
         };
         const finalFormatter = entry.formatter || formatter || defaultFormatter;
         let { value, name } = entry;
-        if (finalFormatter && value && name) {
+        if (finalFormatter ?? value ?? name) {
```
mguida22, line 82: I believe this should check if the value and name are not null or undefined? With `??` value and name can be undefined, which I don't think is the intention here.

```suggestion
        if (finalFormatter && value != null && name != null) {
```

mguida22 (commented):
Thanks for putting up this PR! We're running into the same issue right now, it'd be great to get this fix in.

src/component/DefaultTooltipContent.tsx
jeremy0dell: You're right! I'm still getting used to the `??` operator. Committing this change

jeremy0dell:
@mguida22 Maybe we could run the test suite and get this into master?

mguida22:
I don't have permissions to run anything here, I just ran into the same problem and saw your open PR 😬

jeremy0dell:
Ah. @arcthur thoughts?

PR #3035, opened 2022-10-25
fix: version update date in CHANGELOG.md

WooJunKang:
fixed version update date (2.1.15)


WooJunKang:
already resloved in #3016

PR #3072, opened 2022-11-26
fix: allow dragging brush/traveller when the mouse is outside of brush area

### LINKED ISSUE #2570: Brush does not move if mouse leaves area

### Reproduction link
[https://recharts.org/en-US/examples/BrushBarChart](https://recharts.org/en-US/examples/BrushBarChart)

### Steps to reproduce
Make the brush thinner so that it can be moved horizontally, and then move it horizontally but let your mouse leave the brush area while still holding the button down.

### What is expected?
On most scroll brushes, like the ones on web pages, your mouse can exit the area of the brush while you're holding onto the mouse button and it'll still move the brush. For example, on the left of the minimal reproduction, theres a scroll bar for all the examples, and if you click and drag that bar while exiting the bar's width, the bar still moves properly.

The brush stops responding the instant the mouse leaves the brush, and snaps to the correct position if you enter the brush area.

| Environment | Info |
|---|---|
| Recharts | v2.0.8 |
| React | 16.14.0 |
| System | Windows |
| Browser | Firefox 89 |

tryonelove:
Currently, it's not possible to drag a brush or traveller when mouse leaves traveller area, The fix allows to drag brush and traveller even when the mouse is outside of the brush component.

Fixes #2570

### DIFF
--- a/src/cartesian/Brush.tsx
+++ b/src/cartesian/Brush.tsx
@@ -1,7 +1,7 @@
 /**
  * @fileOverview Brush
  */
-import React, { PureComponent, Children, ReactText, MouseEvent, ReactElement, TouchEvent, SVGProps } from 'react';
+import React, { PureComponent, Children, ReactText, ReactElement, TouchEvent, SVGProps } from 'react';
 import classNames from 'classnames';
 import { scalePoint, ScalePoint } from 'd3-scale';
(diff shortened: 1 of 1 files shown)


aimanmohsin03:
Looking forward.

PR #3075, opened 2022-12-01
Adjust rounding to avoid hairline cracks in cartesian grid

louissobel:
When using vertical striping in a `CartesianGrid`, I noticed two issues that I think will be fixed by the rounding changes in this PR:

Screen Shot 2022-11-30 at 11 26 28 PM

I think this PR is safe because we'll still end up with stripes covering the whole grid, even if they change alignment by a pixel one way or the other.

### Floating point "nearly there" comparison leading to overflow
At some screen widths, the check on line 149 will fire incorrectly, due to what seems like int <> float comparison problems. See for example:
Screen Shot 2022-11-30 at 11 29 28 PM

The "x" of this rectangle is 0 and the width is just a smidge under 5, suggesting that the first entry of `verticalPointsUpdated` was `4.999999999999943`

### Hairline cracks at some widths
The width of each stripe is computed against the unrounded values, but then uses a rounded value as its base. In the picture above, the third stripe starts at 5 and has a width of 142.57 but isn't rendered covering the whole distance to the next stripe, which, starts at 148. Computing the widths of each stripe against the _rounded_ start location of the next stripe (in this case, that would give 143) will lead making sure the whole space is covered.

### This PR
 - I made the change in the `horizonalStripes` function too even though I only am observing this in the vertical striping
 - I kept the `+ x - x` code in the Math.round — I'm not sure why it's there (and naively, I'm not sure it does anything?) but there might be a javascript-specific reason to have that.

### DIFF
--- a/src/cartesian/CartesianGrid.tsx
+++ b/src/cartesian/CartesianGrid.tsx
@@ -144,14 +144,16 @@ export class CartesianGrid extends PureComponent<Props> {
     }
 
     const { fillOpacity, x, y, width, height } = this.props;
-    const verticalPointsUpdated = verticalPoints.slice().sort((a, b) => a - b);
+    const roundedSortedVerticalPoints = verticalPoints.map(e => Math.round(e + x - x)).sort((a, b) => a - b);
 
-    if (x !== verticalPointsUpdated[0]) {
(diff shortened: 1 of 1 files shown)


louissobel:
wonderful, thank you!

Issue #3008, opened 2022-10-11
Breaking change in 2.1.14 type system change for Tooltip formatter prop


ryanolson-aumni:
By changing the type definition (in a patch release!!!) for `formatter`, our build broke. Our formatter uses `(value: string) => JSX.Element` and is now an invalid formatter because of the limiting of the formatter to return a `ValueType` and a `NameType` or just a `ValueType`.

It's wrong to do this in a patch version.

Broken in https://github.com/recharts/recharts/pull/2924

daniel-rodrigue:
Moreover, the goal of a formatter is to provide a different representation for `ValueType`. In my case I have `<Tooltip<number,string> ... />` where the number is an amount (currency). The displayed value of the number is a `string` and not a `number` type (e.g. "$1,000.00").

furrycatherder:
@cravend

bennettdams:
I think this was fixed via https://github.com/recharts/recharts/pull/3618 and is part of the `v2.7.0` release. 
The original types (`[React.ReactNode, React.ReactNode]`) are back, which works for me.

ckifer:
Cool going to close this

PR #3078, opened 2022-12-04
Revert "Fix typing of default tooltip formatter (#2924)"

furrycatherder:
This reverts commit da6b9e2394cd55169ecdf31baf51bfa021eedb18.

See https://github.com/recharts/recharts/issues/3008 for more information. This change does not appear to be working as intended. This sums it up best (emphasis added):

> Moreover, the goal of a formatter is to provide a **different** representation for `ValueType`.


furrycatherder:
@arcthur @cravend

ckifer:
@furrycatherder this will break the build. I'm down to fix this issue but we'll need to do it in a way that isn't strictly reverting this commit (and continues to give some type safety)

PR #3093, opened 2022-12-14
Chore/lint staged

ckifer:
Adding auto-linting to staged files using `husky` and `lint-staged`.

Removed `pre-commit` as I couldn't get it to run `lint-staged`. `pre-push` was never working, add a hook that runs tests and linter on push. - this is kind of annoying, open to feedback on it.

* run husky install on `prepare`
* run `lint-staged` on `pre-commit` (runs eslint and prettier)
* run `npm test` and `npm run lint` on `pre-push`

Also downgraded jest version to 27. Turns out 28 needs TypeScript 4.3+ [docs](https://jestjs.io/docs/28.x/upgrading-to-jest28)


arcthur:
Why jest version dropped

ckifer:
See the link at the bottom of the PR @arcthur. It only worked because a different version of TS was committed to the lock-file. Jest 27 needs TS 4.3+ Jest 28 needs TS 4.5+

Issue #2610, opened 2021-07-28
"points" attribute showing "[object Object]" for all points in the attribute value (looks like bad serialization)


Eric24:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://recharts.org/en-US/examples](https://recharts.org/en-US/examples)

### Steps to reproduce
Using DevTools, show the DOM element of the SVG of the chart data series (i.e. `<g class="recharts-layer recharts-line">`)

### What is expected?
`points` attribute should not contain "[object Object]"

### What is actually happening?
`points` attribute contains "[object Object]"

| Environment | Info |
|---|---|
| Recharts | v2.0.8 |
| React | 16.8.1 |
| System | Mac/Windows |
| Browser | Chrome/Firefox/Safari |

---
```
<path stroke="#8884d8" stroke-width="1" fill="none" points="[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object]" width="468" height="383" class="recharts-curve recharts-line-curve" type="monotone" d="M80,296.08C106,315.26829999999995,132,334.4566,158,334.4566C184,334.4566,210,12.660000000000007,236,12.660000000000007C262,12.660000000000007,288,238.3236,314,238.3236C340,238.3236,366,204.16000000000003,392,204.16000000000003C418,204.16000000000003,444,242.46,470,242.46C496,242.46,522,232.88500000000002,548,223.31000000000003"></path>
```

pabrodez:
This error originates in `Curve.tsx`:
![image](https://user-images.githubusercontent.com/18014700/127627827-e0bd180c-b6aa-4975-bf1d-5c59d2ac4189.png)
`filterProps()` is supposed to filter out those keys of the props that are not attributes of the svg element, but I cannot find `points` as a `path` attribute anywhere in the MDN docs. That function uses an array of compiled attributes of all SVG elements. Because `points` is an attribute of other SVG elements, it'll be included in `path` anyways:
![image](https://user-images.githubusercontent.com/18014700/127629174-2a73fac9-439a-4bbd-a761-eb8eac2c2b16.png)

I believe `filterProps()` should be passed an array of attributes to include and not just use a huge general list of all possible SVG attributes. Or even have an object with names of used svg elements for keys and an array of their attribute for values.

PR #3096, opened 2022-12-14
Removing points attribute from the <path> tag

dhavalsoni2001:
https://github.com/recharts/recharts/issues/2610

### DIFF
--- a/src/shape/Symbols.tsx
+++ b/src/shape/Symbols.tsx
@@ -104,9 +104,12 @@ export class Symbols extends PureComponent<Props> {
     const { className, cx, cy, size } = this.props;
 
     if (cx === +cx && cy === +cy && size === +size) {
+      const filteredProps = filterProps(this.props, true);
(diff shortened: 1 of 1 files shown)


src/shape/Symbols.tsx
```diff
@@ -104,9 +104,12 @@ export class Symbols extends PureComponent<Props> {
     const { className, cx, cy, size } = this.props;
 
     if (cx === +cx && cy === +cy && size === +size) {
+      const filteredProps = filterProps(this.props, true);
+      // removing points from <path> as it's not supported attribute
+      delete filteredProps.points;
       return (
         <path
-          {...filterProps(this.props, true)}
```
ckifer, line 109: this isn't really a sustainable method of preventing unwanted attributes from being spread here. Is there any way `filterProps` can be used?

ckifer:
Going to close this PR in favor of #3101 

Thanks @dhavalsoni2001

PR #3100, opened 2022-12-15
test: refactor and fix piechart jest tests

ckifer:
We want to refactor jest tests that are currently skipped and currently in JS files to run correctly and in TSX files.

* Adjust jest config so ESLint doesn't complain about `expect`, etc.
   *  disable extraneous deps for `test-jest` folder
* add `@types/jest`
* refactor `PieChart.spec.js` -> `PieChart.spec.tsx`
  * remove lint ignores 
  * for consistency use `test` over `it`
  * sinon.spy -> jest.fn
  * fix skipped test
  * use react testing library over enzyme

### DIFF
--- a/test-jest/specs/chart/PieChart.spec.tsx
+++ b/test-jest/specs/chart/PieChart.spec.tsx
@@ -1,16 +1,8 @@
-/* eslint-disable react/no-unused-state */
-/* eslint-disable max-len */
-/* eslint-disable import/order */
-/* eslint-disable no-return-assign */
-/* eslint-disable react/prop-types */
-/* eslint-disable import/no-extraneous-dependencies */
-/* eslint-disable react/jsx-filename-extension */
-/* eslint-disable no-undef */
 import React from 'react';
-import { PieChart, Pie, Sector, Legend, Cell } from '../../../src';
-import { render } from '@testing-library/react';
+import { fireEvent, render } from '@testing-library/react';
+import { PieChart, Pie, Legend, Cell } from '../../../src';
 
-describe.only('<PieChart />', () => {
+describe('<PieChart />', () => {
   const data = [
     { name: 'Group A', value: 400 },
     { name: 'Group B', value: 300 },
@@ -20,7 +12,7 @@ describe.only('<PieChart />', () => {
     { name: 'Group F', value: 189 },
   ];
 
-  it('Renders 6 sectors circles in simple PieChart', () => {
+  test('Renders 6 sectors circles in simple PieChart', () => {
     const { container } = render(
       <PieChart width={800} height={400}>
         <Pie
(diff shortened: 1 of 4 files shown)


test-jest/specs/chart/PieChart.spec.tsx
```diff
@@ -1,16 +1,8 @@
-/* eslint-disable react/no-unused-state */
-/* eslint-disable max-len */
-/* eslint-disable import/order */
-/* eslint-disable no-return-assign */
-/* eslint-disable react/prop-types */
-/* eslint-disable import/no-extraneous-dependencies */
-/* eslint-disable react/jsx-filename-extension */
-/* eslint-disable no-undef */
```
nikolasrieble, line 8: Nice!

test-jest/specs/chart/PieChart.spec.tsx
```diff
@@ -20,7 +12,7 @@ describe.only('<PieChart />', () => {
     { name: 'Group F', value: 189 },
   ];
 
-  it('Renders 6 sectors circles in simple PieChart', () => {
+  test('Renders 6 sectors circles in simple PieChart', () => {
```
nikolasrieble, line 15: It is just an alias for test. Thus this does not make a difference.

test-jest/specs/chart/PieChart.spec.tsx
```diff
@@ -131,18 +123,16 @@ describe.only('<PieChart />', () => {
...
-    se.simulate('mouseEnter');
-    expect(onMouseEnter.calledOnce).to.equal(true);
+    fireEvent.mouseEnter(se);
+    expect(onMouseEnter).toBeCalled();
 
-    se.simulate('mouseLeave');
-    expect(onMouseLeave.calledOnce).to.equal(true);
-    */
+    fireEvent.mouseLeave(se);
+    expect(onMouseLeave).toBeCalled();
```
nikolasrieble, line 136: Either change the name of the test to reflect the scope, or preferably, split this up into two separate tests

`test('click on Sector should invoke onClick callback')`

and 

`test('Mouse enter and mouse leave are triggered'`

and possibly even another for Mouse over and Mouse out?

nikolasrieble (commented):
Instead of testing multiple things in one test, lets split the test up.

test-jest/specs/chart/PieChart.spec.tsx
ckifer: yep! but as a dev I'd rather stick to one or the other. Just a preference change

test-jest/specs/chart/PieChart.spec.tsx
ckifer: Agreed, I was more attempting to get things working as they were before adding net new tests/changing test names (unless it made sense to do so). Since this PR is small I'll make the change, but this might be harder to do until initial changes are made in larger test files like `LineChart`

nikolasrieble (approved):
Great! Thank you!

PR #3101, opened 2022-12-16
fix: points attribute spread onto incorrect svg elements

### LINKED ISSUE #2610: "points" attribute showing "[object Object]" for all points in the attribute value (looks like bad serialization)

ckifer:
Closes #2610 (also closes #3096 PR).

SVG props are spread onto elements after calling `filterProps`. Some SVG attributes only belong on specific SVGs.

This is a short term solution that requires a longterm fix (see TODO inline to use something like https://github.com/wooorm/svg-element-attributes). 

* Remove barely used `isSvg` from `filterProps` and replace with `svgElementType`
  * manually checked to see if there is any usage of `polyline` or `polygon` - no usage confirmed  
* Define svg elements that need specific attributes (only a relevant subset for now due to effort required)
* Map element types to key arrays
* Check to see if `svgElementType` is specified - if it is then check to see if the element has a specific key. If it doesn't continue to act naive and filter to all defined svg props
* Add unit test for `Symbol` to ensure that `points` does not get added to the `path`

Remove `svg` `version` prop on `Surface` as [this is very much deprecated](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/version) and has no effect.


src/util/types.ts
```diff
@@ -1145,14 +1153,20 @@ export const filterProps = (
 
   const out: Record<string, any> = {};
 
+  /**
+   * If the svg element type is explicitly included, check against the filtered element key map
+   * to determine if there are attributes that should only exist on that element type.
+   * @todo Add an internal cjs version of https://github.com/wooorm/svg-element-attributes for full coverage.
```
nikolasrieble, line 1159: Could you please create an issue to follow up with this todo? Please link this PR to provide context for the issue.

src/util/types.ts
```diff
@@ -1145,14 +1153,20 @@ export const filterProps = (
...
   Object.keys(inputProps).forEach(key => {
-    // viewBox only exist in <svg />
     if (
+      (svgElementType && matchingElementTypeKeys.includes(key)) ||
       SVGElementPropKeys.includes(key) ||
-      (isSvg && SVGContainerPropKeys.includes(key)) ||
       (includeEvents && EventKeys.includes(key))
     ) {
-      out[key] = (inputProps as any)[key];
+      out[key] = inputProps[key];
```
nikolasrieble, line 1169: Nice :)

src/util/types.ts
```diff
@@ -1127,7 +1135,7 @@ export type ViewBox = CartesianViewBox | PolarViewBox;
 export const filterProps = (
```
nikolasrieble, line 1135: I wonder, would it make sense to extract this function into a separate file and add a minimal unit test - just to set a precedent and an example for future changes?

src/util/types.ts
```diff
@@ -830,6 +830,14 @@ const SVGElementPropKeys = [
   'angle',
 ];
 
+export type FilteredSvgElementType = 'svg' | 'polyline' | 'polygon';
+
+export const FilteredElementKeyMap: Record<FilteredSvgElementType, string[]> = {
```
nikolasrieble, line 835: Could you please add a small comment? Is this a list of properties not to include per svg element type?

test/specs/shape/SymbolsSpec.js
```diff
@@ -1,36 +1,60 @@
...
+          { name: 'test3', uv: 150 },
+        ]}
+        margin={{ top: 20, right: 20, bottom: 20, left: 20 }}
+      >
+        <XAxis />
+        <YAxis type="category" />
+        <Line type="monotone" dataKey="uv" stroke="#ff7300" />
+      </LineChart>,
+    );
+
```
nikolasrieble, line 55: Can you please add another test to expect points on another element, where it is needed / used?

test/specs/shape/SymbolsSpec.js
```diff
@@ -1,36 +1,60 @@
...
-      </Surface>
+      </Surface>,
     );
 
     expect(wrapper.find('.recharts-symbols').length).to.equal(0);
   });
+
+  it('Expect no points attribute on path', () => {
+    const { container } = render(
+      <LineChart
```
nikolasrieble, line 40: Do you need a full LineChart for this test? 
The other tests just use a Surface and Symbols, 
could you not use a line directly? 

The fact that this test is different in setup makes me wonder wether it belongs in a separate file?

nikolasrieble (commented):
Great stuff, a couple questions on my side. I think this is a great chance to add a unit test for filterProps

ckifer:
Sounds good, will get to the revisions when I get a chance! Thanks

src/util/types.ts
ckifer: created and linked

test/specs/shape/SymbolsSpec.js
ckifer: There are no elements where it is needed or used, but I can add a unit test that specifically calls `filterProps` with this requirement

test/specs/shape/SymbolsSpec.js
ckifer: This is where the issue occurs (since this is where `filterProps` is called). This test asserts something from the element that is being tested. I wouldn't want to put it in `LineChart` since the issue doesn't come from line chart. If I do not render a chart then there is no issue with `Surface` (no "points" spread)

src/util/types.ts
ckifer: This becomes a lot bigger refactor if I do that due to many changing imports. Do you think its acceptable to add a unit test now adjacent to the file it will be in and refactor in a different PR?

ckifer:
Should meet review questions/requirements now! Will follow up by moving the function to ReactUtils

PR #3103, opened 2022-12-16
feat: add ability to set static domains to improve UX for toggling data

### LINKED ISSUE #3099: Allow static domains, to improve UX when toggling datapoints

I'm currently working on a chart that starts with multiple bars being visible, that allows users to toggle them off so they can focus on a smaller set of datapoints. Recharts allows this toggling to be done very easily with the `hide` prop, but I've noticed that the domain of the axis will always adapt to fit only the currently displayed data. This negatively affects the user's ability to make comparisons when toggling datapoints.

[Minimal codesandbox example here](https://codesandbox.io/s/recharts-axis-dynamic-domain-example-msbbmk?file=/src/App.tsx)

It would be great to include a way to set the domain of an axis as static, i.e. always including data even if it's hidden. I've made local changes to `generateCategoricalChart.tsx` and can verify that the behaviour is easy to implement by removing the `!item.props.hide` check when filtering the `graphicalItems`.

If this feature is one you would consider implementing, I'd like to agree on the right API that would update those filter predicates, and I'd be happy to open a pull request with the implementation.

I have two ideas for this API. My preference would be for this first approach, as I feel it's the simplest:

```tsx
<YAxis includeHidden /> // maybe `staticDomain` also works as a prop name?
```

We could also have a boolean flag on the graphical items themselves, so we can still ignore some items if we want:

```tsx
<Bar onDomain hide={booleanState} /> // by default, `onDomain`'s value would be `!hide`
```

neefrehman:
Closes #3099 

This PR introduces the `includeHidden` prop to `Axis` elements. Setting this to true ensures that all datapoints within a chart contribute to its domain calculation, even when they are hidden. This allows developers to improve the end-user experience for charts that allow toggling datapoints, as it can improve their ability to make comparisons on narrower sets of data. Documentation for this feature is added in https://github.com/recharts/recharts.org/pull/223.


.eslintrc
```diff
@@ -8,6 +8,7 @@
     "plugin:prettier/recommended",
     "prettier"
   ],
+  "ignorePatterns": ["test/specs/*"],
```
ckifer, line 11: Fixing tests and refactoring to .tsx files are in progress. Idk if we actually want to ignore the errors we're trying to fix. Might just ask you do remove this for now so we can address it elsewhere

ckifer (changes requested):
Do you mind adding a screenshot of what this looks like in the PR? Also is this a valid prop for all axis types?

nikolasrieble:
Thank you for your contribution! Could you please provide a test plan with a toy dataset, that I can use to verify that this works as expected? 

Further, it would be great if you could split up the PR into (a) refactoring and (b) adding a feature, which would reduce the blast radius and decrease the chance of having to revert everything if a problem comes up.

src/util/types.ts
```diff
@@ -1077,6 +1077,8 @@ export interface BaseAxisProps {
   allowDecimals?: boolean;
   /** The domain of scale in this axis */
   domain?: AxisDomain;
+  /** Whether or not to ensure the domain doesn't change when hiding datapoints (defaults t0 false) */
```
nikolasrieble, line 1080:
```suggestion
  /** Consider hidden elements when computing the domain (defaults to false) */
```

test/specs/cartesian/YAxisSpec.js
```diff
@@ -123,7 +122,7 @@ describe('<YAxis />', () => {
   it('Render 1 ticks when domain={["auto, "auto"]} and dataKey is "noExist" ', () => {
     const wrapper = mount(
       <AreaChart width={600} height={400} data={data}>
-        <YAxis stroke="#ff7300" domain={["auto", "auto"]} />
+        <YAxis stroke="#ff7300" domain={['auto', 'auto']} />
```
nikolasrieble, line 125: Are these redundant formatting checks a result of the linting rules setup in the repo? If yes, we should do a single formatting PR to prevent such changes showing up in PRs.

test/specs/cartesian/YAxisSpec.js
```diff
@@ -194,4 +189,40 @@ describe('<YAxis />', () => {
     expect(wrapper.find('svg').children.length).to.equal(1);
     expect(wrapper.find('svg noscript').children.length).to.equal(1);
   });
+
+  it('Render identical ticks when data is hidden and includeHidden is true', () => {
```
nikolasrieble, line 193: Using the ticks to validate this feature is fragile, because whenever the tick computation would change, this test would break. 

As far as I understand, the feature really is about the domain boundaries. Could you test whether the boundaries are identical? 

This might be possible getting the axis properties.

nikolasrieble, line 193: Would it make sense to add a separate test for the xAxis?

nikolasrieble (commented):
Great work! I have a couple questions and suggestions.

test/specs/cartesian/YAxisSpec.js
neefrehman: Yes i think these were caused by the pre-commit hook on the project. Per @ckifer's comment about upcoming refactors to these tests, i'll revert them and commit without the formatting step

.eslintrc
neefrehman: of course! I hadn't appreciated these upcoming changes, so l've reverted this change

test/specs/cartesian/YAxisSpec.js
neefrehman: I think this test should be resilient to tick computation changes, as it's not asserting on specific tick values, only that the values—whatever they are—stay the same when data items are hidden.

I was trying to keep this test as user-centric as possible as, to the user, static domain boundaries would manifest as consistent ticks on the yAxis, and consistent sizing of the graphical items. I've just added some assertions for bar heights to cover the latter. If you definitely think this is the wrong approach then I'm happy to change it, though I might need a little help as to which axis properties are best to test with!

neefrehman: I did have a look at the XAxis tests, and made the decision to not include them, as I didn't see much crossover in testing responsibilities  between the two files, and the code path should be covered by the YAxis test, since the logic is shared. If you think its the right move to add a test there, then I certainly can.

neefrehman:
@ckifer @nikolasrieble thanks for your comments!

I've just split the changes out into multiple PRs, and have added an interactive demo of these changes to the `demo` directory, so you should be able to see the old and new behaviour side-by-side to verify yourselves. Let me know if you'd like any other changes.

neefrehman:
>Also is this a valid prop for all axis types?

@ckifer it is a valid prop for any axis that extend `BaseAxisProps` which, as it stands, is only `xAxis` and `yAxis`

neefrehman:
@ckifer @nikolasrieble I've just updated this PR to fix conflicts, and move to an RTL/Jest test. I've also replaced the demo with a pair of stories, which should make the behaviour easier to spot. Is there anything else you need from me at this point? I'd love to see this feature added soon!

Also as a side-note, the pre-commit and pre-push hooks are consistently failing for me, with errors seemingly unrelated to the changes in this PR.

ckifer:
hey @neefrehman 

I hear you, but wanted to wait to merge this until the Jest migration was done at least. I will pull down these changes and check them out, hopefully we can get this in for next minor release (2.5).

The  pre-commit and pre-push hooks work for me and assuming they work for Niko as well, in what way are they failing for you?

neefrehman:
@ckifer Thanks!

The pre-push failures I had were all lint related. It looks like the storybook file I added had some issues that my editor took a while to recognise!

The merge commit also triggered a handful of linter errors for test files, which were triggered by lint-staged. My last commit didn't trigger them.

ckifer:
Okay so the hooks themselves are working - good. 

The demo pages have all sorts of lint errors so feel free to skip verification for those but only those. The storybook files should be linted, but there's possibility we miss things on merge.

storybook/stories/API/cartesian/YAxis.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+    /*
+     * Toggles displayed bars when clicking on a legend item
+     */
+    const handleLegendClick: ComponentProps<typeof Legend>['onClick'] = e => {
+      const key: string = e.dataKey;
+      setActiveKeys(prev => (prev.includes(key) ? prev.filter(k => k !== key) : [...prev, key]));
+    };
+
+    return (
+      <ResponsiveContainer width={500} height={500}>
```
ckifer, line 37: I think since last review this changed in the storybook to use a fixed height and width: 100% - can you merge the latest and check what the pattern changed to?

storybook/stories/API/cartesian/YAxis.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <ComposedChart width={500} height={400} data={data} margin={{ top: 20, right: 80, left: 20 }}>
```
ckifer, line 38: Can remove width and height on the chart when responsive container is used

storybook/stories/API/cartesian/YAxis.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+const data = [
+  { name: 'Page A', pv: 800, amt: 1400 },
+  { name: 'Page B', pv: 967, amt: 1506 },
+  { name: 'Page C', pv: 1098, amt: 989 },
+  { name: 'Page D', pv: 1200, amt: 1228 },
+  { name: 'Page E', pv: 1108, amt: 1100 },
+  { name: 'Page F', pv: 680, amt: 1700 },
+];
+
+export const Simple = {
```
ckifer, line 23: Can you add some text to both stories indicating that the legend can be clicked to toggle data visibility?

For the IncludeHidden story can you make the text tell the viewer what the difference between simple and includeHidden is?

ckifer (commented):
Looks good! @neefrehman do you mind merging the latest and making the small suggested changed? Good to go after that

storybook/stories/API/cartesian/YAxis.stories.tsx
neefrehman: Just had a quick look at doing this and I can't see any other stories with text. Do you have a preferred place or way to add some?

nikolasrieble: Have a look at the `storybook/stories/API/cartesian/Line.stories.tsx` for example.

neefrehman: Thanks for the pointer! I've just added some notes to this effect.

ckifer (approved):
🚀

micktg:
Relevant docs:

- https://recharts.org/en-US/api/XAxis#includeHidden
- https://recharts.org/en-US/api/YAxis#includeHidden

PR #3107, opened 2022-12-19
chore: use early return to improve readability of getAxisMapByAxes

neefrehman:
Follow-on from https://github.com/recharts/recharts/pull/3103

I've made a change to `getAxisMapByAxes` in `generateCategoricalChart.tsx`, to do an early return in the `axes.reduce` call when we have cached data in the accumulator, instead of at the end. This reduces the nesting of code to improve readability. It also avoids the function having to call `getDisplayedData` when not necessary—which also iterates over a list—to give a potential performance improvement if there are ever duplicated `axisIdKey`s.


ckifer (approved):
Changes look good to me. Just shifting inverting the return logic which is safe here.

We need to add unit tests to `generateCategoricalChart` but I don't think this PR should be responsible for that. It does get tested via other tests (since every categorical chart calls it and every one with an XAxis calls this specifically).

ckifer:
I pulled this down locally just to do some testing. This early return isn't ever reached in any of the recharts demos - inversion looks safe to merge.

PR #3108, opened 2022-12-19
fix(types): add zIndex as valid AxisType, remove CategoricalChartOptions assertions

neefrehman:
Follow-on from https://github.com/recharts/recharts/pull/3103

Adds 'zIndex' to the AxisType union type, which means we can avoid the type 
assertion in `ComposedChart.tsx` and `ScatterChart.tsx`.

I've also removed the unnecessary type assertion to `CategoricalChartOptions` in other chart components, which should improve type safety a little.


ckifer (approved):
LGTM, thanks for separating these changes

PR #3109, opened 2022-12-21
Developer Experience: Add PR Template

nikolasrieble:
Add a PR template to guide developer attention to better documentation, testing, and provide a template to discuss a PR. 

Auto generated via https://github.com/TalAter/open-source-templates


ckifer (approved):
Looks good to me! Should be helpful!

We can always slightly adjust this overtime if we need to

Issue #3114, opened 2022-12-26
Treemap: use provided node values


mariodsantana:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
When calculating the size of a parent node, the current code adds up the values of all children.  I sometimes want to do something different, for example to size the parents according to the max/min/mean or other aggregate of the children's values.

### What does the proposed API look like?
Compatible with current API, if no value is given for parent nodes.  If a value is given for a parent node, the current API would ignore that value, while the proposed API would adopt it.  The patch could be as simple as:

```diff
diff --git a/src/chart/Treemap.tsx b/src/chart/Treemap.tsx
index ce36070..c04ce67 100644
--- a/src/chart/Treemap.tsx
    b/src/chart/Treemap.tsx
@@ -42,11  42,16 @@ const computeNode = ({
       : null;
   let nodeValue;

-  if (children && children.length) {
+  // TODO need to verify valueKey
+  if (!_.isNaN(node[valueKey as string]) && node[valueKey as string] >= 0) {
+    // nodeValue provided
+    nodeValue = node[valueKey as string];
+  } else if (children && children.length) {
+    // calculate nodeValue by adding up children's nodeValues
     nodeValue = computedChildren.reduce((result: any, child: TreemapNode) => result   child[NODE_VALUE_KEY], 0);
   } else {
-    // TODO need to verify valueKey
-    nodeValue = _.isNaN(node[valueKey as string]) || node[valueKey as string] <= 0 ? 0 : node[valueKey as string];
+    // no children and no nodeValue provided
+    nodeValue = 0;
   }

   return {
```

This would allow me to calculate the parent node sizes outside of Recharts.  Alternatively, we could allow a callback that would calculate the size of a parent instead of the default of adding up all the children's values.  The benefit of the callback is that it's 100% compatible with the current API - if anyone counts on the code ignoring parent values in the passed data, they won't notice any change.  The benefit of the proposed approach is its simplicity.

ckifer:
@mariodsantana a little bit confused on the use-case here. You want to use the average, etc, but still have the same child values? Wouldn't that make the chart incorrect in what its displaying? Would it be one big square?

Agreed this isn't a breaking change but there's still a chance it could effect some consumers. I think I'd want to see some consensus or +1s that this is how the treemap api should work before merging this in.

mariodsantana:
[...]

PR #3115, opened 2022-12-26
WIP: Treemap: use provided node values

mariodsantana:
## Description

Compatible with current API, if no value is given for parent nodes. If a value is given for a parent node, the current API would ignore that value, while the API in this patch would adopt it.

**Work In Progress** - I still need to figure out how to only show the parent names.  As it is, the parent and all child names are superimposed into an unreadable mess.  I plan to get to that over the holiday break, but I'm publishing now in hopes of avoiding wasted work in case this whole approach is wrong.

## Related Issue

#3114 

## Motivation and Context

When calculating the size of a parent node, the current code adds up the values of all children. I sometimes want to do something different, for example to size the parents according to the max/min/mean or other aggregate of the children's values.

## How Has This Been Tested?

I've created a demo with pre-calculated parent values.  I generated that demo with random data many times and observed it to show sizes that match up with parent values as different from the normal Treemap demo that shows parent sizes according to the sum of all child sizes.

Note that this level of detail is not explained in the docs, so I made no changes there.

## Types of changes

## Checklist:

### DIFF
--- a/demo/component/PrecalculatedTreemap.tsx
+++ b/demo/component/PrecalculatedTreemap.tsx
@@ -1,0 +1,482 @@
+import React, { Component } from 'react';
+import { Treemap, Tooltip } from 'recharts';
+import DemoTreemapItem from './DemoTreemapItem';
+import { changeNumberOfData as changeData } from './utils';
+
+const data = [
+  {
+    name: 'analytics',
+    size: 55,
+    children: [
+      {
+        name: 'cluster',
+        size: 21,
+        children: [
+          { name: 'AgglomerativeCluster', size: 3938 },
+          { name: 'CommunityStructure', size: 3812 },
+          { name: 'HierarchicalCluster', size: 6714 },
+          { name: 'MergeEdge', size: 743 },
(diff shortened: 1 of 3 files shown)


nikolasrieble:
This looks like a great improvement. Please do consider to add a storybook story - since we have moved to storybook since. 

Also please lets us know if you need any input or help here. Happy to help however possible.

ckifer:
I also feel like this is a breaking change. It changes behavior of charts that have values in the parent currently.

nikolasrieble:
@mariodsantana We are about to start working on 3.0, this could make it in, if you would like to pick it up again.

ckifer:
@mariodsantana we're at a point we could merge this in if you'd like to update it

mariodsantana:
Apologies for the slow reply.  I was hoping to make time to work on this but I'm off the project I needed it for, and unfortunately, just realistically I'm not going to be able to finish this any time soon.

ckifer:
Going to close this as stale for now

Issue #3012, opened 2022-10-12
Fix dependency vulnerability d3-color < 3.1.0


StanDeveloper:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Updating the package will, fix the Redos vulnerability of the d3-color; https://github.com/advisories/GHSA-36jr-mh4h-2g58

### What does the proposed API look like?
No api needed

timges:
Wanted to post the same. Imo this should have high priority as the vulnerability is of high severity. 👍🏼

zavan:
I think #3009 solves this.

timges:
Recharts downgraded the version's again with today's release, so i guess dependabot's pr is not sufficient, as the checks are failing anyways

vitalii-bulyzhyn:
> Release 2.1.15 downgraded the version's again today. I guess dependabot's pr ( #3009 ) is not sufficient, as the checks are failing anyways

+ 1 for this, still the same problem

nftnftnftgem:
+1

emirefek:
Are there any fix for this? What do you people suggest?
My dependabot and npm audit really going crazy about it.

mark-todd:
Dependabot also has a PR for this it turns out #3009

G-Rath:
@mark-todd it's not that simple because v3 of `d3-color` is ESM based, meaning its not directly compatible with `recharts` - the upgrade you're suggesting has actually already been done and then [reverted](https://github.com/recharts/recharts/commit/a78ed04256a9497cfbc6645ff0522a413daf9f75) in v2.1.15 because it was breaking.

To resolve this properly one of three things needs to happen:
  1. `recharts` does a new major version that switches it to using v3 of `d3-color` (via the other `d3` dependencies)
    - `recharts` itself technically doesn't have to move to ESM, but it's still a breaking change because downstream consumers need to support ESM or transpile the modules
  2. someone forks the related modules, publishes them to npm, and switches recharts over to using that
    - this would be a lot of work, and just be an overall annoyance; but maybe someone out there is able to spare the time?
  3. `d3-color` releases a v2 with the fix
    - this would be the most optimal, though @mbostock has said [they won't be doing this](https://github.com/d3/d3-color/issues/108#issuecomment-1271705762); I have [asked them to reconsider](https://github.com/d3/d3-color/issues/108#issuecomment-1282822581) but I'm not super hopeful

[...]

PR #3120, opened 2022-12-27
fix: patch d3 libraries with victory-vendor to fix security vulnerability

ckifer:
## Description

D3 moved to vending ESM only. Our consumers use jest, next, etc. which requires cjs - if we patch d3 security vulnerability in a patch fix we break our consumers. In order for v2.x.x to be secure we need a cjs version of patched d3.

`victory-vendor` does this for us as per blog post https://formidable.com/blog/2022/victory-esm/

change d3 references to `victory-vendor`, remove direct d3 dependencies

## Related Issue

#3012 #3009, more

## Motivation and Context

Fix dependency vulnerability that will not go away in 2.x.x due to d3 not wanting to upgrade

## How Has This Been Tested?

See comment here https://github.com/recharts/recharts/issues/3012#issuecomment-1366152741

Created two duplicate branches. One I upgraded the dependencies directly, this one I replaced with `victory-vendor`. Due to adding jest tests to our test suite we now have a failing build when directly upgrading the deps.

Many references are also in our demo project - these continue to work as well. Tested manually.
`npm run test`
`npm run build`
`npm run demo`

## Types of changes

## Checklist:


ckifer:
Built and tested code in https://github.com/ckifer/recharts/releases/tag/v2.3-beta.2 my fork. Installed and used in a next project with success.

ckifer:
image
Use with next

nikolasrieble:
Do we have a reproducible example where the upgrade broke an application?

ckifer:
@nikolasrieble yes and no. I'm not sure what the release process is for recharts or else I was going to produce a real alpha release. Otherwise I've been copying my built folders to replace what is in node_modules in a project. I've reproduced both success and fail but it's hard to give that reproduction to others.

package.json
```diff
@@ -59,20 +59,15 @@
...
-    "d3-shape": "^2.0.0",
     "eventemitter3": "^4.0.1",
     "lodash": "^4.17.19",
     "react-is": "^16.10.2",
     "react-resize-detector": "^7.1.2",
     "react-smooth": "^2.0.1",
     "recharts-scale": "^0.4.4",
-    "reduce-css-calc": "^2.1.8"
+    "reduce-css-calc": "^2.1.8",
+    "victory-vendor": "^36.6.8"
```
nikolasrieble, line 70: Could you please document the context of this workaround and the next steps?
You could write a comment in the package.json right above this line.

nikolasrieble (changes requested):
Code wise not much is happening here, this is good to go.

What is missing is documentation on why this was done. This knowledge is right now distributed across the slack channel and multiple issues.

Please do write a summary comment about the motivation for this fix, and the next steps to take here (and when to take them).

package.json
ckifer: Imo we might need a better more visible place for documenting some of the technical decisions we're making like this one. Not sure where though

ckifer: Comments are not permitted in `json` files, I will add my documentation steps in the thread instead of the review so they're easier to see

ckifer:
since comments aren't permitted in json what I did was create an FAQ in the GitHub wiki
https://github.com/recharts/recharts/wiki

then I added a few lines to the README
https://github.com/recharts/recharts/blob/master/README.md?plain=1#L20-L22

I think we need a place to document technical decisions without needing to deploy the website, wiki fits this decently. README itself could get messy but we could add a smaller FAQ to the readme for more visibility. Lets discuss.

This isn't pressing to merge since I can't release without Arthur

ckifer:
Closed by #3167

Issue #2963, opened 2022-09-09
Incorrect type of IconType


rongsen-intuit:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://www.typescriptlang.org/play?#code/C4TwDgpgBAMhDmEB2ATAKuaBeAUFKAPlAORgA2AhgJZJk0TF6El1INNHEDOAjgK4UATu3ydhAY2CNRJcVUHiyI5sXGCA9ly7SVKKhQC261Ds5dgQ0yWCD9SeEqvEA7iGWckxhgG4cOUJBQAJLixhiBWFAA8gZUwAA8cIio4RAANCSebMQAfL4A9Pn4UAB6APx AdAhYZgATFCRAKIAHop8KBCJCMjomBnEWQx5OIXF5ThAA](https://www.typescriptlang.org/play?#code/C4TwDgpgBAMhDmEB2ATAKuaBeAUFKAPlAORgA2AhgJZJk0TF6El1INNHEDOAjgK4UATu3ydhAY2CNRJcVUHiyI5sXGCA9ly7SVKKhQC261Ds5dgQ0yWCD9SeEqvEA7iGWckxhgG4cOUJBQAJLixhiBWFAA8gZUwAA8cIio4RAANCSebMQAfL4A9Pn4UAB6APx AdAhYZgATFCRAKIAHop8KBCJCMjomBnEWQx5OIXF5ThAA)

### Steps to reproduce
1. Go to https://www.typescriptlang.org/play?#code/C4TwDgpgBAMhDmEB2ATAKuaBeAUFKAPlAORgA2AhgJZJk0TF6El1INNHEDOAjgK4UATu3ydhAY2CNRJcVUHiyI5sXGCA9ly7SVKKhQC261Ds5dgQ0yWCD9SeEqvEA7iGWckxhgG4cOUJBQAJLixhiBWFAA8gZUwAA8cIio4RAANCSebMQAfL4A9Pn4UAB6APx AdAhYZgATFCRAKIAHop8KBCJCMjomBnEWQx5OIXF5ThAA
1. You should see that IconType is using the wrong utility type to exclude 'none' from LegendType and the end result is not what we expect.

### What is expected?
1. IconType should list all other values except 'none'
2. Use Exclude instead of Omit in union types

### What is actually happening?
1. IconType does not list all possible values excluding 'none'
2. Wrong use of Omit in a union type

[...]

PR #3122, opened 2022-12-30
Fix IconType type by replacing Omit utility type with Exclude

lukask-proxora:
## Description

The `IconType` type used the `Omit` utility type to remove `'none'` from `LegendType`. This results in a wrong type since `Omit` only works on object types or interfaces. As a result, any string was accepted as value for `IconType`.
Using the `Exclude` utiltiy type instead, the type is defined as expected and allows all values from  `LegendType` apart from `'none'`.

## Related Issue

- #2963

## Motivation and Context

The `iconType` prop was not typesafe since the `IconType` type was wrong.

## How Has This Been Tested?

I did not write unit tests since the interesting ones would not compile.
I tested the fix by making sure that a TS error is thrown if I try to build a `Legend` component with `iconType="notavalidicontype". Before the fix, this did compile.

In addition I ran the demo app to see if the legends still work as expected.

## Types of changes

## Checklist:

### DIFF
--- a/src/component/DefaultLegendContent.tsx
+++ b/src/component/DefaultLegendContent.tsx
@@ -16,7 +16,7 @@ import {
 
 const SIZE = 32;
 export type ContentType = ReactElement | ((props: Props) => ReactNode);
-export type IconType = Omit<LegendType, 'none'>;
+export type IconType = Exclude<LegendType, 'none'>;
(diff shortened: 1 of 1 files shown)


ckifer (approved):
thank you @lukask-proxora!

I checked this one out myself the other day too, this is the right change for sure. Approved and merging :)

Issue #3125, opened 2023-01-02
[Jest Migration] LineChart.spec.js -> LineChart.spec.tsx 


ckifer:
As part of the jest migration we want to move all current jest tests to typescript. Issues will be made to encourage small chunks of refactoring done at once.

Refactor `LineChart.spec.js -> LineChart.spec.tsx` and any existing bar chart tests from karma

Acceptance criteria: 

* All bar chart tests are in TS + RTL + Jest at ` LineChart.spec.tsx`
* `LineChart.spec.js` is deleted 
* All tests pass

turculaurentiu91:
Since the last issue was easy, I would like to work on this one too. Can you assign it to me?

ckifer:
@turculaurentiu91 all you!

ckifer:
Closed by #3141

PR #3141, opened 2023-01-04
LineChart.spec jest migration

turculaurentiu91:
## Description

Migrated unit tests for LineChart.spec to typescript / RTL

## Related Issue

#3125 

## Motivation and Context

Migrate unit tests to jest/RTL/typescript

## How Has This Been Tested?

`npm run test` on ubuntu linux with node 16

## Types of changes

- [x] Refactoring

## Checklist:

### DIFF
--- a/test-jest/chart/LineChart.spec.tsx
+++ b/test-jest/chart/LineChart.spec.tsx
@@ -1,14 +1,18 @@
-/* eslint-disable react/no-unused-state */
-/* eslint-disable max-len */
-/* eslint-disable import/order */
-/* eslint-disable no-return-assign */
-/* eslint-disable react/prop-types */
-/* eslint-disable import/no-extraneous-dependencies */
-/* eslint-disable react/jsx-filename-extension */
-/* eslint-disable no-undef */
-import React from 'react';
-import { LineChart, Line, Curve, XAxis, YAxis, CartesianAxis, Tooltip, Brush, Legend } from '../../src';
+import React, { FC } from 'react';
 import { render, fireEvent } from '@testing-library/react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { JSX } from '@babel/types';
+
+import {
+  LineChart,
+  Line,
+  /* Curve, */ XAxis,
+  YAxis,
+  /* CartesianAxis, */ Tooltip,
+  Brush /* Legend */,
+  CartesianAxis,
+  Legend,
+} from '../../src';
 
 const data = [
   { name: 'Page A', uv: 400, pv: 2400, amt: 2400 },
@@ -61,6 +65,11 @@ describe('<LineChart />', () => {
     const curves = container.querySelectorAll('.recharts-line .recharts-line-curve');
     expect(curves).toHaveLength(1);
     const path = curves[0].getAttribute('d');
+
+    if (!path) {
+      throw new Error('Path is null');
+    }
+
     expect(path.indexOf('C')).not.toEqual(-1);
   });
 
@@ -84,6 +93,11 @@ describe('<LineChart />', () => {
     const curves = container.querySelectorAll('.recharts-line .recharts-line-curve');
     expect(curves).toHaveLength(1);
     const path = curves[0].getAttribute('d');
+
+    if (!path) {
+      throw new Error('Path is null');
+    }
+
     expect(path.length - path.split('M').join('').length).toEqual(2);
   });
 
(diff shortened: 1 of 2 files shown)


test-jest/chart/LineChart.spec.tsx
```diff
@@ -522,100 +412,109 @@ describe('<LineChart /> - Rendering two line charts with syncId', () => {
...
-        <XAxis dataKey="name" />
-      </LineChart>
-    );
+  // type ActiveDotProps = {
+  //   cx?: number;
+  //   cy?: number;
+  // };
 
+  // eslint-disable-next-line max-len
+  test.skip('should show tooltips for both charts synced by index on MouseEnter and hide on MouseLeave/Escape', async () => {
```
turculaurentiu91, line 423: I am having trouble triggering the tool-tip for those tests. Tried different approaches, `fireEvent` with `pageX` & `pageY` manually set, [RTL User Interaction Hover](https://testing-library.com/docs/user-event/convenience#hover) but I cannot make the '.recharts-tooltip-cursor' pop. Can anybody give me a hand?

ckifer: We might hit some limitations of `jsdom` here - try logging and seeing if you're getting any actual values set for pageX and pageY

Usually for `fireEvent` I will do something like the following:

```ts
const firstChart = container.querySelector('.recharts-wrapper');

fireEvent.mouseEnter(firstChart);
```

turculaurentiu91: In this case, pageX will be undefined. Was the same for the Brush test, where I had to set pageX using Object.assign as you cannot set it directly with `fireEvent`. where it works `mousedown` and `mousemove` I cannot make it work for `mouseEnter`. 

https://github.com/recharts/recharts/pull/3141/files#diff-467d743ad34119402d9ef9d760c90b7ba4053cab936f4fa4288db04b7fecef2eR364

I will do some logs later and see what happens.

test-jest/chart/LineChart.spec.tsx
```diff
@@ -61,6 +65,11 @@ describe('<LineChart />', () => {
     const curves = container.querySelectorAll('.recharts-line .recharts-line-curve');
     expect(curves).toHaveLength(1);
     const path = curves[0].getAttribute('d');
+
+    if (!path) {
+      throw new Error('Path is null');
+    }
+
     expect(path.indexOf('C')).not.toEqual(-1);
```
nikolasrieble, line 73: Instead of throwing an error you could simply use conditional invocation. 

```suggestion
    expect(path?.indexOf('C')).not.toEqual(-1);
```

test-jest/chart/LineChart.spec.tsx
nikolasrieble: I also looked into this and had the same issue. 
You could consider to solve the issue in two PRs, and solve the Tooltip tests in another PR, still getting in most changes now.

ckifer: Hmm, this might be why karma was used initially. Karma uses Headless or actual Chrome/Firefox to run tests - Jest uses `jsdom` which doesn't use the actual browser, it just acts like it.

Adding E2E type tests with something like Cypress/Puppeteer/Selenium/etc. could test for these things but that would be a large and separate effort + more things to maintain

turculaurentiu91: I have good news, I can make them pass. The following console.log is set here https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.tsx#L1349 inside the handler

`fireEvent.mouseEnter(firstChart);` logs -> `pageX:  undefined`

```javascript
const mouseEnterEvent = new MouseEvent('mouseenter', { bubbles: true, cancelable: true });
    Object.assign(mouseEnterEvent, { pageX: 100, pageY: height / 2 });
    fireEvent(firstChart, mouseEnterEvent);
```
doesn't log, the event is not triggered, however,

```javascript
const mouseEnterEvent = new MouseEvent('mouseover', { bubbles: true, cancelable: true });
    Object.assign(mouseEnterEvent, { pageX: 100, pageY: height / 2 });
    fireEvent(firstChart, mouseEnterEvent);
``` 

logs -> `pageX:  100`

So I can use `mouseover` instead of `mouseenter`.

There is another complexity, I have to mock the `window.requestAnimationFrame` function, otherwize it will not pop the second popup when using `syncId`. I will finish this PR over the weekend.

test-jest/chart/LineChart.spec.tsx
```diff
@@ -84,6 +79,11 @@ describe('<LineChart />', () => {
     const curves = container.querySelectorAll('.recharts-line .recharts-line-curve');
     expect(curves).toHaveLength(1);
     const path = curves[0].getAttribute('d');
+
+    if (!path) {
+      throw new Error('Path is null');
+    }
```
ckifer, line 85: nit: all of these types of errors could be negative assertions `.toBeNull`, etc.

fine for now

ckifer (approved):

test-jest/chart/LineChart.spec.tsx
turculaurentiu91: yes but you will have to deal with typescript too

image

So you have to do use a lot of null coalescing and optional chaining, I find it more elegant to just throw an error.

```javascript
const path = curves[0].getAttribute('d');

  /* if (!path) {
    throw new Error('Path is null');
  } */

  expect(path).not.toBeNull();

  expect((path?.length ?? 0) - (path?.split('M')?.join('')?.length ?? 0)).toEqual(2);
```

ckifer: Eh to each their own. I would rather use one non-null assertion (!) that it will fail on if it's null (for tests only of course) than optional chaining or throwing an error. It's style so all good

Issue #3126, opened 2023-01-02
[Jest Migration] AreaChart.spec.jsx -> AreaChart.spec.tsx


ckifer:
As part of the jest migration we want to move all current jest tests to typescript. Issues will be made to encourage small chunks of refactoring done at once.

Refactor `AreaChart.spec.jsx -> AreaChart.spec.tsx` and any existing bar chart tests from karma

Acceptance criteria: 

* All bar chart tests are in TS + RTL + Jest at `  AreaChart.spec.tsx`
* ` AreaChart.spec.jsx` is deleted
* All tests pass

Yilun-Sun:
working on it

ckifer:
Closed by #3151 

@Yilun-Sun can you create a follow up issue to migrate `AreaChartSpec.js`, uncomment the skips, etc? Thanks!

Yilun-Sun:
> Closed by #3151
> 
> @Yilun-Sun can you create a follow up issue to migrate `AreaChartSpec.js`, uncomment the skips, etc? Thanks!

Yes, #3161 is the one.

TODOs:
- [ ] brush move test
- [ ] add more test

Yilun-Sun:
all related PRs:
- #3151 
- #3161 
- #3163

PR #3151, opened 2023-01-06
fix: migrate area chart test to typescript

Yilun-Sun:
## Related Issue

#3126

## Motivation and Context

Migrate AreaChart tests to jest with typescript

## Types of changes

## Checklist:

### DIFF
--- a/test-jest/chart/AreaChart.spec.tsx
+++ b/test-jest/chart/AreaChart.spec.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable no-undef */
-import { fireEvent, render } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import React from 'react';
-import sinon from 'sinon';
+import userEvent from '@testing-library/user-event';
 
 import { Area, AreaChart, Brush, CartesianAxis, Tooltip, XAxis, YAxis } from '../../src';
 
@@ -36,7 +36,7 @@ describe('AreaChart', () => {
     expect(container.querySelectorAll('.recharts-area-dot')).toHaveLength(1);
   });
 
-  test('Renders empty path when all the datas are null', () => {
+  test('Renders empty path when all the data are null', () => {
     const { container } = render(
       <AreaChart width={100} height={50} data={data}>
         <Area type="monotone" dataKey="any" stroke="#ff7300" fill="#ff7300" />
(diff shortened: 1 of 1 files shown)


test-jest/chart/AreaChart.spec.tsx
```diff
@@ -55,50 +55,37 @@ describe('AreaChart', () => {
     });
   });
 
-  test.skip('Renders customized active dot when activeDot is set to be a ReactElement', () => {
+  test.skip('Renders customized active dot when activeDot is set to be a ReactElement', async () => {
```
nikolasrieble, line 58: Would you like to try to also get the test to work? Or would you prefer to do that in a separate PR?

test-jest/chart/AreaChart.spec.tsx
```diff
@@ -55,50 +55,37 @@ describe('AreaChart', () => {
...
     // eslint-disable-next-line react/prop-types
```
nikolasrieble, line 59: Is this still needed?

test-jest/chart/AreaChart.spec.tsx
```diff
@@ -55,50 +55,37 @@ describe('AreaChart', () => {
...
-    fireEvent.mouseEnter(container, { clientX: 200, clientY: 200 });
-    // fireEvent.mouseEnter(container.querySelector('.recharts-curve'));
+    const area = container.querySelector('.recharts-area-area');
 
-    expect(container.querySelectorAll('.recharts-active-dot')).toHaveLength(1);
-  });
+    if (area !== null) {
+      await userEvent.hover(area);
+      screen.debug(area);
+    }
```
nikolasrieble, line 85: Is the debug needed in the test?
I use it while writing tests, but always remove it in the end.

test-jest/chart/AreaChart.spec.tsx
ckifer: I think this skip was because of the `jsdom` limitations (can't find things on mouse enter/hover

test-jest/chart/AreaChart.spec.tsx
ckifer: +1

test-jest/chart/AreaChart.spec.tsx
ckifer: I think since its in a skip its fine for now

ckifer (approved):
I think this is fine since we're just moving existing jest tests to TS

We should make a follow up issue to get rid of the skips, make tests better, migrate everything else from `AreaChartSpec.js`, etc. We won't delete `AreaChartSpec.js` until that is done

PR #3152, opened 2023-01-06
ReactUtils – find by type – refactor using generic

marcalexiei:
## Description

In order to reduce the `any` types in the code I change the implementation of `findAllByType` and `findChildByType` to use generic to better infer component props in the return value:

Screenshot 2023-01-06 at 13 55 24

To achieve this now these two functions only accepts `React.ComponentType` parameter instead of `string` in order to retrieve the `Props`. 
All invocation of the function have been changed. Most of them was using `Component.displayName` so since `getDisplayName` accounts for both `string` and `React.ComponentType` this should't be a problem.

---

While doing this I encountered an error in `appendOffsetOfLegend` function which was comparing `verticalAlign` `Legend` props with `"center"`.  It looks like it isn't a valid value so I changed it to `"middle"` since `VerticalAlignmentType ` values are 
```ts
'top' | 'bottom' | 'middle'
```

[Commit reference](https://github.com/recharts/recharts/commit/5a4a9874f2b6f20877cc73b4338dbfd06c2745fe)

---

Let me know if you prefer to have two separate PR for the issue and the refactor.

## Related Issue
N/A

## Motivation and Context
Improve code navigation and reference

## How Has This Been Tested?

`npm run build`
`npm run test`

N/A

## Types of changes

## Checklist:

### DIFF
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -387,7 +387,7 @@ export const appendOffsetOfLegend = (offset: any, items: Array<FormattedGraphica
     const box = legendBox || {};
     const { align, verticalAlign, layout } = legendProps;
 
-    if ((layout === 'vertical' || (layout === 'horizontal' && verticalAlign === 'center')) && isNumber(offset[align])) {
+    if ((layout === 'vertical' || (layout === 'horizontal' && verticalAlign === 'middle')) && isNumber(offset[align])) {
       newOffset = { ...offset, [align]: newOffset[align] + (box.width || 0) };
     }
 
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -11,8 +11,8 @@ import _ from 'lodash';
 import { ReactElement, ReactNode } from 'react';
 import { getNiceTickValues, getTickValuesFixedDomain } from 'recharts-scale';
 
+import { ErrorBar } from '../cartesian/ErrorBar';
 import { Legend } from '../component/Legend';
-
 import { findEntryInArray, getPercentValue, isNumber, isNumOrStr, mathSign, uniqueId } from './DataUtils';
 import { filterProps, findAllByType, findChildByType, getDisplayName } from './ReactUtils';
 // TODO: Cause of circular dependency. Needs refactor.
@@ -24,8 +24,8 @@ export function getValueByDataKey<T>(obj: T, dataKey: DataKey<any>, defaultValue
     return defaultValue;
   }
 
-  if (isNumOrStr(dataKey as string)) {
-    return _.get(obj, dataKey as string, defaultValue);
+  if (isNumOrStr(dataKey)) {
+    return _.get(obj, dataKey, defaultValue);
   }
 
   if (_.isFunction(dataKey)) {
@@ -181,7 +181,7 @@ export const getLegendProps = ({
   legendWidth: number;
   legendContent?: any;
 }) => {
-  const legendItem = findChildByType(children, Legend.displayName);
+  const legendItem = findChildByType(children, Legend);
   if (!legendItem) {
     return null;
   }
@@ -429,7 +429,7 @@ export const getDomainOfErrorBars = (
   axisType?: AxisType,
 ) => {
   const { children } = item.props;
-  const errorBars = findAllByType(children, 'ErrorBar').filter((errorBarChild: any) =>
+  const errorBars = findAllByType(children, ErrorBar).filter(errorBarChild =>
     isErrorBarRelevantForAxis(layout, axisType, errorBarChild.props.direction),
   );
 
--- a/src/util/ReactUtils.ts
+++ b/src/util/ReactUtils.ts
@@ -60,7 +60,7 @@ export const TOOLTIP_TYPES = ['none'];
  * @param  {Object} Comp Specified Component
  * @return {String}      Display name of Component
  */
-export const getDisplayName = (Comp: any) => {
+export const getDisplayName = (Comp: React.ComponentType | string) => {
   if (typeof Comp === 'string') {
     return Comp;
   }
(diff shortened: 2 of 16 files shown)


src/util/ReactUtils.ts
```diff
@@ -97,11 +97,11 @@ export const toArray = <T extends ReactNode>(children: T | T[]): T[] => {
  * Find and return all matched children by type. `type` can be a React element class or
  * string
  */
-export const findAllByType = (
-  children: ReactNode,
-  type: string | string[],
-): React.DetailedReactHTMLElement<any, HTMLElement>[] => {
-  const result: React.DetailedReactHTMLElement<any, HTMLElement>[] = [];
+export function findAllByType<
+  Cmp extends React.ComponentType,
```
nikolasrieble, line 101: Nit: Lets not use abbreviations. Instead, please write out the fields.
```suggestion
  Component extends React.ComponentType,
```

src/util/ReactUtils.ts
```diff
@@ -97,11 +97,11 @@ export const toArray = <T extends ReactNode>(children: T | T[]): T[] => {
...
+  ReturnCmp = React.DetailedReactHTMLElement<React.ComponentProps<Cmp>, HTMLElement>,
```
nikolasrieble, line 102:
```suggestion
  ReturnComponent = React.DetailedReactHTMLElement<React.ComponentProps<Component>, HTMLElement>,
```

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
nikolasrieble, line 1: I do not think we need to add vscode settings to recharts.

nikolasrieble (changes requested):
Please remove the vscode settings. Other than that, the change looks good! Thank you for the refactoring.

.vscode/settings.json
ckifer: +1, haven't looked at the rest of the PR yet but editor settings should be in our gitignore. Each devs editor settings are their own and we'll let eslint/prettier take care of formatting

src/util/ChartUtils.ts
```diff
@@ -385,7 +386,7 @@ export const appendOffsetOfLegend = (offset: any, items: Array<FormattedGraphica
     const box = legendBox || {};
     const { align, verticalAlign, layout } = legendProps;
 
-    if ((layout === 'vertical' || (layout === 'horizontal' && verticalAlign === 'center')) && isNumber(offset[align])) {
+    if ((layout === 'vertical' || (layout === 'horizontal' && verticalAlign === 'middle')) && isNumber(offset[align])) {
```
nikolasrieble, line 389: As this is a clear bug fix, I think we are good to keep this change. Center was never an option :)

marcalexiei:
I have renamed generic to avoid abbreviation usage and added `.vscode` folder to .gitignore

marcalexiei:
I would consider to merge and not squash this PR to have the fix clearly visible in the commit history. :-)

src/util/ChartUtils.ts
ckifer: +1, nice find

src/util/ReactUtils.ts
```diff
@@ -97,11 +97,11 @@ export const toArray = <T extends ReactNode>(children: T | T[]): T[] => {
  * Find and return all matched children by type. `type` can be a React element class or
  * string
```
ckifer, line 98: Can we update this comment? It looks like now `type`  can only be `ComponentType | ComponentType[]` which does not include string

Also does this work with a Function component?

src/util/ReactUtils.ts
```diff
@@ -110,27 +110,28 @@ export const findAllByType = (
...
     }
   });
 
   return result;
-};
+}
+
 /*
  * Return the first matched child by type, return null otherwise.
  * `type` can be a React element class or string.
```
ckifer, line 125: Same here, comment needs updated. String no longer accepted and curious if function components get desired results.

src/util/ReactUtils.ts
```diff
@@ -110,27 +110,28 @@ export const findAllByType = (
...
-): React.DetailedReactHTMLElement<any, HTMLElement> => {
+export function findChildByType<
+  ComponentType extends React.ComponentType,
+  DetailedElement = React.DetailedReactHTMLElement<React.ComponentProps<ComponentType>, HTMLElement>,
+>(children: ReactNode[], type: ComponentType | ComponentType[]): DetailedElement {
   const result = findAllByType(children, type);
 
-  return result && result[0];
-};
+  return (result && result[0]) as never;
```
ckifer, line 133: I would rather avoid the cast to `never` here if we could

Instead of creating `DetailedElement` can we just make the return type
`React.DetailedReactHTMLElement<React.ComponentProps<ComponentType>, HTMLElement>`

so the the cast to never can be removed?

Or if we don't want the return type to be so long to do something like
```ts
type DetailedElement<ComponentType extends React.ComponentType> = React.DetailedReactHTMLElement<
  React.ComponentProps<ComponentType>,
  HTMLElement
>;
...
export function findChildByType<ComponentType extends React.ComponentType>(
  children: ReactNode[],
  type: ComponentType | ComponentType[],
): DetailedElement<ComponentType> {
```

ckifer (changes requested):

src/util/ReactUtils.ts
marcalexiei: > Can we update this comment?

Sure, I haven't noticed that earlier. 😅

> Also does this work with a Function component?

`ComponentType` includes both component types
```ts
type ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>;
```

src/util/ReactUtils.ts
marcalexiei: I removed the return type alongside the `never` casting.
TS infers all correct types from `findAllByType`.

ckifer (approved):
LGTM, thanks for doing this!

Will merge this with a merge commit. 

I'm going to have to mull on if to merge this yet or wait until after I release 2.3.0 - we're going to need to adopt a different branching strategy at some point I think (@nikolasrieble)

marcalexiei:
> I'm going to have to mull on if to merge this yet or wait until after I release 2.3.0

In the end this PR contains a bugfix and mostly refactor, so I think is ok to merge since 2.3.0 will contain many bugfixes.
The only notable change is on [`src/util/DetectReferenceElementsDomain.ts`](https://github.com/recharts/recharts/pull/3152/files#diff-55d36b01ae78f58b3f857d1cd10b1f20a33001bd96f5de913f5672cc80ba71b2R18) where I changed:
```ts
const elements = lines.concat(dots);
```
to
```ts
const elements = [...lines, ...dots];
```
That shouldn't cause any issue.

> we're going to need to adopt a different branching strategy at some point I think

When the work on the 3.0 starts you can create a 2.x branch where the fix can be backported from master. 
Meanwhile master will become the house of 3.x version code. This is the simplest approach I can think of.

Issue #3156, opened 2023-01-06
[Jest Migration] ResponsiveContainer


marcalexiei:
We want to move all tests to jest, RTL, and typescript. Issues will be made to encourage small chunks of refactoring done at once.

Refactor `test/specs/component/ResponsiveContainer.jsx -> test-jest/util/ResponsiveContainer.spec.tsx`

https://github.com/recharts/recharts/blob/master/test/specs/component/ResponsiveContainerSpec.js

Acceptance criteria:

- All tests are in TS + RTL + Jest at `ResponsiveContainer.spec.tsx`
- `ResponsiveContainer.jsx` is deleted
- All tests pass

PR #3157, opened 2023-01-06
test(JestMigration): convert ResponsiveContainer

marcalexiei:
## Description

Migrate ResponsiveContainer test to jest.
- Added a test for resizeObserver change

---

Refactored `ResponsiveContainer` using  hooks to memoize variables in order to reduce renders.

---

Moved `chai.use(chaiEnzyme());` inside old test index from deleted ResponsiveContainerSpec.
Without this `'<XAxis /> – Render ticks with tickFormatter` test will fail.

## Related Issue
#3156

## Motivation and Context
N/A

## How Has This Been Tested?
`npm run build`
`npm run test`
`npm run demo` (used ResponsiveContainer examples)

N/A

## Types of changes

## Checklist:


src/component/ResponsiveContainer.tsx
```diff
@@ -60,28 +67,31 @@ export const ResponsiveContainer = forwardRef(
...
+    }, [getContainerSize]);
 
-    const handleResize = debounce > 0 ? _.debounce(updateDimensionsImmediate, debounce) : updateDimensionsImmediate;
+    const handleResize = useMemo(
+      () => (debounce > 0 ? _.debounce(updateDimensionsImmediate, debounce) : updateDimensionsImmediate),
+      [debounce, updateDimensionsImmediate],
+    );
 
-    const renderChart = () => {
+    const computedSizes = useMemo(() => {
```
nikolasrieble, line 94: Would it make sense to extract this as a unit testable helper function, now that it only computes the size?

test-jest/component/ResponsiveContainer.spec.tsx
```diff
@@ -0,0 +1,157 @@
...
+    return {
+      observe: jest.fn().mockImplementation(),
+      unobserve: jest.fn(),
+      disconnect: jest.fn(),
+    };
+  });
+  const consoleWarnSpy = jest.spyOn(console, 'warn').mockImplementation(() => undefined);
+
+  beforeAll(() => {
+    delete window.ResizeObserver;
```
nikolasrieble, line 27: This is a very interesting test setup. Could you please add a comment as to the motivation and usage? 

Thus others might learn from this for further tests.

test-jest/component/ResponsiveContainer.spec.tsx
```diff
@@ -0,0 +1,157 @@
...
+    const { container } = render(
+      <ResponsiveContainer>
+        <div data-testid="inside" />
+      </ResponsiveContainer>,
+    );
+
+    expect(container.querySelector('.recharts-responsive-container')).not.toHaveAttribute('id');
+  });
+
+  it('Renders without an id attribute when not passed', () => {
```
nikolasrieble, line 131: This test name seems wrong. We are passing an id here.

test-jest/component/ResponsiveContainer.spec.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <ResponsiveContainer aspect={2} height={300} width={100}>
+        <div data-testid="inside" />
+      </ResponsiveContainer>,
+    );
+
+    expect(screen.getByTestId('inside')).toHaveAttribute('width', '100');
+    expect(screen.getByTestId('inside')).toHaveAttribute('height', '50');
+  });
+
+  it('Renders without an id attribute when  passed', () => {
```
nikolasrieble, line 121:
```suggestion
  it('Renders without an id attribute when not passed', () => {
```

nikolasrieble (commented):
Great changes. What do you think about adding a test plan to verify feature of the ResponsiveContainer in the demo?

src/component/ResponsiveContainer.tsx
marcalexiei: I don't think this is necessary, right now the majority of tests of `ResponsiveContainer` involve this function.
We might consider to move it away if we have to use it on a different component, but for the moment I think is fine to leave it here.

ckifer: I think its fine to leave here for now as well

src/component/ResponsiveContainer.tsx
```diff
@@ -60,28 +67,31 @@ export const ResponsiveContainer = forwardRef(
...
       const newSize = getContainerSize();
 
       if (newSize) {
-        const { containerWidth: oldWidth, containerHeight: oldHeight } = sizes;
         const { containerWidth, containerHeight } = newSize;
 
-        if (containerWidth !== oldWidth || containerHeight !== oldHeight) {
-          setSizes({ containerWidth, containerHeight });
-        }
+        setSizes(currentSizes => {
```
ckifer, line 78: this has to be a bug fix since referencing previous values with the state itself is an antipattern and causes weird renders. Nice!

ckifer (commented):
Looks good to me pending the already suggested changes

marcalexiei:
- Added comments about `ResizeObserver` mock
- Test description fixed (😅), probably I should get some sleep...

ckifer (approved):
LGTM, thank you! Get some sleep @marcalexiei  😅

PR #3163, opened 2023-01-09
fix: area chart test part3

Yilun-Sun:
## Description

thrid part for #3151

## Motivation and Context

For the jsdom mouse event problem, thanks for @turculaurentiu91 and the PR #3141 is the reference

## Types of changes

## Checklist:


test-jest/chart/AreaChart.spec.tsx
```diff
@@ -222,5 +220,38 @@ describe('AreaChart', () => {
...
+      const rightBrushTraveler = container.querySelectorAll('.recharts-brush-traveller')[1];
+
+      const mouseDownEvent = mockMouseEvent('mousedown', rightBrushTraveler!, { pageX: 400, pageY: 0 });
+      const mouseMoveEvent = mockMouseEvent('mousemove', window, { pageX: 0, pageY: 0 });
+
+      mouseDownEvent.fire();
+      mouseMoveEvent.fire();
+      fireEvent.mouseUp(window);
+
+      // not sure why the drag ended at x: 65, but it did close to the 1st point but not 2nd
```
ckifer, line 252: just for my understanding this (and the above test) is saying that when brush is moved all the way to the left or the right the attributes stop being calculated after a certain point? Maybe once only one point is present it stops caring?

Would be nice to understand this (not in context of this test but just in general)

ckifer (approved):

Issue #3029, opened 2022-10-21
Recharts 2.1.15 throwing Resize Observer Error 


RajB07:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-k2xqp)

### Steps to reproduce
-

### What is expected?
Bug to be fixed soon

### What is actually happening?
Recharts 2.1.15 throwing Resize Observer Error 
  The above error occurred in the <ResizeDetector> component:      
          at ResizeDetector 
 Error: Uncaught [TypeError: _this.resizeObserver.observe is not a function]
          at reportException 

| Environment | Info |
|---|---|
| Recharts | v2.1.1 |
| React | 17 |
| System | - |
| Browser | - |

---
Jest version - 28.1.2
react testing lib version - 12.1.4

vitalikda:
Didn't get why, but `ResizeObserver` is no longer available during jest tests. This fixed my failed test: 

```js
global.ResizeObserver = jest.fn().mockImplementation(() => {
  return {
    observe: jest.fn(),
    disconnect: jest.fn(),
    unobserve: jest.fn()
  }
})
```

RajB07:
@vitalikda I had tried this but the test cases are still failing. It shows the same Resize Observer error.

cjnoname:
Same error here, anyone can help?

ckifer:
Can anyone effected by this issue give a better description of what the issue is? The CodeSanbox is not a reproduction and it is not clear weather this is an error only thrown in tests or otherwise. Thank you

ckifer:
Closing this as this is a general known thing with `ResizeObserver`

Please refer to:
https://github.com/maslianok/react-resize-detector#testing-with-enzyme-and-jest

https://github.com/maslianok/react-resize-detector/issues/145

To resolve the issue. This is not specific to `recharts`. Thanks!

GreenGremlin:
[...]

PR #3175, opened 2023-01-13
fix: ResponsiveContainer debounce updating early

GreenGremlin:
## Description

Passing a debounce value to `ResponsiveContainer` currently only debounces the `handleResize` callback in `ResponsiveContainer` but not the callback in `ReactResizeDetector`. Since `ReactResizeDetector` tracks the updated height and width locally and passes it to its child, sizes are still updated for every call. Since this can result in expensive state updates, it introduces a race condition where `ResizeObserver` sometimes fires a "ResizeObserver loop limit exceeded" error on the `window` object. 

The fix is to use `ReactResizeDetector`'s debounce behavior so that it's internal callback is also debounced.

## Related Issue

https://github.com/recharts/recharts/issues/3029

## Types of changes

## Checklist:

- [ ] ~~My change requires a change to the documentation.~~
- [ ] ~~I have updated the documentation accordingly.~~

### DIFF
--- a/test-jest/component/ResponsiveContainer.spec.tsx
+++ b/test-jest/component/ResponsiveContainer.spec.tsx
@@ -152,7 +152,7 @@ describe('<ResponsiveContainer />', () => {
     expect(container.querySelector('.recharts-responsive-container')).toHaveAttribute('id', 'testing-id-attr');
   });
 
-  it.only('should resize when ResizeObserver notify a change', () => {
+  it('should resize when ResizeObserver notify a change', () => {
     const { container } = render(
       <ResponsiveContainer id="testing-id-attr" width="100%" height={200}>
         <div data-testid="inside" />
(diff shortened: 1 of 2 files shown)


test-jest/component/ResponsiveContainer.spec.tsx
```diff
@@ -168,4 +168,26 @@ describe('<ResponsiveContainer />', () => {
...
+        <div data-testid="inside" />
+      </ResponsiveContainer>,
+    );
+
+    const element = container.querySelector('.recharts-responsive-container');
+
+    notifyResizeObserverChange([{ contentRect: { width: 50, height: 50 } }]);
+    jest.advanceTimersByTime(100);
+    expect(element).not.toHaveAttribute('width');
+    expect(element).not.toHaveAttribute('height');
```
GreenGremlin, line 185: Before this change, this check fails because `ReactResizeDetector`'s internal callback is not debounced.

ckifer (approved):
Pulled this down and tested it locally. Makes a lot more sense to use ResizeDetector's internal debounce.

Thank you!

Issue #3159, opened 2023-01-06
[Jest Migration]: Bar


nikolasrieble:
As part of the jest migration we want to move all current jest tests to typescript. Issues will be made to encourage small chunks of refactoring done at once.

Refactor test/specs/cartesian/BarSpec.js -> test-jest/cartesian/Bar.spec.ts and any existing bar chart tests from karma

Acceptance criteria:

All bar chart tests are in TS + RTL + Jest at Bar.spec.ts
BarSpec.js is deleted
All tests pass

akib22:
Hi @nikolasrieble, I want to work on this issue. Can you please assign me to this issue?

nikolasrieble:
Sure, it is yours! ❤️ Feel free to tag me for review.

akib22:
Hi @nikolasrieble, I have created a pull request. Not running the workflow because of first-time contributor approval, and I'm not able to make a review request. Can you please check it?

ckifer:
Closed by #3179

@akib22 feel free to make another PR with the small changes suggested

PR #3179, opened 2023-01-15
test(JestMigration): convert Bar

akib22:
## Description

Refactor `test/specs/cartesian/BarSpec.js` to `test-jest/cartesian/Bar.spec.ts`.

## Related Issue

#3159 

## Types of changes

- [x] Jest Migration

## Checklist:

### DIFF
--- a/test-jest/cartesian/Bar.spec.tsx
+++ b/test-jest/cartesian/Bar.spec.tsx
@@ -1,7 +1,8 @@
+import { render } from '@testing-library/react';
+import * as _ from 'lodash';
 import React from 'react';
-import { expect } from 'chai';
-import { Surface, Bar } from 'recharts';
-import { mount, render } from 'enzyme';
+
+import { Bar, Surface } from '../../src';
 
 describe('<Bar />', () => {
   const data = [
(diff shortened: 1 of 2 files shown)


test-jest/cartesian/Bar.spec.tsx
```diff
@@ -1,7 +1,8 @@
+import { render } from '@testing-library/react';
+import * as _ from 'lodash';
```
nikolasrieble, line 2:
```suggestion
import _ from 'lodash';
```
should be sufficient, right?

test-jest/cartesian/Bar.spec.tsx
```diff
@@ -12,34 +13,34 @@ describe('<Bar />', () => {
     { x: 170, y: 50, width: 20, height: 50, value: 100, label: 'test' },
   ];
 
-  it('Render 5 rectangles in a simple Bar', () => {
+  it(`Render ${data.length} rectangles in a simple Bar`, () => {
     const wrapper = render(
```
nikolasrieble, line 17: Nit: Instead of using the wrapper, best use the screen https://stackoverflow.com/questions/61482418/react-testing-library-screen-vs-render-queries

test-jest/cartesian/Bar.spec.tsx
```diff
@@ -65,29 +66,29 @@ describe('<Bar />', () => {
...
+      );
     });
 
     it('Will accept a function for the background prop', () => {
-      const wrapper = mount(
+      const className = 'test-custom-background';
+      const backgroundComponent = () => {
+        return <div key={_.uniqueId()} className={className} />;
+      };
+      const { container } = render(
```
nikolasrieble, line 85: Nit: It could be helpful, and improve readability, if the arrange/act/assert sections of each test were separated by an empty line. 
```suggestion

      const { container } = render(
```

nikolasrieble (approved):
Left a few nits, please do consider them. 
Lets see that the tests run the workflow and 🚢

test-jest/cartesian/Bar.spec.tsx
akib22:
```js
it('Will create a background Rectangle with the passed in props', () => {
  const { container } = render(
    <Surface width={500} height={500}>
      <Bar data={composedDataWithBackground} background={{ fill: '#000' }} dataKey="value" />
    </Surface>,
  );

  expect(container.querySelectorAll('.recharts-bar-background-rectangle')).toHaveLength(
    composedDataWithBackground.length,
  );
})
```

I can't find a way to use `screen` for this test. `screen.getByRole` does not allow to select `<svg>` or `<path>` tags, so need to use `wrapper`'s utility. Do you have any recommendations?

ckifer: The way that we are testing with classes its very difficult not to use `container.querySelectorAll`. In order to not @nikolasrieble we'd need to come up with a mechanism for getting things by role, by text, by title, etc. If we had a `title` tag we could query by that easily enough, but since we don't we'll have to rely on this method.

ckifer (approved):
LGTM overall, lets 🚢 this and come back to it if we need

test-jest/cartesian/Bar.spec.tsx
ckifer: +1 - default to screen if it can be used as opposed to destructuring functions like `getAllByText`, etc.

Issue #3160, opened 2023-01-06
[Jest Migration]: Brush


nikolasrieble:
As part of the jest migration we want to move all current jest tests to typescript. Issues will be made to encourage small chunks of refactoring done at once.

Refactor test/specs/cartesian/BrushSpec.js -> test-jest/cartesian/Brush.spec.ts and any existing bar chart tests from karma

Acceptance criteria:

All bar chart tests are in TS + RTL + Jest at Brush.spec.ts
BrushSpec.js is deleted
All tests pass

tharitRT:
You must have your react code(JSX) in `.tsx` file. I try change `BrushSpec.js` to `BrushSpec.tsx` and it's working fine. If this is ok I will proceed with creating a PR.

ckifer:
Closed by  #3183

PR #3183, opened 2023-01-19
test(JestMigration): convert Brush

Yilun-Sun:
## Description

Refactor `test/specs/cartesian/BrushSpec.js` -> `test-jest/cartesian/Brush.spec.ts`

## Related Issue

#3160 

## Types of changes

## Checklist:


test-jest/cartesian/Brush.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+    expect(container.querySelectorAll('.recharts-line')).toHaveLength(1);
+  });
+
+  test('mouse enter and mouse leave on traveller will set isTextActive true', () => {
+    const { container } = render(barChartWithBrush);
+
+    const brushSlide = container.querySelector('.recharts-brush-slide');
+
+    const mouseDownEvent = mockMouseEvent('mousedown', brushSlide!, { pageX: 0, pageY: 0 });
```
nikolasrieble, line 69: The test title says MouseEnter and MouseLeave, but in the test we use mouseDown and mouseUp. 

While this test adds value it is not equivalent to the previous test. 

My suggestion: Add this tests as a new one, and also add the existing test with MouseEnter and MouseLeave.

test-jest/cartesian/Brush.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+    { date: '2023-01-09', value: 10 },
+    { date: '2023-01-10', value: 70 },
+    { date: '2023-01-11', value: 40 },
+    { date: '2023-01-12', value: 20 },
+    { date: '2023-01-13', value: 10 },
+    { date: '2023-01-14', value: 10 },
+  ];
+
+  // wrap brush with a bar chart to make brush traveler work
+  const barChartWithBrush = (
```
nikolasrieble, line 26: Instead of sharing the setup between tests, please move this into the test. That would prevent potential side effects.

nikolasrieble (changes requested):
Please ensure independent test setup by not sharing a setup, add the missing test and make sure that test content and name are in sync.

test-jest/cartesian/Brush.spec.tsx
ckifer: Why? I think this is fine - especially if the test didn't change any props anyways. The component always gets rendered fresh so there isn't any possible side effects.

test-jest/cartesian/Brush.spec.tsx
ckifer: +1

ckifer (changes requested):
Looks fine after the requested change to the mouseenter/mouseleave test

test-jest/cartesian/Brush.spec.tsx
nikolasrieble: Well, plainly the object barChartWithBrush could be changed, overwritten, etc. 
One could solve that with Object.freeze() or simply setup per test separately. 

But it is more of a nit, fair enough.

ckifer (approved):
Thanks @Yilun-Sun !

PR #3203, opened 2023-01-20
test(JestMigration): convert CartesianAxis

akib22:
## Description

Refactor test/specs/cartesian/CartesianAxisSpec.js to test-jest/cartesian/CartesianAxis.spec.tsx

## Related Issue
#3185 

## Types of changes

- [x] Jest Migration

## Checklist:


test-jest/cartesian/CartesianAxis.spec.tsx
```diff
@@ -66,29 +76,36 @@ describe('<CartesianAxis />', () => {
...
-    const wrapper = mount(
-      <CartesianAxis
-        orientation="bottom"
-        width={400}
-        height={50}
-        viewBox={{ x: 0, y: 0, width: 500, height: 500 }}
-        ticks={ticks}
-      />,
-    );
+    const myStyle = { fontSize: '14px', letterSpacing: '0.5em' } as CSSStyleDeclaration;
```
ckifer, line 83: nit: could this be `const myStyle: CSSStyleDeclaration` instead of a cast?

test-jest/cartesian/CartesianAxis.spec.tsx
```diff
@@ -66,29 +76,36 @@ describe('<CartesianAxis />', () => {
...
+          width={400}
+          height={50}
+          viewBox={{ x: 0, y: 0, width: 500, height: 500 }}
+          ticks={ticks}
+          label="test"
+          interval="preserveStartEnd"
+        />
+      </Surface>,
+    );
+    const container = document.querySelector('#recharts_measurement_span') as HTMLSpanElement;
```
ckifer, line 102: any way we can avoid this cast?

ckifer (approved):
A few nits but over all looks good. Will merge after you respond as long as the reasoning for the casts are good. Thanks!

test-jest/cartesian/CartesianAxis.spec.tsx
akib22: Yes, I have updated this.

test-jest/cartesian/CartesianAxis.spec.tsx
akib22: It is simply not possible. If I want to make it possible, I must declare a massive object that contains every property from CSSStyleDeclaration.

I  am new to TypeScript, Is there anything wrong with casting?

ckifer: You can do 
```typescript
const myStyle: Partial<CSSStyleDeclaration
```

This way you don't have to cast and not all the attributes are required. Casting is unsafe as you could have an incorrect attribute in the object and you are basically saying "I am right" with the cast that you make. I'll merge this though, its fine in a test

Issue #3192, opened 2023-01-19
[JestMigration] ScatterSpec.js


nikolasrieble:
As part of the jest migration we want to move all current tests to typescript + RTL + jest. Issues will be made to encourage small chunks of refactoring done at once.

Refactor test/specs/cartesian/ScatterSpec.js -> test-jest/cartesian/Scatter.spec.tsx

Acceptance criteria:

- [ ] All tests are in TS + RTL + Jest
- [ ] Previous test is deleted
- [ ] All tests pass

akib22:
Sorry, I mistakenly linked this #3223.

ckifer:
Closed by https://github.com/recharts/recharts/pull/3220

PR #3220, opened 2023-01-21
test(JestMigration): ScatterSpec.js -> Scatter.spec.tsx

proke03:
## Description
Migrate ScatterSpec.js -> Scatter.spec.tsx

## Related Issue
#3192

## Types of changes

## Checklist:


test-jest/cartesian/Scatter.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <Surface width={500} height={500}>
+        <Scatter isAnimationActive={false} shape={renderCustomizedShape} points={data as any} />
+      </Surface>,
+    );
+
+    expect(container.querySelectorAll('.customized-shape')).toHaveLength(data.length);
+  });
+
+  test('Render customized line when line is set to be a ReactElement', () => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars, react/prop-types
```
nikolasrieble, line 61: Is this required? Could you please try to not use it?

test-jest/cartesian/Scatter.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <Surface width={500} height={500}>
+        <Scatter isAnimationActive={false} line={<CustomizedLine points={data as any} />} points={data as any} />
+      </Surface>,
+    );
+
+    expect(container.querySelectorAll('.customized-line')).toHaveLength(1);
+  });
+
+  test('Render customized line when line is set to be a function', () => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
nikolasrieble, line 73: Same here.

test-jest/cartesian/Scatter.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const renderCustomizedLine = ({ points }: { points: any }) => <path d="M0,0L200,200" className="customized-line" />;
```
nikolasrieble, line 74: Please generally do not use any if possible. You can infer the type automatically.

test-jest/cartesian/Scatter.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <Scatter
+          isAnimationActive={false}
+          points={data as any}
+          onClick={onClick}
+          onMouseEnter={onMouseEnter}
+          onMouseLeave={onMouseLeave}
+        />
+      </Surface>,
+    );
+    const syb = container.querySelectorAll('.recharts-symbols')[4];
```
nikolasrieble, line 100: Please do not use abbreviations.
```suggestion
    const symbols = container.querySelectorAll('.recharts-symbols')[4];
```

test-jest/cartesian/Scatter.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+    fireEvent.mouseEnter(syb);
+    expect(onMouseEnter).toHaveBeenCalled();
+    fireEvent.mouseLeave(syb);
+    expect(onMouseLeave).toHaveBeenCalled();
+    fireEvent.click(syb);
+    expect(onClick).toHaveBeenCalled();
```
nikolasrieble, line 106: Great test, I think we miss this test for many other components. This is a good example on how to test this feature.

nikolasrieble (approved):
Please do consider my suggestions. Nothing blocking though.

test-jest/cartesian/Scatter.spec.tsx
proke03: I think it's ok in test code.

test-jest/cartesian/Scatter.spec.tsx
proke03: Thank you!

test-jest/cartesian/Scatter.spec.tsx
ckifer: Even in tests, `any` has a chance to make things happen that we don't expect. Unless there is an explicit reason (testing if something breaks with wrong props passed, there is no type in the code itself) we should stick to as strict types as we can. Maybe we can say if we use `any` we have to comment above it as to why we did

ckifer (approved):

PR #3223, opened 2023-01-21
test(JestMigration): convert ZAxisSpec

akib22:
## Description
Refactor `test/specs/cartesian/ZAxisSpec.js` to `test-jest/cartesian/ZAxis.spec.tsx`

## Related Issue
#3195 

## Types of changes
- Jest Migration

## Checklist:


test-jest/cartesian/ZAxis.spec.tsx
```diff
@@ -0,0 +1,17 @@
...
+  it("Don't render anything", () => {
+    render(
+      <Surface width={500} height={500}>
+        <ZAxis dataKey="x" name="stature" unit="cm" />
+      </Surface>,
+    );
+    const svg = document.querySelector('svg');
+
+    expect(svg).toBeInTheDocument();
+    expect(svg?.children).toHaveLength(2);
```
nikolasrieble, line 15: Why is this condition different to the one in the original test? 

```
    expect(wrapper.find('svg').children.length).to.equal(1);
    expect(wrapper.find('svg noscript').children.length).to.equal(1);
```

nikolasrieble (changes requested):
I am not sure whether the tests are equivalent.

akib22:
image
The above image shows the HTML code of the ZAxis component. There is only one SVG element inside that element that contains two children. So, I can't find any element with this pattern `svg noscript` and can't able to test that condition. So, I have tested it like this. Please let me know what should I do now.

test-jest/cartesian/ZAxis.spec.tsx
ckifer: Yeah I tested the output as well - there is no way this test was working properly beforehand. This change works - it just renders an empty `title` and `desc` tag

ckifer (approved):

Issue #2952, opened 2022-08-25
Treemap error: "Warning: Received `true` for a non-boolean attribute `z`."


alextenb:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/simple-treemap-r5o4e)

### Steps to reproduce
Render the Treemap with any data and note the following error logged to the console:

Treemap error: "Warning: Received `true` for a non-boolean attribute `z`."

### What is expected?
No console errors should be generated by the component.

### What is actually happening?
An console error is generated by the component.

| Environment | Info |
|---|---|
| Recharts | v2.1.6 |
| React | 17.0.1   |
| System | linux ubuntu |
| Browser | firefox |

---
I am seeing the issue when using Recharts version 2.1.13 with React version 17.0.2, but it also happens with the CodeSandbox link provided in the "Examples" section of the docs (Treemap example), which uses Recharts version 2.0.6. This is the repro link which I have provided.

Babadzhanov:
Hi 👋 I'm experience the same issue 👀 however the error disappears if you add "content" property as per this example https://recharts.org/en-US/examples/CustomContentTreemap. Hope this helps 👍

alextenb:
So it does! Thanks for the heads up.

ckifer:
Not sure what exactly this is but looks like the error is here:

https://github.com/recharts/recharts/blob/master/src/chart/Treemap.tsx#L112,L141

ckifer:
Will be fixed by #3228 on next release.`z` attribute is only valid on elements we will never use
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/z

Closing this issue

ckifer:
Should be fixed in 2.4.0

PR #3228, opened 2023-01-21
fix: treemap error(non boolean attribute z)

proke03:
## Description
That variable assignment gives the attribute z to the path element in the svg.
However, the expression is not valid because svg cannot set the z-value.
https://www.w3.org/TR/SVG2/render.html#RenderingOrder

## Related Issue
#2952

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/Treemap.tsx
+++ b/src/chart/Treemap.tsx
@@ -109,7 +109,7 @@ const horizontalPosition = (row: any, parentSize: number, parentRect: TreemapNod
     curX += child.width;
   }
   // what's z
-  child.z = true;
+  child.z = 'true';
   // add the remain x to the last one of row
   child.width += parentRect.x + parentRect.width - curX;
 
@@ -138,7 +138,7 @@ const verticalPosition = (row: any, parentSize: number, parentRect: TreemapNode,
     curY += child.height;
   }
   if (child) {
-    child.z = false;
+    child.z = 'false';
     child.height += parentRect.y + parentRect.height - curY;
   }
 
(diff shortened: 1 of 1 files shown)


src/chart/Treemap.tsx
```diff
@@ -108,8 +108,6 @@ const horizontalPosition = (row: any, parentSize: number, parentRect: TreemapNod
     child.width = Math.min(rowHeight ? Math.round(child.area / rowHeight) : 0, parentRect.x + parentRect.width - curX);
     curX += child.width;
   }
-  // what's z
```
nikolasrieble, line 111: Amazing comment. While this seems to be a bug fix, I am yet worried whether some other place in recharts magically expects this to be set. Please do manually test the TreeMap in the Demo, and investigate whether this is used anywhere.

Typescript does not help a lot due to the large amount of any's in the code base.

proke03: Thanks for your review.
Sure, I tested it in the Demo. I checked the result rendered and element attributes.

ckifer: +1 to Niko - I'm going to wait to merge this until I pull it down and play with it (at least just for some more confidence if nothing else). Looks like a non-breaking change from the naked eye (and testing). Thanks

ckifer:
Pulled this down, it looks good to me

ckifer (approved):

ckifer:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/z 

The only valid elements for this we will never use

Issue #3237, opened 2023-01-22
[Jest Migration] polar/RadialBarSpec


proke03:
We want to move all tests to jest, RTL, and typescript. Issues will be made to encourage small chunks of refactoring done at once.

Refactor test/specs/polar/RadialBarSpec.js -> test-jest/polar/RadialBar.spec.ts.

https://github.com/recharts/recharts/blob/master/test/specs/polar/RadialBarSpec.js

Acceptance criteria:

All tests are in TS + RTL + Jest at RadialBar.spec.ts
RadialBarSpec.js is deleted
All tests pass

ckifer:
Closed by  #3238

PR #3238, opened 2023-01-22
test(JestMigration): RadialBarSpec.js -> RadialBar.spec.tsx

proke03:
## Description
Migrate RadialBarSpec.js -> RadialBar.spec.tsx

## Related Issue
#3237

## Types of changes

## Checklist:


test-jest/polar/RadialBar.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const { container } = render(
+      <Surface width={500} height={500}>
+        <RadialBar data={[]} dataKey="value" />
+      </Surface>,
+    );
+
+    expect(container.querySelectorAll('.recharts-pie')).toHaveLength(0);
+  });
+
+  test('Render 5 sectors', () => {
```
james2hey, line 33: It looks like this test name isn't accurate on what its testing. Something like this could be more suitable
```suggestion
  test('RadialBar event handlers', () => {
```

test-jest/polar/RadialBar.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+          onMouseEnter={onMouseEnter}
+          onMouseLeave={onMouseLeave}
+          onClick={onClick}
+        />
+      </Surface>,
+    );
+
+    const sector = container.querySelectorAll('.recharts-sector')[0];
+
+    fireEvent.mouseEnter(sector);
```
james2hey, line 55: Any reason to use `fireEvent` over `userEvent`? My understanding is that `userEvent` is the recommended approach

ckifer: I actually didn't know this, nice! 

> userEvent simulates complete interactions, and not only single events

+1 to using `userEvent` where possible

test-jest/polar/RadialBar.spec.tsx
ckifer: agreed though it was named this before, a non-blocking nit

ckifer: but actually I will call it blocking because we shouldn't have more than 1 test with the same name (and the above difference in test)

ckifer (changes requested):

test-jest/polar/RadialBar.spec.tsx
proke03: @ckifer
Is the one more code PieChart.spec.tsx?

ckifer (approved):

test-jest/polar/RadialBar.spec.tsx
ckifer: I fixed the conflicts, all good now

Issue #3247, opened 2023-01-22
[Jest Migration] shape/RectangleSpec


proke03:
We want to move all tests to jest, RTL, and typescript. Issues will be made to encourage small chunks of refactoring done at once.

Refactor test/specs/shape/RectangleSpec.js -> test-jest/shape/Rectangle.spec.ts.

https://github.com/recharts/recharts/blob/master/test/specs/shape/RectangleSpec.js

Acceptance criteria:

All tests are in TS + RTL + Jest at Rectangle.spec.ts
RectangleSpec.js is deleted
All tests pass

ckifer:
Closed by #3255

PR #3255, opened 2023-01-23
test(JestMigration): convert RectangleSpec

akib22:
## Description

Refactor `test/specs/shape/RectangleSpec.js` to `test-jest/shape/Rectangle.spec.tsx`

## Related Issue
#3247 

## Types of changes

- [x] Jest Migration

## Checklist:


proke03:
LGTM

proke03 (commented):
It's impressive to reduce two tests to one using jest-each. 
I learned one thing here. thank you!

package.json
```diff
@@ -114,6 +114,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "husky": "^8.0.2",
     "jest": "^27.5.1",
+    "jest-each": "^29.3.1",
```
ckifer, line 117: I think this is fine to add, but it should match our current jest number - since we haven't upgraded TS yet we can't upgrade jest past 27. Can you downgrade this to 27.x?

ckifer (changes requested):
Looks good besides the version change! Just want to make sure we are compatible between jest and `jest-each`

package.json
akib22: Sure. I have downgraded the package.

ckifer (approved):
Thanks!

Issue #3215, opened 2023-01-21
[Jest Migration] component/TooltipSpec


proke03:
We want to move all tests to jest, RTL, and typescript. Issues will be made to encourage small chunks of refactoring done at once.

Refactor test/specs/component/TooltipSpec.js -> test-jest/component/Tooltip.spec.ts.

https://github.com/recharts/recharts/blob/master/test/specs/component/TooltipSpec.js

Acceptance criteria:

All tests are in TS + RTL + Jest at Tooltip.spec.ts
TooltipSpec.js is deleted
All tests pass

ckifer:
Closed by #3267

PR #3267, opened 2023-01-23
Jest: Migrate Tooltip.spec.tsx

nikolasrieble:
## Description
This PR does not strictly fulfil the requirements of the issue. 

The existing tests directly modify the state, which is an antipattern, since effectively implementation details are tested. Thus I do not intend to replicate this test.

Instead I wrote 4 tests which document the behaviour of the Tooltip and added comments to guide for test writing. 

This is an improvement over the existing test and still sufficient for the migration. 

## Related Issue
https://github.com/recharts/recharts/issues/3215

## Types of changes

## Checklist:


test-jest/component/Tooltip.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const mouseOverEvent = mockMouseEvent('mouseover', chart!, { pageX: 200, pageY: 200 });
+    mouseOverEvent.fire();
+
+    // After the mouse over event over the chart, the tooltip wrapper still is not set to visible,
+    // but the content is already created based on the nearest data point.
+    const tooltipContentName = container.querySelector('.recharts-tooltip-item-name');
+    const tooltipContentValue = container.querySelector('.recharts-tooltip-item-value');
+    expect(tooltipContentName).toBeInTheDocument();
+    expect(tooltipContentValue).toBeInTheDocument();
+  });
```
nikolasrieble, line 67: If you have any idea / suggestion on how to trigger visibility, I would love to include that in the PR. @ckifer @proke03

ckifer: I think we may need to refactor the Tooltip to use React Portals (theres a FR for that somewhere) in order for this to be easier to test.

I'm okay with merging this and coming back to it

ckifer (approved):
LGTM, maybe @nikolasrieble we can create an issue to re-visit these tests so we don't forget

test-jest/component/Tooltip.spec.tsx
proke03: For now, I think this is the best way.
I tried implementing that code as a userEvent, but it failed.

PR #3269, opened 2023-01-23
test: Fix log errors

marcalexiei:
## Description
There are a lot of error in during test execution. See nom run test log in one of the [latest run](https://github.com/recharts/recharts/actions/runs/3989413884/jobs/6841812358).

E.g.:

image

This PR fixes all this errors expect the one coming from `z` prop on `Treemap.spec.tsx` which will be resolved via #3228

It also adds a deprecation comment on `alwaysShow` property on Reference components (which already issues a warning runtime).

## Related Issue
N/A

## Motivation and Context
Make test logs more easy to read

## How Has This Been Tested?
`npm run test`

N/A

## Types of changes

## Checklist:


test-jest/cartesian/Brush.spec.tsx
```diff
@@ -36,11 +41,13 @@ describe('<Brush />', () => {
 
   test('Render panorama when specified LineChart as child', () => {
     const { container } = render(
-      <Brush x={100} y={50} width={400} height={40} data={data}>
-        <LineChart>
-          <Line />
-        </LineChart>
-      </Brush>,
+      <BarChart width={400} height={100} data={data}>
```
ckifer, line 44: any reason this is being wrapped in a bar chart?

test-jest/cartesian/Brush.spec.tsx
```diff
@@ -22,8 +22,13 @@ describe('<Brush />', () => {
     { date: '2023-01-14', value: 10 },
   ];
 
-  test('Render 2 travellers and 1 slide in simple Brush', () => {
-    const { container } = render(<Brush x={100} y={50} width={400} height={40} data={data} />);
+  test('Render 2 travelers and 1 slide in simple Brush', () => {
+    const { container } = render(
+      <BarChart width={400} height={100} data={data}>
```
ckifer, line 27: any reason this is being wrapped in a bar chart?

test-jest/cartesian/Brush.spec.tsx
```diff
@@ -63,6 +70,7 @@ describe('<Brush />', () => {
     expect(screen.getAllByText(data[data.length - 1].date)).toHaveLength(1);
   });
 
+  // eslint-disable-next-line max-len
   test('mouse down on traveller will trigger the brush text display, and mouse move out will hide the brush text', () => {
```
ckifer, line 74: probably don't need to disable this line? Just let it be formatted

test-jest/component/Customized.spec.tsx
```diff
@@ -1,22 +1,31 @@
...
+    function Custom() {
+      return <rect data-testid="customized-svg-element" />;
     }
-    // eslint-disable-next-line react/jsx-no-bind
-    const { container } = render(<Customized component={rect} />);
-    expect(container.querySelectorAll('.customized-svg-element')).toHaveLength(1);
+
+    render(
+      <svg>
+        {/* eslint-disable-next-line react/jsx-no-bind */}
```
ckifer, line 25: this lint line should be fine to remove

test-jest/cartesian/Brush.spec.tsx
```diff
@@ -95,7 +103,11 @@ describe('<Brush />', () => {
   });
 
   test('render text when alwaysShowText is true', () => {
-    const { container } = render(<Brush x={100} y={50} width={400} height={40} data={data} alwaysShowText />);
+    const { container } = render(
+      <BarChart width={500} height={100} data={data}>
```
ckifer, line 107: any reason this is wrapped in a bar chart?

ckifer (changes requested):
A few of the same comments across the changes. Overall nice changes to have though!

test-jest/cartesian/Brush.spec.tsx
marcalexiei: I already tried. 
The line is too long to be formatted automatically.
Do I need to do something like this?

```tsx
[
  'mouse down on traveller will trigger the brush text display,',
  'and mouse move out will hide the brush text',
].join(' ')
```

ckifer: Shouldn't need to no. Strange. Feel free to keep the ignore there I guess.

test-jest/cartesian/Brush.spec.tsx
marcalexiei: Because Brush render SVG elements in the page. 
If not wrapped in a <svg /> they aren't valid element for the browser.

image

If you try this out in the demo page you see that there isn't any visible Brush in the page.

image

If you prefer I can wrap them in a Surface like in `demo/component/BrushDemo.tsx`:

```tsx
<Surface width={800} height={200}>
          <Brush {...props} />
</Surface>
```

test-jest/cartesian/Brush.spec.tsx
marcalexiei: Same as this [comment](https://github.com/recharts/recharts/pull/3269#discussion_r1084847142)

test-jest/cartesian/Brush.spec.tsx
marcalexiei: Same as this [comment](https://github.com/recharts/recharts/pull/3269#discussion_r1084847142)

test-jest/component/Customized.spec.tsx
marcalexiei: I haven't changed the test implementation which was disabling the rule since `Custom` is a standard function so aslant correctly issues the warning.

To get rid of the comment I can turn `Custom` into an arrow function

test-jest/cartesian/Brush.spec.tsx
ckifer: I think we should keep these simple and yeah wrap in `Surface` instead of an entire chart

PR #3270, opened 2023-01-23
refactor(Label & LabelList): refine types and disable react/no-array-index-key to avoid warnings

marcalexiei:
## Description
Added `// eslint-disable-next-line react/no-array-index-key` to `Label` and `LabelList` to get rid of the two related warnings.
So far there is no known issue so I don't think this should be a problem adding the disabled comment here like has been done in other files.

I also tried to refine types on Label and LabelList components.
I made few changes that shouldn't affect run time results:

1. On `Label` `isPolar` function now use `in` operator to avoid inline casting
2. On `RadialBar` I removed `clockWise` prop from `LabelList.renderCallByParent` invocation since the function uses only `label` and `children` props

## Related Issue
N/A

## Motivation and Context
N/A

## How Has This Been Tested?

`npm run test`
`npm run lint`

N/A

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Label.tsx
+++ b/src/component/Label.tsx
@@ -370,7 +370,8 @@ const getAttrsOfCartesianLabel = (props: Props) => {
   };
 };
 
-const isPolar = (viewBox: CartesianViewBox | PolarViewBox) => isNumber((viewBox as PolarViewBox).cx);
+const isPolar = (viewBox: CartesianViewBox | PolarViewBox): viewBox is PolarViewBox =>
+  'cx' in viewBox && isNumber(viewBox.cx);
 
 export function Label(props: Props) {
   const { viewBox, position, value, children, content, className = '', textBreakAll } = props;
@@ -420,7 +421,7 @@ Label.defaultProps = {
   offset: 5,
 };
 
-const parseViewBox = (props: any) => {
+const parseViewBox = (props: any): ViewBox => {
   const {
     cx,
     cy,
@@ -477,7 +478,7 @@ const parseViewBox = (props: any) => {
   return {};
 };
 
-const parseLabel = (label: any, viewBox: ViewBox) => {
+const parseLabel = (label: unknown, viewBox: ViewBox) => {
   if (!label) {
     return null;
   }
@@ -492,7 +493,7 @@ const parseLabel = (label: any, viewBox: ViewBox) => {
 
   if (isValidElement(label)) {
     if (label.type === Label) {
-      return cloneElement(label as any, { key: 'label-implicit', viewBox });
+      return cloneElement<LabelProps>(label, { key: 'label-implicit', viewBox });
     }
 
     return <Label key="label-implicit" content={label} viewBox={viewBox} />;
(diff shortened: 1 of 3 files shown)


src/polar/RadialBar.tsx
```diff
@@ -398,14 +397,7 @@ export class RadialBar extends PureComponent<Props, State> {
 
         <Layer className="recharts-radial-bar-sectors">{this.renderSectors()}</Layer>
 
-        {(!isAnimationActive || isAnimationFinished) &&
-          LabelList.renderCallByParent(
-            {
-              ...this.props,
-              clockWise: this.getDeltaAngle() < 0,
```
ckifer, line 405: why was `clockWise` removed? Unless there is a reason if should probably remain

ckifer (changes requested):
We should probably come up with element keys that don't include indexes in general instead of disabling the lint rule. 

One comment on the last change

src/polar/RadialBar.tsx
marcalexiei: > On RadialBar I removed clockWise prop from LabelList.renderCallByParent invocation since the function uses only label and children props

from the description

ckifer: Ah sorry I missed that

marcalexiei:
> We should probably come up with element keys that don't include indexes in general instead of disabling the lint rule.

This was done ~9 times before the PR. 
If we don't want to disable the rule we should find a consistent way to do this across the whole codebase so I think it's fine disable also for these two scenario to be consistent 😅

src/component/Label.tsx
```diff
@@ -509,16 +510,21 @@ const parseLabel = (label: any, viewBox: ViewBox) => {
...
     return null;
   }
   const { children } = parentProps;
   const parentViewBox = parseViewBox(parentProps);
 
-  const explicitChildren = findAllByType(children, Label).map((child, index: number) =>
+  const explicitChildren = findAllByType(children, Label).map((child, index) =>
     cloneElement(child, {
       viewBox: viewBox || parentViewBox,
+      // eslint-disable-next-line react/no-array-index-key
```
nikolasrieble, line 527: Instead of ignoring eslint, how about following the rule instead? Does the child have a property from which we could generate a new key string?

src/component/LabelList.tsx
```diff
@@ -92,15 +92,20 @@ function parseLabelList<T extends Data>(label: any, data: Array<T>) {
...
   if (!parentProps || (!parentProps.children && checkPropsLabel && !parentProps.label)) {
     return null;
   }
   const { children } = parentProps;
 
-  const explicitChildren = findAllByType(children, LabelList).map((child, index: number) =>
+  const explicitChildren = findAllByType(children, LabelList).map((child, index) =>
     cloneElement(child, {
       data,
+      // eslint-disable-next-line react/no-array-index-key
```
nikolasrieble, line 108: Same here.

src/component/Label.tsx
marcalexiei: Due to the large use of `any` type it's difficult to find a suitable prop to use as a `key`. 
In this specific scenario I can't find a suitable prop to use.
If you have any idea let me know.

---

Since the warning are there from a certain amount if time and there are 9 additional disabled comment may I suggest you to consider to leave the disabled comment  and create a separate issue to address the issue in the entire codebase.

Otherwise I can remove the disable comment if you are ok with the rest of the code of the PR.

src/component/LabelList.tsx
marcalexiei: [Same here](https://github.com/recharts/recharts/pull/3270#discussion_r1084919755)

src/component/Label.tsx
nikolasrieble: Fair enough, please do create a follow up issue.

nikolasrieble (approved):
Fair enough, lets go. 🚀

ckifer (approved):

Issue #3268, opened 2023-01-23
Storybook: Replace the demo with storybook and use it for testing.


nikolasrieble:
TDLR: 
Integrating storybook into the main repo would (a) enable a better development experience, (b) integrate the documentation in the main repository and (c) enable us to add simple user interaction tests. 

# The problem
We are not very confident to ship bug fixes or new features.

## Documentation and code out of sync: 
As the documentation lives in another repository (the website), it naturally comes out of sync, which complicates developer experience and adds effort for us to maintain. 

## Manual testing: 
While unit tests exist, they are not exhaustive and not sufficient to test user interactions. Thus we must currently manually test using the demo, and the website. Still in the last release we managed to ship two bugs which resulted in two patches. 

# The solution
## Replace the demo code with Storybook
Locally developing with Storybook is a great developer experience and is closer to state of the art of frontend development. This would improve the quality of life for existing contributors and possibly attract new contributors. 

## Integrate the storybook test runner into our CI 
The test runner enables us to simple stories as tests. If they render without an error, the test passes.

## Future: Use storybooks play function to write user interaction stories and thereby unlock integration tests
This would set a path forward for confidence required to fix bugs and add new features. 
https://storybook.js.org/docs/react/writing-stories/play-function
https://storybook.js.org/addons/@storybook/test-runner

nikolasrieble:
I would sure love to hear your thoughts on my proposal. @ckifer @dillonreedy @dsldsl @arcthur @proke03

ckifer:
I think storybook is a nice win for us. I have my opinions in Storybook as a product/software, but I don't think most of my concerns really apply here.

What we need: self-documenting tests/examples as we build. Storybook gives us that mostly out of the box.

My one concern is that this should not completely replace the current docs site. Not sure what the correct solution is, but initially we can host the storybook separate from the docs site and add a link to it as its in-progress.

nikolasrieble:
In how far can Storybook not replace the current docs site? What would you think is the best solution for docs?

ckifer:
I guess I enjoy the current website structure and layout for things that are not updated often - home page, install guide, etc. Could we embed the story book using a similar home page as to what we have now? 

Kind of like https://cultureamp.design/storybook/

proke03:
![캡처_2023_01_24_13_42_50_269](https://user-images.githubusercontent.com/19148682/214215800-2bb3846f-76e1-4f97-bf20-18ae1cdc24f3.png)

[...]

PR #3276, opened 2023-01-24
feat: integrate docs project(WIP)

proke03:
## Description
I add the storybook tab on charts.org like below image. And I integrated it into this repository.
It's been maintained with different lint setups so far, so it's hard to fully integrate them all at once.

## Related Issue
#3268

image
## Types of changes

## Checklist:


nikolasrieble (commented):
Why not just call the folder `storybook` instead of `org`?

org/docs/183.js
```diff
@@ -0,0 +1,2 @@
...se 3:if(!(c=p.sent())||e.isDisposed())return[2];for(l=[],f=0,d=c;f<d.length;f++)h=d[f],l.push({resource:o.Sf.parse(h.fileName),edit:{range:this._textSpanToRange(e,h.textSpan),text:n}});return[2,{edits:l}]}}))}))},t}(g);function N(e){P=W(e,"typescript")}function E(e){I=W(e,"javascript")}function K(){return new Promise((function(e,t){if(!I)return t("JavaScript not registered!");e(I)}))}function R(){return new Promise((function(e,t){if(!P)return t("TypeScript not registered!");e(P)}))}function W(e,t){var n=new u(t,e),r=function(){for(var e=[],t=0;t<arguments.length;t++)e[t]=arguments[t];return n.getLanguageServiceWorker.apply(n,e)},i=new b(r);return o.Mj.registerCompletionItemProvider(t,new y(r)),o.Mj.registerSignatureHelpProvider(t,new S(r)),o.Mj.registerHoverProvider(t,new w(r)),o.Mj.registerDocumentHighlightProvider(t,new k(r)),o.Mj.registerDefinitionProvider(t,new x(i,r)),o.Mj.registerReferenceProvider(t,new C(i,r)),o.Mj.registerDocumentSymbolProvider(t,new M(r)),o.Mj.registerDocumentRangeFormattingEditProvider(t,new D(r)),o.Mj.registerOnTypeFormattingEditProvider(t,new A(r)),o.Mj.registerCodeActionProvider(t,new O(r)),o.Mj.registerRenameProvider(t,new T(r)),new v(i,e,t,r),r}}}]);
```
nikolasrieble, line 1: These are generated files, right? Why commit them? Maybe a git ignore update is required first.

proke03: charts.org also includes files in docs.
I think this is used for deployment using github pages.

proke03:
I totally agree with you.
But I thought I couldn't do it all at once.

proke03:
I think both are possible. And I also think the latter is better. 
And the storybook is shown as an iframe in this PR.

ckifer:
First of all thank you @proke03 for the POC and work here!

I think I am fine with this, but maybe we should take smaller steps at it. We don't want two versions of the docs that are out of sync with each other, that is no fun.

How about maybe lets take a step back, integrate the storybook itself here first, set ourselves up as a monorepo (recharts/recharts-docs or recharts/org doesn't matter to me), and then we can move the docs site over into the same sub-repo the storybook is in?

ckifer:
Hmm... 

This is actually interesting. We can only host one GH pages site per repository. If we move the storybook here and host it we can't host the docs site here and if we host the docs site here we can't host the storybook. 

I can't seem to find any other way other than iframe to embed a storybook. Can you?

proke03:
> We can only host one GH pages site per repository.

I didn't know that. I will try putting both in one folder.

ckifer:
Even if you put them in one folder it doesn't solve the issue. In order for the iframe to work it has to be deployed separately no?

proke03:
Since it is not going to be distributed separately, I think I need to find a way other than iframe.

proke03:
In my opinion, it is better to minimize the use of tools. 
Ideally, only use github pages for org distribution.

ckifer:
Agreed with @proke03 - minimize the tools minimize the things people need access to. 

If we deploy the storybook here we get the benefit of updating documentation in the same repo. When it deploys it will automatically be updated in the iframe on the org site. So you could go to it directly or via the org site and we wouldn't need to update the org site very often once the storybook is in a good spot.

Yilun-Sun:
Agreed with you

proke03:
I will continue to develop this PR.
However, as shown above, the history is too messy, so I am experimenting with github actions in other branch.

ckifer:
Sounds good. Thank you!

Issue #3278, opened 2023-01-25
Change defaultProps in ReferenceArea and ReferenceLine to use Javascript default parameters


AlexMcClay:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
React will remove support for default props in a future major release, so it's best to change it to use Javascript default parameters instead. There are already some errors that show up when using these features in React, so it would remove the errors and warnings.

### What does the proposed API look like?
changing defaultProps to Javascript default parameters is pretty self-explanatory.

ckifer:
Thanks @AlexMcClay 

Labeled and ready to go

aushwin:
@ckifer can I pick this up  ?

ckifer:
Absolutely - assigned

aushwin:
@ckifer Could you please guide me here as it's my first time contributing. I have refactored and pushed the changes.

ckifer:
@aushwin Looks good. Fix the build/test failure and we should be good to go

aushwin:
Should I refactor more such that instead of combining the destructured values back to an object, it can be passed into the functions wherever needed ?  How can I resolve this one  @ckifer 
Thanks 
![image](https://user-images.githubusercontent.com/54993680/214925246-85fd69b7-1023-4d59-b7b4-b4f92df0f50b.png)

aushwin:
> Should I refactor more such that instead of combining the destructured values back to an object, it can be passed into the functions wherever needed ? How can I resolve this one @ckifer Thanks ![image](https://user-images.githubusercontent.com/54993680/214925246-85fd69b7-1023-4d59-b7b4-b4f92df0f50b.png)

got the error fixed

ckifer:
just an `npm install` should install cross-env. Try to delete node_modules

aushwin:
@ckifer please re-review

ckifer:
closed by  #3283

leo-terrier:
Hello, I've seen that this commit has been reverted and that defaultProps is back. 

I was wondering what could be the impact of defaultProps usage (other than the console warning), since its going to be deprecated on functional component. Thanks for your reply and for the great lib!

ckifer:
Please see https://github.com/recharts/recharts/issues/3438#issuecomment-1467419256

ckifer:
Options to move forward on this are essentially:

[...]

PR #3283, opened 2023-01-26
Refactored: added default values in params for ReferenceArea and ReferenceLine since default props is deprecated

aushwin:
- updated ReferenceArea
- updated ReferenceLine

## Description

Have destructed the props in the params added the value and combined them into a prop object so that it can be passed down to other functions with the default value.

## Related Issue

https://github.com/recharts/recharts/issues/3278

## Motivation and Context
Since defaultProps is deprecated, we would avoid future potential warnings/errors

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/ReferenceArea.tsx
+++ b/src/cartesian/ReferenceArea.tsx
@@ -67,7 +67,32 @@ const getRect = (hasX1: boolean, hasX2: boolean, hasY1: boolean, hasY2: boolean,
   return rectWithPoints(p1, p2);
 };
 
-export function ReferenceArea(props: Props) {
+export function ReferenceArea({
+  isFront = false,
+  ifOverflow = 'discard',
+  xAxisId = 0,
+  yAxisId = 0,
+  r = 10,
+  fill = '#ccc',
+  fillOpacity = 0.5,
+  stroke = 'none',
+  strokeWidth = 1,
+  ...restProps
+}: Props) {
+  // Props with default values added.
+  const props: Props = {
+    isFront,
+    ifOverflow,
+    xAxisId,
+    yAxisId,
+    r,
+    fill,
+    fillOpacity,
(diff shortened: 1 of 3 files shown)


src/cartesian/ReferenceArea.tsx
```diff
@@ -67,7 +67,32 @@ const getRect = (hasX1: boolean, hasX2: boolean, hasY1: boolean, hasY2: boolean,
   return rectWithPoints(p1, p2);
 };
 
-export function ReferenceArea(props: Props) {
+export function ReferenceArea({
+  isFront = false,
```
ckifer, line 71: destructuring props here doesn't really do anything for us if they're just added back into the props object below. I guess a future refactor can be to use the destructured values directly where possible.

ckifer (approved):
🚀  - thanks!

ckifer (changes requested):
Looks like build is failing though, gotta fix that

src/cartesian/ReferenceArea.tsx
aushwin: Will work on this and the build failing issue

aushwin: > 

only when am destructuring the xAxis and yAxis are becoming undefined . Any reason why this is happening
image

ckifer (approved):
I think this is fine as an initial refactor, we can come back to it if we need. Thanks @aushwin

src/cartesian/ReferenceArea.tsx
ckifer: probably because `0` is falsy, but not sure

ckifer:
Hey @aushwin 

Sorry to say but I'm going to have to revert this change. Please see https://github.com/recharts/recharts/issues/3438

Issue #3214, opened 2023-01-21
[Jest Migration] component/TextSpec


proke03:
We want to move all tests to jest, RTL, and typescript. Issues will be made to encourage small chunks of refactoring done at once.

Refactor test/specs/component/TextSpec.js -> test-jest/component/Text.spec.ts.

https://github.com/recharts/recharts/blob/master/test/specs/component/TextSpec.js

Acceptance criteria:

All tests are in TS + RTL + Jest at Text.spec.ts
TextSpec.js is deleted
All tests pass

ckifer:
Closed by  #3286

PR #3286, opened 2023-01-27
test: refactor Text component tests from karma to jest

ckifer:
## Description

Migration of TextSpec -> Text.spex.tsx

Last test I cannot get to pass due to having to mock `getBoundingClientRect` - there's just no way to mock is so that the same behavior as the real browser is triggered from what I can tell...

## Related Issue

 #3214

## How Has This Been Tested?

`npm test` 

## Types of changes

## Checklist:


test-jest/component/Text.spec.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+      const text = screen.getByRole('img');
+      expect(text).toBeInTheDocument();
+      const lastChild = text?.children[text?.children?.length - 1];
+      const lastLetter = lastChild.textContent?.[lastChild?.textContent?.length - 1];
+
+      expect(lastLetter).toEqual('…');
+    });
+
+    // test('adds an ellipsis at the end of a very long word', () => {
```
ckifer, line 198: @nikolasrieble @Yilun-Sun @proke03 any ideas on how to get this to work?

As you can see above I mock `getBoundingClientRect` to specific widths - the problem with that is `Text` has a lot of logic relying on `getStringSize` - apparently `maxLines={1}` with a long string is supposed to result in ellipses but I can't get it to work as it should because the strings can't be measured properly.

Something in here https://github.com/recharts/recharts/blob/master/src/component/Text.tsx#L33

as well as where its called https://github.com/recharts/recharts/blob/master/src/component/Text.tsx#L130-L155
(`originalResult` is always returned here which doesn't have ellipsis)

nikolasrieble: Well, you could mock getStringSize as well, and then we could separately unit test getStringSize.

proke03: Although Text has several logics inside, I don't think it is necessary to test the output of text according to the values of those logics. You can write tests according to the applicable case.

ckifer: We already separately test `getStringSize` (we can't really test it though, we just have to assert it gives back the mocked value). 

The same issue will occur if we mock `getStringSize` instead of `getBoundingClientRect` - there is a loop that processes the characters in the string. We would need to change the value on every loop iteration which I think is impossible.

nikolasrieble: I am fine with creating a separate issues to track this test, and unblock the migration. Same with the two missing tests for the xAxis in your PR @proke03

PR #3293, opened 2023-01-27
feat: Short-circuiting axis map creation with specified domain to improve performance

### LINKED ISSUE #2762: Performance improvement by short-circuiting axis map creation with specified domain

There's an easy/small change to hugely improve performance when the user provides a specified domain.

The creation of an axis map is really expensive ([see the function docs of `updateStateOfAxisMapsOffsetAndStackGroups`](https://github.com/recharts/recharts/blob/2a3405ff64a0c050d2cf94c36f0beef738d9e9c2/src/chart/generateCategoricalChart.tsx#L846)).
Right now, no matter if a specified domain is provided via props, the chart data is always used to calculate the axis map. This is done to automatically extend the axis when the data is smaller/bigger than a specified domain value.
This can be prohibited via the `allowDataOverflow` prop. In my case, I set `allowDataOverflow` to `true` and also provide the domain start and the domain end. For this scenario, the chart data never has to be taken into account when creating the axis map.

[There are two ways to create the axis map](https://github.com/recharts/recharts/blob/2a3405ff64a0c050d2cf94c36f0beef738d9e9c2/src/chart/generateCategoricalChart.tsx#L481), `getAxisMapByItems` and `getAxisMapByAxes`. Let's take `getAxisMapByAxes` as an example. Right now, `domain` is filled based on the calculation with the chart data ("step 1") and then `domain` is overwritten via the specified domain ("step 2").

We could short-circuit the domain creation in one scenario:
- `allowDataOverflow` is set to `true`
- the specified domain (start AND end) was provided via props with two values

I'm not sure where to do this, because I'm missing the knowledge of how the code flow and `domain` variable is used. Instead of multiple variables, the one `domain` variable is reused in different situations in the code, so I can't really follow the needed values.

To test the proposal out, I modified [`getAxisMapByAxes`](https://github.com/recharts/recharts/blob/2a3405ff64a0c050d2cf94c36f0beef738d9e9c2/src/chart/generateCategoricalChart.tsx#L266) and **even before "step 1"** (see text above), I set the domain via the specified domain:

```ts
const getAxisMapByAxes = (
  props: CategoricalChartProps,
  { axes, graphicalItems, axisType, axisIdKey, stackGroups, dataStartIndex, dataEndIndex }: any,
) => {
  ...
    if (!result[axisId]) {
      let domain, duplicateDomain, categoricalDomain;

      /*
       * We short-circuit the domain creation here to enhance performance.
       * Usually, the data is used to determine the domain, but when the user specifies
       * a domain upfront (via props), there is no need to calculate the domain start and end,
       * which is very expensive for a larger amount of data.
       * The only thing that would prohibit short-circuiting is when the user doesn't allow data overflow,
       * because the axis is supposed to ignore the specified domain that way.
       */
      const specifiedDomain: AxisDomain[] = child.props.domain;
      if (allowDataOverflow === true && type === 'number' && !!specifiedDomain?.[0] && !!specifiedDomain?.[1]) {
        domain = parseSpecifiedDomain(specifiedDomain, null, allowDataOverflow);
      } else {
        // "step 1" as described above
        if (dataKey) {
          // has dataKey in <Axis />
          domain = getDomainOfDataByKey(displayedData, dataKey, type);
        } ...
  ...
}
```

This small change hugely improves the performance for my use case.

# Without the change : 1.7 seconds

image

# With the change: 890 ms
image

## By the way
Did you already investigate whether the axis map creation is optimized based on the input data? I can't really tell why, but I feel like it is executed too often, because the results are not saved/reused.

bennettdams:
## Description

The creation of an axis map is really expensive ([see the function docs of `updateStateOfAxisMapsOffsetAndStackGroups`](https://github.com/recharts/recharts/blob/2a3405ff64a0c050d2cf94c36f0beef738d9e9c2/src/chart/generateCategoricalChart.tsx#L846)).
Right now, no matter if a specified domain is provided via props, the chart data is always used to calculate the axis map. This is done to automatically extend the axis when the data is smaller/bigger than a specified domain value.
This can be prohibited via the `allowDataOverflow` prop. In my case, I set `allowDataOverflow` to `true` and also provide the domain start and the domain end. For this scenario, the chart data never has to be taken into account when creating the axis map.

[There are two ways to create the axis map](https://github.com/recharts/recharts/blob/2a3405ff64a0c050d2cf94c36f0beef738d9e9c2/src/chart/generateCategoricalChart.tsx#L481), `getAxisMapByItems` and `getAxisMapByAxes`. Let's take `getAxisMapByAxes` as an example. Right now, `domain` is filled based on the calculation with the chart data ("step 1") and then `domain` is overwritten via the specified domain ("step 2").

We could short-circuit the domain creation in one scenario:
- `allowDataOverflow` is set to `true`
- the specified domain (start AND end) was provided via props with two values

This change brings the idea of short-circuiting to the `getAxisMapByAxes` function. It hugely improves the performance for my use case. See the linked issue for an example.

I didn't change the functionality of how axes are created. If you look at the code, **everything starting from `if (dataKey)` remains untouched**. I only added `if` blocks above/before that, so we can use the specified domain. 

## Related Issue

Fixes #2762 

## How Has This Been Tested?

Written a new test to make sure a specified domain is respected.
It uses random integers as specified domain values to make sure that the axis is not representing the test data.

## Types of changes

## Checklist:


src/chart/generateCategoricalChart.tsx
```diff
@@ -288,87 +288,115 @@ const getAxisMapByAxes = (
...
+     * Usually, the data is used to determine the domain, but when the user specifies
+     * a domain upfront (via props), there is no need to calculate the domain start and end,
+     * which is very expensive for a larger amount of data.
+     * The only thing that would prohibit short-circuiting is when the user doesn't allow data overflow,
+     * because the axis is supposed to ignore the specified domain that way.
+     */
+    if (type === 'number' && allowDataOverflow === true) {
+      const specifiedDomain: unknown[] | null | undefined = child.props.domain;
+      const domainStart: unknown | null | undefined = specifiedDomain?.[0];
+      const domainEnd: unknown | null | undefined = specifiedDomain?.[1];
```
ckifer, line 302: nit: can we destructure this?
```suggestion
      const [domainStart, domainEnd]: [unknown | null | undefined, unknown | null | undefined] = specifiedDomain;
```
pretty sure JS doesn't throw errors even if the array is `[]` or `[200, undefined]`

ckifer (changes requested):
This looks fine, one tiny nit.

Sorry I just merged changes for moving test files around if you want to fix the conflicts / merge master.

Thanks!

src/chart/generateCategoricalChart.tsx
bennettdams: Just using destructuring does not work, as `specifiedDomain` can be `undefined` there.
I could add another check in the `if` if you want, but using but using optional chaining takes care of all cases I think:

image

ckifer: Yeah it does, idk what I was thinking 😅

bennettdams:
> if you want to fix the conflicts / merge master.

When I merge the current commit into my fork, there are a lot of Prettier warnings, so I cannot `git push`:

image

ckifer:
Looks like your local might not be running lint staged correctly? I'll PR to your fork

bennettdams:
Done, thanks so much for the support!

ckifer:
@bennettdams Just FYI we'll probably leave this open for a bit to give some time to test it - there's been a history of merging things too quickly and introducing bugs since the code is pretty wild in some spots like this one. Thanks for the contribution!

ckifer (approved):

src/chart/generateCategoricalChart.tsx
```diff
@@ -288,87 +288,115 @@ const getAxisMapByAxes = (
...
+
+      /*
+       * The `isNumber` check is needed because the user could also provide strings like "dataMin" via the domain props.
+       * In such case, we cannot short-circuit, as we need to look at the data.
+       */
+      if (!!domainStart && !!domainEnd && isNumber(domainStart) && isNumber(domainEnd)) {
```
nikolasrieble, line 308: Nit: It would be nice to wrap the two if statements into a single function named 

```
function isDomainComputationFromDataRequired(domain): boolean
```

That would be nice, because such a function could independently be tested and it would improve readability of the larger function.

bennettdams: Done! I switched to logic to `isDomainSpecifiedByUser` though, made more sense to me.

bennettdams:
Just to be sure, as the Git diff is a bitmisleading: Everything after `if (dataKey) {` remains untouched - that's why I called this whole PR "short-circuit", as I only added this check above the normal axis generation to skip it.

nikolasrieble (approved):
Wonderful, you even added a test! Thank you for your contribution!

ckifer (approved):

ckifer:
@bennettdams - looks like this broke some stuff in #3369

src/chart/generateCategoricalChart.tsx
```diff
@@ -288,87 +313,105 @@ const getAxisMapByAxes = (
...
+     * which is very expensive for a larger amount of data.
+     * The only thing that would prohibit short-circuiting is when the user doesn't allow data overflow,
+     * because the axis is supposed to ignore the specified domain that way.
+     */
+    if (isDomainSpecifiedByUser(child.props.domain, allowDataOverflow, type)) {
+      domain = parseSpecifiedDomain(child.props.domain, null, allowDataOverflow);
+    }
+
+    // we didn't create the domain from user's props above, so we need to calculate it
+    if (!domain || domain.length === 0) {
```
ckifer, line 329: this breaks highlight and zoom functionality - confirmed. Commented out and tested, functionality works again.

ckifer: Looks like if we have a categorical chart we still need to find and set a categorical domain, else the `categoricalDomain` is undefined and this happens

PR #3295, opened 2023-01-27
Storybook: Minimal setup running locally. 

nikolasrieble:
## Description
Adds storybook following the tutorials. 
Removed all the storybook toy examples. 
Added a LineChart example instead. 

TODO: 
Verify that the stories are not considered when building. 

## Related Issue
https://github.com/recharts/recharts/issues/3294

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #3296, opened 2023-01-28
feat: integrate storybook

proke03:
## Description
Taking @ckifer 's advice earlier, I worked on a storybook integration.
As an example, I only added the brush story. Please review and I will add all other examples in the same way.

Why not distribute the storybook in chromatic and the org on github pages?

Chromatic workflow requires environment variable settings to work.

## Related Issue
#3268

## How Has This Been Tested?
https://63d24aee84ee7cc215239dcc-yrlgloqamw.chromatic.com/?path=/story/recharts-brush--simple

## Types of changes

## Checklist:


nikolasrieble:
https://github.com/recharts/recharts/pull/3295 :D 
Happy to close mine, and review yours again.

storybook/main.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  stories: ['../src/**/*.stories.@(js|jsx|ts|tsx)'],
```
nikolasrieble, line 2: Lets remove js and jsx here, to clearly communicate our preference for TS

proke03:
@nikolasrieble 
Thanks for your review!
Sorry for opening a new PR without a word.
With some changes, I thought it would be better to do something new.

src/stories/cartesian/brush/BrushComponent.jsx
```diff
@@ -0,0 +1,84 @@
+/* eslint-disable react/prop-types */
+/* eslint-disable no-useless-constructor */
+/* eslint-disable import/no-default-export */
```
nikolasrieble, line 3: Please disable to eslint rule `import/no-default-export` for the stories folder.

src/stories/cartesian/brush/BrushComponent.jsx
```diff
@@ -0,0 +1,84 @@
...
+import React, { Component } from 'react';
+import { Surface, Brush } from '../../..';
+
+export default class BrushComponent extends Component {
```
nikolasrieble, line 7: I think it is preferable to only have a single file per component. (Brush, in this case). 
See https://github.com/recharts/recharts/blob/34ccf85ec4681f95813d78ed52dad1e79e51f695/src/stories/LineChart.stories.tsx for example.

src/stories/cartesian/brush/Brush.stories.jsx
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable import/no-default-export */
+import React from 'react';
+import BrushComponent from './BrushComponent';
+
+export default {
+  title: 'Recharts/Brush',
+  component: BrushComponent,
+};
+
+const data = [
```
nikolasrieble, line 10: Please move the data into a shared data file. That encourages sharing datasets between stories, which makes the stories easier understandable.

src/stories/cartesian/brush/BrushComponent.jsx
nikolasrieble: Really, this is not the BrushComponent, but a story template and should instead live here 
```
function Template(args) {
  return <BrushComponent {...args} />;
}
```

nikolasrieble (commented):
Great idea to use chromatic! Also thank you for your work! Please consider my suggestions and lets establish a nice pattern which others can also follow to add stories.

src/stories/cartesian/brush/BrushComponent.jsx
proke03: Include the other two shown above?

src/stories/cartesian/brush/BrushComponent.jsx
```diff
@@ -0,0 +1,84 @@
+/* eslint-disable react/prop-types */
+/* eslint-disable no-useless-constructor */
+/* eslint-disable import/no-default-export */
+import React, { Component } from 'react';
+import { Surface, Brush } from '../../..';
```
Yilun-Sun, line 5: Not sure if it's better to replace the import with an explicit path, just an idea. I faced a problem in my project with import path end with `'..'`. So to avoid potential problem, please consider the suggestion.
```typescript
import { Surface } from '../../../container/Surface';
import { Brush } from '../../../cartesian/Brush';
```

src/stories/cartesian/brush/BrushComponent.jsx
nikolasrieble: After validating that their violation is part of the storybook best practice setup, sure. I am not certain.

nikolasrieble:
@proke03 I assigned you to the first issues of the storybook milestone.

src/stories/cartesian/brush/BrushComponent.jsx
proke03: I was thinking of just integrating what dillonreedy had been working on.
I think that's better than starting with a storybook from nothing.

src/stories/cartesian/brush/BrushComponent.jsx
proke03: I thought the path ending in .. was odd, too. But I couldn't find a way.
Your suggestion sounds good, but I'd like to hear more.

src/stories/cartesian/brush/Brush.stories.jsx
proke03: Where do you think the data files should be located? 

Looking at our test code and examples, most of the data used for each chart is different, so I think it will be less reusable in storybooks.

nikolasrieble: In the story folder would be nice. If you look at the examples on the website, the mostly use the exact same dataset everywhere, with the 'pv' and 'uv' values.

src/stories/cartesian/brush/BrushComponent.jsx
nikolasrieble: Well each has pros and cons. Integrating @dillonreedy s work gives us a nice set a baseline stories, but we will have to refactor it. 

I think we should integrate his work indeed, but only after agreeing on a best practice setup with a single story. We need to be aligned on the baseline / expectation and then we can write issues to integrate in the same manner.

package.json
```diff
@@ -77,6 +80,16 @@
     "@babel/preset-react": "^7.6.3",
     "@babel/preset-typescript": "^7.6.0",
     "@babel/runtime": "^7.6.3",
+    "@storybook/addon-actions": "^6.5.16",
+    "@storybook/addon-essentials": "^6.5.16",
+    "@storybook/addon-interactions": "^6.5.16",
+    "@storybook/addon-links": "^6.5.16",
+    "@storybook/builder-webpack4": "^6.5.16",
```
nikolasrieble, line 87: I wonder, why do we need the builder-webpack4, if we use only webpack5?

proke03: It seems to have been created by storybook init.

nikolasrieble (changes requested):
Thank you! Please do (a) remove the BrushComponent in favour of an actual story template, which currently is unused: 

```

function Template(args: any) {
  return <BrushComponent {...args} />;
}
```

Secondly, please extend the README to share information about the chromatic setup, the existence of storybook, and what else you find helpful.

proke03:
@nikolasrieble
Removed the BrushComponent.

I don't think we need to edit the readme file.
Because distribution of storybooks and docs is not a change for people using our library.

nikolasrieble:
Code looks great, thank you. I have one last question - I just saw that Storybook 7 is a complete rewrite. 
As we start clean, should we right away start with Storybook 7 and save us later migration effort? 
It is in development since early 2022 and currently in beta. 

https://storybook.js.org/releases/7.0

https://storybook.js.org/blog/storybook-csf3-is-here

What do you think @proke03 @ckifer

nikolasrieble:
> @nikolasrieble Removed the BrushComponent.
> 
> I don't think we need to edit the readme file. Because distribution of storybooks and docs is not a change for people using our library.

I meant the readme for developers. The one that explains how to start the demo and where to find it. It should also have a section on how to start storybook. Everything that makes a developers life easier (i.e. discoverability of tools) is important.

ckifer:
Agreed with Niko here, I think we should go with 7. Worst that can happen is they make a few breaking changes on release which wouldn't be awful on our side.

package.json
```diff
@@ -77,6 +80,14 @@
     "@babel/preset-react": "^7.6.3",
     "@babel/preset-typescript": "^7.6.0",
     "@babel/runtime": "^7.6.3",
+    "@storybook/addon-actions": "^6.5.16",
+    "@storybook/addon-essentials": "^6.5.16",
+    "@storybook/addon-interactions": "^6.5.16",
+    "@storybook/addon-links": "^6.5.16",
+    "@storybook/builder-webpack5": "^6.5.16",
+    "@storybook/manager-webpack5": "^6.5.16",
+    "@storybook/react": "^6.5.16",
```
nikolasrieble, line 89: The last piece of the puzzle is upgrading the dependencies here: 

```
npm i @storybook/react@7.0.0-alpha.34
```

https://www.npmjs.com/package/@storybook/react?activeTab=versions

nikolasrieble:
@proke03 Should I take this over again? Happy to continue from here. Please let me know if you would like to land this still.

proke03:
@nikolasrieble 
Can you give me the slack link?
> As discussed in the recharts slack (feel free to join)

proke03:
Thank you!

PR #3297, opened 2023-01-28
feat: integrate docs

proke03:
## Types of changes

## Checklist:


nikolasrieble:
Amazing. It seems we can after all have a monorepo, as we host storybook via chromatic. @ckifer

nikolasrieble (commented):
Great work! Could you please extend the top level README file to explain about the structure of the repository, and how the parts play together (i.e. hosting storybook via chromatic, also integrated in the website)

Yilun-Sun:
Well done! Thanks for the first step to be able to change all three related project in a single PR. 

I'm trying to figure out why the refresh and examples not working in deployment. 

And one more thing, could you please make this scss class scope to the `iframe` in `Storybook.js`?
image
This scss contaminate the iframe in our home page which has a incorrect height and width.

proke03:
@Yilun-Sun 
Thank you!
Fixed iframe src.

I think the refresh issue is a problem caused by github pages not supporting SPA.
If we are hosting on the recharts.org address as before, changing the value below to 0 will solve it.
https://github.com/recharts/recharts/blob/72a080b3d45bf2f9eaeeaf55782d776e8e631e52/org/docs/404.html#L25

.gitignore
```diff
@@ -12,3 +12,7 @@ npm-debug.log
 .justconf/
 types
 build
+org/docs/*
+!org/docs/examples/
+!org/docs/index.html
+!org/docs/404.html
```
ckifer, line 18: nit: newline

org/src/views/Storybook.js
```diff
@@ -0,0 +1,42 @@
...
+      return <Installation locale={locale} />;
+    }
+    if (page === 'getting-started') {
+      return <GettingStarted locale={locale} />;
+    }
+    if (page === 'customize') {
+      return <Customize locale={locale} />;
+    }
+    return null;
+  }
```
ckifer, line 32: this isn't doing anything it seems

org/src/views/Storybook.js
```diff
@@ -0,0 +1,42 @@
...
+
+  render() {
+    const { page } = this.props;
+    const locale = getLocaleType(this.props);
```
ckifer, line 36: these aren't referenced, I wonder if we can still do translation in the storybook...

org/src/views/IndexView.js
```diff
@@ -0,0 +1,169 @@
...
+            </Link>
+          </p>
+          <iframe
+            title="star"
+            src="https://63d24aee84ee7cc215239dcc-yrlgloqamw.chromatic.com/"
+            frameBorder="0"
+            scrolling="0"
+            width="120px"
+            height="22px"
+          />
```
ckifer, line 82: this iframe rendering can be removed

org/src/views/Storybook.js
```diff
@@ -0,0 +1,42 @@
...
+
+    return <iframe src="https://63d24aee84ee7cc215239dcc-yrlgloqamw.chromatic.com/"/>;
```
ckifer, line 38: We have the new URL now so lets use https://63da8268a0da9970db6992aa-srbatovpky.chromatic.com/

instead

ckifer (changes requested):
left a few new comments.

The initial integration here is going to feel... gross I think.

If we want to set this up as a true monorepo we should use something like lerna or npm workspaces to make our lives a lot easier. Searching, building, etc. is going to be really tough without that

org/src/views/Storybook.js
nikolasrieble: https://storybook.js.org/addons/storybook-i18n/

ckifer:
Thanks for the changes - going to check this out again soon but don't want to merge it just yet. I'll pull it down at some point and make the the repo has everything setup to deploy pages correctly

nikolasrieble:
@proke03 Really this PR has two changes: (1) Functional - Show the storybook on the website, and (2) Organisational - Move the website in the monorepo. 

We might want to do some more preparations and invest into tooling before merging this PR due to (2), but (1) is independent. Would you consider to create a separate PR on the website for (1)? Happy to get that in right away.

ckifer:
+1 to the above, lets merge this first https://github.com/recharts/recharts.org/pull/252

nikolasrieble:
Given our current progress with storybook, we might consider to instead replace the website with storybook completely:
- pending the branding (i.e. theme storybook in alignment with the website)
- pending a deployment to recharts.org

ckifer:
Yeah I think I'm going to close this for now due to the above

PR #3313, opened 2023-02-02
Storybook: Refactor ComposedChart.stories.tsx

nikolasrieble:
## Description
- Use a single shared render function.
- Collect all stories for the component in a single file. 
- Add a README to refer to. 

- Delete the time scale example of a ComposedChart
The motivation to delete the time scale example is that the time scale behaviour is a function the XAxis. This behaviour should be shown in great detail on the XAxis story.

## Types of changes

## Checklist:


nikolasrieble:
@proke03 Please consider my refactoring. I followed the pattern described in https://storybook.js.org/blog/storybook-csf3-is-here to defined a template rendering function.

If we define such a template, then every story differs only in details, which helps the user better understand what is happening.

storybook/stories/chart/ComposedChart.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+export default {
+  title: 'Recharts/Chart/Composed Chart',
+  component: Composed,
+};
+
+// This render template can be shared across multiple stories.
+// All stories use the same data, but pass different children.
+const HorizontalChartTemplate = {
+  render: (args: any) => {
```
ckifer, line 29: I hate this `any` here but know why its needed 😅 

Could we maybe maybe this `Record<string, any>`? Just for a little bit of a stricter type

ckifer (approved):
Looks fine, my comment is a nit 🚀

proke03:
Thanks for making a good standard.

storybook/stories/chart/ComposedChart.stories.tsx
nikolasrieble: It was late, and I had to submit for review :D I agree, and will iterate.

PR #3321, opened 2023-02-03
docs(storybook): PieChart

proke03:
## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/chart/piechart/PieChart.stories.tsx
+++ b/storybook/stories/chart/piechart/PieChart.stories.tsx
@@ -11,7 +11,7 @@ export const Simple = {
   render: (args: Record<string, any>) => {
     return (
       <ResponsiveContainer width="100%" height={400}>
-        <PieChart width={800} height={400} startAngle={180} endAngle={0} outerRadius={80}>
+        <PieChart>
           <Pie data={args.data} dataKey="uv" />
         </PieChart>
       </ResponsiveContainer>
--- a/storybook/stories/chart/piechart/PieChart.stories.tsx
+++ b/storybook/stories/chart/piechart/PieChart.stories.tsx
@@ -12,7 +12,7 @@ export const Simple = {
     return (
       <ResponsiveContainer width="100%" height={400}>
         <PieChart>
-          <Pie data={args.data} dataKey="uv" />
+          <Pie dataKey="uv" />
         </PieChart>
       </ResponsiveContainer>
     );
--- a/storybook/stories/chart/piechart/PieChart.stories.tsx
+++ b/storybook/stories/chart/piechart/PieChart.stories.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { Pie, PieChart, ResponsiveContainer } from '../../../../src';
+import { pageData } from '../../data';
+
+export default {
+  title: 'Recharts/Chart/Pie Chart',
+  component: PieChart,
+};
+
+export const Simple = {
+  render: (args: Record<string, any>) => {
+    return (
+      <ResponsiveContainer width="100%" height={400}>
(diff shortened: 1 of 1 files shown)


storybook/stories/chart/piechart/PieChart.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+export default {
+  title: 'Recharts/Chart/Pie Chart',
+  component: PieChart,
+};
+
+export const Simple = {
+  render: (args: Record<string, any>) => {
+    return (
+      <ResponsiveContainer width="100%" height={400}>
+        <PieChart width={800} height={400} startAngle={180} endAngle={0} outerRadius={80}>
```
nikolasrieble, line 14: The simplest possible chart would be this one, right? 

```suggestion
        <PieChart>
```

storybook/stories/chart/piechart/PieChart.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+          <Pie data={args.data} dataKey="uv" />
```
nikolasrieble, line 15: Interesting, why do we pass the data to the Pie instead of to the Chart? 
I wonder which case we should suggest as the default / recommend / basic usage?

nikolasrieble: Consistent we the other basic stories, lets rather pass it to the PieChart. I think passing the data to the Pie should be an advanced story - because it begs the question what happens if we pass different data to different Pies

storybook/stories/chart/piechart/PieChart.stories.tsx
nikolasrieble, line 14:
```suggestion
        <PieChart {...args}>
```
Lets spread the args to make the controls functional.

storybook/stories/chart/piechart/PieChart.stories.tsx
nikolasrieble, line 15:
```suggestion
          <Pie dataKey="uv" />
```

storybook/stories/chart/piechart/PieChart.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+    data: pageData,
+  },
+};
+
+export const Donut = {
+  render: (args: Record<string, any>) => {
+    return (
+      <ResponsiveContainer width="100%" height={400}>
+        <PieChart width={800} height={400} startAngle={180} endAngle={0} outerRadius={80}>
+          <Pie data={args.data} dataKey="uv" cx={200} cy={200} innerRadius={50} outerRadius={80} />
```
nikolasrieble, line 30:
```suggestion
        <PieChart {...args}>
          <Pie dataKey="uv" innerRadius={50} outerRadius={80} />
```

nikolasrieble: This seems the simplest possible showcase.

nikolasrieble (changes requested):

storybook/stories/chart/piechart/PieChart.stories.tsx
proke03: Thanks for your review.
I will reflect what you said in the next stories.

nikolasrieble (approved):
🚀

nikolasrieble (approved):

storybook/stories/chart/piechart/PieChart.stories.tsx
nikolasrieble: Sorry, I was wrong. The interface is different on the PieChart, and it does not handle the data itself. 🤔

nikolasrieble: Fixed it in https://github.com/recharts/recharts/pull/3329

Issue #3310, opened 2023-02-02
Invalid value for prop `type` on <path> tag


minhTheGreat:
I got an issue when I set the type of lineChart is stepAround, it shows a warning
![Screenshot_2](https://user-images.githubusercontent.com/45730043/216309068-440a9b30-566a-45ad-a193-b0ed81b9edc0.png)
How can i fix it ?

ckifer:
@minhTheGreat can you create a codesandbox reproducing the issue?

ckifer:
`stepAround` is not a valid option for `type`.

See https://github.com/recharts/recharts/blob/1216e06412fd896da484bc0a84ca85d2b415fbd9/src/shape/Curve.tsx#L45-L58

ckifer:
Going to close this as resolved, please comment back if there is something I'm missing

minhTheGreat:
oh this is custom curve that implements step curve interpolators d3JS

ckifer:
Well type definitely isn't valid on path https://developer.mozilla.org/en-US/docs/Web/SVG/Element/path

It probably isn't getting filtered out correctly somewhere. I'll check this tomorrow. Thanks!

ckifer:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/type

Yeah.. basically because props are just blindly spread onto the element and `type` happens to be a valid SVG property it doesn't get filtered out like it should. This is fine when the type is a string or a number cause the DOM ignores it - not fine when its a function. 

None of the actual elements `type` is valid for are used in the project. Going to add a check for a function and remove that prop from allowed attributes

ckifer:
@minhTheGreat   #3327

ckifer:
Merged in #3327 

Resolving this, will be released in next minor version

ckifer:
Should be fixed in 2.4.0

PR #3327, opened 2023-02-03
fix: filter out functions in filterProps, remove type from valid SVGElementPropKeys

ckifer:
## Description

Recharts has a `filterProps` function which was intended to filter out all properties that are not valid SVGProps or events that need passed down to children.

Multiple times we have come across issues where Recharts defined props such as `type` (in the past `points`) overlap with traditional SVGProps (https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute)

In these cases props get spread onto SVG elements that aren't intended or valid.

In this case the `type` prop [is a valid SVG attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/type). This never caused error in the past because the DOM will ignore attributes that are strings or numbers, but when adding a line or area `type` property that is custom `CurveFactory` from d3 such as
```
const stepAround = curveCardinal.tension(0.5);
```
the DOM will throw an error "Invalid value for prop type on <path> tag" because functions are not valid on HTML elements in the DOM.

- remove `type` from valid `SVGElementAttributes` - we do not use any elements that need it
- add an `isFunction` check within `filterProps` for attributes that are supposed to be one of `SVGElementAttributes` - these should never be functions
- detail comment within `filterProps`
- adds an example using a `CurveFactory` in AreaChart demo

## Related Issue

https://github.com/recharts/recharts/issues/3310

## Motivation and Context

Resolve above described error. Prevent more errors with `type` in the future.

## How Has This Been Tested?

The true `type` SVG attribute can only be used with the following tags: 

image

None of which we use in Recharts.

- In demo slightly modify an example to use a `CurveFactory` - ensure the new curve factory type is displayed correctly and that the error is not thrown.
  - `type` is passed down explicitly by the chart so it is okay that it gets filtered out in `filterProps`
- check functionality of other chart types and `CurveType` values such as `monotone` and `linear`

image

^ No error thrown like in issue but curve still reflects the given `CurveFactory`

## Types of changes

## Checklist:

### DIFF
--- a/demo/component/AreaChart.tsx
+++ b/demo/component/AreaChart.tsx
@@ -15,6 +15,7 @@ import {
   Label,
 } from 'recharts';
 import { changeNumberOfData } from './utils';
+import { curveCardinal } from 'victory-vendor/d3-shape';
 
 const data = [
   { name: 'Page A', uv: 4000, pv: 2400, amt: 2400, time: 1 },
@@ -118,6 +119,9 @@ const renderLabel = (props: any) => {
     </text>
   );
 };
+
+// custom curve cardinal `type` prop
+const stepAround = curveCardinal.tension(0.5);
 
 export default class AreaChartDemo extends React.Component<any, any> {
   static displayName = 'AreaChartDemo';
@@ -269,16 +273,14 @@ export default class AreaChartDemo extends React.Component<any, any> {
 
         <p>Tiny AreaChart</p>
         <div className="area-chart-wrapper">
-          <AreaChart width={100} height={50} data={data.slice(0, 1)} margin={{ top: 5, right: 0, left: 0, bottom: 5 }}
-          >
+          <AreaChart width={100} height={50} data={data.slice(0, 1)} margin={{ top: 5, right: 0, left: 0, bottom: 5 }}>
             <Area type="monotone" dataKey="uv" stroke="#ff7300" fill="#ff7300" />
           </AreaChart>
         </div>
 
         <p>AreaChart with three y-axes</p>
         <div className="area-chart-wrapper">
-          <AreaChart width={600} height={400} data={data} margin={{ top: 20, right: 30, left: 20, bottom: 5 }}
-          >
+          <AreaChart width={600} height={400} data={data} margin={{ top: 20, right: 30, left: 20, bottom: 5 }}>
             <YAxis type="number" yAxisId={0} stroke="#ff7300">
               <Label position="top" offset={10}>
                 uv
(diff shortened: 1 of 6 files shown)


demo/component/AreaChart.tsx
```diff
@@ -313,27 +315,23 @@ export default class AreaChartDemo extends React.Component<any, any> {
...
             <XAxis dataKey="name" />
             <YAxis tickCount={7} />
             <Tooltip content={<CustomTooltip external={data} />} />
             <CartesianGrid stroke="#f5f5f5" />
             <ReferenceArea x1="Page A" x2="Page E" />
             <ReferenceLine y={7500} stroke="#387908" />
             <ReferenceDot x="Page C" y={1398} r={10} fill="#387908" isFront />
-            <Area type="monotone" dataKey="pv" stroke="#ff7300" fill="#ff7300" fillOpacity={0.9}
-            />
+            <Area type={stepAround} dataKey="pv" stroke="#ff7300" fill="#ff7300" fillOpacity={0.9} />
```
nikolasrieble, line 334: As we already have a story for the AreaChart, could you please add an example with this custom type to storybook?
```
<AreaChart data={data}>
    <Area type={stepAround} dataKey="pv" stroke="#ff7300" fill="#ff7300" fillOpacity={0.9} />
</AreaChart>
```

src/util/ReactUtils.ts
```diff
@@ -305,10 +305,17 @@ export const filterProps = (
...
+   *   - functions in properties that are SVG attributes
+   *   - props that are SVG attributes but don't matched the passed svgElementType
+   *   - any prop that is not in SVGElementPropKeys (or in EventKeys if includeEvents is true)
+   */
   Object.keys(inputProps).forEach(key => {
     if (
-      (svgElementType && matchingElementTypeKeys.includes(key)) ||
-      SVGElementPropKeys.includes(key) ||
+      (!_.isFunction(inputProps?.[key]) &&
+        ((svgElementType && matchingElementTypeKeys.includes(key)) || SVGElementPropKeys.includes(key))) ||
```
nikolasrieble, line 318: Thanks for adding the comment. How about refactoring this into a separate function? Such a separate function could then be unit tested nicely.

demo/component/AreaChart.tsx
nikolasrieble: That would also provide a test for this feature, once we have the storybook test runner up.

proke03: Opened PR
https://github.com/ckifer/recharts/pull/1

src/util/types.ts
```diff
@@ -584,7 +584,6 @@ export const SVGElementPropKeys = [
   'name',
   'style',
   'target',
-  'type',
```
nikolasrieble, line 587: Please do add a comment as to why type is not used here. 
Otherwise another engineer might one day add this back, arguing `Well type exists as a prop, see https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/type`.

And yes, it does, but it conflicts with an existing recharts prop, thus we do not allow it.

nikolasrieble (approved):
Good cleanup, thank you! It would be nice to capture in a comment as to why TYPE in no longer in the list of SVGElementPropKeys.

Either way this already is an improvement.

src/util/ReactUtils.ts
```diff
@@ -305,10 +305,17 @@ export const filterProps = (
    */
   const matchingElementTypeKeys = FilteredElementKeyMap?.[svgElementType] ?? [];
 
+  /**
+   * Props are blindly spread onto SVG elements. This loop filters out properties that we don't want to spread.
+   * Items filtered out are as follows:
+   *   - functions in properties that are SVG attributes
```
nikolasrieble, line 311:
```suggestion
   *   - functions
```

nikolasrieble: All functions are filtered out.

nikolasrieble (approved):
I do not see any issue and could not find any while testing. 
Please do consider my refactoring suggestion.

Good to go.

proke03 (approved):

src/util/types.ts
ckifer: > conflicts with an existing recharts prop, thus we do not allow it.

I don't think this is quite the right way to frame it as we do have some conflicting props that we have to allow (can't think off the top of my head but I know they're there). More like we can remove it until there is a use-case for an SVG element that utilizes the attribute (if ever).

Not sure that adding comments inline for every removed attribute is sustainable... I would love to utilize that SVG attribute type library I have mentioned in the past. But will add something here for now

src/util/ReactUtils.ts
ckifer: if `includeEvents` is true then there are functions that are passed down. `includeEvents` is only set to true when props are spread via our components (such as Curve) and never when props are spread onto an SVG element

src/util/ReactUtils.ts
ckifer: Hmm, the point of the function is to filter props and this is the part that does the filtration. (imo) sometimes factoring out too much can lead to more confusion, but in this case something like `isValidSpreadableProp` maybe would do for the check itself?

ckifer:
Should have requested updates now!

PR #3328, opened 2023-02-05
fix: remove axis domain default prop and calculate it dynamically

### LINKED ISSUE #2593: Empty BarChart shows "0 auto" as X Axis ticks

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/simple-bar-chart-empty-x-axis-issue-tfep4)

### Steps to reproduce
Simply having a BarChart with an empty data array. Doesn't matter what other components are used within other than having the XAxis component. 

Just use any version of recharts 2.0.0 or higher including latest 2.0.10

### What is expected?
The chart will display with the 2 axis borders but no axis ticks.

The chart displays 2 X Axis ticks:
- 0
- auto

| Environment | Info |
|---|---|
| Recharts | v2.0.8 |
| React | 17.0.1 |
| System | Linux |
| Browser | Firefox 89 or Chrome 91 |

---
This behavior has changed from version 1.X. Previously it did not display the X Axis at all if the data was empty.

See <https://codesandbox.io/s/recharts-issue-template-forked-45fdr> for example.

lukask-proxora:
## Description

Fixes #2593.

I removed the `domain` prop from the default props for all axis components. As described in the issue comments, those default props do not take the axis type into account and thereby use unsuitable domains in some cases. This results in the bug.

Instead of using the component default props, the fallback domain is now determined when the axis map is created. At this point the type of the axis is known and is used to determine which fallback domain to use.

## Related Issue

#2593

## Motivation and Context

Solves a bug.

## How Has This Been Tested?

I added a unit test for the BarChart component, as this specific case was mentioned in the bug ticket.
I also tested charts with all the modified axis in the demo and storybook. Both with and without specified domain.

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -197,6 +197,10 @@ function isDomainSpecifiedByUser(domain: AxisDomain, allowDataOverflow: boolean,
   return false;
 }
 
+function getDefaultDomainByAxisType(axisType: 'number' | string) {
+  return axisType === 'number' ? [0, 'auto'] : undefined;
+}
+
 /**
  * Get the content to be displayed in the tooltip
  * @param  {Object} state          Current state
@@ -327,7 +331,7 @@ const getAxisMapByAxes = (
 
     // we didn't create the domain from user's props above, so we need to calculate it
     if (!domain || domain.length === 0) {
-      const childDomain = child.props.domain ?? (type === 'number' ? [0, 'auto'] : undefined);
+      const childDomain = child.props.domain ?? getDefaultDomainByAxisType(type);
 
       if (dataKey) {
         // has dataKey in <Axis />
@@ -468,7 +472,7 @@ const getAxisMapByItems = (
   const axisMap = graphicalItems.reduce((result: any, child: any) => {
     const axisId = child.props[axisIdKey];
 
-    const originalDomain = Axis.defaultProps.domain ?? [0, 'auto'];
+    const originalDomain = getDefaultDomainByAxisType('number');
 
     if (!result[axisId]) {
       index++;
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -327,6 +327,8 @@ const getAxisMapByAxes = (
 
     // we didn't create the domain from user's props above, so we need to calculate it
     if (!domain || domain.length === 0) {
+      const childDomain = child.props.domain ?? (type === 'number' ? [0, 'auto'] : undefined);
+
       if (dataKey) {
         // has dataKey in <Axis />
         domain = getDomainOfDataByKey(displayedData, dataKey, type);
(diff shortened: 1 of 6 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -327,6 +327,8 @@ const getAxisMapByAxes = (
 
     // we didn't create the domain from user's props above, so we need to calculate it
     if (!domain || domain.length === 0) {
+      const childDomain = child.props.domain ?? (type === 'number' ? [0, 'auto'] : undefined);
```
nikolasrieble, line 330: It might be nice to make this a bit more verbose, to improve readability. 
This line define the default domain, given a type. 

Please do instead extract a function called 

```
getDefaultDomainByType(type) {
    return type === 'number' ? [0, 'auto'] : undefined
}

```
```suggestion
      const childDomain = child.props.domain ?? getDefaultDomainByType(type);
```

This is only required to not further complicate this largest of files with most of complexity.

test/cartesian/XAxis.spec.tsx
```diff
@@ -162,4 +162,16 @@ describe('<XAxis />', () => {
     const bar = container.querySelector('.recharts-rectangle');
     expect(parseInt(bar?.getAttribute('x') as string, 10)).toEqual(66);
   });
+
+  test('Render no ticks if type is category and data is empty', () => {
```
nikolasrieble, line 166: ❤️ Thank you for adding a test as well.

nikolasrieble (commented):
This is an improvement and makes sense. It does not break the existing API, but merely fixes a bug.

Thank you for your contribution. I will still have to test this, to be confident, and will do that separately. Please consider my refactoring suggestion.

nikolasrieble (changes requested):
It seems there is yet another place where the default is set. 

```
  // The default type of x-axis is category axis,
  // The default contents of x-axis is the serial numbers of data
  // The default type of y-axis is number axis
  // The default contents of y-axis is the domain of data
  const axisMap = graphicalItems.reduce((result: any, child: any) => {
    const axisId = child.props[axisIdKey];

    const originalDomain = Axis.defaultProps.domain ?? [0, 'auto'];
```

I guess we should also use the getDefaultDomainByType in this place as well.

lukask-proxora:
Thanks for the feedback @nikolasrieble :+1: 

I will extract a function for the default domain based on the axis type and use this function within `getAxisMapByAxes`.

However, in the second case (`getAxisMapByItems`) we don't have the type of the axis, since the axis map is only created from the axis component types. Actually, since the default domain props were removed now we could ignore them here and always use the default `const originalDomain = [0, 'auto']`. What do you think?

nikolasrieble:
> Thanks for the feedback @nikolasrieble 👍
> 
> I will extract a function for the default domain based on the axis type and use this function within `getAxisMapByAxes`.
> 
> However, in the second case (`getAxisMapByItems`) we don't have the type of the axis, since the axis map is only created from the axis component types. Actually, since the default domain props were removed now we could ignore them here and always use the default `const originalDomain = [0, 'auto']`. What do you think?
> 
> Edit: Of course we could also use the function and pass `'number'` as fixed type in that case. Then the default domain is only defined once in the new function.

Right, I would prefer the latter, merely to have defaults only defined in a single place.

ckifer (approved):
Code itself looks good to me. @nikolasrieble - is there more action required here before merge?

nikolasrieble:
> Code itself looks good to me. @nikolasrieble - is there more action required here before merge?

No, all good, we can refactor this in a follow up.

nikolasrieble (approved):
🚀

ckifer:
@lukask-proxora 

We missed one thing here that broke some stuff https://github.com/recharts/recharts/issues/2593

Fixed now

lukask-proxora:
Sorry, was not expecting that the originalDomain would be used to calculate the ticks. I should have tested the default case more intensively.

@ckifer Thanks for the fix

PR #3336, opened 2023-02-07
refactor: tooltip to function component

ckifer:
## Description

Refactor `Tooltip` to be a function component

## Motivation and Context

Refactoring library to be more maintainable - testing refactorability

https://user-images.githubusercontent.com/25180830/233861482-13dd6acb-c8cc-4058-99e5-cf3c8231fd79.mov

## Types of changes

## Checklist:


ckifer:
@nikolasrieble - did this as a draft for fun, definitely not done or tested. Not sure if we want to take tasks like this yet but oh well.

I actually think I notice some performance improvement over the class based approach but I'm not 100% sure. Feel free to pull down and play

nikolasrieble:
Amazing, will test it today.

test/chart/LineChart.spec.tsx
```diff
@@ -648,7 +648,7 @@ describe('<LineChart /> - Rendering two line charts with syncId', () => {
         <XAxis dataKey="name" />
       </LineChart>
     );
-    const { container, getByText } = render(
+    const { container, getByText, debug } = render(
```
ckifer, line 651: oops, left these here. Not planning on merging this so will leave them for now

storybook/stories/API/chart/BarChart.stories.tsx
```diff
@@ -12,6 +12,7 @@ export const Simple = {
       <ResponsiveContainer width="100%" height={400}>
         <BarChart {...args}>
           <Bar dataKey="uv" />
+          <Tooltip reverseDirection={{ x: true }} />
```
ckifer, line 15: same with this, just testing. Would remove in non-draft

ckifer:
I mean it works 😅  - probably not a priority rn though

nikolasrieble:
> I mean it works 😅 - probably not a priority rn though

Well, a refactoring is always important. I will test this thoroughly and then lets see whether the behaviour replicates.

src/component/Tooltip.tsx
```diff
@@ -202,101 +196,120 @@ export class Tooltip<TValue extends ValueType, TName extends NameType> extends P
...
+Tooltip.displayName = 'Tooltip';
+
+/**
+ * needs to be set so that renderByOrder can access an have default values for
+ * children.props when there are no props set by the consumer
+ * doesn't work if using default parameters
+ */
+Tooltip.defaultProps = tooltipDefaultProps;
+
+export { Tooltip };
```
nikolasrieble, line 315: Nit: Why export it in an object? Why not simply `export Tooltip`?

src/component/Tooltip.tsx
```diff
@@ -202,101 +196,120 @@ export class Tooltip<TValue extends ValueType, TName extends NameType> extends P
...
+              ...prev.dismissedAtCoordinate,
+              x: coordinate.x,
+              y: coordinate.y,
+            },
+          }));
+        }
+      }}
+      className={cls}
+      style={outerStyle}
+      ref={wrapperNode}
```
nikolasrieble, line 295: I am not sure whether this change equivalent.

src/component/Tooltip.tsx
```diff
@@ -202,101 +196,120 @@ export class Tooltip<TValue extends ValueType, TName extends NameType> extends P
...
+    </div>
+  );
+};
 
-    return (
-      // ESLint is disabled to allow listening to the `Escape` key. Refer to
-      // https://github.com/recharts/recharts/pull/2925
-      // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
-      <div
-        tabIndex={-1}
```
nikolasrieble, line 278: Why remove the tabIndex?

src/component/Tooltip.tsx
```diff
@@ -90,82 +90,78 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Def
...
+  filterNull: true,
+  useTranslate3d: false,
+  cursor: true,
+};
 
-  state = {
+const Tooltip = <TValue extends ValueType, TName extends NameType>(
+  props: TooltipProps<TValue, TName> & { children?: React.ReactNode },
+) => {
+  const [state, setState] = useState({
```
nikolasrieble, line 120: While this works, it is recommended to split the state into separate variables. 
https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables

src/component/Tooltip.tsx
```diff
@@ -90,82 +90,78 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Def
...
-    animationDuration: 400,
-    filterNull: true,
-    useTranslate3d: false,
-  };
+const tooltipDefaultProps = {
+  active: false,
+  allowEscapeViewBox: { x: false, y: false },
+  reverseDirection: { x: false, y: false },
+  offset: 10,
+  viewBox: { x: 100, y: 100, height: 50, width: 50 },
```
nikolasrieble, line 98: What is the reasoning for these specific values? - as opposed to the existing ones?

src/component/Tooltip.tsx
```diff
@@ -90,82 +90,78 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Def
...
-    viewBox: { x1: 0, x2: 0, y1: 0, y2: 0 },
-    coordinate: { x: 0, y: 0 },
-    cursorStyle: {},
-    separator: ' : ',
-    wrapperStyle: {},
-    contentStyle: {},
-    itemStyle: {},
-    labelStyle: {},
-    cursor: true,
-    trigger: 'hover',
```
nikolasrieble, line 112: Why remove these two?

src/component/Tooltip.tsx
```diff
@@ -90,82 +90,78 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Def
...
         boxHeight: -1,
-      });
+      }));
     }
-  }
+  };
+
+  useEffect(() => {
+    updateBBox();
+  }, []);
```
nikolasrieble, line 162: Previously `updateBBox()` was called both on onComponentMount as well as onComponentUpdate.

This is not equivalent. The useEffect will only fire onMount, but not onUpdate. 
If the behaviour should replicate, then state would have to be added into the dependency array.

nikolasrieble (commented):
Sorry that the review took so long, I was heads down in storybook. Have a few questions.

src/component/Tooltip.tsx
ckifer: `{ x1: 0, x2: 0, y1: 0, y2: 0 }` - these are not valid values on `viewBox` - there is no such thing as x1, y1, x2, y2

As for the values I think these are the default viewBox values for the tooltip but I honestly forget. Will check

src/component/Tooltip.tsx
ckifer: agreed, was doing this for the sake of speed when originally refactoring. Can revise

src/component/Tooltip.tsx
ckifer: which two?
`cursor: true,` is still there

`trigger: 'hover',` - I forget, might just be a miss (though it does get set without this defaulted)

will add it back

src/component/Tooltip.tsx
ckifer: This should be in a separate PR sorry - but imho it shouldn't be there. It was added with accessibility in mind and its only made accessibility worse by blurring focused elements (say in a form) on chart hover.

Will add it back for this PRs sake

src/component/Tooltip.tsx
ckifer: oh, habit I guess -  I'm used to exporting multiple things at the bottom of a file

src/component/Tooltip.tsx
ckifer: Not sure either - this is the approach I would take with refs in general (and how they should be done now), but might be a breaking change. Will test to see if it is. Can revert to a legacy ref if needed

ckifer (commented):
Thanks for the review, will make some changes

ckifer:
@nikolasrieble I think I addressed all comments here, this _should_ be ready for re-review and testing. On the storybook all looks good to me. 

I'm going to leave the `useRef` change, I'm confident that the changes are equivalent

src/chart/generateCategoricalChart.tsx
```diff
@@ -2157,6 +2157,7 @@ export const generateCategoricalChart = ({
       }
 
       const { children, className, width, height, style, compact, title, desc, ...others } = this.props;
+
```
nikolasrieble, line 2160:
```suggestion
```
Unrelated change.

src/component/Tooltip.tsx
```diff
@@ -90,99 +99,97 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Def
...
+  contentStyle: {},
+  itemStyle: {},
+  labelStyle: {},
+  isAnimationActive: !Global.isSsr,
+  animationEasing: 'ease',
+  animationDuration: 400,
+  filterNull: true,
+  useTranslate3d: false,
+  cursor: true,
+  trigger: 'hover',
```
nikolasrieble, line 124: Nit: Any reason these a ordered differently now? If you want to change the order, alphabetically might be nice. Else keeping a PR as small as possible (i.e. no redundant changes) would set a great example.

src/component/Tooltip.tsx
nikolasrieble: I understand the change, but it seems an unrelated bug fix. I did not find these default values anywhere, I believe this might be changing behaviour. Please consider doing this change in a prefactoring PR, keeping the refactoring class --> function a pure refactoring without behavioural change.

src/util/ReactUtils.ts
```diff
@@ -416,8 +416,8 @@ export const renderByOrder = (children: React.ReactElement[], renderMap: any) =>
       elements.push(child);
     } else if (child) {
       const displayName = getDisplayName(child.type);
-      const { handler, once } = renderMap[displayName] || {};
 
+      const { handler, once } = renderMap[displayName] || {};
```
nikolasrieble, line 420: Nit: Ideally no redundant changes.

src/component/Tooltip.tsx
nikolasrieble: I will test the Tooltip behaviour on this PR. This is the only change that I suspect might be an issue / change behaviour.

src/chart/generateCategoricalChart.tsx
ckifer: oh sorry, unintentional

src/component/Tooltip.tsx
ckifer: Hmm okay, I think I got these values from somewhere but I don't remember exactly where. Either way these x1, y1 etc. fields are incorrect and need addressed somewhere. I can ts-ignore them here and fix later

src/component/Tooltip.tsx
ckifer: No reason, just how it happened. Will move back to original order

ckifer (commented):
Will put out one more revision

src/component/Tooltip.tsx
```diff
@@ -90,99 +99,100 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Def
...
-        dismissed: true,
-        dismissedAtCoordinate: {
-          ...this.state.dismissedAtCoordinate,
-          x: this.props.coordinate.x,
-          y: this.props.coordinate.y,
-        },
-      });
-    }
-  };
+const tooltipDefaultProps: TooltipProps<number, string> = {
```
ckifer, line 102: typed as number, string for default props. Doesn't really affect anything except for typing this explicitly

src/component/Tooltip.tsx
```diff
@@ -90,99 +99,100 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Def
...
+  active: false,
+  allowEscapeViewBox: { x: false, y: false },
+  reverseDirection: { x: false, y: false },
+  offset: 10,
+  // TODO: replace invalid properties
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  viewBox: { x1: 0, x2: 0, y1: 0, y2: 0 },
```
ckifer, line 110: these are incorrect default values - fix for this out of scope of PR

ckifer (commented):
@nikolasrieble requested changes made. defaultProps back in original order and original content. Check out the video in the PR description for a glance at functionality without having to pull the branch down

nikolasrieble (approved):
Great, thank you for addressing all comments! 🚀 Would you like to create a follow up issue for the invalid default values of the viewBox?

ckifer:
https://github.com/recharts/recharts/issues/3548

ckifer:
Lets test this in the storybook deployment

Issue #3344, opened 2023-02-07
LabelList - Y value is undefined on content props


benedekdaniel:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/tender-bose-6eh5cs)

### Steps to reproduce
Try simply changing the recharts version from 1.8.5 to 2.3.2 in the package.json.

### What is expected?
It's expected that the y value on the LabelList content props is not undefined but returns a dynamic value relative to the chart's y and x-axis so it always fits on the x axis.

### What is actually happening?
Notice that the custom label component rendered in LabelList doesn't have an y value (it's undefined) in version 2.3.2. After setting the version back to 1.8.5 the noDataLabel chart appears as y gets a value. 

| Environment | Info |
|---|---|
| Recharts | v2.3.2 |
| React | 18.0.0 |
| System | MacOs Monterey version 12.4 |
| Browser | Chrome Version 109.0.5414.119 (Official Build) (arm64) |

ckifer:
Hey there @benedekdaniel 

From what I can test the `y` value breaks as of version 2.0.10 - it is present in `2.0.9`.

Try upgrading to 2.0.9 and in the meantime I'll look into the breaking change.
Upgraded CodeSandbox: https://codesandbox.io/s/kind-lovelace-f0bzs2?file=/src/App.tsx

ckifer:
Diff between versions - https://github.com/recharts/recharts/compare/v2.0.9...v2.0.10

Probably this PR  https://github.com/recharts/recharts/pull/2477

ckifer:
I can't really tell what is happening based on the diff alone - nothing really looks out of place there. 

I do find it odd that you are using `LabelList` to render a bar item.
There might be a better approach for this.

ckifer:
Really confused on why this broke specifically in 2.0.10.

This line: https://github.com/recharts/recharts/blob/master/src/cartesian/Bar.tsx#L195
Has potential to make height `NaN` (makes sense. 300 - undefined = `NaN`). That causes some calculations down the line to use `NaN` and therefore make `y` undefined or `NaN`. Defaulting this to 0... breaks more things because it results in the bars that are `null` being totally full. 

i.e.
[...]

PR #3346, opened 2023-02-08
fix: NaN for Bar y and height attributes if data is null

ckifer:
## Description

If bar data is `null` then bar `y` and `height` value come back as `NaN` due to some bad optional handling.

This PR attempts to add good default values if this is the case.

(will consider this a draft but feel free to review)

## Related Issue

#3344

image

## Types of changes

## Checklist:

### DIFF
--- a/demo/component/BarChart.tsx
+++ b/demo/component/BarChart.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-unresolved */
 /* eslint-disable react/no-multi-comp */
 /* eslint-disable max-classes-per-file */
 import React, { Component } from 'react';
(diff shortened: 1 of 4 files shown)


demo/component/BarChart.tsx
```diff
@@ -126,40 +127,36 @@ const pageData = [
   { name: 'Page G', uv: 189, pv: 4800, amt: 2400 },
 ];
 
-const RenderLabel = (props: any) => {
```
nikolasrieble, line 129: I would suggest to keep the diff in a fix PR as small as possible. This seems to be an unrelated cleanup.

src/cartesian/Bar.tsx
```diff
@@ -226,7 +227,7 @@ export class Bar extends PureComponent<Props, State> {
         x,
         y,
         width,
-        height,
+        height: Number.isNaN(height) ? 0 : height,
```
nikolasrieble, line 230: This line really is the only relevant change, right? The rest is just refactoring.

nikolasrieble: The height is also used for the Tooltip position a few lines below this one. We should move `const validHeight = Number.isNaN(height) ? 0 : height` up a few lines and then use in both places in the return type.

nikolasrieble (changes requested):
Lets use the validHeight also for the Tooltip

demo/component/BarChart.tsx
ckifer: yeah sorry I couldn't help myself - I can't stand all the errors in these demo files LOL

src/cartesian/Bar.tsx
ckifer: line 194 is actually more relevant. 

line 194:
If we don't have a currentValueScale (aka, current value is null or undefined) then use the baseValueScale (aka whatever y value is computed at first without taking the value into account, and if we don't have that then use undefined and allow `NaN` down the line because something is wrong.

ckifer: will check the tooltip comment out, I'm not sure

ckifer: It turns out that height does not affect the tooltip because when the tooltip isn't shown (since there's no data) these values aren't used. But better safe than sorry so moved the check up.

ckifer:
Requested changes made, this should be good to go

nikolasrieble (approved):
Thank you!

Issue #3299, opened 2023-01-28
Storybook: Test interaction using the play function


nikolasrieble:
With the https://storybook.js.org/docs/react/writing-stories/play-function we can easily simulate interaction and then verify that it runs without error.

Acceptance criteria: 
- A single story has a play function

PR #3365, opened 2023-02-14
docs(storybook): test interaction using the play function

Yilun-Sun:
## Description

- add a test interaction demo to `Brush` using the play function.

## Related Issue

#3299

image

## Types of changes

## Checklist:


package.json
```diff
@@ -88,7 +88,7 @@
     "@storybook/addon-links": "^7.0.0-beta.47",
     "@storybook/react": "^7.0.0-beta.47",
     "@storybook/react-webpack5": "^7.0.0-beta.47",
-    "@storybook/testing-library": "^0.0.13",
+    "@storybook/testing-library": "^0.0.14-next.1",
```
ckifer, line 91: I think this is fine but just curious, what comes with this upgrade?

ckifer (approved):
Looks good, just a question

storybook/stories/API/chart/LineChart.stories.tsx
```diff
@@ -20,3 +23,28 @@ export const Simple = {
     data: pageData,
   },
 };
+
+export const PlayMoveBrush = {
```
nikolasrieble, line 27: Love the story and the interaction. Yet I think this story is better fitting in the Brush API. Instead of adding a brush example to every chart type, please move this over to: 

`storybook/stories/API/cartesian/Brush.stories.tsx`

nikolasrieble (changes requested):
Great, thank you! Please do move the brush story over the Brush stories though, instead of adding it to the LineChart.

nikolasrieble (approved):
Given that this is already functional it is good to go, we can also move this afterwards.

package.json
Yilun-Sun: We got an error and the solution from: https://github.com/storybookjs/testing-library/issues/32

> It looks like this is because you are using storybook v7. The release notes for @storybook/testing-library show that support for storybook v7 is being added in @storybook/testing-library v0.0.14.
To get it working right now you can install the next release, currently it is @storybook/testing-library@v0.0.14-next.1.

proke03 (changes requested):
Looks good.
Rather than creating a separate Number.ts file, I think it's better to put that data in the Page.ts file.

nikolasrieble (approved):
🚀

ckifer (approved):

PR #3371, opened 2023-02-14
Storybook: Add BarInBar story

nikolasrieble:
## Description
Answering a request on SO https://stackoverflow.com/questions/75401714/is-it-possible-to-stack-2-bars-layer-on-top-of-each-other-not-above-each-other

Screenshot 2023-02-14 at 22 42 17

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/API/chart/BarChart.stories.tsx
+++ b/storybook/stories/API/chart/BarChart.stories.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Bar, BarChart, ResponsiveContainer } from '../../../../src';
+import { Bar, BarChart, ResponsiveContainer, XAxis } from '../../../../src';
 import { pageData } from '../../data';
 
 export default {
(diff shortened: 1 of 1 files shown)


proke03 (commented):
Looks good.
But I don't know if this type of bar chart is used enough to be added as a story.
If we need to compare multiple values, it seems more common to put the bars side by side like below without overlapping.
column-chart-excel

ckifer:
Yeah +1, the SOF use-case is very niche

nikolasrieble:
> Looks good. But I don't know if this type of bar chart is used enough to be added as a story. If we need to compare multiple values, it seems more common to put the bars side by side like below without overlapping. column-chart-excel

I agree, one could argue this belongs into examples more than into the API folder, being an edge case. Should I move it there?

proke03:
If you move to an example, how about changing the example data to be closer to the real example like the data representing the achievement of the target number.

PR #3386, opened 2023-02-21
Refactoring: Extract TicksUtils from CartesianAxis and add unit tests

nikolasrieble:
## Description
In preparation of further work on the ticks, we add unit tests and refactor. 
This PR does not change any behaviour. 

In this PR:
- Extract getTicks, getTicksStart, getTicksEnd, getNumberIntervalTicks from CartesianAxis into TickUtils.
- Add multiple unit tests to the getTicks function to document existing behaviour. 

This unblocks:
- Further cleanup and refactoring of the getTicks function, which largely is duplicate code.
- Adding a new method to compute ticks.

## Types of changes

## Checklist:


nikolasrieble:
TODO: Add a story for the Axis with controls for interval.

test/util/TickUtils.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+        {
+          coordinate: 250,
+          isShow: true,
+          tickCoord: 20,
+          value: 90,
+        },
+      ]);
+    });
+
+    test('0', () => {
```
petii, line 120: You could consider parametrizing the numeric interval tests or give more meaningful names, like "interval 0 - includes every tick", "interval N - every Nth" and then it might also be worthwhile to write test cases for inputs like _-1_ or when the number is bigger than the number of ticks (and maybe some from other edge cases like NaN, not integers, the maximum number value, infinity etc)

petii (commented):
One issue I see with moving these functions out from under "cartesian" is that polar charts also have axis components ([PolarAngleAxis](https://recharts.org/en-US/api/PolarAngleAxis), [PolarRadiusAxis](https://recharts.org/en-US/api/PolarRadiusAxis)) which themselves have ticks as well and I'm not entirely sure whether this logic is shared between the two.

nikolasrieble:
> One issue I see with moving these functions out from under "cartesian" is that polar charts also have axis components ([PolarAngleAxis](https://recharts.org/en-US/api/PolarAngleAxis), [PolarRadiusAxis](https://recharts.org/en-US/api/PolarRadiusAxis)) which themselves have ticks as well and I'm not entirely sure whether this logic is shared between the two.

Good point, it might make sense to call this file CartesianTickUtils instead to communicate the context.

test/util/TickUtils.spec.ts
ckifer: +1 for more meaningful names. Else could run these tests in a loop using `each`

ckifer (approved):
LGTM to start - would consider the suggestions from @petii !

src/util/TickUtils.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+  }
+
+  return result.filter(entry => entry.isShow);
+}
+
+// todo Array<Tick>
+// TODO: Step 1 - Add unit tests. Document existing behaviour
+// Step 2 - add unit tests for equidistant behaviour.
+// Do it.
```
nikolasrieble, line 158:
```suggestion
```

test/util/TickUtils.spec.ts
nikolasrieble: Good point, done.

storybook/stories/data/Tick.ts
```diff
@@ -1,9 +1,16 @@
-const ticks = [
+import { TickItem } from '../../../src/util/types';
+
+const ticks: TickItem[] = [
   { value: 10, coordinate: 50 },
```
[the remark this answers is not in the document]
nikolasrieble: The coordinates is something that is generated internally and to my knowledge always equidistant. Thus I am not sure this case would represent something that can happen on the happy path. Good point though to add it as an edge case.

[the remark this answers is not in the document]
petii: Right, then it could also be extended by an additional first point. My point being that this way there is no visual difference between _preserveEnd_ and _preserveStartEnd_

Issue #2562, opened 2021-06-02
There is no way to display the XAxis values equidistant for each resolution.


tobiasAuzmendi:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/determined-easley-vjr7s?file=/src/index.js)

### Steps to reproduce
No steps needed.

### What is expected?
I expect to have at least the first and the latest value on the XAxis always visible independently of the resolution or the container size. 
I also expect to have equidistant XAxis values.

### What is actually happening?
I expect to have at least the first and the latest value on the XAxis always visible independently of the resolution or the container size. 
I also expect to have equidistant XAxis values, sometimes it seems to work inconsistently because there is no logical relation between shown label positions. For example, in the second chart that you can see in the code sandbox I'm providing I expect to see the 1st, 4th, and 7th labels, but I just see the 3rd and 7th, which seems really inconsistent from the user point of view.

| Environment | Info |
|---|---|
| Recharts | v2.0.8 |
| React | 17.0.1 |
| System | MacOs |
| Browser | Chrome |

nikolasrieble:
## The current state 

The **interval** property on the CartesianAxis influences the tick positions. 

With the three option: "preserveStart", "preserveEnd", "preserveStartEnd" a single tick is hidden, if it overflow the allowed area. 

If the interval is set to a number, the tick width is not considered at all, but ticks are drawn overlapping. The interval number signifies the amount of ticks to be hidden between two subsequent ticks. 

0 --> show all
1 --> show every second
...

## The enhancement
Instead of changing the behaviour of all existing options, we could add a new option: 

"start-equidistant"

This could later be extended to also support "end-equidistant".

Start equidistant would position all ticks such that:
- no tick is overlapping with another 
- every nTH tick is shown (i.e. equidistant)
- the first tick is always shown. 

How does that sound? @ckifer @proke03 @Yilun-Sun

ckifer:
Released in 2.5!

ckifer:
enhanced in 2.9

PR #3392, opened 2023-02-22
CartesianAxis: Extend interval option by equidistantPreserveStart

nikolasrieble:
## Description

Add a new option to the interval setting of the CartesianAxis.
This change is not breaking, does not change any existing behaviour, but extends the API.

I heavily reuse existing logic to identify which ticks can be shown, and then filter the resulting array.

With **'equidistantPreserveStart'**, ticks are positioned according to the following rule:

From the first tick onwards, show every Nth tick, with N being the smallest possible integer for which all visible ticks are non overlapping (relying on the isShow computation).

All existing rules are respected (i.e. min tick distance, etc.).

## Related Issue
https://github.com/recharts/recharts/issues/2562

## Types of changes

## Checklist:


nikolasrieble:
TODO: Split PR up into
- Test improvements
- New Feature

nikolasrieble:
All unrelated changes contained in this PR are to be found in a separate PR instead: 
https://github.com/recharts/recharts/pull/3393/files

ckifer:
Just to make reviewing easier on myself I'm going to prefer to wait to approve until the other PR is merged

src/cartesian/ticks/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+ * information whether they can be shown without overlapping with their neighbour isShow.
+ * @returns {CartesianTickItem[]} Every nTh tick in an array.
+ */
+export function getEveryNThTick(ticks: CartesianTickItem[]) {
+  let N = 1;
+  let previous: CartesianTickItem[] = ticks;
+  while (N < ticks.length) {
+    const everyNthTick = getEveryNth(ticks, N);
+    const visibleTicks = everyNthTick.reduce((previousValue, currentValue) => {
+      if (currentValue.isShow === true) {
```
ckifer, line 37: nit: same as if (currentValue.isShow) {

src/cartesian/ticks/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+    result.push(array[i]);
+  }
+  return result;
+}
+
+/**
+ * Given an array of ticks, find N, the lowest possible number for which every
+ * nTH tick in the ticks array isShow == true and return the array of every nTh tick.
+ * @param {CartesianTickItem[]} ticks An array of CartesianTickItem with the
+ * information whether they can be shown without overlapping with their neighbour isShow.
```
ckifer, line 28: nit nit: 
```suggestion
 * @param {CartesianTickItem[]} ticks An array of CartesianTickItem with 
 * information about whether they can be shown without overlapping with their neighbour.
```

src/cartesian/ticks/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @returns {CartesianTickItem[]} Every nTh tick in an array.
+ */
+export function getEveryNThTick(ticks: CartesianTickItem[]) {
```
ckifer, line 31: nit nit: personally think `getEveryNthTick` looks better

ckifer (commented):
Nice!

src/cartesian/ticks/utils.ts
petii: For me the name itself is a bit off. Having N in the function name suggests that it is a parameter. I do not have a good suggestion though.

src/cartesian/ticks/utils.ts
nikolasrieble: I do not like to rely on truthy and falsy, because they are partially surprising / inconsistent. 
0, for example is falsy, 1 is truthy. In this case only true is valid. 
Thus, I would always nit the the other way around :D

nikolasrieble:
Ready for final review. Now with much smaller diff.

src/cartesian/getTicks.ts
```diff
@@ -1,10 +1,47 @@
...
+ * Given an array of ticks, find N, the lowest possible number for which every
+ * nTH tick in the ticks array isShow == true and return the array of every nTh tick.
+ * @param {CartesianTickItem[]} ticks An array of CartesianTickItem with the
+ * information whether they can be shown without overlapping with their neighbour isShow.
+ * @returns {CartesianTickItem[]} Every nTh tick in an array.
+ */
+export function getEveryNThTick(ticks: CartesianTickItem[]) {
+  let N = 1;
+  let previous: CartesianTickItem[] = ticks;
+  while (N < ticks.length) {
```
petii, line 19: Isn't this loop simpler with something like `getEveryNthWithCondition`?

```typescript
let N = 1;
let previous = getEveryNthWithCondition(ticks, N, (tickItem) => tickItem.isShow));
while (previous.length === Math.ceil(ticks.length / N)) {
  N++;
  previous = getEveryNthWithCondition(ticks, N, (tickItem) => tickItem.isShow));
}
return previous;
```

nikolasrieble: It certainly is, good point.

src/cartesian/ticks/utils.ts
peterklingelhofer: Screen Shot 2023-02-23 at 6 28 54 AM

`isShow` is an optional `boolean`, so via TypeScript we know that it can only ever be `true`, `false`, or `undefined`, so as the value would never be `0`. If we were worried about `undefined` being conflated with `false` that would be one thing, but we're not in this case, as we're checking truthiness and not falsyness so @ckifer's suggestion seems pretty safe here.

nikolasrieble: @peterklingelhofer Fair enough, in this specific case that is true. Yet with using any, Typescript is muted, and my reservations against truthy and falsy are general. I prefer to avoid it because that preference has in edge cases prevented bugs. 

Yet this is not going to be the hill to die on for me, happy to change this here. :D

ckifer: No worries - it's personal preference. Agreed with your reservations though also agreed with the counter above. Not the hill anyone should die on. Being explicit is good

nikolasrieble:
This solution is a conservative estimate. 
We can compare the behaviour when we see what the interval set to an integer provides.

Although it would be possible to show every second tick (as achieved via interval=1), this solution only shows every 4th tick (as achieved via interval=3). 

That is because we reuse the isShow result, which only computes whether a tick would overlap with its direct neighbour.

From here on we can:
(1) Directly implement a better variant
(2) Merge this as adding value, and iterate

@ckifer Would you consider it a breaking change if we were to later improve the behaviour of this new interval setting? While it would change the behaviour, it would only change it to the better, thus i would argue it is fine improvement to make without a major version bump. 

interval=1
Screenshot 2023-02-23 at 16 33 07
interval=2
Screenshot 2023-02-23 at 16 33 12
interval=3
Screenshot 2023-02-23 at 16 33 17
interval="equidistantPreserveStart"
Screenshot 2023-02-23 at 16 33 21

ckifer:
How about we release the variant as experimental?

ckifer (approved):
LGTM 🚀 

Lets figure out how to indicate to people that this key is experimental and it's behavior is subject to slight modification

nikolasrieble:
> LGTM 🚀
> 
> Lets figure out how to indicate to people that this key is experimental and it's behavior is subject to slight modification

For now we could do that simply by marking it as such in the documentation on the website.

ckifer:
Sounds good - we'll want to make that update before the next release

frontier159:
@nikolasrieble @ckifer -- I think `equidistantPreserveStart` is missing from: https://github.com/recharts/recharts/blob/master/src/util/types.ts#L1113

Could you please confirm?

ckifer:
Looks like it is missing from that type. On the component though, it is defined correctly. 

We should correct this type and use it in CartesianAxis instead of defining things multiple places. Nice catch.

nikolasrieble:
> @nikolasrieble @ckifer -- I think `equidistantPreserveStart` is missing from: https://github.com/recharts/recharts/blob/master/src/util/types.ts#L1113
> 
> Could you please confirm?

Great catch. @frontier159 , Would you like to make a contribution and fix this as described by @ckifer above?

mmirus:
Very cool to see this first iteration. Thank you! :partying_face: 

I have a little feedback for future iterations that I'll share here. I'm happy to also create a new issue with the same information if that is preferable. Just let me know.

My expectation--which could be wrong!--would be that when using this interval the number of visible ticks on the axis should only increase as the width of the chart increases, at least given fairly uniform tick text widths.

In other words, if at width _w_ the chart shows 3 ticks, for every width greater than _w_ the number of ticks would be 3 or greater.

In practice, the new interval bounces back and forth between showing larger and much smaller numbers of ticks. Here's an example:

https://user-images.githubusercontent.com/524994/235784230-6c7a7772-43e0-4630-a524-4ed26c86dc47.mp4

Please let me know if I can provide any more information or otherwise help.

nikolasrieble:
> Very cool to see this first iteration. Thank you! 🥳
> 
> I have a little feedback for future iterations that I'll share here. I'm happy to also create a new issue with the same information if that is preferable. Just let me know.
> 
> My expectation--which could be wrong!--would be that when using this interval the number of visible ticks on the axis should only increase as the width of the chart increases, at least given fairly uniform tick text widths.
> 
> In other words, if at width _w_ the chart shows 3 ticks, for every width greater than _w_ the number of ticks would be 3 or greater.
> 
> In practice, as the width of the chart increases, the new interval setting bounces back and forth between showing larger and much smaller numbers of ticks. Here's an example:
> 
>  recharts.mp4 
> Please let me know if I can provide any more information or otherwise help.

Your expectation is correct! The current implementation is the simplest possible solution to enable equidistant ticks, but it is not the optimal solution for equidistant ticks. Your expectation would only be correct for the optimal solution.

mmirus:
Awesome! Looking forward to seeing future iterations, whenever they come. Thanks for the progress that's already been made, and all your work on this package.

PR #3393, opened 2023-02-22
Refactoring: Adapt getTicks to increase reusability of code

nikolasrieble:
## Description

This PR only refactors and does not change the behaviour. 

Refactoring:
- Move `CartesianTickItem` type into types file 
- Improve getNumberIntervalTicks by using getEveryNth instead of filtering. (This function will be reused in an upcoming PR).
- Change getTicks to return the tick in both getTicksStart and getTicksEnd before filtering the result. Doing so enables me to reuse the function for an upcoming PR. The filtering is moved to a single share place.
- Adapt tests to use a mock of getStringSize that estimates the string size based on the string length. That enables me to write better tests in the upcoming PR. 

The refactoring makes most sense when viewed in context of the next PR https://github.com/recharts/recharts/pull/3392. 

To keep the blast radius small of the changes, I yet decided to split it into a separate PR. 

## Types of changes

## Checklist:


petii:
I'm not sure I understand the reasoning for
> Adapt tests to use a mock of getStringSize that estimates the string size based on the string length. That enables me to write better tests in the upcoming PR.

This seems a bit strange from the perspective that "This PR only refactors and does not change the behaviour." but the tests change that are supposed to validate that behaviour is kept the same.

Also why do the tests need to be adapted and not new ones written in the upcoming PR?

nikolasrieble:
> This seems a bit strange from the perspective that "This PR only refactors and does not change the behaviour." but the tests change that are supposed to validate that behaviour is kept the same.



> I'm not sure I understand the reasoning for
> 
> > Adapt tests to use a mock of getStringSize that estimates the string size based on the string length. That enables me to write better tests in the upcoming PR.
> 
> This seems a bit strange from the perspective that "This PR only refactors and does not change the behaviour." but the tests change that are supposed to validate that behaviour is kept the same.
> 
> Also why do the tests need to be adapted and not new ones written in the upcoming PR?

Well, basically this change was required to be able to properly test equidistant ticks at all. Thus I changed it in the upcoming PR. 

Yet I wanted to strip the PR with the new feature from everything must not necessarily be part of it, to make it as small as possible and reviewing as easy as possible. 

If you mean to argue that I could also split this PR into two - refactoring - and changes to the test - then sure, I can.

src/cartesian/ticks/utils.ts
```diff
@@ -0,0 +1,26 @@
+import { CartesianTickItem } from '../../util/types';
+
+/**
+ * Given an array and a number N, return a new array which contains every nTh
+ * element of the input array. For n below 1, an empty array is returned.
+ * @param {any[]} array An input array.
+ * @param {integer} n A number
+ * @returns {any[]} The result array of the same type as the input array.
+ */
+function getEveryNth<Type>(array: Type[], n: number): Type[] {
```
petii, line 10: This could be on it's own. Maybe in something like `src/util/ArrayUtils.ts`?

src/cartesian/ticks/getTicks.ts
```diff
@@ -163,8 +154,10 @@ export function getTicks(props: CartesianAxisProps, fontSize?: string, letterSpa
     return getNumberIntervalTicks(ticks, typeof interval === 'number' && isNumber(interval) ? interval : 0);
   }
 
-  if (interval === 'preserveStartEnd') {
-    return getTicksStart(
+  let candidates: CartesianTickItem[] = [];
+
+  if (interval === 'preserveStart' || interval === 'preserveStartEnd') {
```
petii, line 159: My personal preference is for the early returns as in before the changes, but since you moved the _isShow_ filtering here it makes sense to try and reduce code duplication.

However it could also make sense that to not stop here but create and use an additional layer like
```typescript
function getShownTicks(...) {
  return getTicks(...).filter(entry => entry.isShow);
}
```

But then you have to go and change every usage of getTicks to keep existing behaviour (and it probably does not work too well with the following PR?).

Alternatively, if we would want to go one step further in reducing code duplication we could introduce a holder for these like
```typescript
const tickGeneratorForInterval: Record<
  CartesianAxisProps['interval'],
  (args: Omit<CartesianAxisProps, 'tickMargin'>) => CartesianTickItem[]
> = { 
  'preserveEnd': getTicksEnd, 
  'preserveStart': getTicksStart, 
  'preserveStartEnd': (args) => getTicksStart(args, true),
}
```
and then just
```typescript
  const candidates = tickGeneratorForInterval[interval]({ ... });
  return candidates.filter(entry => entry.isShow);
```
to ~~get rid of~~ hide the branching altogether

petii:
> If you mean to argue that I could also split this PR into two - refactoring - and changes to the test - then sure, I can.

I'm not entirely sure about refactoring best practices whether it makes sense to refactor and the change the tests or change the tests and then refactor, however my question is more directed towards setting precedence I guess.

src/cartesian/ticks/utils.ts
```diff
@@ -0,0 +1,26 @@
+import { CartesianTickItem } from '../../util/types';
+
+/**
+ * Given an array and a number N, return a new array which contains every nTh
+ * element of the input array. For n below 1, an empty array is returned.
+ * @param {any[]} array An input array.
```
ckifer, line 6: nit: comment says `any[]` but it has a generic type (even if that type can be anything, much better than `any`). Same two lines down.

src/cartesian/ticks/utils.ts
ckifer: Yeah this isn't 'tick' specific per se.

Also if we're going to move to different naming conventions for `utils` and for our files in general lets discuss.

We'll end up having files in `util/SomeUtilFile.ts` as well as places like `cartesian/ticks/utils`. Which pattern do we follow? imo we should stick to something consistent for ease of contribution

test/cartesian/ticks/getTicks.spec.ts
```diff
@@ -27,9 +28,9 @@ const EXAMPLE_INPUT = {
   y: 100,
 };
 
-jest.mock('../../src/util/DOMUtils', () => ({
+jest.mock('../../../src/util/DOMUtils', () => ({
   // We mock string size measurement, because getStringSize else returns 0 in these tests.
-  getStringSize: jest.fn(() => ({ width: 20, height: 20 })),
+  getStringSize: jest.fn((text: string) => ({ width: text.length, height: 20 })),
```
ckifer, line 33: I don't think this adds any indication of actual text length. Better maybe than no variation, but certainly not accurate.

ckifer (approved):
few nits, but overall looks good.

Curious about util placement moving forwards

test/cartesian/ticks/getTicks.spec.ts
nikolasrieble: Agree, but this enables me to test with larger labels on different positions (i.e. first tick, second tick, third tick, etc.)

src/cartesian/ticks/utils.ts
nikolasrieble: I find that things should be as close to their usage as possible. If used all across the places, a top level utils folder makes sense. But if used only within CartesianAxis, I would argue that is where it should live.

src/cartesian/ticks/getTicks.ts
nikolasrieble: I like the ideas, and will most probably come back to them - but only when following up with this file after landing the feature enhancement. For now I want to only refactor as much as I need to, in order to nicely implement the equidistant ticks.

src/cartesian/ticks/utils.ts
petii: _Right now_ it's only used in CartesianAxis. But when the need for the functionality comes along outside of `tick` or `cartesian` it's much easier to miss it if it's hidden a few subfolders deep. (Which then leads to just duplicating some similar logic ad-hoc where and when it's needed)

src/cartesian/ticks/utils.ts
petii: There might even be support for generic types in jsdoc. https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template

src/cartesian/ticks/utils.ts
nikolasrieble: Fair enough. For consistency I will move it back to a shared utils.

src/cartesian/ticks/getTicks.ts
petii: Looking at the following PR this also has the benefit that adding the new option is just
```typescript
  'equidistantPreserveStart': (args) => getEveryNthTick(getTicksStart(args)),
```
instead of adding back the duplicated code that this PR removed. With that context I would argue even more so that the conditions should still be separate and not share the branch for 'preserveStart' and 'preserveStartEnd'.

Issue #3406, opened 2023-02-27
Storybook: Migrate all documentation from the website into MDX files. 


nikolasrieble:
Following the best practice as established by https://reaviz.io/?path=/story/docs-intro--page

Acceptance criteria: 
- Create a landing page / welcome page 
- Create an mdx file for the BLOG page from the website
- Create an mdx file for the GUIDE page from the website

ckifer:
@nikolasrieble this is good to close yeah?

nikolasrieble:
I believe so.

PR #3417, opened 2023-03-01
Storybook: Create Welcome page

nikolasrieble:
## Description
- Create a Welcome page
- Add Blog page
- Migrate all pages from websites GUIDE section 
Screenshot 2023-03-02 at 21 59 14

## Related Issue
https://github.com/recharts/recharts/issues/3406

Screenshot 2023-03-02 at 22 09 15

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/Getting started.mdx
+++ b/storybook/stories/Getting started.mdx
@@ -1,0 +1,129 @@
+import { pageData as data } from './data/Page';
+import { LineChart, Line, CartesianGrid, XAxis, YAxis, Tooltip } from '../../src';
+
+# Getting started
+
+1. Choose the type of chart you need
+With the help of babel-plugin-recharts, we can import components we actually need, making the project smaller than otherwise.
+Generate a simple chart by using plain javascript data (usually an array of objects).
+
+```tsx
+const data = [{name: 'Page A', uv: 400, pv: 2400, amt: 2400}, ...];
+```
+
+```tsx
+import { LineChart, Line } from 'recharts';
+const renderLineChart = (
+<LineChart width={400} height={400} data={data}>
+  <Line type="monotone" dataKey="uv" stroke="#8884d8" />
+</LineChart>
+);
+```
+
+<LineChart width={400} height={400} data={data}>
+  <Line type="monotone" dataKey="uv" stroke="#8884d8" />
+</LineChart>
+
+2. Add components that you want to be drawn
+LineChart can have XAxis, YAxis, Legend, CartesianGrid and so on.
+
+```tsx
+import { LineChart, Line, CartesianGrid, XAxis, YAxis } from 'recharts';
+const renderLineChart = (
+<LineChart width={600} height={300} data={data}>
+  <Line type="monotone" dataKey="uv" stroke="#8884d8" />
+  <CartesianGrid stroke="#ccc" />
+  <XAxis dataKey="name" />
+  <YAxis />
+</LineChart>
+);
+```
+
+<LineChart width={600} height={300} data={data}>
+  <Line type="monotone" dataKey="uv" stroke="#8884d8" />
+  <CartesianGrid stroke="#ccc" />
+  <XAxis dataKey="name" />
+  <YAxis />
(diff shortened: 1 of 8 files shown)


storybook/stories/Getting started.mdx
```diff
@@ -0,0 +1,129 @@
...
+<LineChart width={600} height={300} data={data} margin={{ top: 5, right: 20, bottom: 5, left: 0 }}>
+  <Line type="monotone" dataKey="uv" stroke="#8884d8" />
+  <CartesianGrid stroke="#ccc" strokeDasharray="5 5" />
+  <XAxis dataKey="name" tick={<CustomizedAxisTick />} />
+  <YAxis />
+  <Tooltip />
+</LineChart>
+);
+```
+
```
nikolasrieble, line 126: todo: Make this example show up.

nikolasrieble:
@proke03 With this, we might not even need to migrate the website into the repository at all, since storybook could simply replace the website completely. 

What do you think?

nikolasrieble:
While doing this, i realised that the code and the pictures on the https://recharts.org/en-US/guide/customize page never were in sync 👀 

Do we even want to keep them @ckifer ?

ckifer:
lol 😅 - I think as long as we structure a few of out examples like this (i.e. can see the code, can see the result, can see a small blurb about what this example accomplishes). I think we are fine to remove this section

nikolasrieble:
For now I will suggest to merge this, merely as an example of how to do docs, use code.
It also contains the Welcome page. 

Medium term this should go 😅

ckifer (approved):

ckifer:
Yeah this all looks fine - probably a lot of things that don't need to be here but for the sake of copying existing docs

proke03:
@nikolasrieble
If charts.org continues to play its role now, it can replace that completely.
However, I think the role of the site should be expanded. (forum, showcases)

storybook/stories/Welcome.mdx
```diff
@@ -0,0 +1,44 @@
...
+  <CartesianGrid stroke="#eee" strokeDasharray="5 5" />
+  <XAxis dataKey="name" />
+  <YAxis />
+  <Line type="monotone" dataKey="uv" stroke="#8884d8" />
+  <Line type="monotone" dataKey="pv" stroke="#82ca9d" />
+  <Tooltip />
+</LineChart>
+
+```typescript jsx
+  <LineChart width={500} height={300} data={data}>
```
Yilun-Sun, line 24: I think we could indent one space to the left to make it has a better look 
image

Yilun-Sun (approved):

nikolasrieble:
> @nikolasrieble If charts.org continues to play its role now, it can replace that completely. However, I think the role of the site should be expanded. (showcases, forum)

Could you please expand on what kind of cases you think could not live in storybook? A forum is a good example but I would doubt the value of it. After all we build a charting library, our users are engineers, and Github Discussions exist. 
If we focus on the core of Recharts, that will always be API documentation and actual feature work (and bug fixing 😅 )

nikolasrieble:
> @nikolasrieble If charts.org continues to play its role now, it can replace that completely. However, I think the role of the site should be expanded. (showcases, forum)

Also, removing the website would massively reduce complexity, i.e. one thing less to deploy, less code to maintain, etc.

nikolasrieble:
> Agreed, don't think we need a forum.
> 
> I don't think storybook results in documentation that I love 1000%, but it does result in readable and usable docs as long as we structure things correctly. 1 less thing to maintain (especially when that thing is old) is always beneficial to those maintaining.
> 
> My few concerns about totally removing the current website lie in 1. Branding - do our consumers care? And 2. Web domain - will we be able to get the storybook hosted as recharts.org?

Great question! 
re 1 - Branding wise I find storybook much more targeted to our community - developers. Although the welcome page should be improved design wise. 
re 2 - We most certainly can host it there.

arcthur:
I agree that storybook is great document tool. but I want keep the brand, and use same style of website

PR #3421, opened 2023-03-03
feat: Expose verticalAnchor on XAxis (and CartesianAxis)

petii:
## Description
This PR adds the option to manually override the _verticalAnchor_ prop in order to be able to define the rotation anchor without using a custom tick component.

## Related Issue
https://github.com/recharts/recharts/issues/3469

## Motivation and Context
The _XAxis_ component has the _angle_ prop, which lets the user add a rotation to the tick labels. The horizontal anchor point for this rotation can be set through the _textAnchor_ prop, but the vertical anchor is calculated in _CartesianAxis_ internally.
```typescript
  getTickVerticalAnchor() {
    const { orientation, mirror } = this.props;
    let verticalAnchor = 'end';

    switch (orientation) {
      case 'left':
      case 'right':
        verticalAnchor = 'middle';
        break;
      case 'top':
        verticalAnchor = mirror ? 'start' : 'end';
        break;
      default:
        verticalAnchor = mirror ? 'end' : 'start';
        break;
    }

    return verticalAnchor;
  }
```
Thus for the X axis, the vertical anchor can only be `start` or `end` which can result in some visual discrepancies:
code | result
--|--
`<XAxis dataKey="name" angle={90} textAnchor="start" />` | Screenshot 2023-03-21 at 11 20 03

code | result
--|--
`<XAxis dataKey="name" angle={90} textAnchor="start" verticalAnchor="middle" />` |  Screenshot 2023-03-21 at 11 22 27

https://user-images.githubusercontent.com/2715418/226579948-9270987d-6358-42db-95c8-cb23fed8f056.mov

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/CartesianAxis.tsx
+++ b/src/cartesian/CartesianAxis.tsx
@@ -44,6 +44,7 @@ export interface CartesianAxisProps {
   ticksGenerator?: (props?: CartesianAxisProps) => CartesianTickItem[];
   interval?: number | 'preserveStart' | 'preserveEnd' | 'preserveStartEnd' | 'equidistantPreserveStart';
   angle?: number;
+  verticalAnchor?: 'start' | 'middle' | 'end';
 }
 
 interface IState {
@@ -177,7 +178,10 @@ export class CartesianAxis extends Component<Props, IState> {
   }
 
   getTickVerticalAnchor() {
-    const { orientation, mirror } = this.props;
+    const { orientation, mirror, verticalAnchor: verticalAnchorProp } = this.props;
+    if (verticalAnchorProp) {
+      return verticalAnchorProp;
+    }
     let verticalAnchor = 'end';
 
     switch (orientation) {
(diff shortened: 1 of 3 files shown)


src/cartesian/CartesianAxis.tsx
```diff
@@ -43,6 +43,8 @@ export interface CartesianAxisProps {
   tickFormatter?: (value: any, index: number) => string;
   ticksGenerator?: (props?: CartesianAxisProps) => CartesianTickItem[];
   interval?: number | 'preserveStart' | 'preserveEnd' | 'preserveStartEnd' | 'equidistantPreserveStart';
+  angle?: number;
```
nikolasrieble, line 46: Why add the angle here? Seems like an unrelated change.

storybook/stories/API/cartesian/CartesianAxis.stories.tsx
```diff
@@ -77,6 +77,78 @@ export const TickPositioning = {
   },
 };
 
+export const TickAlignment = {
```
nikolasrieble, line 80: Nit: Given that this PR adds a new property, it would be great to also add a description to the Argument Table - as done in the Line.stories.tsx
Not required, but would be very well appreciated :)

src/cartesian/CartesianAxis.tsx
petii: I was going off of the property being present in the documentation (https://recharts.org/en-US/api/XAxis#angle) but now that I looked at it again this is only the case for the XAxis component.

Also I do wonder whether it should be 🤔 Because it might be a bit misleading to expose properties that are limited to affecting the ticklabels on the main axis components.

ckifer:
would love to get this merged at some point if we can

ckifer:
Would love to get this change in but this PR is stale - please re-open if we want to continue here

Issue #3382, opened 2023-02-20
Storybook: Create stories for the Bar component


nikolasrieble:
For the Line component we have a story that does not even use a chart, but shows the Line all alone within a Surface.

Lets follow the same pattern for the Bar

Acceptance criteria: 
- storybook/stories/API/cartesian/Bar.stories.tsx

Create stories for:
- Simple
- Stacked/Grouped Bars (exposing stackID as controls for two bar groups)
- Bar with Labels
- Bar with Background
- Bar with custom shape
- Bar with animation

Each story should only expose the minimal controls relevant to the story

Yilun-Sun:
Closes #3435

PR #3435, opened 2023-03-04
Storybook: add Bar stories

Yilun-Sun:
## Related Issue

#3382

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/API/cartesian/Bar.stories.tsx
+++ b/storybook/stories/API/cartesian/Bar.stories.tsx
@@ -1,0 +1,333 @@
+import React, { useState } from 'react';
+import { ComposedChart, Bar, ResponsiveContainer, Cell } from '../../../../src';
+import { coordinateWithValueData, pageData } from '../../data';
+
+export default {
+  component: Bar,
+  argTypes: {
+    background: {
+      control: {
+        type: 'boolean',
+      },
+    },
+    stroke: {
+      control: { type: 'color' },
+    },
+    fill: {
+      control: { type: 'color' },
+    },
+    animationEasing: {
+      options: ['ease', 'ease-in', 'ease-out', 'ease-in-out', 'linear'],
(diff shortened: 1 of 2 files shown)


nikolasrieble (approved):
Love the stories, especially the one with the custom event. 
Two things that would be great to add, but could also be added in another PR: 
1 - A doc file with an ArgsTable
2 - Using the exact same dataset in all stories

But this is already a great addition either way.

storybook/stories/API/cartesian/Bar.stories.tsx
```diff
@@ -0,0 +1,333 @@
...
+          <Bar dataKey="y" isAnimationActive={false} {...areaArgs} />
+        </ComposedChart>
+      </ResponsiveContainer>
+    );
+  },
+  args: {
+    data: coordinateWithValueData,
+    dataKey: 'y',
+  },
+  parameters: { controls: { include: ['data'] } },
```
nikolasrieble, line 57: Thinking about this again, I suggest to take out `data` from the controls, because really, data is a property of the ComposedChart. 

Instead I suggest to hardcode a single dataset across all stories. Preferably pageData as used in most examples.

storybook/stories/API/cartesian/Bar.stories.tsx
```diff
@@ -0,0 +1,333 @@
...
+          margin={{
+            top: 20,
+            right: 20,
+            bottom: 20,
+            left: 20,
+          }}
+          data={data}
+        >
+          {defs}
+          <Bar dataKey="y" isAnimationActive={false} {...areaArgs} />
```
nikolasrieble, line 48: You could consider adding other components here that interact with the Bar or rather - are influenced by the Bar props. 

That might possibly be the Tooltip, the Legend, and the XAxis and YAxis

Issue #3447, opened 2023-03-10
Moving Brush when XAxis has ticks and interval={0} set leads to NaN errors


jackcannon:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/competent-forest-k11ts4?file=/src/App.tsx)

### Steps to reproduce
Move the brush traveller

### What is expected?
No javascript errors

### What is actually happening?
The following javascript errors:
* Warning: Received NaN for the `x1` attribute. If this is expected, cast the value to a string.
* Warning: Received NaN for the `x2` attribute. If this is expected, cast the value to a string.

| Environment | Info |
|---|---|
| Recharts | v2.4.3 |
| React | react 18.2.0 |
| System | Mac OS |
| Browser | Chrome |

---
This is only present when all three of these are present:
* `ticks` (on `XAxis`)
* `interval={0}` (on `XAxis`)
* `Brush`.

If any of the three are omitted, there are no errors, however, it would very useful to be able to keep all manual ticks visible when using a Brush.
There is initially no problem until you try to move the Brush travellers.

ckifer:
Thanks for the repro, will look into it

nikolasrieble:
Investigation so far: 

The CartesianAxis produces the finalTicks. 

```
    console.log('before', finalTicks);

    if (_.isFunction(ticksGenerator)) {
      finalTicks = ticks && ticks.length > 0 ? ticksGenerator(this.props) : ticksGenerator(noTicksProps);
    }

    console.log('after', finalTicks);
```

In this specific case, the output of final ticks is an object 

```
{
    coordinate: Nan,
    values: [...], 
    offset: 0,
}
```

The log message is caused by computing x values for rendering using the coordinate of Nan. 
This indicates that the error originates from the ticksGenerator.

ckifer:
fixed by https://github.com/recharts/recharts/pull/3454

Will be released in next minor version

ckifer:
Released in 2.5!

jackcannon:
Works great! Thanks both!

PR #3454, opened 2023-03-13
ChartUtils: GetTicks only returns valid ticks that are part of the current domain.

nikolasrieble:
## Description
Description of the issue is found here. https://github.com/recharts/recharts/issues/3447#issuecomment-1465206498

The solution is simple: 
Filter all ticks that have NaN as a coordinate. 

This PR additionally adds unit tests and a few comments.

## Related Issue
https://github.com/recharts/recharts/issues/3447

## Types of changes

## Checklist:


src/cartesian/CartesianAxis.tsx
```diff
@@ -315,10 +315,14 @@ export class CartesianAxis extends Component<Props, IState> {
     const { ticks, ...noTicksProps } = this.props;
     let finalTicks = ticks;
 
+    console.log('before', finalTicks);
+
```
nikolasrieble, line 319:
```suggestion
```

src/cartesian/CartesianAxis.tsx
```diff
@@ -315,10 +315,14 @@ export class CartesianAxis extends Component<Props, IState> {
...
     if (_.isFunction(ticksGenerator)) {
       finalTicks = ticks && ticks.length > 0 ? ticksGenerator(this.props) : ticksGenerator(noTicksProps);
     }
 
+    console.log('after', finalTicks);
+
```
nikolasrieble, line 325:
```suggestion
```

src/util/ChartUtils.ts
```diff
@@ -573,27 +573,44 @@ export const getCoordinatesOfGrid = (ticks: Array<TickItem>, min: number, max: n
...
+    realScaleType?: 'scaleBand' | 'band' | 'point' | 'linear';
+    scale?: any;
+    axisType?: AxisType;
+    ticks?: any;
+    niceTicks?: any;
+    isCategorical?: boolean;
+    categoricalDomain?: any;
+  },
+  isGrid?: boolean,
+  isAll?: boolean,
```
nikolasrieble, line 588: While I do not like `any` at all, this is a marginal improvement over the current type.

src/util/ChartUtils.ts
```diff
@@ -573,27 +573,44 @@ export const getCoordinatesOfGrid = (ticks: Array<TickItem>, min: number, max: n
...
+): TickItem[] | null => {
   if (!axis) return null;
   const { scale } = axis;
   const { duplicateDomain, type, range } = axis;
 
   const offsetForBand = axis.realScaleType === 'scaleBand' ? scale.bandwidth() / 2 : 2;
   let offset = (isGrid || isAll) && type === 'category' && scale.bandwidth ? scale.bandwidth() / offsetForBand : 0;
 
-  offset = axis.axisType === 'angleAxis' ? mathSign(range[0] - range[1]) * 2 * offset : offset;
+  offset = axis.axisType === 'angleAxis' && range?.length >= 2 ? mathSign(range[0] - range[1]) * 2 * offset : offset;
```
nikolasrieble, line 597: Fixing an unrelated bug for when the range does not contain range[0], and range[1]

src/util/ChartUtils.ts
```diff
@@ -573,27 +573,44 @@ export const getCoordinatesOfGrid = (ticks: Array<TickItem>, min: number, max: n
...
       return {
+        // If the scaleContent is not a number, the coordinate will be NaN.
+        // That could be the case for example with a PointScale and a string as domain.
         coordinate: scale(scaleContent) + offset,
         value: entry,
         offset,
       };
     });
+
+    return result.filter((row: TickItem) => !_.isNaN(row.coordinate));
```
nikolasrieble, line 613: Finally the fix.

src/util/ChartUtils.ts
ckifer: totally

src/util/ChartUtils.ts
ckifer: why `>=` and not just `=== 2`?

src/util/ChartUtils.ts
ckifer: So there is no way to fix this upstream just because of the way that `PointScale` works?

ckifer (approved):
Niceeee!

src/util/ChartUtils.ts
nikolasrieble: Upstream nothing is broken. This is about the domain being filtered by the brush - as it should - and the PointScale simply has no coordinates for items outside of the domain. 

That is how the point scale works.

nikolasrieble: To my best understanding, all works as it should - but this function.

src/util/ChartUtils.ts
nikolasrieble: Fair enough, ===2 would also work, but it also is stricter, thus I am not certain it might break something.

nikolasrieble: >= on the other only prevents a bug without restricting beyond the requirement.

ckifer: yeah I was curious if there ever was going to be more than 2. Hopefully not but agreed

src/util/ChartUtils.ts
ckifer: Got it. Logically in my head I think I'd want the data itself to be filtered and then everything trickle down into domain and ticks from there. Makes sense, but these are the kind of fragile pieces we have to keep an eye on and harden/test

ckifer: Also do we want to use `_.isNaN(row.coordinate)` as opposed to `Number.isNaN(row.coordinate)`? The more we use lodash the more we have dependency on it (I know its used a ton but just future thinking)

PR #3463, opened 2023-03-18
refactor: `Text` to component function

akamfoad:
## Description

- add: react-hooks eslint plugin:
  React hooks eslint plugin was installed but wasn't used. From now on, it'll be extended from in the eslint config, so migrating components to funtional components and usage of hook will be smoother as it can detect misuse of hooks.
- refactor: `Text` to function component
- fix: missing hook dependency in `ResponsiveContainer`

## Motivation and Context

## Types of changes

## Checklist:


src/component/ResponsiveContainer.tsx
```diff
@@ -90,7 +90,7 @@ export const ResponsiveContainer = forwardRef(
           return currentSizes;
         });
       }
-    }, [getContainerSize]);
+    }, [getContainerSize, onResize]);
```
nikolasrieble, line 93: Nice, this might as well be an unrelated bug fix! ❤️

src/component/Text.tsx
```diff
@@ -183,109 +183,113 @@ const getWordsByLines = (props: Props, needCalculate: boolean) => {
...
+};
 
-  state: State = {};
+export const Text = (props: Props) => {
+  const [state, setState] = useState<State>({
+    prevWidth: props.width,
+    prevScaleToFit: props.scaleToFit,
+    prevChildren: props.children,
+    prevStyle: props.style,
+    wordsByLines: getWordsByLines(props, true),
```
nikolasrieble, line 203: Instead of building on state object, you could simply have separate state variables.

nikolasrieble: It would also be nice to avoid recreating the initial state
https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state

src/component/Text.tsx
```diff
@@ -183,109 +183,113 @@ const getWordsByLines = (props: Props, needCalculate: boolean) => {
...
-      };
+        props.children !== state.prevChildren ||
+        props.style !== state.prevStyle ||
+        props.breakAll !== state.prevBreakAll;
+
+      setState({
+        prevWidth: props.width,
+        prevScaleToFit: props.scaleToFit,
+        prevChildren: props.children,
+        prevStyle: props.style,
```
nikolasrieble, line 223: I am not sure I understand why these are in the state at all. 🤔 
Would it not be sufficient to only have them as props?

nikolasrieble: I am raising this as a question, not as a thing to fix. Your PR already adds value and seems good to go.

src/component/ResponsiveContainer.tsx
nikolasrieble: Could you please do this change, and the addition of `"plugin:react-hooks/recommended",` as a separate PR? If issues ever come up, we might revert the changes to the Text, and i would love for these two changes to remain independent.

nikolasrieble (commented):
Great work, thank you for your contribution! Would love to hear @ckifer opinion as well, as he recently also worked on refactoring to functional components.

src/component/Text.tsx
akamfoad: Now I think we don't need the state at all, its just a trigger for recalculation, I tried to keep the code as is mostly, but now I think a useMemo would be sufficient, it calculates on first render, and when the props changed it'll recalculate the wordsByLine. not need to state+effect.

src/component/Text.tsx
akamfoad: I dig a bit and all those was for computing whether we should do a recalculation or not, I simplified the code and removed those states, a useMemo would be sufficient IMO.

src/component/Text.tsx
nikolasrieble: 💪

src/component/Text.tsx
```diff
@@ -160,132 +151,106 @@ const getWordsWithoutCalculate = (children: React.ReactNode): Array<Words> => {
...
+  scaleToFit: false,
+  textAnchor: 'start',
+  verticalAnchor: 'end', // Maintain compat with existing charts / default SVG behavior
+  fill: '#808080',
+};
 
+export const Text = (props: Props) => {
+  const wordsByLines: Array<Words> = useMemo(() => {
+    return getWordsByLines(props);
+  }, [props.width, props.scaleToFit, props.children, props.style, props.breakAll]);
```
ckifer, line 189: can we disable the exhaustive dependencies hook here with a justification if it doesn't apply?

src/component/Text.tsx
```diff
@@ -160,132 +151,106 @@ const getWordsWithoutCalculate = (children: React.ReactNode): Array<Words> => {
   return [{ words }];
 };
 
-const getWordsByLines = (props: Props, needCalculate: boolean) => {
+const getWordsByLines = (props: Props) => {
   // Only perform calculations if using features that require them (multiline, scaleToFit)
   if ((props.width || props.scaleToFit) && !Global.isSsr) {
     let wordsWithComputedWidth: Array<WordWithComputedWidth>, spaceWidth: number;
 
-    if (needCalculate) {
```
ckifer, line 168: does the `useMemo` account for `needCalculate`? Looks like it does, just confirming

ckifer (changes requested):
LGTM, but can we disable the exhaustive hooks lint rule with a justification if it doesn't apply? It gets noisy if it is there without the explicit disable

src/component/Text.tsx
akamfoad: I think it does, previously the diffing was done by keeping prev props in state and in the subsequent re-renders, the code was comparing new props to older ones to see the need of recalculating, but now that whole logic is done by React itself.

I also refactored the utility function types so each function takes only desired props, which makes it easy to narrow down the props that is necessary to trigger the recalculation.

src/component/Text.tsx
akamfoad: Btw for the disabling, I do not recommend it, in my workspace I've seen some nasty hidden bugs related to stale closures, I think its essential to have the rule enabled.

ckifer: Agreed, but at times there are valid use-cases for disabling the rule. Wanted the justification for the warning if it was warranted, the refactoring solves this. Thanks!

src/component/Text.tsx
```diff
@@ -160,132 +156,121 @@ const getWordsWithoutCalculate = (children: React.ReactNode): Array<Words> => {
...
-      } else {
-        return getWordsWithoutCalculate(props.children);
-      }
-
-      return calculateWordsByLines(props, wordsWithComputedWidth, spaceWidth, props.width, props.scaleToFit);
+      wordsWithComputedWidth = wcw;
+      spaceWidth = sw;
+    } else {
+      return getWordsWithoutCalculate(children);
     }
```
ckifer, line 175: out of scope for this PR, but we could invert this and not need to set 
```
 wordsWithComputedWidth = wcw;
spaceWidth = sw;
```

ckifer (approved):
LGTM, thank you!

ckifer:
@nikolasrieble @akamfoad Lets make sure to test these merged refactors in storybook and otherwise to try and uncover any bugs before next release

ckifer:
Looks like this broke some CI - @akamfoad do you mind taking a look?
https://github.com/recharts/recharts/pull/3486/files#annotation_10016950433

akamfoad:
Sorry I've been very busy this last week, looks like #3487 might have fixed the issue, can we update #3486 I had that issue in my one of my PRs as well I think but it went away, I had no idea.

ckifer:
Yeah no worries, just want to make sure everything is green before merges, not sure how that got through CI. Thanks!

PR #3466, opened 2023-03-21
Eslint react hooks plugin

akamfoad:
## Description

React hooks eslint plugin was installed but wasn't used. From now on, it'll be extended from in the eslint config, so migrating components to funtional components and usage of hook will be smoother as it can detect misuse of hooks.

## Motivation and Context

React hooks has some rules, when followed they are great utilities to re-use code between function components and they're very powerful. But when these ruled are ignored, they can cause regression, unnecessary computation, or give unexpected results. `eslint-plugin-react-hooks` is the eslint plugin to enforce these rules and make sure the code is correct.

## Types of changes

## Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,7 @@
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:react/recommended",
+    "plugin:react-hooks/recommended",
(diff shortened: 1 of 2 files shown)


nikolasrieble (approved):
Thank you!

PR #3471, opened 2023-03-21
refactor: `PolarGrid` to function component

akamfoad:
### rm: JSDoc eslint rule in favor of typescript

JSDoc is great for documenting code in JS, but since the library built with typescript natively, we shall use JSDoc only for documenting purposes. For everything related to typing is better to utilize TypeScript to keep the library source code consistent and correct. The ESLint rule for valid JSDoc comments have been removed since it requires all comments to have some properties that would be the duplicate of the TS types and adds maintenance overhead with no value. Also this rule [has been deprecated 5y ago](https://eslint.org/blog/2018/11/jsdoc-end-of-life/)

### refactor: `PolarGrid` to function component

Converted the PolarGrid component to functional component and converted the static functions to in-component functions, then refactored them out of the component itself. After that converted all outside functions which rendered JSX to be React components instead.

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/PolarGrid.tsx
+++ b/src/polar/PolarGrid.tsx
@@ -18,7 +18,12 @@ interface PolarGridProps {
 }
 export type Props = SVGProps<SVGPathElement> & PolarGridProps;
 
-type ConcentricProps = Props & { radius: number; index: number };
+type ConcentricProps = Props & {
+  // The radius of circle
+  radius: number;
+  // The index of circle
+  index: number;
+};
 
 const polarGridDefaultProps = {
   cx: 0,
@@ -46,10 +51,7 @@ const getPolygonPath = (radius: number, cx: number, cy: number, polarAngles: num
   return path;
 };
 
-/**
- * Draw axis of radial line
- * @return The lines
- */
+// Draw axis of radial line
 const PolarAngles: React.FC<Props> = props => {
   const { cx, cy, innerRadius, outerRadius, polarAngles, radialLines } = props;
 
@@ -82,12 +84,7 @@ const PolarAngles: React.FC<Props> = props => {
   );
 };
 
-/**
- * Draw concentric circles
- * @param radius The radius of circle
- * @param index  The index of circle
- * @return circle
- */
+// Draw concentric circles
 const ConcentricCircle: React.FC<ConcentricProps> = props => {
   const { cx, cy, radius, index } = props;
   const concentricCircleProps = {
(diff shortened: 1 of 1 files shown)


ckifer:
Looks like the build id failing

ckifer (changes requested):
The rest looks good to me, but build is failing and lets keep the ESLint rule for now. Agreed with you, but we can modify in a different PR

ckifer:
I am hesitant to merge this one due to it existing in this map https://github.com/recharts/recharts/blob/e7858c926b01e05f846168eed093821a89aa2c68/src/chart/generateCategoricalChart.tsx#L2140-L2161

This means that the static property `displayName` is used by the `generateCategoricalChart` function. That means more potential for bugs (imo) in the components in that map as compared to the others. Just a callout

akamfoad:
@ckifer I understand, tbh I tried to refactor some other components but I hesitated too since it or some of it's static members were used in this function to generate the charts. I think we need to refactor that somehow, I might try again once more, will keep this map in mind for the next refactor.

ckifer:
Totally agreed, the more I understand how it's working the more I consider it an anti-pattern. 

Recharts in general needs an architectural deep dive and maybe some flow diagrams so we can figure out a decently replacement for the static member issue among others.

nikolasrieble:
Please separate this into two PRs, one for the refactoring and one for the eslint rule. Both are separate concerns. Removing the eslint rule still seems a reasonable improvement.

akamfoad:
@nikolasrieble I removed the ESLint change at the time, this PR only changes PolarGrid

src/polar/PolarGrid.tsx
```diff
@@ -18,159 +18,146 @@ interface PolarGridProps {
...
-      <g className="recharts-polar-grid-concentric">
-        {polarRadius.map((entry: number, i: number) =>
-          gridType === 'circle' ? this.renderConcentricCircle(entry, i) : this.renderConcentricPolygon(entry, i),
-        )}
-      </g>
-    );
+  return (
+    <g className="recharts-polar-grid-concentric">
+      {polarRadius.map((entry: number, i: number) => {
+        const key = i;
```
ckifer, line 139: nit: why assign a variable to `i` which is already defined?

ckifer (approved):
Overall looks good. Can we do the same kind of comparison as with `Curve` before merge?

src/polar/PolarGrid.tsx
akamfoad: Actually just to trick ESLint regarding the `no-array-index-key` rule, that otherwise shows a warning.

akamfoad:
Of course @ckifer here it is:

```
recharts on refactor/polar-grid-component ➜ pixelmatch .\master-polar-grid.png .\refactored-polar-grid.png diff.png 0.1

matched in: 14.547ms
different pixels: 0
error: 0%
```

### On `master`
master-polar-grid



### On `akamfoad:refactor/polar-grid-component`
refactored-polar-grid


### `Diff`
diff

ckifer:
Wonderful, thank you!

PR #3475, opened 2023-03-22
refactor: `Cross` to component function

akamfoad:
## Description

Refactor `Cross` component from class to function component.

## Motivation and Context

## Types of changes

## Checklist

### DIFF
--- a/src/shape/Cross.tsx
+++ b/src/shape/Cross.tsx
@@ -18,15 +18,6 @@ interface CrossProps {
 
 export type Props = SVGProps<SVGPathElement> & CrossProps;
 
-const crossDefaultProps = {
-  x: 0,
-  y: 0,
-  top: 0,
-  left: 0,
-  width: 0,
-  height: 0,
-};
-
 const getPath = (x: number, y: number, width: number, height: number, top: number, left: number) => {
   return `M${x},${top}v${height}M${left},${y}h${width}`;
 };
@@ -47,4 +38,11 @@ export const Cross: React.FC<Props> = props => {
   );
 };
 
-Cross.defaultProps = crossDefaultProps;
+Cross.defaultProps = {
+  x: 0,
+  y: 0,
+  top: 0,
+  left: 0,
+  width: 0,
+  height: 0,
+};
--- a/src/shape/Cross.tsx
+++ b/src/shape/Cross.tsx
@@ -18,7 +18,7 @@ interface CrossProps {
 
 export type Props = SVGProps<SVGPathElement> & CrossProps;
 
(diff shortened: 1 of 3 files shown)


nikolasrieble:
Love it! Can you please test by adding a storybook story that uses the Cross component? 
Or a unit test with RTL doing so? 

Same please for all other refactorings. We need to be confident that this does not break anything. And manual testing is not promising 😓

akamfoad:
@nikolasrieble Sure, the current tests are passing which means the current tested behavior worked as expected, I don't think most of the refactors will have any UI changes as I try to avoid that. I'll try to add jest test cases for the functions and codes that are refactored out of the components.

akamfoad:
@nikolasrieble 

As I see, snapshots are best strategy for these small components, making sure their result is consistent when changes happen. Let me know if you'd have a different approach, I updated all other PRs too.

nikolasrieble:
> @nikolasrieble
> 
> As I see, snapshots are best strategy for these small components, making sure their result is consistent when changes happen. Let me know if you'd have a different approach, I updated all other PRs too.

I think storybook is best, because we then can with chromatic visually compare whether the behaviour is identical.

akamfoad:
@nikolasrieble I'd have added a storybook for it too but since these components don't have storybooks it will not make any difference to add it in this PR tbh. But I can add some storybook entries for these shapes in a different one.

nikolasrieble:
see https://github.com/recharts/recharts/pull/3539

akamfoad:
@nikolasrieble The build broke due to some special characters in `Cross` story, I fixed those as well, I think now we can be confident to merge this. I'm working on adding a similar story to other Components I refactored so we can have a story before merging the refactor PRs.

ckifer:
@akamfoad fixing the build issue separately from this PR since it is unrelated, you may have to reconcile that change after I merge https://github.com/recharts/recharts/pull/3543

akamfoad:
@ckifer Sure I'll

akamfoad:
@nikolasrieble  @ckifer 

I noticed that the Chromatic action will run only on master pushes (I guess) but not on PRs, does that mean these PRs should be merged then tested whether they're correct or not, or am I missing something?

https://github.com/recharts/recharts/blob/a32f1ee85ecae4ccf18cd3352f164a707c737f03/.github/workflows/chromatic.yml#L6-L7

akamfoad:
Makes sense, I'd be happy to help where I can. I'd love to see progress on this project.

nikolasrieble (approved):
Tested manually and works well in storybook. Thank you for your contribution, and please excuse the very slow reaction time on our side ❤️

PR #3484, opened 2023-03-24
Storybook: add `ErrorBar` stories

Yilun-Sun:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #3487, opened 2023-03-27
fix: `CI` error

Yilun-Sun:
image

CI error blocks all new PRs.

image

## Related Issue

PR has ci error:
#3484 

## Types of changes

## Checklist:

### DIFF
--- a/src/util/DataUtils.ts
+++ b/src/util/DataUtils.ts
@@ -72,7 +72,7 @@ export const getAnyElementOfObject = (obj: any) => {
   return null;
 };
 
-export const hasDuplicate = (ary: Array<unknown>) => {
+export const hasDuplicate = (ary: Array<any>) => {
   if (!_.isArray(ary)) {
     return false;
   }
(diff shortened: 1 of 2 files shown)


src/util/DataUtils.ts
```diff
@@ -72,7 +72,7 @@ export const getAnyElementOfObject = (obj: any) => {
   return null;
 };
 
-export const hasDuplicate = (ary: Array<unknown>) => {
+export const hasDuplicate = (ary: Array<any>) => {
```
ckifer, line 75: why do we need to change this? Probably better to keep it `unknown` no?

ckifer (changes requested):

src/util/DataUtils.ts
ckifer: oh nevermind I lost context since I looked at this. Can't be used as an index type, got it

ckifer:
thank you! This isn't really ideal because `width` is still able to be a number or string and we're casting it to a Number. But better than before.

PR #3539, opened 2023-04-17
Storybook: Add story for Cross

nikolasrieble:
## Description
Strangely I could not get the cross to work without the Customized wrapper, although I do not even use what Customized provides, namely access to chart content. 

Either way, this does work, all props are usable controls, and it documents the Cross component, lets iterate. 

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/API/shapes/Cross.mdx
+++ b/storybook/stories/API/shapes/Cross.mdx
@@ -1,0 +1,8 @@
+import { ArgTypes } from '@storybook/blocks';
+import * as CrossStories from './Cross.stories';
+
+# Cross
+
+TODO
+
+<ArgTypes of={CrossStories} sort={'requiredFirst'} />
--- a/storybook/stories/API/shapes/Cross.mdx
+++ b/storybook/stories/API/shapes/Cross.mdx
@@ -3,6 +3,14 @@ import * as CrossStories from './Cross.stories';
 
 # Cross
 
-TODO
+The Cross is a shape that itself is not a valid child in any component. 
+It can be used to implement custom elements, returning it from within a
+
+\```jsx
+<Customized component={<Cross {...crossProps} />}>
+\```
+
+or other custom components (i.e. shape of a bar) 
+
 
 <ArgTypes of={CrossStories} sort={'requiredFirst'} />
--- a/storybook/stories/API/shapes/Cross.stories.tsx
+++ b/storybook/stories/API/shapes/Cross.stories.tsx
@@ -1,0 +1,85 @@
+import React from 'react';
+import { Cross, Customized, ComposedChart, ResponsiveContainer } from '../../../../src';
+import { GeneralStyle } from '../props/Styles';
+
+export default {
+  component: Cross,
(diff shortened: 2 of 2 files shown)


nikolasrieble:
@akamfoad Have a look at this example. After this is merged, when can see whether the component behaves the same after refactoring in storybook. Also, please do review this PR. I could not get the Cross to work without the Customized wrapper.

Yilun-Sun:
I think you can only use shapes like `Cross` in `Customized`, since the chart only accept components of:
image

storybook/stories/API/shapes/Cross.mdx
```diff
@@ -0,0 +1,8 @@
+import { ArgTypes } from '@storybook/blocks';
+import * as CrossStories from './Cross.stories';
+
+# Cross
+
+TODO
```
Yilun-Sun, line 6: The parent component would be `Customzied` and no child support.

Yilun-Sun (approved):

storybook/stories/API/shapes/Cross.mdx
akamfoad, line 6:
```suggestion
## Parent Component

The Cross can be used in combination with `<Customized />` to render inside all charts that Customized can be used. Rendering `Cross` directly inside charts doesn't have any effect.

You can use Cross with Customized like below: (please remove the `\` I added it to escape GitHub automatic matching)

\```jsx
<Customized component={<Cross {...crossProps} />}>
\```

## Child component

The Cross component is a leaf component, it doesn't render anything as its child.

## Properties

Properties in the groups Other and Internal are not recommended to be used.

```

storybook/stories/API/shapes/Cross.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+            top: 5,
+            right: 30,
+            left: 20,
+            bottom: 5,
+          }}
+        >
+          {/* I have no idea why this is needed. */}
+          <Customized component={() => CustomizedCross(args)} />
+          {/* But this does not work. */}
+          {/* <Cross {...args} /> */}
```
akamfoad, line 67: As @Yilun-Sun mentioned these components (I think all shapes) need `<Customized />` to render, I think we can remove these comments.

Also the customized story can be improved I might open a PR later, so the component prop can be either a function that returns a React element or it can be React element directly, so you can do this

```suggestion
          <Customized component={<Cross {...args} />} />
```

storybook/stories/API/shapes/Cross.mdx
akamfoad: You're right, please feel free to dismiss the Paren components part.

nikolasrieble, line 6:
```suggestion
The Cross is a shape that itself is not a valid child in any component. 
It can be used to implement custom elements, returning it from within a

\```jsx
<Customized component={<Cross {...crossProps} />}>
\```

or other custom components (i.e. shape of a bar) 

```

PR #3544, opened 2023-04-21
Add displayName for ResponsiveContainer

seiyab:
## Description
Add displayName for ResponsiveContainer

## Related Issue
N/A

## Motivation and Context
As of master branch, `<ResponsiveContainer />` is shown as anonymous component in React devtools.
`displayName` should help maintainer and application developer debugging.

And, to be honest, an application that I develop heavily depends on Recharts internal and removing name / displayName by `forwardRef` breaks our app 😅 .

## How Has This Been Tested?

I looked storybook and opened React devtools.
And I think new test for this change is not necessary.

image

## Types of changes

## Checklist:

- [ ] ~My change requires a change to the documentation.~
- [ ] ~I have updated the documentation accordingly.~
- [ ] ~I have added tests to cover my changes.~
- [x] All ~new and~ existing tests passed.

### DIFF
--- a/src/component/ResponsiveContainer.tsx
+++ b/src/component/ResponsiveContainer.tsx
@@ -180,3 +180,5 @@ export const ResponsiveContainer = forwardRef(
     );
   },
 );
+
+ResponsiveContainer.displayName = 'ResponsiveContainer';
--- a/src/container/Layer.tsx
+++ b/src/container/Layer.tsx
@@ -22,5 +22,3 @@ export const Layer = React.forwardRef((props: Props, ref: any) => {
     </g>
   );
 });
-
-Layer.displayName = 'Layer';
--- a/src/container/Layer.tsx
+++ b/src/container/Layer.tsx
@@ -22,3 +22,5 @@ export const Layer = React.forwardRef((props: Props, ref: any) => {
     </g>
   );
 });
+
+Layer.displayName = 'Layer';


seiyab:
ResponsiveContainer looks to be shown as `"ResponsiveContainer"` in Storybook on master branch, too.
I'm misunderstanding something.
I made it draft.

ckifer:
static `displayName` is used for debugging purposes and shouldn't have anything to do with how devtools interprets the component name itself (at least not that I know of). Unfortunately `displayName` was used in implementation of recharts but shouldn't break anything here

seiyab:
Thank you, I understand🙌.
https://legacy.reactjs.org/docs/react-component.html#displayname

Anyway, it should be better to have a displayName here.

ckifer:
I would say that this is a deprecated attribute since it isn't in the current react docs at all. 

"Usually, you don’t need to set it explicitly because it’s inferred from the name of the function or class that defines the component"

So I don't think this PR is needed, if anything we should be removing displayName from the codebase

ckifer:
Going to close this as won't fix as this would be moving recharts backwards. I think instead we could fix it like some of the things suggested here https://stackoverflow.com/a/71692429/8050147

seiyab:
Sorry for late response.
Closing this PR is OK because I'm not eager to merge it.

I'm still investigating personally because I'm still getting confused about devtools on Recharts document shows `Anonymous`, neither desired name `ResponsiveContainer` nor minified name like `t` for `ResonsiveContainer`.
image

PR #3546, opened 2023-04-21
[WIP] [a11y] Keyboard navigation support for categorical charts

julianna-langston:
## Types of changes

## Checklist:

### DIFF
--- a/src/chart/AccessibilityManager.tsx
+++ b/src/chart/AccessibilityManager.tsx
@@ -1,3 +1,10 @@
+import { LayoutType } from '../util/types';
+
+interface ContainerOffset {
+  top: number;
+  left: number;
+}
+
 /* eslint-disable no-underscore-dangle */
 export class AccessibilityManager {
   private _coordinateList: any[] = [];
@@ -8,6 +15,10 @@ export class AccessibilityManager {
 
   private _container: any;
 
+  private _layout: LayoutType = 'horizontal';
+
+  private _offset: ContainerOffset | null = null;
+
   public focus() {
     this.spoofMouse();
   }
@@ -15,12 +26,34 @@ export class AccessibilityManager {
   public keyboardEvent(e) {
     switch (e.key) {
       case 'ArrowRight': {
+        if (this._layout !== 'horizontal') {
+          return;
+        }
         this._activeIndex = Math.min(this._activeIndex + 1, this._coordinateList.length - 1);
         this.spoofMouse();
         break;
       }
       case 'ArrowLeft': {
+        if (this._layout !== 'horizontal') {
+          return;
+        }
         this._activeIndex = Math.max(this._activeIndex - 1, 0);
+        this.spoofMouse();
+        break;
+      }
+      case 'ArrowUp': {
+        if (this._layout !== 'vertical') {
+          return;
+        }
+        this._activeIndex = Math.max(this._activeIndex - 1, 0);
+        this.spoofMouse();
+        break;
+      }
+      case 'ArrowDown': {
+        if (this._layout !== 'vertical') {
+          return;
+        }
+        this._activeIndex = Math.min(this._activeIndex + 1, this._coordinateList.length - 1);
         this.spoofMouse();
         break;
       }
@@ -32,12 +65,12 @@ export class AccessibilityManager {
 
   private spoofMouse() {
     const { x, y } = this._container.getBoundingClientRect();
-    const newX = x + this._coordinateList[this._activeIndex].coordinate;
+    const { coordinate } = this._coordinateList[this._activeIndex];
 
-    this._mouseHandlerCallback({
-      pageX: newX,
-      pageY: y + 100,
-    });
+    const pageX = x + (this._layout === 'horizontal' ? coordinate : 100);
+    const pageY = y + (this._layout === 'vertical' ? coordinate : this._offset.top);
+
+    this._mouseHandlerCallback({ pageX, pageY });
   }
 
   set coordinateList(list: any[]) {
@@ -53,5 +86,13 @@ export class AccessibilityManager {
 
   set container(el: any) {
     this._container = el;
+  }
+
+  set layout(newLayout: LayoutType) {
+    this._layout = newLayout;
+  }
+
+  set offset(newOffset: ContainerOffset) {
+    this._offset = newOffset;
   }
 }
--- a/src/chart/AccessibilityManager.tsx
+++ b/src/chart/AccessibilityManager.tsx
@@ -23,7 +23,7 @@ export class AccessibilityManager {
     this.spoofMouse();
   }
 
-  public keyboardEvent(e) {
+  public keyboardEvent(e: any) {
     switch (e.key) {
       case 'ArrowRight': {
         if (this._layout !== 'horizontal') {
--- a/src/chart/AccessibilityManager.tsx
+++ b/src/chart/AccessibilityManager.tsx
@@ -1,0 +1,57 @@
+/* eslint-disable no-underscore-dangle */
+export class AccessibilityManager {
+  private _coordinateList: any[] = [];
+
+  private _activeIndex = 0;
+
+  private _mouseHandlerCallback: any;
+
+  private _container: any;
+
+  public focus() {
+    this.spoofMouse();
+  }
+
+  public keyboardEvent(e) {
+    switch (e.key) {
+      case 'ArrowRight': {
+        this._activeIndex = Math.min(this._activeIndex + 1, this._coordinateList.length - 1);
+        this.spoofMouse();
+        break;
+      }
+      case 'ArrowLeft': {
+        this._activeIndex = Math.max(this._activeIndex - 1, 0);
+        this.spoofMouse();
+        break;
+      }
+      default: {
+        break;
+      }
+    }
+  }
+
+  private spoofMouse() {
+    const { x, y } = this._container.getBoundingClientRect();
+    const newX = x + this._coordinateList[this._activeIndex].coordinate;
+
+    this._mouseHandlerCallback({
+      pageX: newX,
+      pageY: y + 100,
+    });
+  }
+
+  set coordinateList(list: any[]) {
+    this._coordinateList = list;
+    if (this._activeIndex >= this._coordinateList.length) {
+      this._activeIndex = this._coordinateList.length - 1;
+    }
+  }
+
+  set mouseHandlerCallback(cb: any) {
+    this._mouseHandlerCallback = cb;
+  }
+
+  set container(el: any) {
+    this._container = el;
+  }
+}
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -69,6 +69,7 @@ import {
   GeometrySector,
   AxisDomain,
 } from '../util/types';
+import { AccessibilityManager } from './AccessibilityManager';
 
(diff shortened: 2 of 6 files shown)


src/chart/AccessibilityManager.tsx
```diff
@@ -0,0 +1,98 @@
+import { LayoutType } from '../util/types';
+
+interface ContainerOffset {
+  top: number;
+  left: number;
+}
+
+/* eslint-disable no-underscore-dangle */
+export class AccessibilityManager {
```
ckifer, line 9: since this doesn't have any JSX can it be a TS file instead of TSX?

src/chart/AccessibilityManager.tsx
```diff
@@ -0,0 +1,98 @@
...
+  private _coordinateList: any[] = [];
+
+  private _activeIndex = 0;
+
+  private _mouseHandlerCallback: any;
+
+  private _container: any;
```
ckifer, line 16: I know these are coming from generateCategoricalChart which has terrible typing but if we could type them any more specifically that would be awesome

src/chart/generateCategoricalChart.tsx
```diff
@@ -2058,6 +2072,19 @@ export const generateCategoricalChart = ({
         return [graphicalItem, null, null];
       }
 
+      if (typeof this.props.defaultActiveIndex !== 'undefined') {
```
ckifer: can't we just check for `this.props.defaultActiveIndex !== undefined` instead of checking `typeof`?

src/chart/generateCategoricalChart.tsx
```diff
@@ -1027,6 +1032,15 @@ export const generateCategoricalChart = ({
...
       }
+
+      this.accessibilityManager.container = this.container;
+      this.accessibilityManager.offset = {
+        left: this.props.margin.left ?? 0,
+        top: this.props.margin.top ?? 0,
+      };
+      this.accessibilityManager.coordinateList = this.state.tooltipTicks;
+      this.accessibilityManager.mouseHandlerCallback = this.handleMouseMove;
+      this.accessibilityManager.layout = this.props.layout;
```
ckifer, line 1042: can we init the `AccessibilityManager` here and set all of this in a constructor instead of one at a time?

src/chart/generateCategoricalChart.tsx
```diff
@@ -2191,6 +2218,17 @@ export const generateCategoricalChart = ({
...
 
+      if (this.props.accessibilityLayer) {
+        attrs.tabIndex = 0;
+        attrs.role = 'img';
+        attrs.onKeyDown = (e: any) => {
+          this.accessibilityManager.keyboardEvent(e);
+        };
+        attrs.onFocus = () => {
+          this.accessibilityManager.focus();
+        };
```
ckifer, line 2215: these are all props that will get overridden if the user specifies `accessibilityLayer` - ideally the user wouldn't really be able to specify most of them themselves (recharts extends SVGProps for everything), but we could have issues where users report that their `tabIndex`, `role`, `onKeyDown`, and `onFocus` are being overridden. We would have to call these out in the docs as a caveat to `accessibilityLayer`

ckifer (commented):
Really awesome work! Some suggestions and questions but will let you keep going on the draft here 🎉

src/chart/AccessibilityManager.tsx
nikolasrieble: Agree, the goal should be: No any. Leading by example that would slowly help us pave the way to a better future :D

src/chart/AccessibilityManager.tsx
```diff
@@ -0,0 +1,98 @@
+import { LayoutType } from '../util/types';
+
+interface ContainerOffset {
+  top: number;
+  left: number;
+}
+
+/* eslint-disable no-underscore-dangle */
```
nikolasrieble, line 8: Lets either follow the rule, or disable it. I am fine with both.

src/chart/AccessibilityManager.tsx
```diff
@@ -0,0 +1,98 @@
...
+export class AccessibilityManager {
+  private _coordinateList: any[] = [];
+
+  private _activeIndex = 0;
```
nikolasrieble, line 12: The activeIndex having a default of 0 is certainly coming from the categorical chart. Given that it make sense both have the same default, how about making sure that they do, by extracting a named constant and importing in both places? 

```
export const DEFAULT_ACTIVE_INDEX = 0
```

src/chart/generateCategoricalChart.tsx
nikolasrieble: Good point. I guess this could also be done in the description of the accessibilityLayer in the API documentation.

storybook/stories/API/cartesian/Area.stories.tsx
```diff
@@ -358,3 +358,40 @@ export const Points = {
...
+      <ResponsiveContainer width="100%" height={300}>
+        <ComposedChart
+          margin={{
+            top: 20,
+            right: 20,
+            bottom: 20,
+            left: 20,
+          }}
+          data={pageData}
+          accessibilityLayer
```
nikolasrieble, line 374: Strictly speaking, the accessibilityLayer is not the API of the Area (thus does not belong into this folder) but the API of the ComposedChart.

Could you please move it to the ComposedChart API documentation, and add a description to the ArgsTable for accessibilityLayer

julianna-langston: I moved the story to the ComposedChart storybook section, but I couldn't figure out where the ArgsTable was.

nikolasrieble: Fair enough! I will do a follow up PR and tag you on it, after this merged.

src/chart/AccessibilityManager.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  private mouseHandlerCallback: InitiableOptions['mouseHandlerCallback'];
+
+  private container: InitiableOptions['container'];
+
+  private layout: InitiableOptions['layout'] = 'horizontal';
+
+  private offset: InitiableOptions['offset'];
+
+  public init({ coordinateList = [], container, layout, offset, mouseHandlerCallback }: InitiableOptions) {
```
ckifer, line 29: nit: could this not be an actual constructor or is there a reason we need to create an instance before we add these properties?

src/chart/AccessibilityManager.ts
```diff
@@ -0,0 +1,94 @@
...
+  mouseHandlerCallback: (e: any) => void;
+  container: HTMLElement;
+  layout: LayoutType;
+  offset: ContainerOffset;
+}
+
+export class AccessibilityManager {
+  private activeIndex = 0;
+
+  private coordinateList: InitiableOptions['coordinateList'] = [];
```
ckifer, line 19: personal preference would just be to define these types as their true types and without `InitiableOptions` (can still exist for the sake of the `init` function but these are easier to read if their explicit types are here)

src/chart/AccessibilityManager.ts
```diff
@@ -0,0 +1,94 @@
+import type { LayoutType, TickItem } from '../util/types';
+
+interface ContainerOffset {
+  top: number;
+  left: number;
+}
+
+interface InitiableOptions {
+  coordinateList: TickItem[];
+  mouseHandlerCallback: (e: any) => void;
```
ckifer, line 10: we should still be able to get more specific with this no?
```
  mouseHandlerCallback: (e: { pageX: number, pageY: number }) => void;
```

src/chart/generateCategoricalChart.tsx
```diff
@@ -1027,6 +1031,19 @@ export const generateCategoricalChart = ({
...
+        this.accessibilityManager.init({
+          container: this.container,
+          offset: {
+            left: this.props.margin.left ?? 0,
+            top: this.props.margin.top ?? 0,
+          },
+          coordinateList: this.state.tooltipTicks,
+          mouseHandlerCallback: this.handleMouseMove,
+          layout: this.props.layout,
+        });
```
ckifer, line 1045: why not use a constructor?
```
accessibilityManager = new AccessibilityManager({ 
          container: this.container,
          offset: {
            left: this.props.margin.left ?? 0,
            top: this.props.margin.top ?? 0,
          },
          mouseHandlerCallback: this.handleMouseMove,
          layout: this.props.layout,
});
```

ckifer (changes requested):
Some more comments, but should be fine after making the event handler types a little more specific. Others are just suggestions

src/chart/generateCategoricalChart.tsx
```diff
@@ -2191,6 +2243,23 @@ export const generateCategoricalChart = ({
         );
       }
 
+      if (this.props.accessibilityLayer) {
+        // Set tabIndex to 0 by default (can be overwritten)
+        attrs.tabIndex = 0 ?? this.props.tabIndex;
+        // Set role to img by default (can be overwritten)
+        attrs.role = 'img' ?? this.props.role;
+        attrs.onKeyDown = (e: any) => {
```
ckifer, line 2251: this event should be able to be typed rather than `any`

src/chart/generateCategoricalChart.tsx
```diff
@@ -1027,6 +1035,50 @@ export const generateCategoricalChart = ({
...
+          left: this.props.margin.left ?? 0,
+          top: this.props.margin.top ?? 0,
+        },
+        coordinateList: this.state.tooltipTicks,
+        mouseHandlerCallback: this.handleMouseMove,
+        layout: this.props.layout,
+      });
+    }
+
+    getSnapshotBeforeUpdate(
```
ckifer, line 1051: just a callout that using this means not refactoring this to a more functional component any time soon. Probably weren't going to anyways, but `getSnapshotBeforeUpdate` doesn't exist in the world of hooks

ckifer (approved):
Looks great, thank you! Left a minor comment and a callout but I think this is relatively ready for merge

src/chart/generateCategoricalChart.tsx
ckifer: @nikolasrieble just tagging you for visibility

nikolasrieble: Admittedly I do not think this makes a large difference in context of an overall complex component that will be hard to take apart. We might consider simply rewriting it from scratch in a functional way, and use the existing component as a comparison / test.

test/chart/AccessibilityLayer.spec.tsx
```diff
@@ -0,0 +1,342 @@
...
+    );
+  };
+
+  // Q: What is the point of this test?
+  // A: Well, I was thinking, what was a way this could break and would be hard to figure out?
+  //    If the elements and their event listeners got disconnected. But how would that happen
+  //    outside of something weird happening with React rendering? I don't know, but if it DID
+  //    happen, it would take a while for someone to notice, and could be potentially hard to
+  //    reproduce or debug. It's not that hard to test, and could potentially save someone some
+  //    stress down the line. That trade-off feels worth it!
```
nikolasrieble, line 317: ❤️

ckifer:
Merging this in - thank you @julianna-langston !

src/chart/generateCategoricalChart.tsx
ckifer: Sweet, sounds good

Issue #3548, opened 2023-04-23
[Tooltip] Invalid defaultProps for viewBox


ckifer:
## Do you want to request a feature or report a bug?
Bug

## What is the current behavior?
Currently the recharts tooltip has some incorrect viewBox default props set https://github.com/recharts/recharts/blob/master/src/component/Tooltip.tsx#L103
```
  viewBox: { x1: 0, x2: 0, y1: 0, y2: 0 },
```
current viewBox properties are `x, y, height, width`

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/ndLhnegs/).
https://github.com/recharts/recharts/blob/master/src/component/Tooltip.tsx#L103

## What is the expected behavior?
Use the correct props + test that the correct props don't break anything

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
latest version, code issue

PR #3554, opened 2023-04-25
fix: defaultProps of tooltip, fix viewBox

ckifer:
## Description

`viewBox` default props are incorrect in `Tooltip`

## Related Issue
https://github.com/recharts/recharts/issues/3548

## Motivation and Context

these props are incorrect, x1, x2, y1, y2 do not exist on viewBox.

This gives a default for when no default is provided (but a default is always provided when Tooltip is used)

## How Has This Been Tested?

- Run storybook + console log viewBox
  - the default is never used
  - Output: `{ brushBottom: 50, top: 20, bottom: 50, left: 20, right: 20, width: 1822, height: 330, x: 20, y: 20 }`
- Check `generateCategoricalChart` code
  - https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.tsx#L1877-L1894
    - x and y are always set from `renderTooltip` 
    - offset always contains (chart) height and width
 - viewBox is used here https://github.com/recharts/recharts/blob/master/src/component/Tooltip.tsx#L214-L227
   - confirm that it is always set

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Tooltip.tsx
+++ b/src/component/Tooltip.tsx
@@ -104,10 +104,7 @@ const tooltipDefaultProps: TooltipProps<number, string> = {
   allowEscapeViewBox: { x: false, y: false },
   reverseDirection: { x: false, y: false },
   offset: 10,
-  // TODO: replace invalid properties
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
-  viewBox: { x1: 0, x2: 0, y1: 0, y2: 0 },
+  viewBox: { x: 0, y: 0, height: 0, width: 0 },
   coordinate: { x: 0, y: 0 },
   // this doesn't exist on TooltipProps
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment


src/component/Tooltip.tsx
```diff
@@ -104,10 +104,7 @@ const tooltipDefaultProps: TooltipProps<number, string> = {
   allowEscapeViewBox: { x: false, y: false },
   reverseDirection: { x: false, y: false },
   offset: 10,
-  // TODO: replace invalid properties
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
-  viewBox: { x1: 0, x2: 0, y1: 0, y2: 0 },
+  viewBox: { x: 0, y: 0, height: 0, width: 0 },
```
ckifer, line 107: zeros should be good defaults. If Tooltip doesn't go through `generateCategoricalChart` it won't get rendered and doesn't need a viewBox anyways

nikolasrieble (approved):
🚀

PR #3561, opened 2023-04-30
Storybook: add custom tooltip

CoffeeGeek101:
## Description

Adding Tooltip implementation to the storybook, Simple tooltip, Custom single valued tooltip and Custom multiple valued tooltip

A Tooltip story has been added to the API section of the Storybook

**Before :**
Screenshot 2023-04-30 at 11 36 15 PM
**After :**
Screenshot 2023-04-30 at 11 36 29 PM

## Motivation and Context

The default tooltip of the library is good enough, but user might want to have the tooltip matches with their UI, and there is no direct prop available to customised the component, might lead to a bad UX.

## How Has This Been Tested?

This is an initial PR just to get the work started toward it, so the changes haven't tested yet.

This is directory structure and the custom tooltip implementation preview:

Screenshot 2023-04-30 at 11 44 07 PM

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/API/tooltip/Tooltip.stories.tsx
+++ b/storybook/stories/API/tooltip/Tooltip.stories.tsx
@@ -7,19 +7,15 @@ export default {
 };
 
 export const SimpleTooltip = {
-  render: (tootipArgs: Record<string, any>) => {
+  render: (tooltipArgs: Record<string, any>) => {
     return (
       <ResponsiveContainer width="100%" height={400}>
-        <LineChart {...tootipArgs}>
+        <LineChart data={pageData}>
           <Line dataKey="uv" />
-          <Tooltip />
+          <Tooltip {...tooltipArgs} />
         </LineChart>
       </ResponsiveContainer>
     );
-  },
-  args: {
-    data: pageData,
-    dataKey: 'uv',
   },
 };
 
@@ -52,18 +48,8 @@ const CustomTooltip: React.FC = ({ active, payload, label }: any) => {
 };
 
 export const CustomSingleValued = {
-  render: (tooltipArgs: Record<string, any>) => {
-    return (
-      <ResponsiveContainer width="100%" height={400}>
-        <LineChart {...tooltipArgs}>
-          <Line dataKey="uv" />
-          <Tooltip content={<CustomTooltip />} />
-        </LineChart>
-      </ResponsiveContainer>
-    );
-  },
+  ...SimpleTooltip,
   args: {
-    data: pageData,
-    dataKey: 'uv',
+    content: <CustomTooltip />,
   },
 };
--- a/storybook/stories/API/tooltip/Tooltip.stories.tsx
+++ b/storybook/stories/API/tooltip/Tooltip.stories.tsx
@@ -1,0 +1,71 @@
+import React from 'react';
+import { pageData } from '../../data';
+import { Line, LineChart, ResponsiveContainer, Tooltip } from '../../../../src';
+
+export default {
+  component: LineChart,
+};
+
+export const SimpleTooltip = {
+  render: (args: Record<string, any>) => {
+    return (
+      <ResponsiveContainer width="100%" height={400}>
+        <LineChart {...args}>
+          <Line dataKey="uv" />
+          <Tooltip />
+        </LineChart>
+      </ResponsiveContainer>
+    );
+  },
+  args: {
+    data: pageData,
+    dataKey: 'uv',
+  },
+};
+
+const CustomTooltip = ({ active, payload, label }: any) => {
+  if (active && payload) {
+    return (
+      <div
+        // className="custom-tooltip"
+        style={{
+          backgroundColor: '#5b63ffe7',
+          padding: '5px',
+          borderRadius: '10px',
+          boxShadow: '1px 2px 10px -2px #7873ff',
+        }}
+      >
+        <p
+          // className="label"
+          style={{
+            borderStyle: 'solid 1px',
+            fontSize: '13px',
+            fontWeight: '600',
+            fontFamily: 'sans-serif',
+            color: '#fff',
+          }}
+        >
+          {`${label} : ${payload[0].value}`}
+        </p>
+      </div>
+    );
+  }
+  return null;
+};
+
+export const CustomSingleValued = {
+  render: (args: Record<string, any>) => {
+    return (
+      <ResponsiveContainer width="100%" height={400}>
+        <LineChart {...args}>
+          <Line dataKey="uv" />
+          <Tooltip content={<CustomTooltip />} />
+        </LineChart>
+      </ResponsiveContainer>
+    );
+  },
+  args: {
+    data: pageData,
+    dataKey: 'uv',
+  },
+};
--- a/storybook/stories/API/tooltip/Tooltip.stories.tsx
+++ b/storybook/stories/API/tooltip/Tooltip.stories.tsx
@@ -23,11 +23,10 @@ export const SimpleTooltip = {
   },
 };
 
-const CustomTooltip = ({ active, payload, label }: any) => {
+const CustomTooltip: React.FC = ({ active, payload, label }: any) => {
   if (active && payload) {
     return (
       <div
-        // className="custom-tooltip"
         style={{
           backgroundColor: '#5b63ffe7',
           padding: '5px',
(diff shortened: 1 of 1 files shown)


CoffeeGeek101:
Need a feedback on the work till now, If the implementation is alright or not.

storybook/stories/API/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+          // className="label"
+          style={{
+            borderStyle: 'solid 1px',
+            fontSize: '13px',
+            fontWeight: '600',
+            fontFamily: 'sans-serif',
+            color: '#fff',
+          }}
+        >
+          {`${label} : ${payload[0].value}`}
```
ckifer, line 48: would do some validation that `${payload[0].value}` exists or use `payload?.[0]?.value}`

storybook/stories/API/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+      </ResponsiveContainer>
+    );
+  },
+  args: {
+    data: pageData,
+    dataKey: 'uv',
+  },
+};
+
+const CustomTooltip = ({ active, payload, label }: any) => {
```
ckifer, line 26: Can you type this? Should return `JSX.Element | null` I think (or similar)?

storybook/stories/API/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+  if (active && payload) {
+    return (
+      <div
+        // className="custom-tooltip"
```
ckifer, line 30: either add these or remove them. I think they are fine to be there as classNames for an example

ckifer (commented):
Thanks for the contribution, looking good so far

CoffeeGeek101:
Heyy, @ckifer thank you feedback, I have fixed the type and added the validation to the payload array.
Also, removed the 'className' comments, which are not required here.

If this is alright, I'll start working on the multiple valued tooltip story and docs mdx file for them.

storybook/stories/API/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <Tooltip />
+        </LineChart>
+      </ResponsiveContainer>
+    );
+  },
+  args: {
+    data: pageData,
+    dataKey: 'uv',
+  },
+};
```
nikolasrieble, line 24:
```suggestion
export const SimpleTooltip = {
  render: (tooltipArgs: Record<string, any>) => {
    return (
      <ResponsiveContainer width="100%" height={400}>
        <LineChart data={pageData}>
          <Line dataKey="uv" />
          <Tooltip {...tooltipArgs} />
        </LineChart>
      </ResponsiveContainer>
    );
  },
};
```

In the API section of storybook, we want to document the API of a single component, in this case, the Tooltip. 

Thus every story should only have args that belong to the component in question.

nikolasrieble: You could consider the Line stories as an example. There, we group the props by categories (i.e. animation, style, general, ...) and create a story for each, with non default values, which show the behaviour and are changeable per story.

nikolasrieble (commented):
Great first step! Please consider my suggestion

storybook/stories/API/tooltip/Tooltip.stories.tsx
CoffeeGeek101: okay, let me explain the suggested changes, so that I can be certain if I understood the required changes to be made.

- the tooltip story currently having args that is not relevant to the component, so I need to change the args that are specific to the tooltip component and will displayed in the controls section of the storybook playground.

- If I understand the problem correctly, I need to use "tooltipArgs" for that case.

nikolasrieble: Yes, makes sense.

CoffeeGeek101: I was checking out the Line stories, and understood the problem in this PR, but I couldn't find any tooltipArgs on the API props, do I need to or can I import **TooltipProps** from the tooltip component

nikolasrieble: tooltipPops is not a type just an alias to make clear that we only want to pass it to the tooltip in api stories

storybook/stories/API/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const SimpleTooltip = {
+  render: (tootipArgs: Record<string, any>) => {
+    return (
+      <ResponsiveContainer width="100%" height={400}>
+        <LineChart {...tootipArgs}>
+          <Line dataKey="uv" />
+          <Tooltip />
+        </LineChart>
```
nikolasrieble, line 16:
```suggestion
        <LineChart data={pageData} >
          <Line dataKey="uv" />
          <Tooltip  {...tootipArgs} />
        </LineChart>
```
The args must still be passed to the Tooltip component.

nikolasrieble: And the chart needs data.

storybook/stories/API/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+      </ResponsiveContainer>
+    );
+  },
+  args: {
+    data: pageData,
+    dataKey: 'uv',
+  },
```
nikolasrieble, line 23:
```suggestion
```

Here no args are needed.

storybook/stories/API/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <Line dataKey="uv" />
+          <Tooltip content={<CustomTooltip />} />
+        </LineChart>
+      </ResponsiveContainer>
+    );
+  },
+  args: {
+    data: pageData,
+    dataKey: 'uv',
+  },
```
nikolasrieble, line 68: If you change the first story following my suggestion, you might then simply reuse the render function as such: 

```suggestion
  ... SimpleTooltip,
  args: {
    content: <CustomTooltip />,
  },
```

nikolasrieble: Because between the first and the second story, the only difference is the value passed to the content prop.

storybook/stories/API/tooltip/Tooltip.stories.tsx
CoffeeGeek101: Thank you, I really made a massive mistake there. Working on it!

storybook/stories/API/tooltip/Tooltip.stories.tsx
CoffeeGeek101: This really helped me, I understood, what I was doing wrong. I have made the possible changes.

ckifer (approved):

storybook/stories/API/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+};
+
+const CustomTooltip: React.FC = ({ active, payload, label }: any) => {
```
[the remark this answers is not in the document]
CoffeeGeek101: I have added some types for the props

Screenshot 2023-05-06 at 11 07 17 AM

storybook/stories/API/tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <p
+          style={{
+            borderStyle: 'solid 1px',
+            fontSize: '13px',
+            fontWeight: '600',
+            fontFamily: 'sans-serif',
+            color: '#fff',
+          }}
+        >
+          {`${label} : ${payload[0]?.value}`}
```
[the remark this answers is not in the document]
CoffeeGeek101: I have handle the case when the payload array is empty and updated the label with the key of value that is coming from the payload array.

Screenshot 2023-05-06 at 11 08 31 AM

CoffeeGeek101:
If the single value is tooltip is ready, can I start with the multi - valued tooltip?

CoffeeGeek101:
I'm getting this eslint error, can you give some context why this might be happening?
Screenshot 2023-05-11 at 1 28 01 PM

kolserdav:
Is there an example of use this feature?

ckifer:
@kolserdav  not sure what you mean. This is a PR that added some changes to the storybook rather than a feature. Those examples are deployed on the storybook if you want to review them there. Link in the readme

Issue #3468, opened 2023-03-21
`angle` prop on XAxis is not considered when deciding which ticks to show


petii:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-hji7ig?file=/src/index.js)

### Steps to reproduce
1. Create a chart which does not have enough space for all X axis ticks to be shown
2. Rotate X axis tick labels through the usage of the `angle` prop

### What is expected?
More ticks are shown if they are able to be rendered without overlap.

### What is actually happening?
The same ticks are rendered as if they were not rotated.

| Environment | Info |
|---|---|
| Recharts | v2.3.0 |
| React | 16.8.6 |
| System | N/A |
| Browser | N/A |

---
The methods in `src/cartesian/getTicks.ts` are all under the impression that the text is displayed horizontally, because _getStringSize_ in `src/util/DOMUtils.ts` only considers the font styling and not the transformations.

This is a tricky thing to consider however, as with non right angles, the rotated SVG elements could be allowed to overlap, so the _getBoundingClientRect_ call in _getStringSize_ is probably not sufficient.

ckifer:
+1 to the tricky nature of this. Also hard to test (thank you Storybook for interaction tests/running in the browser).

Totally on board

Sergiiivzhenko:
workaround for me was to set interval={0}

petii:
> workaround for me was to set interval={0}

@Sergiiivzhenko While that usually works, if you have many ticks (or a small chart) that can lead to the ticks being rendered on top of each other.

bernardobelchior:
I have been facing this problem at work, so I've implemented a fix [here](https://github.com/recharts/recharts/pull/3576).

ckifer:
released in 2.7, resolving this

moaazbhnas228:
This is still an issue in version `2.15.1`.

``` js
              <XAxis
                dataKey="step_description"
                tickLine={false}
                tickMargin={0}
                axisLine={true}
                interval={0}

                
                angle={45}
                tick={{ dy: 20 }}
                // tickSize={2}
              />
```

![Image](https://github.com/user-attachments/assets/fced2a14-760b-4aad-9744-78804e131904)

ckifer:
[...]

PR #3576, opened 2023-05-09
fix(#3468): fix tick visibility calculation not accounting for angle prop on XAxis

bernardobelchior:
## Description

The XAxis component accepts an `angle` prop which renders the ticks at an angle. When calculating which ticks to show, Recharts does not take the `angle` prop into account, which results in ticks being hidden when they could have been shown without overlap. 

This PR fixes this issue by taking the angle at which ticks are rendered into account so that more ticks are shown if they don't overlap.

## Related Issue

https://github.com/recharts/recharts/issues/3468

## Motivation and Context

As explained above, this PR fixes ticks being hidden too conservatively because the angle at which they were displayed wasn't being taken into account.

## How Has This Been Tested?

I have added unit tests and have tested in our product that the changes do not cause ticks to overlap when displayed at an angle and `minTickGap` is 0. 

You can also check the Storybook in [this branch](https://github.com/rows/recharts/tree/chore/storybook-x-axis-angle), which was used to take the screenshots below.

| Before | After |
|--------|-------|
| image      |  image |

## Types of changes

## Checklist:


test/cartesian/getTicks.spec.ts
```diff
@@ -31,7 +32,7 @@ const EXAMPLE_INPUT = {
 
 jest.mock('../../src/util/DOMUtils', () => ({
   // We mock string size measurement, because getStringSize else returns 0 in these tests.
-  getStringSize: jest.fn((text: string) => ({ width: text.length, height: 20 })),
+  getStringSize: jest.fn((text: string) => ({ width: text.length, height: 12 })),
```
bernardobelchior, line 35: Reduce height to 12 so that the `value`s in the `ticks` didn't need to be as large since their width is equal to the number of characters.

src/util/CartesianUtils.ts
```diff
@@ -238,3 +238,18 @@ export const createLabeledScales = (options: Record<string, any>): LabeledScales
     },
   } as LabeledScales<Record<string, any>>;
 };
+
+/** Calculates the width of the largest horizontal line that fits inside a rectangle that is displayed at an angle.
+ * @param {Object} size Width and height of the text in a horizontal position.
+ * @param {number} angle Angle in degrees in which the text is displayed.
+ * @return {number} The width of the largest horizontal line that fits inside a rectangle that is displayed at an angle.
+ */
+export const getAngledRectangleWidth = ({ width, height }: Size, angle: number | undefined = 0) => {
```
bernardobelchior, line 247: This calculates the width of the red arrow shown below:

image

This width is the effective width of the tick and is what will be used to decide whether to show a tick.

nikolasrieble: Could you please a a few unit tests to this function, to document the behaviour and test edge cases?

nikolasrieble:
First: Great contribution. ❤️ This was a missing puzzle piece, and it very well appreciated. 

To test this, I checked out this branch, and added the angle to the CartesianAxis story of TickPositioning as such

```
export const TickPositioning = {
  render: (args: Record<string, any>) => {
    return (
      <ResponsiveContainer width="100%" height={surfaceHeight}>
        <Surface
          width={surfaceWidth}
          height={surfaceHeight}
          viewBox={{
            x: 0,
            y: 0,
            width: surfaceWidth,
            height: surfaceHeight,
          }}
        >
          <CartesianAxis
            width={surfaceHeight}
            height={surfaceWidth}
            viewBox={{
              x: 0,
              y: 0,
              width: surfaceWidth,
              height: surfaceHeight,
            }}
            ticks={args.data}
            interval={args.interval}
            angle={args.angle}
          />
          <CartesianAxis
            width={surfaceHeight}
            height={surfaceWidth}
            viewBox={{
              x: 0,
              y: 0,
              width: surfaceWidth,
              height: surfaceHeight,
            }}
            ticks={args.data.map((tick: TickItem) => ({ ...tick, coordinate: tick.coordinate / 2 }))}
            interval={args.interval}
            angle={args.angle}
            orientation="left"
          />
        </Surface>
      </ResponsiveContainer>
    );
  },
  args: {
    data: ticks,
  },
  parameters: {
    controls: { include: ['interval', 'angle'] },
    docs: {
      description: {
        story:
          'The ticks of a cartesian axis are positioned using the interval field. ' +
          'The interval computation can be overridden by setting interval to a number, ' +
          'which signifies how many ticks should not be shown between two visible ticks. ' +
          'By default, ticks that overflow their assigned space, are not shown. ',
      },
    },
  },
};
```

Then I manually rotated the ticks and looked at their behaviour: 

1 - It seems this does not work correctly with a vertical axis - but it also seems it did not work before - so that is okay. 
2 - It works great for the horizontal axis. As the degree increases, more labels are shown. That is expected, and an improvement in behaviour. 

What is strange is, that with a rotation of 1 degree, first all ticks disappear. That seems to be a bug in this implementation. I did not pin it down in the code, but it seems a bug in the video: 

https://github.com/recharts/recharts/assets/27995606/52d1f90c-01ee-4f92-80c0-139b3bc10c7d

nikolasrieble (commented):
Although I see a possible bug in the implementation, this behaviour is already much better and adds value than the current buggy behaviour which simply ignores the angle.

I am curious to hear your opinion on the behaviour, and whether you can find a fix. If not, we can still get this in, and simply iterate.

nikolasrieble (approved):

nikolasrieble (approved):

ckifer (approved):
Nice, thank you - changes LGTM - curious about Niko's comment above as well though

bernardobelchior:
> First: Great contribution. ❤️ This was a missing puzzle piece, and it very well appreciated.

Thank you 😄 

Thanks for your detailed writeup, @nikolasrieble! I'll address you comments below.

> 1 - It seems this does not work correctly with a vertical axis - but it also seems it did not work before - so that is okay.

According to [the docs](https://recharts.org/en-US/api/YAxis), the Y axis does not accept an angle, so I didn't implement it. Plus, for my work, I only need this implemented for the X axis, so the lack of a practical use case also led me not to implement it. 

I can take a look, if it's not too hard, but wouldn't that require an `angle` prop in `YAxis` component? Do we want that?

> What is strange is, that with a rotation of 1 degree, first all ticks disappear. That seems to be a bug in this implementation. I did not pin it down in the code, but it seems a bug in the video:

That is indeed weird. Tomorrow I'll take a look and let you know my findings.

nikolasrieble:
> > First: Great contribution. ❤️ This was a missing puzzle piece, and it very well appreciated.
> 
> Thank you 😄
> 
> Thanks for your detailed writeup, @nikolasrieble! I'll address you comments below.
> 
> > 1 - It seems this does not work correctly with a vertical axis - but it also seems it did not work before - so that is okay.
> 
> According to [the docs](https://recharts.org/en-US/api/YAxis), the Y axis does not accept an angle, so I didn't implement it. Plus, for my work, I only need this implemented for the X axis, so the lack of a practical use case also led me not to implement it.
> 
> I can take a look, if it's not too hard, but wouldn't that require an `angle` prop in `YAxis` component? Do we want that?
> 
> > What is strange is, that with a rotation of 1 degree, first all ticks disappear. That seems to be a bug in this implementation. I did not pin it down in the code, but it seems a bug in the video:
> 
> That is indeed weird. Tomorrow I'll take a look and let you know my findings.

No need to look into the yAxis angle now. Good point.

bernardobelchior:
Thanks for your comments! I've found the issue. 

The algorithm I was using didn't account for an edge case. 

As you can see in the image below, the sole contributor to the red line's width is the rectangle's height. In that case, dividing the height by the sine of the angle is enough. 

image

However, at a certain angle, the sole contributor to the red line's width is actually the rectangle's width. It was this case that I forgot to account for.

image

In this case, the red line's width is the rectangle's width divided by the cosine of the angle.

----

Now, how do we decide which angle to use? 

That is dependent on the ratio of height to width of the rectangle. To determine that threshold we use the inverse tangent (`Math.atan`). If `angle > threshold && angle < 180 - threshold`, then we know that the contributor to the red line width is height, so we will use the first formula. If not, then we'll use the second one. 


From my experiment, it seems to fix the problems you found, @nikolasrieble. However, if you could test it again, I'd appreciate it 🙏 

Thanks!

ckifer:
Can confirm that this fixes the behavior based on doing the same test as @nikolasrieble above and doing my own test with the SB story added in your branch.

Awesome improvement overall - I wish there wasn't so many conditionals in `getTicksStart` and `getTicksEnd`, but that mess was already there.

ckifer:
Going to merge! This won't be released until next minor release

PR #3589, opened 2023-05-17
Storybook: Show code for Area component via StorySource in Examples and via Source in API 

nikolasrieble:
### The goal: 
We want to show the code in Storybook and enable users to copy paste both examples, as well as the result of using controls directly into their applications. 

### The tools: 
Storybook offers two ways to show code:

- [Storysource](https://storybook.js.org/addons/@storybook/addon-storysource) shows the complete code of the story. This works does not work well with using controls. 

- [Source](https://storybook.js.org/docs/react/api/doc-block-source#page-top) shows the code that is used for rendering the story. Works well with controls (i.e. does not show control setup, but only the value selected in the control). 

### The proposal:
Instead of using only one, lets use both, each where it fits best: 

- API  --> One story, in an MDX file, with an ArgsTable, showing the code via Source
- Examples  --> Multiple complex stories possibly with external dependencies, no MDX, no ArgsTable, showing the code via Storysource

### The result:

In this PR I implemented the proposal for the component `Area`:

We now have a single API story for the Area component, with an ArgsTable that shows all controls. The code can be shown in the MDX. 

Screenshot 2023-05-21 at 15 23 49
Screenshot 2023-05-21 at 15 24 10

In the Examples section, all Area usage stories went which are not easily usable with controls such as where functions are passed as arguments (for custom rendering of label/dot/etc.). Here, the story have no doc, and the code is visible in the story. 

Screenshot 2023-05-21 at 15 24 26

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/API/cartesian/Area.mdx
+++ b/storybook/stories/API/cartesian/Area.mdx
@@ -1,7 +1,13 @@
-import { ArgTypes, Meta, Stories } from '@storybook/blocks';
+import { ArgTypes, Meta, Story, Canvas, Controls } from '@storybook/blocks';
 import * as ComponentStories from './Area.stories';
 
 # Area
+
+<Meta of={ComponentStories} />
+
+<Canvas>
+  <Story name="API" of={ComponentStories} />
+</Canvas>
 
 ## Parent Component
 The Area can be used within a `<ComposedChart/>` or a `<AreaChart/>`.
@@ -10,11 +16,14 @@ The Area can be used within a `<ComposedChart/>` or a `<AreaChart/>`.
 The Area can be used with the following child components: `<LabelList/>`
 
 ## Properties
+The following table shows a complete list of all properties of the Area.
 Properties in the groups Other and Internal are not recommended to be used.
 
-<Meta of={ComponentStories} />;
-<ArgTypes of={ComponentStories} sort={'requiredFirst'} />;
-<Stories />;
+<Controls of={ComponentStories.API} />
+
+
+
+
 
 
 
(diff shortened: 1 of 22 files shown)


ckifer:
Just to clarify, this adds all of the stories from the `.stories` files to the bottom of the docs in a row?

Its better than not having them there at all (we can see the code now), but I think a separate file would be confusing if it was just some duplicated stories in a row. Maybe lets add them in this format to start and see how we feel about it for a little bit

nikolasrieble:
> Just to clarify, this adds all of the stories from the `.stories` files to the bottom of the docs in a row?
> 
> Its better than not having them there at all (we can see the code now), but I think a separate file would be confusing if it was just some duplicated stories in a row. Maybe lets add them in this format to start and see how we feel about it for a little bit


Exactly. Stories need to be shown in a Canvas in an MDX document in order for the code to be shown. Do you have an alternative suggestion as to how this should be structured? 🤔

nikolasrieble:
Lets not merge this yet. I want to look into the https://storybook.js.org/addons/@storybook/addon-storysource as an alternative before.

nikolasrieble:
Updated this based on the discussions in slack. 
Looking for feedback: @ckifer @Yilun-Sun @proke03

storybook/stories/API/cartesian/Area.mdx
```diff
@@ -1,15 +1,31 @@
...
+<Controls of={ComponentStories.API} />
+
+
+
+
+
+
+
+
+
```
nikolasrieble, line 31:
```suggestion








```

nikolasrieble:
Next steps: If we agree on this pattern, I shall roll it out across the whole API section - component per component.

ckifer (approved):
Looks great! I think this pattern will be the best bet. Allows for great readability on the API side as well as great copy-paste resources / views of source for those trying to replicate an example. Thank you @nikolasrieble!

nikolasrieble:
> Looks great! I think this pattern will be the best bet. Allows for great readability on the API side as well as great copy-paste resources / views of source for those trying to replicate an example. Thank you @nikolasrieble!

Great! I will then spend the next weekend 😅 (lets see whether that is enough) to migrate all existing stories to this pattern and update the README accordingly.

Issue #3591, opened 2023-05-19
className not propagated to axis components


mitrotasios:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Currently, passing a className to the axis components is not possible, since the className is not appended in the `renderAxis` method when the `axisOptions` are forwarded to the `CartesianAxis` component [here](https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.tsx#L1831).

This behaviour prohibits styling axis components with CSS classes.

### What does the proposed API look like?
There are no changes to the API, the className would merely be appended to the `CartesianAxis` component.

severinlandolt:
+1

ckifer:
this is merged and released in the latest alpha version. Will be included with next release, resolving!

PR #3592, opened 2023-05-19
fix: propagate className to CartesianAxis

mitrotasios:
This PR propagates the `className` prop of the `XAxis` and `YAxis` components to the root of the internal `CartesianAxis` component by appending it to its `className`.

## Description
Currently, passing a `className` to the axis components is not possible, since the `className` is not appended in the `renderAxis` method when the `axisOptions` are forwarded to the `CartesianAxis` component [here](https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.tsx#L1831).

This PR appends the `className` value from the `XAxis` and `YAxis` components to the `CartesianAxis`'s `className`.

## Related Issue

#3591

## Motivation and Context

The current behaviour prohibits styling axis components with CSS classes. Given that our library is using tailwind for styling, this change is necessary for us to enable global theming.

## How Has This Been Tested?

* tested the forwarding of classNames in storybook by inspecting the rendered HTML.
* ran unit tests to make sure this does not break other behaviours.

image

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1828,10 +1828,7 @@ export const generateCategoricalChart = ({
       return (
         <CartesianAxis
           {...(axisOptions as any)}
-          className={classNames(
-            `recharts-${axisOptions.axisType} ${axisOptions.axisType}`,
-            (axisOptions as any).className,
-          )}
+          className={classNames(`recharts-${axisOptions.axisType} ${axisOptions.axisType}`, axisOptions.className)}
           key={element.key || `${displayName}-${index}`}
           viewBox={{ x: 0, y: 0, width, height } as any}
           ticksGenerator={this.axesTicksGenerator}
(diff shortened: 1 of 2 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -1828,7 +1828,10 @@ export const generateCategoricalChart = ({
       return (
         <CartesianAxis
           {...(axisOptions as any)}
-          className={`recharts-${axisOptions.axisType} ${axisOptions.axisType}`}
+          className={classNames(
+            `recharts-${axisOptions.axisType} ${axisOptions.axisType}`,
+            (axisOptions as any).className,
```
ckifer, line 1833: would love for less things to be cast to `any` rather than more. These types are a mess but would prefer `className` added explicitly to `axisOptions` rather than cast here

ckifer (changes requested):

ckifer (approved):
LGTM

PR #3603, opened 2023-05-24
add: ResponsiveContainer story

akamfoad:
## Description
Add ResponsiveContainer story and a simple example of how to use it.

## Related Issue

Follow up of #3596

## Motivation and Context

Document ResponsiveContainer and how its used.

## Types of changes

- [X] Documentation & Storybook

## Checklist:

### DIFF
--- a/storybook/stories/API/ResponsiveContainer.mdx
+++ b/storybook/stories/API/ResponsiveContainer.mdx
@@ -1,7 +1,13 @@
-import { ArgTypes } from '@storybook/blocks';
-import * as ResponsiveContainerStories from './ResponsiveContainer.stories';
+import { Meta, Story, Canvas, Controls } from '@storybook/blocks';
+import * as ComponentStories from './ResponsiveContainer.stories';
 
 # ResponsiveContainer
+
+<Meta of={ComponentStories} />
+
+<Canvas>
+  <Story name="API" of={ComponentStories} />
+</Canvas>
 
 This is a component that can be used to render charts that will be responsive to its parent.
 When you render a chart you have to specify its **width** and **height** for the chart to be
@@ -31,8 +37,8 @@ and only then, renders the child Chart and passes the calculated width to it. So
 rendering a Chart inside ResponsiveContainer, you can omit its width and height and specify
 those on the ResponsiveContainer.
 
-If you're using ResponsiveContainer and you're rendering on the server (SSR) as most JavaScript
-frameworks do, then initially on the Browser, the chart will not be shown until after hydration.
+If you're using ResponsiveContainer and you're rendering on the server (SSR), then initially on
+the Browser, the chart will not be shown until after hydration.
 As mentioned, this is becase while rendering on server, the dimensions cannot be calculated,
 but if you need to have the chart rendered before hydration, you can pass `initialDimension`
 prop, which is an object like `{ width: 500, height: 500 }`, when passed, the chart will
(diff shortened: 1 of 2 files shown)


storybook/stories/API/ResponsiveContainer.mdx
```diff
@@ -0,0 +1,66 @@
...
+like [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to
+first calculate what the percentage equals to in terms of its parent element. For example
+if you specify `width="80%"`, ResponsiveContainer should first calculate the width of the
+parent, say its `720px`, now it has to calculate 80 percent of that 720 which will be **576px**
+and only then, renders the child Chart and passes the calculated width to it. So when you're
+rendering a Chart inside ResponsiveContainer, you can omit its width and height and specify
+those on the ResponsiveContainer.
+
+If you're using ResponsiveContainer and you're rendering on the server (SSR) as most JavaScript
+frameworks do, then initially on the Browser, the chart will not be shown until after hydration.
```
ckifer, line 35: Nit: "as most JavaScript frameworks do" can probably be removed. I would still say most JavaScript frameworks/libraries render on the client as opposed to the server. Can just call out SSR but not be specific

storybook/stories/API/ResponsiveContainer.mdx
```diff
@@ -0,0 +1,66 @@
...
+<ResponsiveContainer width="100%" height={400} initialDimension={{ width: 520, height: 400 }}>
+  <LineChart {...args}>
+    <Line dataKey="uv" />
+  </LineChart>
+</ResponsiveContainer>
+```
+
+
+> Note: Try to pass initial dimensions with accurate numbers, otherwise you might add some
+> [content layout shift](https://web.dev/cls) that can be annoying for your users.
```
ckifer, line 52: nit nit nit: wording, change `that can` to `which could`?

storybook/stories/API/ResponsiveContainer.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+## Parent Component
+
+It can be used anywhere you would render another React component.
+
+## Child component
+
+The ResponsiveContainer can be used to render all types of charts
```
ckifer, line 60: nit nit: wording, `all types of charts` -> `recharts charts as well as any child that needs to know the dimensions of its parent container`

storybook/stories/API/ResponsiveContainer.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+## Properties
+
+Properties in the groups Other and Internal are not recommended to be used.
+
+<ArgTypes of={ResponsiveContainerStories} sort={'requiredFirst'} />
```
ckifer, line 66: We're following a bit of a slightly different pattern now, can you follow this example?

https://github.com/recharts/recharts/blob/master/storybook/stories/API/cartesian/Area.mdx?plain=1#L6-L22

ckifer (changes requested):
Few nits and one change requested - we want to follow the linked pattern I think as opposed to strictly using ArgTypes

akamfoad:
@ckifer  Thanks for the suggestions, I think its ready for another round.

ckifer (approved):
LGTM, I would just say that in general this is a lot of text. Text tends to become outdated over time so if we make changes to behavior we need to make sure we read the docs back to ourselves and ensure the things we say here are still true.

Issue #2585, opened 2021-07-04
ErrorBar doesn't work with stacked bars


danvc:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/line-chart-specified-domain-forked-h3k8v?file=/src/App.tsx)

### Steps to reproduce
ErrorBar works perfectly while the bar is not stacked with others.

- In the source code mentioned as example, we can see 2 charts;
- Both charts contains ErrorBars;
- The chart in the bottom is not stacked.

### What is expected?
The ErrorBar should be displayed to the bar which is connected to.

### What is actually happening?
After stacking the bars, the ErrorBar disappear

| Environment | Info |
|---|---|
| Recharts | v2.0.8 |
| React | 17.0.1 |
| System | MacOS 11.4 |
| Browser | Firefox 89.0.2 |

ckifer:
Looks like this is still an issue with current recharts 2.6.2. Labeled and new reproduction link:
https://codesandbox.io/s/line-chart-specified-domain-forked-e7xs7u?file=/src/App.tsx:2628-2650

ckifer:
Released in 2.7

PR #3612, opened 2023-06-04
fix: check for stacked data in barchart errorbar dataPointFormatter

ckifer:
## Description

When using `ErrorBar` in a stacked bar chart, the error bars do no appear. Refactor types to correctly reflect what could happen. Fix bug in the `Bar` `dataPointFormatter`

This _might_ be considered a breaking change because of the type changes of `dataPointFormatter`? I don't really know though... I would have to create an example of how it would break someone to see how intrusive it might be. If it does break someone it will direct them to the correct way of using `dataPointFormatter`...

## Related Issue

https://github.com/recharts/recharts/issues/2585

## Motivation and Context

Fix a bug that has been around for a while. It is caused by this line of code https://github.com/recharts/recharts/blob/master/src/cartesian/Bar.tsx#L235 - this causes `value` to be a `[number, number]` when a chart is stacked.

## How Has This Been Tested?

Uncover the bug via failing unit test (no x values in rendered ErrorBar lines pre-change) and refactor until the test passes. 

Before (see left side):
image

After:
image

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -7,7 +7,7 @@ import Animate from 'react-smooth';
 import _ from 'lodash';
 import { Rectangle, Props as RectangleProps } from '../shape/Rectangle';
 import { Layer } from '../container/Layer';
-import { ErrorBar, Props as ErrorBarProps } from './ErrorBar';
+import { ErrorBar, Props as ErrorBarProps, ErrorBarDataPointFormatter } from './ErrorBar';
 import { Cell } from '../component/Cell';
 import { LabelList } from '../component/LabelList';
 import { uniqueId, mathSign, interpolateNumber } from '../util/DataUtils';
@@ -36,8 +36,8 @@ import {
 } from '../util/types';
 import { ImplicitLabelType } from '../component/Label';
 
-interface BarRectangleItem extends RectangleProps {
-  value?: number;
+export interface BarRectangleItem extends RectangleProps {
+  value?: number | [number, number];
   /** the coordinate of background rectangle */
   background?: {
     x?: number;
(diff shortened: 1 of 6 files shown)


src/cartesian/Bar.tsx
```diff
@@ -430,14 +430,19 @@ export class Bar extends PureComponent<Props, State> {
 
     const offset = layout === 'vertical' ? data[0].height / 2 : data[0].width / 2;
 
-    function dataPointFormatter(dataPoint: BarRectangleItem, dataKey: Props['dataKey']) {
+    const dataPointFormatter: ErrorBarDataPointFormatter = (dataPoint: BarRectangleItem, dataKey) => {
+      /**
+       * if the value coming from `getComposedData` is an array then this is a stacked bar chart.
+       * arr[1] represents end value of the bar since the data is in the form of [startValue, endValue].
+       * */
```
nikolasrieble, line 437: Thank you for the explanation! ❤️

src/cartesian/ErrorBar.tsx
```diff
@@ -7,20 +7,28 @@ import { Props as XAxisProps } from './XAxis';
...
+  entry: BarRectangleItem | LinePointItem | ScatterPointItem,
+  dataKey: DataKey<any>,
+) => ErrorBarDataItem;
+
 interface InternalErrorBarProps {
   xAxis?: Omit<XAxisProps, 'scale'> & { scale: D3Scale<string | number> };
   yAxis?: Omit<YAxisProps, 'scale'> & { scale: D3Scale<string | number> };
   data?: any[];
   layout?: 'horizontal' | 'vertical';
-  dataPointFormatter?: (entry: any, dataKey: DataKey<any>) => ErrorBarDataItem;
```
nikolasrieble, line 23: If we define a breaking change as one that can breaks a users TS project, removing `any` will always be breaking.

storybook/stories/Examples/BarChart.stories.tsx
```diff
@@ -92,6 +93,39 @@ export const Stacked = {
   },
 };
 
+const pvErrorData = pageData.map(d => ({ ...d, pvError: [100, 200] }));
```
akamfoad, line 96: Added this 
```js
const pvErrorData = pageData.map(d => ({ ...d, uvError: [100, 200], pvError: [100, 200] }));
```
and then add an ErrorBar to the first Bar too and this is how it looks, is this expected?

image

ckifer: Ehh, yeah we could make a separate issue for this probably. Most likely this was also what happened before this bug was introduced. At least now we can get error bars on stacked bars at all. In general I think they would mostly be used on the end bars

akamfoad: Alright, great work, that was the only concern, will approve.

src/cartesian/ErrorBar.tsx
ckifer: yeah true that, if we get complaints we'll just have to say that its for the best of the project as we increase type safety

Issue #2976, opened 2022-09-20
Type for Tooltip formatter function limits return type to input type


dhulme:
## Do you want to request a feature or report a bug?
Bug

## What is the current behavior?
Type for Tooltip formatter function limits return type to input type:
`(value: TValue, name: TName ....) => [TValue, TName] | TValue;`

## What is the expected behavior?
For formatting labels you may want to return a `string` when a `number` is the input. 

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Change was introduced by https://github.com/recharts/recharts/pull/2916 in 2.1.14. Prior to this version returning a different format to the input value was supported.

natBizitza:
Having the same issue. I temporally allowed a union type there.

PR #3618, opened 2023-06-08
refactor: `DefaultTooltipContent` to be functional component

akamfoad:
## Description

1. The DefaultTooltipComponent is now a functional component.
2. The result of formatter function is either
   `[React.ReactNode, React.ReactNode]` or `React.ReactNode`, instead of
   being the same type as the inputs, because inputs can be transformed
   and the result of the format is eventually is rendered in the JSX, so
   its meaningful to le the user return anything that react accepts for
   rendering.
3. A story is added for a case of Tooltips that use click to show
    tooltip instead of hover with pointer.

## Related Issue

Fixes: #2976

## Types of changes

## Checklist:

### DIFF
--- a/src/component/DefaultTooltipContent.tsx
+++ b/src/component/DefaultTooltipContent.tsx
@@ -2,7 +2,7 @@
  * @fileOverview Default Tooltip Content
  */
 import _ from 'lodash';
-import React, { PureComponent, CSSProperties, ReactNode } from 'react';
+import React, { CSSProperties, ReactNode } from 'react';
 import classNames from 'classnames';
 import { isNumOrStr } from '../util/DataUtils';
 
@@ -19,7 +19,7 @@ export type Formatter<TValue extends ValueType, TName extends NameType> = (
   item: Payload<TValue, TName>,
   index: number,
   payload: Array<Payload<TValue, TName>>,
-) => [TValue, TName] | TValue;
+) => [React.ReactNode, React.ReactNode] | React.ReactNode;
 
 export interface Payload<TValue extends ValueType, TName extends NameType> {
   type?: TooltipType;
@@ -50,21 +50,24 @@ export interface Props<TValue extends ValueType, TName extends NameType> {
   itemSorter?: (item: Payload<TValue, TName>) => number | string;
 }
 
-export class DefaultTooltipContent<TValue extends ValueType, TName extends NameType> extends PureComponent<
-  Props<TValue, TName>
-> {
-  static displayName = 'DefaultTooltipContent';
+export const DefaultTooltipContent = <TValue extends ValueType, TName extends NameType>(
+  props: Props<TValue, TName>,
+) => {
+  const {
+    separator = ' : ',
+    contentStyle = {},
+    itemStyle = {},
+    labelStyle = {},
+    payload,
+    formatter,
+    itemSorter,
+    wrapperClassName,
+    labelClassName,
+    label,
+    labelFormatter,
+  } = props;
 
-  static defaultProps = {
-    separator: ' : ',
-    contentStyle: {},
-    itemStyle: {},
-    labelStyle: {},
-  };
-
-  renderContent() {
-    const { payload, separator, formatter, itemStyle, itemSorter } = this.props;
-
+  const renderContent = () => {
     if (payload && payload.length) {
       const listStyle = { padding: 0, margin: 0 };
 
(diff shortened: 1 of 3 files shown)


src/component/DefaultTooltipContent.tsx
```diff
@@ -19,7 +19,7 @@ export type Formatter<TValue extends ValueType, TName extends NameType> = (
   item: Payload<TValue, TName>,
   index: number,
   payload: Array<Payload<TValue, TName>>,
-) => [TValue, TName] | TValue;
+) => [React.ReactNode, React.ReactNode] | React.ReactNode;
```
ckifer, line 22: `[TValue, TName] | React.ReactNode;` would be more appropriate no? Based on the conditionals below, if formatter returns an array the name will not be rendered if it is not a string or a number. If someone returns a react element in the array `[<div>{name}</div>, <div>{value}</div>]` only value will be rendered

src/component/DefaultTooltipContent.tsx
```diff
@@ -81,21 +84,24 @@ export class DefaultTooltipContent<TValue extends ValueType, TName extends NameT
...
         }
+
         return (
           // eslint-disable-next-line react/no-array-index-key
           <li className="recharts-tooltip-item" key={`tooltip-item-${i}`} style={finalItemStyle}>
-            {isNumOrStr(name) ? <span className="recharts-tooltip-item-name">{name}</span> : null}
-            {isNumOrStr(name) ? <span className="recharts-tooltip-item-separator">{separator}</span> : null}
-            <span className="recharts-tooltip-item-value">{value}</span>
+            {isNumOrStr(finalName) ? <span className="recharts-tooltip-item-name">{finalName}</span> : null}
+            {isNumOrStr(finalName) ? <span className="recharts-tooltip-item-separator">{separator}</span> : null}
```
ckifer, line 103: these checks prevent a react element from being rendered here

src/component/DefaultTooltipContent.tsx
```diff
@@ -50,21 +50,24 @@ export interface Props<TValue extends ValueType, TName extends NameType> {
...
-  renderContent() {
-    const { payload, separator, formatter, itemStyle, itemSorter } = this.props;
+export const DefaultTooltipContent = <TValue extends ValueType, TName extends NameType>(
+  props: Props<TValue, TName>,
+) => {
+  const {
+    separator = ' : ',
+    contentStyle = {},
+    itemStyle = {},
+    labelStyle = {},
```
ckifer, line 60: Just a note, anything that is called within a `cloneElement` within `generateCategoricalChart` has the potential to lose the values of their default props (we've discussed this issue elsewhere I believe I forget where atm though). I think we're good since this is within the function though https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.tsx#L1929-L1946

ckifer (changes requested):
some callouts/things to double check but nothing too major. Thanks!

src/component/DefaultTooltipContent.tsx
akamfoad: Since the value always rendered regardless of the type of name, isn't better to make it:

`[React.ReactNode, TName] | React.ReactNode` so the value type will be consistent?

src/component/DefaultTooltipContent.tsx
akamfoad: You're right, I think I should revert the name type back to `TName` type, I'd love your feedback here too
https://github.com/recharts/recharts/pull/3618#discussion_r1224190579

src/component/DefaultTooltipContent.tsx
akamfoad: Yeah I believe the issue was with `defaultProps` property on the component, but since these are in the function are fine.

Its a bit hard to apply this style of defaultProps since the `props` object has been passed around a lot in all components, its better to make the functions only receive the props they need in new refactors (I'd do them as I got time if no one picked it up).

src/component/DefaultTooltipContent.tsx
ckifer: I guess, but if someone is providing a ReactNode imo they would just use only a react node and do

```
<>
  {myName}: {myValue}
</>
```

instead of

`[<div>{someValue}</div>, 'some name'`]`

ckifer: but the suggested change is also fine, it removes my concern

akamfoad: The motivation for changing this was raised in #2976 you don't need to be bound to the input type, and while the `TName` keeps the name in check, the value can be more flexible since it can be anything from users perspective, but here the types keep you from doing passing back anything outside of inputs scope, I say ReactNode because the value is rendered as its, it enables people to do whatever they need since it doesn't concern recharts what the rendered value is.

ckifer (approved):
Thanks @akamfoad 🚀

Issue #3387, opened 2023-02-21
Tooltip not getting overlapping values if XAxis is set to scale time


bandinopla:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-jz8yup?file=/src/index.js)

### Steps to reproduce
rendering 2 Line components with 2 distinct data sets, using the X axis as time, if both overlap, and you set a tooltip, the tooltip doesn't display the combined data on that point like it does when you set the XAxis to "category" with allowDuplicates = false.

### What is expected?
if i put the mouse on an X Point, i will expect the tooltip to be provided with only the data on that X Point. 

### What is actually happening?
i get the data from one line in the tooltip then if i move the mouse a little i see the other overlapping point data.

| Environment | Info |
|---|---|
| Recharts | v2.1.2 |
| React | ^17.0.2 |
| System | Windows |
| Browser | Brave v1.48.167 |

ckifer:
Can you post a screenshot of what you mean? Its not clear based on the issue description.

Are you talking about how the example is including `uv` from `data` in the tooltip as opposed to nothing?

![image](https://user-images.githubusercontent.com/25180830/220430610-8f49f8b6-a10f-4f98-b22e-8785cc586968.png)

bandinopla:
@ckifer yes, if i hover the cursor over an X Point that has only 1 point in it, i would expect the "payload" from the tooltip to only contains the points in that position. This works fine if you use an XAxis of type category, but when using a numeric X Axis for some reason the tooltip only gets one value at a time. 

One quick fix im doing right now is, when i hover, in the custom tooltip, i get the label of the XAxis which in my case is a timestamp, then i use that to manually relocate all points in all series that have that same X value and build the tooltip from that data. 

I was expecting to get this automatically vía the payload property, but i always get all series even if they dont contain any point in the X position and the payload of those irrelevant series contain also a payload pointing to a point that is on a diferent X...

dkstrong:
I'm having the same issue, I believe the core problem is that its trying to find the other payload items by their index. however since each data set has  different number of items you can't use the index to match/compare them. I think it needs to be looking at the dataKey for the x axis to figure out the matching payloads.

dusan-w3lab:
[...]

PR #3621, opened 2023-06-12
Tooltip multiple datasource

GMer78:
Increased tooltip compatibility with multiple data sources

## Description

In function "getTooltipContent", after getting the value selected by the cursor via and activeIndex, the program runs a "find" function on the array of data to make sure it matches the tooltip selection.
A similar approach is taken in the function "renderGraphicChild" which highlights selected datapoint.

Added allowSelectNearestValue boolean to YAxis and XAxis, which will look for the nearest value if the exact matching datapoint can not be found in the two aformentioned functions. This is particularly useful with time series where datapoints can be very close yet not overlapping.

Changed how tooltip content generation is handled. Code used to render the title, and the items seperately. As allowSelectNearestValue now allows values which are close but different to be shown in the tooltip, changes have been made for grouping to be handled. The code groups items by titles, to output multiple title/items if multiple titles exist.

## Related Issue
 Fixes: [#3387](https://github.com/recharts/recharts/issues/3387)

## Motivation and Context

Changes stem from recharts handling of multiple datasets on the same graph. I believe recharts used to need a single array of objects as input data. Later it allowed data to be input on on a ```<Line/>``` per ```<Line/>``` basis.

When working with a single array, tooltip content would be calculated by finding the nearest index, and extracting all the items associated with that index. This clearly breaks down in the presence of multiple data inputs as the value associated with the index for one dataset will most likely not match the value associated that same index in the other dataset.

This brings along the related feature of finding the nearest point to the tooltip, if an exact match isn't found.

I think there can be fundamental opinions in terms of whether finding the nearest value is a parameter of the tooltip, or whether it the axis which has more "linear" categories. Applying this characteristic to the axis seemed to make more sense to me. 

## How Has This Been Tested?

Added data similar to the issue at the bottom of the demo page, and demonstrated behaviour as expected. The other examples in the demo page looked unaffected

## Types of changes

## Checklist:

Potentially a change needed to document allowSelectNearestValue if accepted.


ckifer:
Hey @GMer78 thanks for the contribution.

Will review this in more detail, but we're going to need to be a bit more strict about testing changes like this. We are no longer using the demo project and instead using storybook - can you by chance add a storybook example to the tooltip section of the storybook which demonstrates these changes? 

This is a large change and large changes which have been merged as of the past while have introduced unintended breaking changes (even when from code review it doesn't seem like they would) due to the state of recharts code (which I'm sure you've seen is messy, if not gross). That being said it might be a while before accepting something like this if not need to go through a few rounds of implementation changes. My goal rather than shoehorning new features into the existing bad code, would be to make the code better in general first so its easier to understand and contribute to. 

If we merged this as is, documented a new property, then went to refactor in a few months, we may unintentionally break this (or other) functionality because we didn't totally understand every inch of what was happening in these large, multi-branch functions if that makes sense. Not saying I don't want new features and functionality, but breaking consumers at random is a pattern I'd like to break / avoid.

That said, lets review this and discuss what can be done from there

GMer78:
Hi @ckifer,

I have updated the storybook. Under the Examples/TimeSeries/Custom Multiple Source Tooltip, you should be able to see/modify allowSelectNearestValue and see why it might be important for some time-series.

I entirely agree with not wanting to breaking previous versions of the software with this update. I think that to a large degree this issue came about because multi-data source inputs has been added without considering all the ramifications. I definitely that unifying how that is/should be handled would be beneficial to avoid things breaking moving forward.

With that said, I do not mind this pull request being paused until the handling of multiple data sources is figured out. I do think that moving away from index based logic might be required for things to be elegantly handled.

However it seems that this PR solved an issue that at least 3 people (me included) has run into, so I figured my fix for it might help people out although it might not immediately be integrated in a recharts release.

src/cartesian/XAxis.tsx
```diff
@@ -49,4 +49,5 @@ XAxis.defaultProps = {
   scale: 'auto',
   reversed: false,
   allowDuplicatedCategory: true,
+  allowSelectNearestValue: false,
```
ckifer, line 52: nit: I know we already have `allowDuplicatedCategory` and `allowDataOverflow`, but this might be better as just `selectNearestValue`

src/chart/generateCategoricalChart.tsx
```diff
@@ -238,6 +239,21 @@ const getTooltipContent = (
       // graphic child has data props
       const entries = data === undefined ? displayedData : data;
       payload = findEntryInArray(entries, tooltipAxis.dataKey, activeLabel);
+    } else if (tooltipAxis.dataKey && tooltipAxis.allowSelectNearestValue && tooltipAxis?.categoricalDomain) {
```
ckifer, line 242: nit: you destructure `dataKey` above but use `tooltipAxis.dataKey` here. Could maybe use destructured values for all three checks here?

src/chart/generateCategoricalChart.tsx
```diff
@@ -238,6 +239,21 @@ const getTooltipContent = (
...
+      const entries = data === undefined ? displayedData : data;
+      const value = tooltipAxis?.categoricalDomain?.[activeIndex];
+
+      const matchingData: any = entries.reduce((acc: any, entry: any) =>
```
ckifer, line 246: what is `acc` here? can you name that variable a bit better?

Would also love to avoid more casts and `any` where possible (I know its tough in this code, but if we don't enforce it it will keep getting worse)

src/chart/generateCategoricalChart.tsx
```diff
@@ -238,6 +239,21 @@ const getTooltipContent = (
...
+        Math.abs(value - entry[dataKey as string]) < Math.abs(value - acc[dataKey as string]) ? entry : acc,
```
ckifer, line 247: why does `dataKey` need to be cast?

src/chart/generateCategoricalChart.tsx
```diff
@@ -750,6 +768,9 @@ export interface CategoricalChartState {
   /** active tooltip payload */
   activePayload?: any[];
 
+  /** datakey queried for tooltip */
+  activeDataKey?: any;
```
ckifer, line 772: can we accurately type this?

src/chart/generateCategoricalChart.tsx
```diff
@@ -2087,6 +2109,19 @@ export const generateCategoricalChart = ({
               : 'payload.'.concat(tooltipAxis.dataKey.toString());
           activePoint = findEntryInArray(points, specifiedKey, activeLabel);
           basePoint = isRange && baseLine && findEntryInArray(baseLine, specifiedKey, activeLabel);
+        } else if (tooltipAxis.dataKey && tooltipAxis.allowSelectNearestValue && tooltipAxis?.categoricalDomain) {
```
ckifer, line 2112: this seems like the same logic as above? Can we modularize a bit better and extract some logic into a helper function?

src/chart/generateCategoricalChart.tsx
```diff
@@ -1934,14 +1955,15 @@ export const generateCategoricalChart = ({
...
-      const { isTooltipActive, activeCoordinate, activePayload, activeLabel, offset } = this.state;
+      const { isTooltipActive, activeCoordinate, activePayload, activeLabel, offset, activeDataKey } = this.state;
 
       return cloneElement(tooltipItem, {
         viewBox: { ...offset, x: offset.left, y: offset.top },
         active: isTooltipActive,
         label: activeLabel,
         payload: isTooltipActive ? activePayload : [],
         coordinate: activeCoordinate,
+        tooltipDatakey: activeDataKey,
```
ckifer, line 1966: casing inconsistency `tooltipDatakey` vs `activeDataKey`
```suggestion
        tooltipDataKey: activeDataKey,
```

src/component/DefaultTooltipContent.tsx
```diff
@@ -46,6 +46,7 @@ export interface Props<TValue extends ValueType, TName extends NameType> {
   labelStyle?: CSSProperties;
   labelFormatter?: (label: any, payload: Array<Payload<TValue, TName>>) => ReactNode;
   label?: any;
+  tooltipDatakey?: any;
```
ckifer, line 49: casing inconsistency as mentioned above + please type if at all possible

src/component/DefaultTooltipContent.tsx
```diff
@@ -63,56 +64,107 @@ export class DefaultTooltipContent<TValue extends ValueType, TName extends NameT
...
+        },
+
+        labelGroupedPayload => {
+          return _.map(labelGroupedPayload, (payloadPayload, payloadLabel) => {
+            // Build label from datakey value
+            const hasLabel = !_.isNil(payloadLabel);
+            let finalLabel: ReactNode = hasLabel ? payloadLabel : '';
+            const labelCN = classNames('recharts-tooltip-label', labelClassName);
+
+            if (hasLabel && labelFormatter && payload !== undefined && payloadPayload !== null) {
```
ckifer, line 100: logic was already here but can use loose equals to cover both undefined and null in one go
```suggestion
            if (hasLabel && labelFormatter && payloadPayload != null) {
```

src/util/types.ts
```diff
@@ -1110,6 +1110,13 @@ export interface BaseAxisProps {
   label?: string | number | ReactElement | object;
   /** The HTML element's class name */
   className?: string;
+  /**
+   * When the cursor selects a value on the axis which is not present on another dataset,
+   * allow the nearest value to be selected instead.
+   */
+  allowSelectNearestValue?: boolean;
+
+  categoricalDomain?: any;
```
ckifer, line 1119: what is this and why is it needed? Can it be typed?

ckifer (changes requested):
Some feedback - would really love if we could be stricter on types (even if previously the code wasn't)

ckifer:
Also found a breaking change - not sure where it is in the code but if you look at the storybook on master [here](https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/story/examples-tooltip--locked-by-click) and then view the same story locally with your changes the labels are now `undefined` which is an unintended breaking change

src/chart/generateCategoricalChart.tsx
GMer78: Agreed on the variable naming.

I think I would need help figuering out how to avoid ```any``` for these variables. This is a single slice of data from the data array. The data array is of type ```any[]```, to allow the user to input any sort of array of data. Please let me know if you have any thoughts on this.

src/chart/generateCategoricalChart.tsx
GMer78: If is the same logic as above, since finding the nearest active point for the tooltip, and finding the nearest active point to visually highlight on the plot are at different parts of the code. 

In the future it probably would make sense to have these in similar places.

Happy to make 2 helper functions for each of these tasks.

src/util/types.ts
GMer78: categoricalDomain is an array of the min/max values of the axis. 

This looks like it is calculated if the axis is of type ```'number'```, or if the axis is a data.

(apparently this variable also gets set if for the axis the ```scale !== 'auto'```, but I am not sure I visualize what that would mean for non numerical/date data, which cannot be easily ordered.

Just thought about adding it in the types as it wasn't documented there yet. 

I am not sure if a refactor on axis is in the works, but this is definitely a variable which at first glance seems like it is not well defined.

GMer78:
Good catch on the breaking changes, I will have a look into it.

GMer78:
looking at the issues linked in [API/ComposedChart/Line Bar Area Scatter Time Scale](https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/story/api-chart-composedchart--line-bar-area-scatter-time-scale) , it looks like this is due to some inconsistencies in how chart properties are handled. 

in [src/chart/generateCategoricalChart.tsx](https://github.com/recharts/recharts/blob/f1c52c16139f501b7fadc5d2a8a2329a9aed61a8/src/chart/generateCategoricalChart.tsx#LL239C1-L239C65) there is this line : 
```const entries = data === undefined ? displayedData : data;```
So for line charts, and area charts, this should work as intended. However it looks like bar charts initialises data[] as an empty array ( see: [src/cartesian/Bar.tsx](https://github.com/recharts/recharts/blob/f1c52c16139f501b7fadc5d2a8a2329a9aed61a8/src/cartesian/Bar.tsx#LL53C1-L53C29) ). Line/area charts seem to use points[] instead.


I think fundamentally it would make sense to either move bar charts to use points[] as opposed to data[] to unify how charts can hold data. However this might be a breaking change for people using bar charts. 
I am happy to modify the line:
```const entries = data === undefined ? displayedData : data;``` to ```const entries = data === (undefined || []) ? displayedData : data;```
but this very much seems to be putting a band-aid over the problem rather than addressing a core design issue of the library. 

Let me know your thoughts on this. I think the best way forward is for me to create that fix, and for a rework of the structure of bar charts to be on the radar for the next big library rework.

File: src/chart/generateCategoricalChart.tsx (hunk 3 above)
Comment (ckifer) [reply src/chart/generateCategoricalChart.tsx:246]: yeah this is fine until Recharts implements generics

File: src/chart/generateCategoricalChart.tsx (hunk 6 above)
Comment (ckifer) [reply src/chart/generateCategoricalChart.tsx:2112]: right, they can still be called in the same place as they are but just extract the duplicate logic into an (easier to test) helper function)

File: src/chart/generateCategoricalChart.tsx (hunk 7 above)
Comment (ckifer) [reply src/chart/generateCategoricalChart.tsx:1966]: that is my preference since it is two separate words words

File: src/util/types.ts (hunk 10 above)
Comment (ckifer) [reply src/util/types.ts:1119]: oh this was here before? sorry didn't see that 😅

**Comment by ckifer** [comment]
Hmm... Bar charts don't really have points so it makes sense to me why `data` was used. I don't really have enough knowledge of things to suggest one way or the other. 

Make the change for now to account for the empty array, we'll see what it looks like and go from there. Thanks!

File: src/chart/generateCategoricalChart.tsx (hunk 3 above)
Comment (nikolasrieble) [reply src/chart/generateCategoricalChart.tsx:246]: To be on the safe side then, I would suggest to replace `any` with `unknown`. It would cause all sorts of checks to be required, but we would be on the safe side.

File: src/chart/generateCategoricalChart.tsx (hunk 4 above)
Comment (nikolasrieble) [reply src/chart/generateCategoricalChart.tsx:247]: The dataKey can be an accessor function as well, right? Do we have an existing helper function to get the value from a row/entry using a dataKey?

File: src/chart/generateCategoricalChart.tsx (hunk 5 above)
Comment (nikolasrieble) [reply src/chart/generateCategoricalChart.tsx:772]: Or resort to `unknown` else.

File: src/chart/generateCategoricalChart.tsx (hunk 6 above)
Comment (nikolasrieble) [reply src/chart/generateCategoricalChart.tsx:2112]: Making helper function with intention revealing names would also help us not increase the complexity of this file further.

File: src/chart/generateCategoricalChart.tsx (hunk 4 above)
Comment (nikolasrieble) [reply src/chart/generateCategoricalChart.tsx:247]: I believe the function you would prefer to use here, you can find in `src/util/ChartUtils.ts`

```
export function getValueByDataKey<T>(obj: T, dataKey: DataKey<any>, defaultValue?: any) {
  if (_.isNil(obj) || _.isNil(dataKey)) {
    return defaultValue;
  }

  if (isNumOrStr(dataKey)) {
    return _.get(obj, dataKey, defaultValue);
  }

  if (_.isFunction(dataKey)) {
    return dataKey(obj);
  }

  return defaultValue;
}
```

nikolasrieble (commented):
This is a great contribution, unlocking user value, and will specifically enable me to throw away quite some wrappers I have built for this exact behaviour. 

- Please use `unknown` instead of `any`
- Please use the `getValueByDataKey` instead of treating the dataKey like a key to an object (could be a function)

ckifer:
If no updates to this soon I may close this as stale - theres been a decent amount of refactoring since these changes and I think they'll be easier to make later on than they currently are

nikolasrieble:
@GMer78 This is a highly valuable contribution, we would love to see this land. Should we take this over, or would you like to drive this to finish?

GMer78:
Hi, I have updated the code to account for the requested changes + the general changes the master branch has undergone. It's been a while since I dug into it, but I seem to remember that this ended up having strange behaviour on plots with multiple axis. Because of the large ammount of refactoring done on the master branch, I will have to see if this is still the case.

ckifer:
due to the sheer amount of merge conflicts and amount of refactoring done in both master and 3.x I'm going to close this as stale for now. I think this is a great candidate change for once we get Tooltip on context in 3.x

PR #3627, opened 2023-06-18
Address issue where tooltips and AccessibilityManager can get misaligned

julianna-langston:
Addresses bug with AccessibilityManager where, when the user interacts with the AccessibilityManager and *then* the `<Tooltip/>` component changes, there is a javascript runtime error. (Note that I could only reproduce this if the container chart used the `margin` prop.)

## Description

- I added error handling to prevent the javascript runtime error from throwing.
- TDD test

## Related Issue

@ckifer - Was the issue filed? 

For reference, the issue can be reproduced using this sandbox:
https://codesandbox.io/s/stacked-area-chart-forked-8ipx36?file=/src/App.tsx:951-1010

## Motivation and Context

Solves a javascript runtime error

## How Has This Been Tested?

I generated a reproducible example within the jest testing framework and in Storybook. Once I added in defensive checks for the error condition, I confirmed that this solved the error in both a browser and jest.

## Types of changes

## Checklist:

- [NA] My change requires a change to the documentation.
- [NA] I have updated the documentation accordingly.

### DIFF
--- a/src/chart/AccessibilityManager.ts
+++ b/src/chart/AccessibilityManager.ts
@@ -39,9 +39,8 @@ export class AccessibilityManager {
     this.offset = offset ?? this.offset;
     this.mouseHandlerCallback = mouseHandlerCallback ?? this.mouseHandlerCallback;
 
-    if (this.activeIndex >= this.coordinateList.length) {
-      this.activeIndex = this.coordinateList.length - 1;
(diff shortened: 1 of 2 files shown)


ckifer (approved):
Changes look good, thank you! and thanks for adding the test

Issue #3635, opened 2023-06-20
TypeError: Cannot read properties of null (reading 'clipDot')


zahir-dct:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Steps to reproduce
It comes from /static/node_modules/recharts/es6/cartesian/Line.js:350:31

### What is expected?
Previously the same worked, Now it's not working.

### What is actually happening?
Uncaught TypeError: Cannot read properties of null (reading 'clipDot')
    at a.value (Line.js:350:31)
    at Wa (react-dom.production.min.js:187:188)
    at Qa (react-dom.production.min.js:186:173)
    at Gs (react-dom.production.min.js:269:427)
    at Ic (react-dom.production.min.js:250:347)
    at jc (react-dom.production.min.js:250:278)
    at Pc (react-dom.production.min.js:250:138)
    at bc (react-dom.production.min.js:243:163)
    at react-dom.production.min.js:123:115
    at t.unstable_runWithPriority (scheduler.production.min.js:18:343)

| Environment | Info |
|---|---|
| Recharts | v2.3.2 |
| React | 17.0.2 |
| System | Windows/Linux |
| Browser | All |

---
"recharts": "^2.3.2"
By removing "^" from "^2.3.2", it works

ckifer:
Sorry this happened, will fix asap. This happens when? Can you make a sandbox to reproduce?

ckifer:
@zahir-dct also, `clipDot` does not exist in 2.3.2 - i'm not sure how you are using the property but it was only released in 2.7.0

zahir-dct:
> Sorry this happened, will fix asap. This happens when? Can you make a sandbox to reproduce?

https://codesandbox.io/s/vigilant-blackwell-gg5rkq?file=/package.json

ckifer:
`dot` doesn't have `null` as an acceptable type - if you use `undefined`, don't define it at all, or supply `DotProps` it will render correctly

ckifer:
I'll fix this by making sure null doesn't throw an error like this, but I don't consider this a breaking change since it follows the types defined since v2

ckifer:
Released in 2.7.2

PR #3637, opened 2023-06-21
fix: dont throw when dot is null, use typeguard in line and improve in area

ckifer:
## Description

When the `clipDots` PR was merged I added a typeguard to `Area` but I did not see to do the same for `Line`. If `dot` was something other than `DotProps` this cast is incorrect and wrong. 

Linked issue is caused when passing `null` to `dot`.  `null` is not an acceptable type to `dot` and therefore this wasn't a breaking change, but it still uncovered a need for this check.

## Related Issue

fixes https://github.com/recharts/recharts/issues/3635

## Motivation and Context

See issue, removes need to cast to `DotProps`

## How Has This Been Tested?

Added unit tests that reproduce the issue and are green after fixing it

N/A no visual changes

## Types of changes

- [Y] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [Y] My code follows the code style of this project.
- [N] My change requires a change to the documentation.
- [Y] I have added tests to cover my changes.
- [Y] All new and existing tests passed.

### DIFF
--- a/src/util/ReactUtils.ts
+++ b/src/util/ReactUtils.ts
@@ -1,10 +1,12 @@
 import _ from 'lodash';
 import React, { Children, Component, FunctionComponent, isValidElement, ReactNode } from 'react';
 import { isFragment } from 'react-is';
-
+import { DotProps } from '..';
 import { isNumber } from './DataUtils';
 import { shallowEqual } from './ShallowEqual';
 import { FilteredSvgElementType, FilteredElementKeyMap, SVGElementPropKeys, EventKeys } from './types';
+import { AreaDot } from '../cartesian/Area';
+import { LineDot } from '../cartesian/Line';
 
 const REACT_BROWSER_EVENT_MAP: Record<string, string> = {
   click: 'onClick',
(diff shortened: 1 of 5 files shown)


ckifer:
going to merge this to get a release out before I'm on vacation for a bit. Please review in retrospect for red-flags.

src/util/ReactUtils.ts
```diff
@@ -260,6 +262,9 @@ const SVG_TAGS: string[] = [
 
 const isSvgElement = (child: any) => child && child.type && _.isString(child.type) && SVG_TAGS.indexOf(child.type) >= 0;
 
+export const isDotProps = (dot: LineDot | AreaDot): dot is DotProps =>
+  dot && typeof dot === 'object' && 'cx' in dot && 'cy' in dot && 'r' in dot;
```
nikolasrieble, line 266: Nit: It might be nice going forward to start smaller files with smaller scope to move away from generic collect all files. 

This function for example could live in a separate file. It seems unrelated to all other functions. 

But that really is just personal preference.

nikolasrieble (commented):
👍 I see no issue, thank you for your quick reaction!

src/util/ReactUtils.ts
ckifer: Sure - to me at the time it seemed meta enough that 'ReactUtil' fit, but in reality its more of a TypeScript util or even a specific Dot util

Issue #3656, opened 2023-07-14
Server side rendering warning with useLayoutEffect


sieroaoj:
When rendering a chart with server side render we get this warning in version 2.7.2:

> Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
>     at Rectangle (app/node_modules/recharts/lib/shape/Rectangle.js:81:35)
> 

Switching to useEffect seems to solve the problem and the chart renders well without any warnings.

It seems that "useLayoutEffect is put forward as the safest upgrade path for class components":
https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a

Could we update the code to use useEffect instead?
https://github.com/search?q=repo%3Arecharts%2Frecharts%20useLayoutEffect&type=code

ckifer:
yeah I think this is fine

seems like the guidance from the react team is that most of the time it isn't needed. Difference in visual performance is negligible if any in this case

ckifer:
released in 2.8

PR #3657, opened 2023-07-14
Switches useLayoutEffect to useEffect

sieroaoj:
## Description

When rendering a chart on server side we get this warning:

Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
    at Rectangle (app/node_modules/recharts/lib/shape/Rectangle.js:81:35)

Switching to useEffect seems to solve the problem and the chart renders well without any warnings.

## Related Issue

https://github.com/recharts/recharts/issues/3656

## Motivation and Context

This change is required when rendering on server side with `ReactDOMServer.renderToString`

## How Has This Been Tested?

This change was tested by server side rendering a chart that used the rectangle shape.

## Types of changes

## Checklist:

### DIFF
--- a/src/shape/Rectangle.tsx
+++ b/src/shape/Rectangle.tsx
@@ -1,7 +1,7 @@
 /**
  * @fileOverview Rectangle
  */
-import React, { SVGProps, useLayoutEffect, useRef, useState } from 'react';
+import React, { SVGProps, useEffect, useRef, useState } from 'react';
 import classNames from 'classnames';
 import Animate from 'react-smooth';
 import { AnimationTiming } from '../util/types';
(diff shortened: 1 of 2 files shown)


ckifer:
@sieroaoj want to mark this as ready for review?

sieroaoj:
done 👍

nikolasrieble (approved):
I tested this in storybook and could not find any issues. Thank you for your contribution!

PR #3673, opened 2023-07-30
Storybook and Type Information: Move description next to type and reuse existing

nikolasrieble:
In the ideal setup, storybook use [automatic arg types inference](https://storybook.js.org/docs/react/api/arg-types#automatic-argtype-inference) via [react-docgen-typescript](https://github.com/styleguidist/react-docgen-typescript). 

We instead have mostly duplicated documentation by creating separate props documentation in the storybook folder, such as AnimationDuration. 

Instead, we should define our own types, and add a comment right above to describe/explain. To do so, requires us to define a type once, and reuse it elsewhere. 

For `animationDuration: number`, to not redundantly redefine the description per comment, I define an alias `type AnimationDuration = number` with a description above. Using this type everywhere, leads to the description being added to every component in the ArgsTable in storybook. 

This PR is a suggestion how to move forward and move documentation from storybook into the codebase. 

This PR only tackles two properties: 
- AnimationDuration
- AnimationTiming

If we agree this is a reasonable way forward, I will follow up with other properties. 

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/API/props/AnimationProps.ts
+++ b/storybook/stories/API/props/AnimationProps.ts
@@ -15,20 +15,12 @@ export const animationBegin = {
   },
 };
 export const animationDuration = {
-  description: 'Specifies the duration of animation, the unit of this option is ms.',
-  type: { name: 'number' },
-  defaultValue: 1500,
   table: {
-    defaultValue: { summary: '1500' },
     category: 'Animation',
   },
 };
 export const animationEasing = {
-  description: 'The type of easing function.',
-  type: { name: 'ease | ease-in | ease-out | ease-in-out | linear' },
-  defaultValue: 'ease',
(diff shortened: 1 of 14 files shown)


storybook/stories/API/props/AnimationProps.ts
```diff
@@ -15,20 +15,12 @@ export const animationBegin = {
   },
 };
 export const animationDuration = {
-  description: 'Specifies the duration of animation, the unit of this option is ms.',
-  type: { name: 'number' },
-  defaultValue: 1500,
```
nikolasrieble, line 20: Funny enough, this was misleading. It was correct for some components, but not for others.

ckifer (approved):
nice, LGTM!

Issue #3683, opened 2023-08-05
Allow disabling Sankey sorting


daformat:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
When using Sankey charts to create cashflow charts, we want to group expenses or income by category. The automatic sorting makes chart links overlap and cross each other, when we want to actually group things by topic to keep the cashflow analysis legible. The fact that one expense in one category is smaller than others in other categories shouldn't move the link around and mix it with other categories, as it makes the chart more complicated to read. The goal is to give control to the user on the grouping of the branch, while still providing with the option to sort data by default

### What does the proposed API look like?
I propose to add a boolean `sort` prop to the Sankey component, with a default value of true (to no break existing implementation), so the user can control how their chart will behave.
I've prepared the [code changes](https://github.com/recharts/recharts/compare/master...daformat:recharts:feat/allow-ignoring-sankey-sorting) for that, ready to submit a PR.

daformat:
Here is a visual example with the current sorting, and one without, to better explain the use-case:
With sorting | Without sorting
--|--
<img width="813" alt="image" src="https://github.com/recharts/recharts/assets/1694893/db511197-d045-4252-b83a-f093afd1bc3e"> | <img width="810" alt="image" src="https://github.com/recharts/recharts/assets/1694893/5941b06a-d2ec-460e-9bcd-47a20ccf4cc2">

daformat:
It also seems that the sorting doesn't really sort the data, at least not the way one would expect, as you can see on this example made via the storybook:
<img width="1135" alt="image" src="https://github.com/recharts/recharts/assets/1694893/6b6459fe-6ed0-40a3-9c93-90724d746755">

Reproduction data:
```json
{
  "nodes": [
    {
      "name": "Income"
    },
    {
      "name": "Budget"
    },
    {
      "name": "Investment"
    },
    {
      "name": "Real Estate"
    },
    {
      "name": "Crypto"
    },
    {
      "name": "Stocks & Funds"
    },
    {
      "name": "Saving"
    },
    {
      "name": "Scpi"
    },
    {
      "name": "BTC"
    },
    {
      "name": "ETH"
    },
    {
      "name": "SOL"
    },
    {
      "name": "Housing"
    },
    {
      "name": "Food"
    },
    {
      "name": "Rent"
    },
    {
      "name": "Utility"
    },
    {
      "name": "Mortgage"
    },
[...]

PR #3690, opened 2023-08-08
feat(Sankey): add prop to control sankey sorting

daformat:
This PR adds a prop to the Sankey component to allow disabling the y sorting

## Description

The Sankey component sorts nodes by y value and doesn't expose a way to control this behavior. This PR adds a boolean prop that allow to disable that sorting behavior. This prop default to true so that existing implementation keep rendering the same.

Here's a visual to showcase the current behavior and the new one with the prop set to false:

With sorting | Without sorting
--|--
image | image

The storybook has been updated to reflect the new prop

## Related Issue

#3683

## Motivation and Context

When using Sankey charts to create cashflow charts, we want to group expenses or income by category. The automatic sorting makes chart links overlap and cross each other, when we want to actually group things by topic to keep the cashflow analysis legible. The fact that one expense in one category is smaller than others in other categories shouldn't move the link around and mix it with other categories, as it makes the chart more complicated to read. The goal is to give control to the user on the grouping of the branches, while still providing with the option to sort data by default for backward compatibility

## How Has This Been Tested?

- [x] used the prop on a existing implementation
- [x] manually tested through storybook
- [x] ran the jest test suites
- [x] ran the storybook tests  

See description

## Types of changes

- ~[ ] Bug fix (non-breaking change which fixes an issue)~
- ~[ ] Breaking change (fix or feature that would cause existing functionality to change)~

## Checklist:

- [x] I have updated the documentation accordingly. (https://github.com/recharts/recharts.org/pull/275)
- ~[ ] I have added tests to cover my changes.~ (I don't see how I can automatically test this since no data-attributes are present on the nodes / links and the existing Sankey test are just counting nodes)

### DIFF
--- a/src/chart/Sankey.tsx
+++ b/src/chart/Sankey.tsx
@@ -147,13 +147,15 @@ const updateYOfTree = (depthTree: any, height: number, nodePadding: number, link
   return links.map((link: any) => ({ ...link, dy: getValue(link) * yRatio }));
 };
 
-const resolveCollisions = (depthTree: any[], height: number, nodePadding: number) => {
+const resolveCollisions = (depthTree: any[], height: number, nodePadding: number, sort = true) => {
   for (let i = 0, len = depthTree.length; i < len; i++) {
     const nodes = depthTree[i];
     const n = nodes.length;
 
     // Sort by the value of y
-    nodes.sort(ascendingY);
+    if (sort) {
+      nodes.sort(ascendingY);
+    }
 
     let y0 = 0;
     for (let j = 0; j < n; j++) {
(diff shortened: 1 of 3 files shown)


ckifer (approved):
LGTM, will wait another review though

akamfoad (approved):
Looks good to me

akamfoad:
NIT: @daformat I'd love to see the example in the PR description as a storybook story, more complex and real-world stories with more data helps spot differences and bugs easier.

ckifer:
+1 to that for sure

daformat:
@akamfoad done! I've added the CustomNodeAndLink the the API stories

akamfoad (approved):
Thanks, good work @daformat
looks good to me.

ckifer (approved):

PR #3700, opened 2023-08-13
add: default tabIndex and accept it as prop

akamfoad:
## Description

Make `<Pie />` component accept `rootTabIndex` as a prop and use it in the root
`Layer` in the render function.

## Motivation and Context

Previously the root layer of `Pie` was always keyboard focus-able which
means it wasn't possible to control its behavior. This changes that to
accept an optional `rootTabIndex` prop which by default is `0` the old value
to make sure not introduce a breaking change, except now users can
control this via a prop.

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -133,7 +133,7 @@ export class Pie extends PureComponent<Props, State> {
 
   static displayName = 'Pie';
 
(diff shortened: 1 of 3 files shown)


ckifer (approved):
LGTM

Issue #3704, opened 2023-08-15
Testing Recharts: Jest --> Vitest 


nikolasrieble:
We have in the past introduced victory-vendor to import d3, fixing a critical security issue with the lowest possible effort. 

Now we want to properly fix it, reverting the initial change, and upgrading d3. 

One of the major issues we are facing after the upgrade is most of the jest tests fail, 
https://github.com/recharts/recharts/actions/runs/5860401123/job/15888379614?pr=3702

We found a [blog](https://blog.isquaredsoftware.com/2023/08/esm-modernization-lessons/) describing both vitest as a solution as well as the ease of migration. 

Lets verify that vitest solves the problem, and if yes, lets switch to vitest.

ckifer:
Caveat to this is that a lot of our consumers will continue to use jest and aren't going to switch just to solve test execution problems. If we do release an update to vitest, we need a dummy project that builds in our CI/CD and runs Jest tests that import (basic rendering) of recharts components. This sounds like a pain but it will for sure come up as a pain point for our consumers

ckifer:
I mean vitest is objectively faster and a better user experience. I would advocate for it either way. But if it's easier for us to mitigate risk by sticking with jest then that's probably okay (just a pain)

PR #3705, opened 2023-08-15
[WIP] Vitest to the rescue?

nikolasrieble:
## Related Issue
https://github.com/recharts/recharts/issues/3704

## Types of changes

## Checklist:


ckifer (commented):
Love how easy the setup is and how straightforward the changes are

nikolasrieble:
@akamfoad @Yilun-Sun @ckifer I am making good progress so far, but am stuck on the last few tests. If you would like to give it a try, or provide some feedback / direction, that would be very appreciated.

ckifer:
Will take a look when I can 🚀

ckifer:
```
beforeEach(() => {
  vi.spyOn(window, 'requestAnimationFrame').mockImplementation((callback): number => {
    callback(0);
    return 0;
  });
});

afterEach(() => {
  window.requestAnimationFrame.mockRestore();
});
```
in `vitest.setup.ts` seems to work for mocking `requestAnimationFrame` 

Then a bunch of tests break though

ckifer:
Well its actually quite hilarious that this refactor highlights the bug thats been present in `react-smooth` for ages. https://github.com/recharts/react-smooth/issues/44#issuecomment-999020364

If I `npm link` my local react-smooth and add under this line https://github.com/recharts/react-smooth/blob/master/src/Animate.js#L127
```jsx
    if (onAnimationEnd) {
      onAnimationEnd();
    }
```

And then in a test specifically tested with the failing `FunnelChart` test https://github.com/recharts/recharts/blob/4582d905033776f71e88fa25a94294e4aabf80cb/test/chart/FunnelChart.spec.tsx#L55
and replaced the current call with timers with
```jsx
    // wait animation end
    await vi.runAllTimersAsync();
```

The test succeeds and renders with [the following successful Funnel chart](https://testing-playground.com/#markup=DwEwlgbgfKkAQGMA2BDAzmgvAIgE4FMEALFXAFzQFoB3XFAB3v12zjTIE8l8d6B7NGDJg+AOwBccAqmER8AbkQBXXGj65JIfADMUSpGUXUwIMkUkBWAAxX6AD0VF8YAOZEykgMw3781rj5uHAIXEVFsGDQIF0RUDGDCEnIqNBVdBHxWY1MiHGsrVidXdxxvArgIMHxqACE+OxwrOCb8uDKI4GEybhgAei6e0Hw0BD6tEZgtbTQYZDB6AAUUMzgTBOJSCgAmSjn6DoIEMjgG7AtWDjzC5zcyHC2ATnLssxwAFieO3sOyPr2lsxjHQzYAxZDoLB4RKbKioDjMKTQ5KUMh0JgALz4JjQHTBcUhhySFEocOYuNiEPWRNhKHhuERG2R2iUolE+CQKLR+ExJn8gR42DAAFsXLiYPRlkQTqVypcztdinczgA2LImV7YN7eViiFBCgWAVR1AA3OrHYAQA1gKAMTaW2sbRgJBIHBWgAcVnd7tY4PiUMZxNRDG5WJArBAOAAsm0rAAaCwAGTgb2VsYTSYAjAB2GPK9OJzOeHN56NxuAALS+EsBwF6Lj6ddBFN9hJhJNpCJbTJZbI5gYxIb5QUFIrFwCrUtOBYuOFzCtueVVcBeuWwnjebx1eoFgE6HQDfiqbUXxLS7bdp7Y7ndg3R6b978VTW33g7y4OHsFGC0XE28s1+2hYtgAOgsGN0yzRN008TNgNA8C4E-XNy0rSV61QxsfQJJFiVJelO2JZlWXZTkgx5UMpH5HBhVFGAG3HaVsEg6DzjgOUwMzOcShVNUcnudNmN1fUcEAfujADvUg8LWtU9zydF1PVvJtMP9Kgn1IsNIzgRiYLYxNPAArS4M8Tx01AgsIIsB4TM8CCoP0zMkL6cc0NrGiFIfZEcIZalKAIntiP7F8AiHKjRzo04+IeacGKnOAinnLil3VFcsy2TdBOwQAKlUAI+jxKPSS7TgB0ZKvOSvVcv0vJUgdX3U8LLJ0oy6ujKxAIsrZHgg10HkAqwHl6vr+t67M2oeMyLKzTx7JrRyawbZz0PvcrWw8vCqB8ojKoCijh2o0FxUlej006wD00i4aOKVCxF2XHAwPTQDXVSgVACLowBVZRy48rykgqL1km9SowtyAy5VTqvfDSjuM4adJsFqYyhuAtnXFqBpRh4htdNNEbeZHUf69G00OrrIceSbemmuaKdm2aolm8BoBrOmoCAA)

No idea how our previous tests were passing unless jest does something special on component unmount/swallows the errors.

ckifer:
On the other hand... this doesn't solve the issue for LineChart :/ but I think its still the same problem

ckifer:
got it - we need to make this change and add `await vi.runOnlyPendingTimersAsync();` in between all mouseOver, mouseOut events

ex:

```jsx
 test('should show tooltips using syncMethod: [function] for both charts on MouseEnter and hide on MouseLeave', async () => {
    const syncMethodFunction = () => {
      return 1;
    };

    const chart1 = (
      <LineChart
        width={width}
        height={height}
        data={data}
        margin={margin}
        syncId="test"
        syncMethod={syncMethodFunction}
      >
        <Line type="monotone" dataKey="uv" stroke="#ff7300" />
        <Tooltip />
        <XAxis dataKey="name" />
      </LineChart>
    );
    const chart2 = (
      <LineChart
        width={width}
        height={height}
        data={data2}
        margin={margin}
        syncId="test"
        syncMethod={syncMethodFunction}
      >
        <Line type="monotone" dataKey="uv" stroke="#ff7300" />
        <Tooltip />
        <XAxis dataKey="name" />
      </LineChart>
    );
    const { container } = render(
      <div>
        {chart1}
        {chart2}
      </div>,
    );

    const chartWidth = width - margin.left - margin.right;
    const dotSpacing = chartWidth / (data.length - 1);

    // simulate entering just past Page A of Chart1 to test snapping of the cursor line
    expect(container.querySelectorAll('.recharts-tooltip-cursor')).toHaveLength(0);

    const firstChart = container.querySelector('.recharts-wrapper');

    fireEvent.mouseOver(firstChart!, { bubbles: true, clientX: margin.left + 0.1 * dotSpacing, clientY: height / 2 });

    await vi.runOnlyPendingTimersAsync();

    screen.logTestingPlaygroundURL();
    // There are two tooltips - one for each LineChart as they have the same syncId
    expect(container.querySelectorAll('.recharts-tooltip-cursor')).toHaveLength(2);

    // make sure tooltips display the correct values, synced by data value
    expect(screen.queryByText('400')).toBeTruthy();
    expect(screen.queryByText('550')).toBeTruthy();

    // Check the activeDots are highlighted
    expect(container.querySelectorAll('.recharts-active-dot')).toHaveLength(2);

    // simulate leaving the area
    fireEvent.mouseLeave(firstChart!);

    await vi.runOnlyPendingTimersAsync();

    expect(container.querySelectorAll('.recharts-active-dot')).toHaveLength(0);
  });
```

ckifer:
For whatever reason the change to `react-smooth` makes animation super laggy 😕

nikolasrieble:
re: react-smooth 

A test exists that verifies the onAnimationEnd is called. 

```
  it('Should called onAnimationEnd', (done) => {
    let num = 0;
    const handleAnimationEnd = () => {
      num += 1;
    };

    render(
      <Animate
        from="1"
        to="0"
        attributeName="opacity"
        duration={500}
        onAnimationEnd={handleAnimationEnd}
      >
        <div />
      </Animate>,
    );

    expect(num).toEqual(0);
    setTimeout(() => {
      expect(num).toEqual(1);
      done();
    }, 700);
  });
```

So it seems that the test here indeed unmounts the component before the animation is finished. Which means that instead we must wait further. 

The fix in react-smooth would only be required if we wanted to test for early unmount, which imho is an unrelated case.

ckifer:
Hmm, I have not the slightest clue then. onAnimationEnd does not look called in react-smooth. I wonder how this test passes. 


Tbh React-smooth isn't the most sustainable library either as it's api is pretty confusing. I wonder if we could get similar animations with a different lib

nikolasrieble:
> Hmm, I have not the slightest clue then. onAnimationEnd does not look called in react-smooth. I wonder how this test passes.
> 
> Tbh React-smooth isn't the most sustainable library either as it's api is pretty confusing. I wonder if we could get similar animations with a different lib

Right, deprecating react-smooth for something else would be lovely - reducing the overall surface of what we have to maintain.

nikolasrieble:
It seems this we have a conflict here.

According to the [README](https://github.com/callstack/reassure), `reassure` requires jest. 
That is presumably why `npm run test:perf` is failing now with 

```
 FAIL  test/chart/AreaChart.perf-test.tsx
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /Users/nikolasrieble/code/recharts/node_modules/d3-scale/src/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){export {
                                                                                      ^^^^^^

    SyntaxError: Unexpected token 'export'

    > 1 | import * as d3Scales from 'd3-scale';
        | ^
      2 | import {
      3 |   Series,
      4 |   stack as shapeStack,

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1728:14)
      at Object.<anonymous> (src/util/ChartUtils.ts:1:1)

----------|---------|----------|---------|---------|-------------------
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------|---------|----------|---------|---------|-------------------
All files |       0 |        0 |       0 |       0 |                   
----------|---------|----------|---------|---------|-------------------
Test Suites: 2 failed, 2 total
Tests:       0 total
Snapshots:   0 total
Time:        10.325 s
Ran all test suites.

❌  Test runner (/Users/nikolasrieble/code/recharts/node_modules/jest/bin/jest.js) exited with error code 1
```

@PavelVanecek @ckifer @branberry

ckifer:
Yeah they definitely use it https://github.com/callstack/reassure/blob/ce92794e2821b80941710a993fcb08d508d2654b/packages/reassure-cli/src/commands/measure.ts#L153

I guess tradeoffs - what do we value more? Reassure was an easy add, vitest was a lot of effort. Vitest is modern, fast, easy to use, but still has similar issues to jest at the end of the day. 

We could easily leave jest in `devDependencies` but we might as well just keep using jest then 😓. Not really sure how I feel about this one

PavelVanecek:
reassure reads the test runner path from environment, have we tried that? Might be an easy fix: https://github.com/callstack/reassure/blob/ce92794e2821b80941710a993fcb08d508d2654b/packages/reassure-cli/src/commands/measure.ts#L43

other than that ... well we still have the storybooks for manual testing. And browser devtools. So if we remove reassure we're still okay.

This is the kind of thing that happens when you switch from the "standard" tool (jest) to a newcomer.

ckifer:
Seems that vitest needs WASM and it can't find it when directly pointing to the test bin file. 

Created https://github.com/callstack/reassure/issues/421 though I'm sure that'll be a while if ever

akamfoad:
Now that all the tests are passing I'm quite happy with the Vitest migration.

Is the perf is the only blocker here?

I'm not quite familiar with reassure, is it possible we can replace it with vitest built-in benchmarking mode @PavelVanecek ? 
https://vitest.dev/advanced/api.html#benchmark

nikolasrieble:
Really, we only need a decision to be made here. 

Given our previous research, and the fact that a popular reference project also [migrated jest -> vitest for the same reason](https://github.com/reduxjs/redux-toolkit/pull/3102), I am still confident that vitest is the right choice. 

The next steps here are: 

- [ ] Rebase the migration PR on master, and already migrate to vitest before 3.0
- [ ] Revert the addition of reassure
- [ ] As @akamfoad mentioned: Investigate the benchmark feature of vitest
- [ ] Add a test in the demo project, that recharts can still be tested with jest for customers

Do you agree, @ckifer? If yes, lets define follow up tickets and get going.

ckifer:
Agreed! 

I already did merge master into this branch once, should be good to do so again without issue. I don't have much time this weekend but will keep an eye out for updates to the branch

PavelVanecek:
I haven't used the benchmark mode in vitest personally but as far as I can see it doesn't support re-render count so it won't be direct replacement for reassure.

Still, I think we can remove reassure for now and search for alternatives later.

ckifer:
Closing this as the other branch is merged

Issue #3615, opened 2023-06-07
Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead. at {Component}


marco-digennaro:
Using a library that include recharts v2.6.2 i get the following error in the console.

> Curve: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead. at Curve

the charts are rendered normally thou.

ckifer:
Thanks, noted! Unfortunately moving away from defaultProps causes unintended bugs due to the way recharts works. But this is known and we're trying to get away from them where we can.

cdimitroulas:
FYI this also happens for the Tooltip component

ckifer:
Reference for why this is an issue https://github.com/recharts/recharts/issues/3438#issuecomment-1467419256

ckifer:
@akamfoad all of our functional refactors are going to have this issue. We may have to prepared to handle soon as I'm going to release 2.7. Let me know if you have any thoughts

akamfoad:
@ckifer 
I was always feeling bad about the defaultProps but in my refactors I kept them just in case and to reduce the number of changes.

But I think the sooner we embrace JS default params the better, I see many teams are going in this direction, especially those who use TypeScript, I see that is quite the default practice.

This is not going to break semver if we do it in v2, because we just change how we receive the default values, internally, for the users of the library everything will be the same and the warning goes away. What do you think?

ckifer:
@akamfoad I totally agree, but we still have the cloneElement issue linked above (https://github.com/recharts/recharts/issues/3438#issuecomment-1467419256).

I believe all of the components we have refactored besides Tooltip are safe from that issue so we should be able to move them over

ckifer:
I believe [this list ](https://github.com/recharts/recharts/blob/4ee4124f3d73f9487b4b74d6aedac0f87af8d3a2/src/chart/generateCategoricalChart.tsx#L2161-L2182) holds those that we cannot touch without fear of breaking things

akamfoad:
So only those components will cause issue if defaultProps is removed, that another component/function needs to access them, including `cloneElement`, before the component is initialized.

I'm guessing there is a lot of them, especially those from the list in `generateCategoricalChart`.

Which means they in fact cause breaking change, when migrated to default params. So we might need to hold off here and revert the Tooltip refactor, that is an easy one.

[...]

PR #3706, opened 2023-08-17
fix(default-props): polar grid to default params

ckifer:
## Description

Fix default props of PolarGrid - this is the last remaining easy refactor to defaultParams

## Related Issue

https://github.com/recharts/recharts/issues/3615

## Motivation and Context

error logged for users on render about defaultProps deprecation

N/A

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/PolarGrid.tsx
+++ b/src/polar/PolarGrid.tsx
@@ -14,7 +14,7 @@ interface PolarGridProps {
   polarAngles?: number[];
   polarRadius?: number[];
   gridType?: 'polygon' | 'circle';
-  radialLines: boolean;
+  radialLines?: boolean;
 }
 export type Props = SVGProps<SVGPathElement> & PolarGridProps;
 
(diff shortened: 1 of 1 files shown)


src/polar/PolarGrid.tsx
```diff
@@ -14,7 +14,7 @@ interface PolarGridProps {
   polarAngles?: number[];
   polarRadius?: number[];
   gridType?: 'polygon' | 'circle';
-  radialLines: boolean;
+  radialLines?: boolean;
```
ckifer, line 17: this is defaulted so this change is fine, tests need this to be optional as it isn't specified in some tests

src/polar/PolarGrid.tsx
```diff
@@ -144,20 +135,41 @@ const ConcentricPath: React.FC<Props> = props => {
...
+        {...props}
+      />
+      <ConcentricPath
+        cx={cx}
+        cy={cy}
+        innerRadius={innerRadius}
+        outerRadius={outerRadius}
+        gridType={gridType}
+        radialLines={radialLines}
+        {...props}
```
akamfoad, line 169: NIT: Since the de-structured values are props also, what do you think about renaming this to `otherProps`

akamfoad (approved):

src/polar/PolarGrid.tsx
ckifer: Yeah otherProps or restProps works - I always forget that spreading the rest doesn't contain the defaulted params. I can change on my next PR but will merge this for now

PR #3709, opened 2023-08-20
migrate generateCategoricalChart to functional component

akamfoad:
> extremly WIP

This is an attempt to migrate the component inside
generateCategoricalChart to a functional component.
I cloned the file and renamed it so I can get one of the
charts to work as expected as the current implementation

NOTE: Currently only LineChart uses the new implementation
I use it as a reference, later on we can only switch the function
call and that will be it, since the input and output is identical
to the current one, the difference is in the implementation
detail.

### DIFF
--- a/src/chart/ComposedChart.tsx
+++ b/src/chart/ComposedChart.tsx
@@ -1,7 +1,8 @@
 /**
  * @fileOverview Composed Chart
  */
-import { generateCategoricalChart } from './generateCategoricalChart';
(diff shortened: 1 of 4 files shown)


ckifer:
Hey @akamfoad I'm going to close this I think because its a bit too much in one go. Nice work though, lets use it as a reference when we tackle in smaller chunks

akamfoad:
Sure, thanks @ckifer

Issue #3716, opened 2023-08-22
Throw useful errors - use the invariant pattern


ckifer:
### Feature request

Throw more useful errors to users utilizing recharts

## What is the current behavior?

Recharts just breaks with no clear indication of why, even for known behavior

## What is the expected behavior?

We should try, where we can (where we have clear cases we need to return error to the user), to use [the invariant pattern](https://www.strictmode.io/articles/invariant) and throw errors early instead of when things break.

For example, when a user specifies an `xAxisId` on all of their `XAxis` components without specifying a corresponding `xAxisId` on their chart components, we should throw a clear error to tell them what has happened. This is easy to do if we fail and fail early instead of letting things break. See https://github.com/recharts/recharts/issues/3712 as an example of this problem

[tiny-invariant](https://github.com/alexreardon/tiny-invariant) can be used as a tool to provide these errors to the user when in dev mode and not change any behavior in production build

CC: @nikolasrieble @akamfoad 
Curious for input on this, I think for some of the known things that will break users using this pattern will reduce churn on issues, etc.

Yilun-Sun:
Good idea!

ckifer:
Looks like somewhere around here https://github.com/recharts/recharts/blob/526644a1fc8b0f9bbfe3c906feedbe0bde56fcb9/src/chart/generateCategoricalChart.tsx#L846-L847

Although looks like ZAxis is also included as an Axis component even when we don't specify one? Maybe we can do a bit of a speed enhancement here by not processing axes that aren't provided as well.

ckifer:
Not going to close this because there are lots of other instances this could be used. Any thoughts?

One that comes to mind is using string data for a number axis

PavelVanecek:
So I was thinking about this a bit. I'm not a fan of adding dozens of small throws and then compiling it away somehow. The less complexity we need to add during the compilation the better (or rather, better to move the complexity outside of the library, to consumers and their preferences and build process).

What perhaps we could do instead is:

1. Add eslint plugin: then we could point to a specific line and show errors. This plugin would be optional and configurable. Downside: it won't be able to catch problems across multiple files/components, such as "Line item has xAxisId pointing to a non-existing axis".
2. Add internal linter, a component that people can insert into the chart, something like this:

```jsx
import { RechartsLinter } from 'recharts-linter'

(
  <BarChart>
    <RechartsLinter />
  </BarChart>
)
```

[...]

Issue #3712, opened 2023-08-21
Multiple XAxis components with `xAxisId` set causes an error


caterpaula:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/bar-chart-with-double-xaxis-dfug7?file=/src/App.tsx)

### Steps to reproduce
* Create a graph with two X axis
* Give each of those axis a unique `xAxisId` string as a prop

### What is expected?
The graph would render as expected, with each axis having it's own unique ID

### What is actually happening?
The graph is unable to render and throws errors

| Environment | Info |
|---|---|
| Recharts | v2.7.3 |
| React | React version 17.0.1 |
| System | macOS Ventura 13.4.1 |
| Browser | Chrome Version 116.0.5845.96 |

---
Comment out line 110 of the supplied reproduction link to see the graph working when only one axis gets an `xAxisId` prop

Depending on which axis has the `xAxisId` prop set the axis render in different places as well -
When the second defined XAxis has the `xAxisId` prop set both XAxis render as expected
![image](https://github.com/recharts/recharts/assets/5556891/0b54460e-8989-442c-8c86-def7e670a27a)
When the first defined XAxis has the `xAxisId` prop set both XAxis render above where would be expected (behind the bars, in the case of the repro link above)
![image](https://github.com/recharts/recharts/assets/5556891/6feb8638-63c8-4920-a347-721f9d15a600)
When both defined XAxis have the `xAxisId` prop set the graph won't render
![image](https://github.com/recharts/recharts/assets/5556891/3cc02587-5600-423b-b68b-8694d8efbe3b)

caterpaula:
Thanks for your reply @ckifer 

That makes sense, thank you - I agree that the error message could be improved and also potentially updating the example to show a use case which doesn't use the default xAxisId for either of the xAxis?

ckifer:
https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/story/examples-cartesian-yaxis-withleftandrightaxes--with-left-and-right-axes

ckifer:
Added example to the storybook and created https://github.com/recharts/recharts/issues/3716 out of this

Going to close this one, thank you!

ckifer:
[...]

PR #3720, opened 2023-08-29
feat: add invariant check for when axisIds to not match between chart…

ckifer:
… and axis

## Description

Add invariant pattern to recharts.

Throw an error only in dev mode when the user provides invalid configuration for `AxisId`

## Related Issue

https://github.com/recharts/recharts/issues/3716

## Motivation and Context

https://github.com/recharts/recharts/issues/3712

User feedback that we need to throw better errors when in development mode. After looking into how this can be done `tiny-invariant` is the perfect case that adds next to nothing to bundle size.

## How Has This Been Tested?

- tests continue to succeed
- test locally in storybook with `xAxisId` `yAxisId` and `zAxisId`

image
image
image

## Types of changes

- [x] New feature (non-breaking change which adds functionality) - dev experience improvement

## Checklist:

- [ ] My change requires a change to the documentation. - no but subsequently supports documentation

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -858,7 +858,7 @@ export const generateCategoricalChart = ({
           axisMap && axisMap[id],
           `Specifying a(n) ${entry.axisType}Id requires a corresponding ${
             entry.axisType
-          }Id on the targeted graphical component ${item?.type?.displayName ? item?.type?.displayName : ''}`,
+          }Id on the targeted graphical component ${item?.type?.displayName ?? ''}`,
         );
 
         // the axis we are currently formatting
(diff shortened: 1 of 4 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -839,12 +840,29 @@ export const generateCategoricalChart = ({
...
+        /**
+         * tell the user in dev mode that their configuration is incorrect if we cannot find a match between
+         * axisId on the chart and axisId on the axis
+         */
+        invariant(
+          axisMap && axisMap[id],
+          `Specifying a(n) ${entry.axisType}Id requires a corresponding ${
+            entry.axisType
+          }Id on the targeted graphical component ${item?.type?.displayName ? item?.type?.displayName : ''}`,
+        );
```
Yilun-Sun, line 862:
```suggestion
        invariant(
          axisMap && axisMap[id],
          `Specifying a(n) ${entry.axisType}Id requires a corresponding ${
            entry.axisType
          }Id on the targeted graphical component ${item?.type?.displayName ?? ''}`,
        );
```

ckifer: Thanks for the review. Yeah makes sense!

Yilun-Sun (approved):

Issue #3721, opened 2023-08-29
ResponsiveContainer should accept "style" prop


abforce:
When we do `React.cloneElement` on an `ResponsiveContainer`  and adding `style`, it simply disregards the prop.

ckifer:
Fixed by #3726

Thanks @d-gottlieb!

@abforce this will be in the next release

ckifer:
released in 2.9

PR #3726, opened 2023-09-01
To resolve Issue 3721

d-gottlieb:
## Description

Per [issue 3721](https://github.com/recharts/recharts/issues/3721), this adds the ability to pass the `style` prop without it being ignored.

## Related Issue

[Issue #3721](https://github.com/recharts/recharts/issues/3721)

## Motivation and Context

It allows the user to pass their own styles and have them rendered with the `ResponsiveContainer` component.

## How Has This Been Tested?

Unit tests have been added and have confirmed `style` being rendered.

## Types of changes

## Checklist:

### DIFF
--- a/src/component/ResponsiveContainer.tsx
+++ b/src/component/ResponsiveContainer.tsx
@@ -167,7 +167,7 @@ export const ResponsiveContainer = forwardRef(
       }
     }, [getContainerSize]);
 
-    const styles: React.CSSProperties = style ?? { width, height, minWidth, minHeight, maxHeight };
+    const styles: React.CSSProperties = { ...style, width, height, minWidth, minHeight, maxHeight };
 
     return (
       <ReactResizeDetector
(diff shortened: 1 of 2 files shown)


src/component/ResponsiveContainer.tsx
```diff
@@ -57,6 +58,7 @@ export const ResponsiveContainer = forwardRef(
       id,
       className,
       onResize,
+      style,
```
ckifer, line 61: looks like this might need defaulted to `{}` since it is optional? 

I don't think `undefined` can be spread

ckifer (changes requested):
Looks good! just one small thing

src/component/ResponsiveContainer.tsx
d-gottlieb: > looks like this might need defaulted to `{}` since it is optional?
> 
> I don't think `undefined` can be spread

Good catch. Fixed!

Issue #3669, opened 2023-07-25
The tooltip does not include data from all charts


eno-tech0:
The tooltip only shows data for the Line graph, but nothing for the Area graph. How to display data for both Area and Line graphs in the tooltip? Here's an example with a problem: https://codesandbox.io/s/recharts-tooltip-with-different-data-kksx5s?file=/src/App.js
---

Edit from maintainer:

The tooltip does not include data from all charts _when a separate dataset is passed to chart prop `data` and specified on `Line/Area/etc` prop `data`_

ckifer:
Hi,

This looks like its because you are using a separate dataset for each chart. Combining these will allow the domain to be calculated correctly for both datasets.

https://codesandbox.io/s/recharts-tooltip-with-different-data-forked-ww6hpj

Ideally your example should work, so I'll label this as a bug

eno-tech0:
@ckifer Thanks for your help)

ckifer:
Merged in #3733, will be out in next minor

ckifer:
Released, closing!

PR #3733, opened 2023-09-07
fix: The tooltip does not include data from all charts when a separate dataset is passed to chart prop data and specified on Line/Area/etc prop data 

andrewangelle:
## Description
Set a default argument for the data source of `getTooltipContent` in `generateCategoricalChart.ts` that points to `chartData`. In this case `child.props.data` is looking at the individual graph element(Line/Area/etc.). Logic is essentially "if there's no data there, check chart data." 

## Related Issue
#3669 

## Motivation and Context

The tooltip does not include data from all charts when a separate dataset is passed to chart prop data and specified on Line/Area/etc prop data

## How Has This Been Tested?

Unit tests and storybook.

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -232,7 +232,7 @@ const getTooltipContent = (
       return result;
     }
 
-    const { data } = child.props;
+    const { data = chartData } = child.props;
     let payload;
 
     if (tooltipAxis.dataKey && !tooltipAxis.allowDuplicatedCategory) {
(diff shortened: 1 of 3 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -232,7 +232,7 @@ const getTooltipContent = (
       return result;
     }
 
-    const { data } = child.props;
+    const { data = chartData } = child.props;
```
ckifer, line 235: could we add a comment here explaining why this needs defaulted? I feel this is something we'll loose context on later and break accidentally

ckifer (changes requested):
Thanks for the fix! one tiny thing

src/chart/generateCategoricalChart.tsx
andrewangelle: @ckifer Added comment. Thanks!

Issue #2778, opened 2022-02-08
Tooltip is displayed incorrectly when the parent container is transformed with CSS "transform: scale()" property


MateuszTrN:
## Do you want to request a feature or report a bug?
Bug

## What is the current behavior?
If the parent (or any container above) div is scaled via css `transform: scale(x)` property - the mouse events work incorrectly. 
The tooltip is displayed for the data that is not hovered. Trying to display tooltip for the few last bars is impossible. 

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via 
https://codesandbox.io/s/simple-bar-chart-forked-enk3g?file=/src/App.tsx

## What is the expected behavior?
regardless of parent div transformation, the mouse events / tooltip work the same

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Recharts 2.1.8 (I quickly tested previous versions  1.x, 2.0.x - same behaviour)
OS: MacOS Mojave
Browser Chrome Version 97.0.4692.99 (Official Build) (x86_64)

MateuszTrN:
Hi,
I did some digging and I found the cause of the issue. I applied some quickfixes to ensure if my thinking was right and it seems to be working with those changes. See below:

file: ` ../chart/getCategoricalChart.js ` lines `1209-1229`
Original code snippet
```
inRange(x: number, y: number): any {
      const { layout } = this.props;

      if (layout === 'horizontal' || layout === 'vertical') {
        const { offset } = this.state;
        const isInRange =
          x >= offset.left && x <= offset.left + offset.width && y >= offset.top && y <= offset.top + offset.height;

        return isInRange ? { x, y } : null;
      }

      const { angleAxisMap, radiusAxisMap } = this.state;

      if (angleAxisMap && radiusAxisMap) {
        const angleAxis = getAnyElementOfObject(angleAxisMap);

        return inRangeOfSector({ x, y }, angleAxis);
      }

      return null;
    }
```


my hotfix:
```
  inRange(x: number, y: number): any {
      const { layout } = this.props; 
      //calculating scale
      const element = this.container; 
     
     
       //the line below is the most dirty part of this and I wouldn't implement it here in final solution. 
       const scale = element.getBoundingClientRect().width / element.offsetWidth;
       //calculating scaled coordinates
        const [scaledX, scaledY] = [x / scale, y / scale]; 
       
      if (layout === 'horizontal' || layout === 'vertical') {    
[...]

PR #3748, opened 2023-09-10
Bugfix/b(#2778) tooltip position when container uses transform scale

MateuszTrN:
Fixes detection of a tooltip mouse event

## Description

When a parent container used transform: scale(Number) Css - the tooltip mouse event was fired up in the incorrect mouse position 

## Related Issue

https://github.com/recharts/recharts/issues/2778

## Motivation and Context

This changes allows user to style parent container with transform style css property without affecting tooltip behaviour

## How Has This Been Tested?

npm run test, npm run storybook and did some manual testing. I also created a dedicated story.
The changes should not affect other parts of the code - it was a change within 1 function, that already had tests around it. It introduces new parameter to a "inRange" function that have default value if wasn't passed. The API of that function was not changed, only extended. 

## Types of changes

## Checklist:

- [x] I have added a story to the storybook

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1255,7 +1255,9 @@ export const generateCategoricalChart = ({
       const e = calculateChartCoordinate(event, containerOffset);
 
       const element = this.container;
-      const scale = element.getBoundingClientRect().width / element.offsetWidth;
+      const boundingRectWidth = element?.getBoundingClientRect()?.width;
+      const { offsetWidth } = element;
+      const scale = boundingRectWidth / offsetWidth || 1;
 
       const rangeObj = this.inRange(e.chartX, e.chartY, scale);
       if (!rangeObj) {
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1253,7 +1253,11 @@ export const generateCategoricalChart = ({
 
       const containerOffset = getOffset(this.container);
       const e = calculateChartCoordinate(event, containerOffset);
-      const rangeObj = this.inRange(e.chartX, e.chartY);
+
+      const element = this.container;
+      const scale = element.getBoundingClientRect().width / element.offsetWidth;
+
+      const rangeObj = this.inRange(e.chartX, e.chartY, scale);
       if (!rangeObj) {
         return null;
       }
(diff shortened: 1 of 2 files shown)


ckifer:
Thanks for the contribution! Do you mind resolving the conflict?

src/chart/generateCategoricalChart.tsx
```diff
@@ -1253,7 +1253,13 @@ export const generateCategoricalChart = ({
 
       const containerOffset = getOffset(this.container);
       const e = calculateChartCoordinate(event, containerOffset);
-      const rangeObj = this.inRange(e.chartX, e.chartY);
+
+      const element = this.container;
+      const boundingRectWidth = element?.getBoundingClientRect()?.width;
+      const { offsetWidth } = element;
+      const scale = boundingRectWidth / offsetWidth || 1;
```
ckifer, line 1260: Just checking, this division is safe if it results in `NaN` correct?

src/chart/generateCategoricalChart.tsx
```diff
@@ -1343,23 +1349,28 @@ export const generateCategoricalChart = ({
       ];
     }
 
-    inRange(x: number, y: number): any {
+    inRange(x: number, y: number, scale = 1): any {
       const { layout } = this.props;
 
+      const [scaledX, scaledY] = [x / scale, y / scale];
```
ckifer, line 1355: what happens if these result in NaN?

ckifer (commented):
Nice! Just a few questions

src/chart/generateCategoricalChart.tsx
MateuszTrN: yes, line 1260 returns 1 if the div is NaN (or 0 - which I also assumed is a value that does not make sense)

src/chart/generateCategoricalChart.tsx
MateuszTrN: I can make another check here too, but I assumed I dealt with that in the parent scope, before passing the arg here. as this is internal function that is not exposed so I did not do a check again, but I can add it if you think it's better that way

ckifer: Nah it should be fine. I do feel though like this function could just be totally taken out of this file and probably have some different, more targeted tests added to it. Probably an improvement for another time though

ckifer (approved):
LGTM, will wait for @nikolasrieble or someone else to take a look as well

nikolasrieble (approved):
A great contribution that solves a long standing problem! Thank you a lot. 
I tested it and could not find any edge cases where this would not work. 

The only constructive feedback I have here is: 
I wish we could extract useful helper function instead of extending the complexity on the existing level of abstraction in `src/chart/generateCategoricalChart.tsx`. Ideally every PR from now on would only take code out of generateCategoricalChart (and make it nicely unit testable). 

That being said, this is great to go, and we should focus on making steps forward. This certainly is a step forward, thank you ❤️

Issue #3749, opened 2023-09-10
dataKey - "Functions are not valid as a React child"


chris-mcdonald-dev:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-h5mxx9)

### Steps to reproduce
Use a callback function as the value for the dataKey prop in any of the Cartesian components (Bar/Line/etc.) as implied in the [API docs](https://recharts.org/en-US/api/Line#dataKey).
*(Update: The error only occurs when also rendering the `<Legend/>` component)*

```js
<Line dataKey={(row) => row.pv} />
```

### What is expected?
I expected to be able to access values using the callback function without any errors logged to the console.

### What is actually happening?
The chart is rendering the values correctly. However, there is a console warning from React stating the following:  
`Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it.`

| Environment | Info |
|---|---|
| Recharts | v2.8.0 |
| React | 18.2.0 |
| System | Ubuntu 22.04 |
| Browser | Edge 116.0.1938.69 |

---
The documentation that led me to running into this is shown here: https://recharts.org/en-US/api/Line#dataKey

chris-mcdonald-dev:
Addressed in #3750, closing :)

PR #3750, opened 2023-09-10
Fix: "Functions are not valid as a React child" error in <Legend/> when a function is passed as the payload #3749

chris-mcdonald-dev:
## Description

Small changes checking if the payload provided to `<Legend/>` is a function and calling it as needed. Otherwise, the original value is rendered. Also added accompanying tests.

## Related Issue
#3749 

## Motivation and Context
When a function was provided as a payload to the `<Legend/>` component, it was not getting called before being rendered resulting in the error described in #3749. This ensures the payload is called if passed as a function.

## How Has This Been Tested?
Unit tests included. Checks that values are correctly rendered both when passed as static values and as a function.

## Types of changes

## Checklist:

### DIFF
--- a/src/component/DefaultLegendContent.tsx
+++ b/src/component/DefaultLegendContent.tsx
@@ -3,6 +3,7 @@
  */
 import React, { PureComponent, ReactNode, MouseEvent, ReactText, ReactElement } from 'react';
 import classNames from 'classnames';
+import _ from 'lodash';
 import { Surface } from '../container/Surface';
 import { Symbols } from '../shape/Symbols';
 import {
@@ -169,6 +170,7 @@ export class DefaultLegendContent extends PureComponent<Props> {
         return null;
       }
 
+      const entryValue = _.isFunction(entry.value) ? entry.value() : entry.value;
       const color = entry.inactive ? inactiveColor : entry.color;
 
       return (
@@ -182,7 +184,7 @@ export class DefaultLegendContent extends PureComponent<Props> {
             {this.renderIcon(entry)}
           </Surface>
           <span className="recharts-legend-item-text" style={{ color }}>
-            {finalFormatter ? finalFormatter(entry.value, entry, i) : entry.value}
+            {finalFormatter ? finalFormatter(entryValue, entry, i) : entryValue}
           </span>
         </li>
       );
--- a/test/component/Legend.spec.tsx
+++ b/test/component/Legend.spec.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { Legend, LineChart, Line } from '../../src';
 
 describe('<Legend />', () => {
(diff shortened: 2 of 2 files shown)


src/component/DefaultLegendContent.tsx
```diff
@@ -3,6 +3,7 @@
  */
 import React, { PureComponent, ReactNode, MouseEvent, ReactText, ReactElement } from 'react';
 import classNames from 'classnames';
+import _ from 'lodash';
```
ckifer, line 6: nit nit, me thinking aloud: I know we use `lodash` in this lib everywhere, but I wonder if there is any risk in just `typeof entry.value === 'function'` or `entry.value instanceof Function`

chris-mcdonald-dev: @ckifer I don't see any risk in it and usually prefer native JS operations as well! I noticed we used `_.isFunction()` far more often than the native operation throughout the repo and just went with this in the name of consistency. _(lodash used 75 times vs native operation used 12 times)_

I can make that quick change if we'd prefer native JS readability over consistency with lodash's method here.

ckifer: It's all good, I just was mostly making an observation for myself. Eventually we may want to move away from lodash but the consistency is perfect for now

src/component/DefaultLegendContent.tsx
```diff
@@ -169,6 +170,7 @@ export class DefaultLegendContent extends PureComponent<Props> {
         return null;
       }
 
+      const entryValue = _.isFunction(entry.value) ? entry.value() : entry.value;
```
MateuszTrN, line 173: hi, I'm not sure if I understood correctly, but I don't think this is correct fix. the fn that is provided expects an argument which is not provided here - yeah, it would fix the warning but I believe it won't work correct @ckifer don't you agree?
image

ckifer: I'm confused. This should work fine - I'll pull it down later and check

MateuszTrN: shouldn't it be something like
`const entryValue = _.isFunction(entry.value) ? entry.value(entry) : entry.value;`
or
`const entryValue = _.isFunction(entry.value) ? entry.value(entry.payload) : entry.value;` 
?

in given example - in the issue, the reporter set the handler as below:
`row => row.pv`

in this example and current code - what is row? wouldn't it throw 'row is undefined' ?

I might be wrong though.

ckifer: yeah you're right, thanks for taking the extra care to check. Just need to fix this before the next release and should be okay

@chris-mcdonald-dev fyi

chris-mcdonald-dev: @MateuszTrN Yeah, thanks for catching that. Looking into a solution now. We won't be able to get it from the entry or payload for the Legend component, since that doesn't seem to contain the chart data and just the payload for the legend elements.

ckifer: we're really just trying to get the actual dataKey name here, not any actual values. When you have `data => data.pv` you don't really have great reference to the text `pv`

ckifer:
I'm thinking we don't actually want the `value` prop of `Legend` to be able to be a function. We want it as a string, however we have to get it as a string somehow before the dataKey is passed to legend

src/component/DefaultLegendContent.tsx
chris-mcdonald-dev: Good point. I can't see the point of this being a function directly passed to the `Legend` component either. 

I'll spend some time thinking of a solution. The only initial idea I have is rendering an empty string when a function is derived from a chart component's `dataKey` and printing a warning message to the console saying to pass a payload value as a string to `<Legend/>` whenever that's the case.

Otherwise, I'm not sure how we'd have any way of knowing what that legend is supposed to be labeled since functions won't necessarily have to follow the format of `() => someObject.theValue`

ckifer: It does also use the "name" property before the dataKey property in the Legend. If the dataKey is a function we could print a warning that name is required for a label to appear in the legend

chris-mcdonald-dev: Ah true! I'll create an issue and open a matching PR with tests in a bit.

Issue #3723, opened 2023-08-30
Consistent interface: Line / Bar - activeDot/ activeBar


nikolasrieble:
The Line exposes activeDot, which enables the user to customise how the tooltip interaction is rendered. 
Naively I would expect the Bar to have a similar interface for activeBar.

Suggestion: 
- Align the interface across components, and add to the bar the identical functionality and interface as exists for the Line, see https://recharts.org/en-US/api/Line#activeDot

andrewangelle:
@nikolasrieble I'm taking a stab at this if you want to assign it to me.

ckifer:
merged #3756 - will be released next minor barring no issues

ckifer:
released in 2.9

PR #3756, opened 2023-09-13
feat: Implement activeBar for Bar component

andrewangelle:
## Description
Implementing support for `activeBar` prop on Bar component. I attempted to follow the same behavior of activeDot for a Line. 

## Related Issue
#3723 

## Motivation and Context
Feature request for a consistent interface between Line/Bar for customizing the active element that renders via the Tooltip

## How Has This Been Tested?
Unit tests and storybook

Screenshot 2023-09-13 at 9 19 29 AM

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -73,7 +73,7 @@ export interface BarProps extends InternalBarProps {
   stackId?: string | number;
   barSize?: number;
   unit?: string | number;
-  name?: string;
+  name?: string | number;
   dataKey: DataKey<any>;
   tooltipType?: TooltipType;
   legendType?: LegendType;
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -1,7 +1,7 @@
 /**
  * @fileOverview Render a group of bar
  */
-import React, { PureComponent, ReactElement, SVGProps } from 'react';
+import React, { Key, PureComponent, ReactElement, SVGProps } from 'react';
 import classNames from 'classnames';
 import Animate from 'react-smooth';
 import _ from 'lodash';
@@ -64,15 +64,16 @@ type RectangleShapeType =
   | RectangleProps
   | boolean;
 
-interface BarProps extends InternalBarProps {
+export interface BarProps extends InternalBarProps {
   className?: string;
+  index?: Key;
   layout?: 'horizontal' | 'vertical';
   xAxisId?: string | number;
   yAxisId?: string | number;
   stackId?: string | number;
   barSize?: number;
   unit?: string | number;
-  name?: string | number;
+  name?: string;
   dataKey: DataKey<any>;
   tooltipType?: TooltipType;
   legendType?: LegendType;
@@ -80,7 +81,11 @@ interface BarProps extends InternalBarProps {
   maxBarSize?: number;
   hide?: boolean;
   shape?: ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>);
-  activeBar?: ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>) | boolean | SVGProps<SVGElement>;
+  activeBar?:
+    | ReactElement<SVGProps<SVGElement>>
+    | ((props: BarProps) => ReactElement<SVGProps<SVGElement>>)
+    | boolean
+    | SVGProps<SVGElement>;
   background?: RectangleShapeType;
   radius?: number | [number, number, number, number];
 
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -1,7 +1,7 @@
 /**
  * @fileOverview Render a group of bar
  */
-import React, { PureComponent, ReactElement } from 'react';
+import React, { PureComponent, ReactElement, SVGProps } from 'react';
 import classNames from 'classnames';
 import Animate from 'react-smooth';
 import _ from 'lodash';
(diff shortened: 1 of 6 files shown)


andrewangelle:
#3723

andrewangelle:
Question for maintainers:

When i checked off `I have updated the documentation accordingly.`, I was referring the prop documentation in storybook. I noticed that there is a separate repo for the site documentation for recharts.org. Will there need to be a separate PR to update documentation there?

charpeni:
Yes, API definitions are defined here: https://github.com/recharts/recharts.org/tree/master/src/docs/api.

Here's an example of a corresponding recharts.org pull request: https://github.com/recharts/recharts.org/pull/271.

ckifer:
Yeah sorry for not getting back. For changes like this for now we'll have to update both - the storybook will eventually replace the custom docs site because the docs site is a pain to update.

src/chart/generateCategoricalChart.tsx
```diff
@@ -2040,6 +2041,24 @@ export const generateCategoricalChart = ({
       );
     };
 
+    static renderActiveBar = (option: any, props: any): React.ReactElement => {
```
ckifer, line 2044: So I know the other active components and a lot of others have many uses of `any` - is there any probable way that we are able to type this correctly?

Trying to enforce not using `any` where possible

ckifer, line 2044: We might also want to start moving away from static functions like this @nikolasrieble  thoughts on moving this to `Bar` or a new `BarUtils`? It doesn't really use anything within the `generateCategoricalChart` closure. 

We could refactor all of them in a separate PR I guess

src/chart/generateCategoricalChart.tsx
```diff
@@ -2078,19 +2097,48 @@ export const generateCategoricalChart = ({
...
+        ...item.item.props,
+        ...item.props.data[childIndex],
+      };
+
+      const barProps = {
+        index: childIndex,
+        dataKey,
+        fill: getMainColorOfGraphicItem(item.item),
+        strokeWidth: 2,
+        stroke: '#fff',
```
ckifer, line 2118: Does this mean the outline of the bar will always be black and a fixed width?

people are probably going to want to customize these - do we have ability to modify them at all?

ckifer (commented):
LGTM overall, though I wish we could use less `any` (I know it isn't entirely feasible with what you have to work with).

One clarification needed on if we're able to change style properties

src/chart/generateCategoricalChart.tsx
andrewangelle: I agree with you. This was a good ask. Take another look. I think I have something with a little more type safety?

andrewangelle: Moved this to a `BarUtils` file.

src/chart/generateCategoricalChart.tsx
andrewangelle: Yes we have the ability to modify. Below these lines I spread in props from the parent bar, then props from the activeBar, in that order.

So declaring attributes on `activeBar` will take precedent
Screenshot 2023-09-13 at 9 08 19 PM

No attributes on `activeBar` will then inherit from values declared on the parent bar
Screenshot 2023-09-13 at 9 07 46 PM

And finally if both `activeBar` and parent bar don't have these values defined then we'll hit these two hardcoded values you are referencing

src/cartesian/Bar.tsx
```diff
@@ -64,22 +64,28 @@ type RectangleShapeType =
...
+export interface BarProps extends InternalBarProps {
   className?: string;
+  index?: Key;
   layout?: 'horizontal' | 'vertical';
   xAxisId?: string | number;
   yAxisId?: string | number;
   stackId?: string | number;
   barSize?: number;
   unit?: string | number;
-  name?: string | number;
```
ckifer, line 75: is this removal safe? I wouldn't ever expect `name` to be a number but downstream consumers could have it as a number? this would break anyone that does (unfortunately).

https://recharts.org/en-US/api/Bar
Old docs still say `String | Number` and same with the rest of the cartesian components. I think we should keep this for now

src/chart/generateCategoricalChart.tsx
ckifer: cool thank you

ckifer (changes requested):
Concerned about the type change to name, otherwise LGTM

src/cartesian/Bar.tsx
andrewangelle: Fixed

ckifer (approved):
🚀

ckifer:
Is it just me or is this story _realllly slow_?

nikolasrieble:
Great contribution, thank you @andrewangelle! If you would like to follow up, you could consider adding this ability also to other charts. To my best knowledge we know have activeDot and activeBar. It seems consistent to also extend this to other elements such as Funnel.

src/chart/generateCategoricalChart.tsx
```diff
@@ -2078,19 +2087,51 @@ export const generateCategoricalChart = ({
...
```
nikolasrieble, line 2113: Nit: It would be nice to extract hardcoded constants named to the top of the file (we might even want to reuse constant across files as consistent defaults).

src/cartesian/Bar.tsx
```diff
@@ -115,6 +121,7 @@ export class Bar extends PureComponent<Props, State> {
     hide: false,
     data: [] as BarRectangleItem[],
     layout: 'vertical',
+    activeBar: true,
```
ckifer, line 124: I'm thinking this shouldn't be true by default https://github.com/recharts/recharts/issues/4103

Issue #3757, opened 2023-09-15
<Legend/> should warn users to pass a name to a chart's cartesian components when a function is used as a dataKey.


chris-mcdonald-dev:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-h5mxx9)

### Steps to reproduce
See https://github.com/recharts/recharts/issues/3749

### What is expected?
Need to prevent callback function from being called without the payload as a parameter. Instead, it should prompt API users to use the `name` property on the cartesian component being used and render and empty string instead.

### What is actually happening?
See https://github.com/recharts/recharts/issues/3749

| Environment | Info |
|---|---|
| Recharts | v2.8.0 |
| React | 18.2 |
| System | Ubuntu 22.04 |
| Browser | Edge 116 |

chris-mcdonald-dev:
Have PR fix ready. Pushing soon.

ckifer:
Fixed! Closing this up

PR #3758, opened 2023-09-15
Always render static string as <Legend/> value

chris-mcdonald-dev:
## Description
Always renders static string in `<Legend/>` when functions are passed as a dataKey to sibling cartesian components. Prompts the API user to include the `name` property when `entry.value` is a function.

## Related Issue
#3757

## Motivation and Context

Addresses problems brought up in #3750.
## How Has This Been Tested?
Includes unit tests checking that name prop is rendered when functions are passed and when the name prop is missing.

## Types of changes

## Checklist:

### DIFF
--- a/src/component/DefaultLegendContent.tsx
+++ b/src/component/DefaultLegendContent.tsx
@@ -4,6 +4,7 @@
 import React, { PureComponent, ReactNode, MouseEvent, ReactText, ReactElement } from 'react';
 import classNames from 'classnames';
 import _ from 'lodash';
+import { warn } from '../util/LogUtils';
 import { Surface } from '../container/Surface';
 import { Symbols } from '../shape/Symbols';
 import {
@@ -170,13 +171,12 @@ export class DefaultLegendContent extends PureComponent<Props> {
         return null;
       }
 
-      let entryValue = entry.value;
-      if (_.isFunction(entry.value)) {
-        entryValue = null;
-        console.warn(
-          `The name property is also required when using a function for the dataKey of a chart's cartesian components. Ex: <Bar name="Name of my Data"/>`,
-        );
-      }
+      // Do not render entry.value as functions. Always require static string properties.
+      const entryValue = !_.isFunction(entry.value) ? entry.value : null;
+      warn(
+        !_.isFunction(entry.value),
+        `The name property is also required when using a function for the dataKey of a chart's cartesian components. Ex: <Bar name="Name of my Data"/>`, // eslint-disable-line max-len
+      );
 
       const color = entry.inactive ? inactiveColor : entry.color;
       return (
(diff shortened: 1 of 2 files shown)


src/component/DefaultLegendContent.tsx
```diff
@@ -170,9 +171,14 @@ export class DefaultLegendContent extends PureComponent<Props> {
         return null;
       }
 
-      const entryValue = _.isFunction(entry.value) ? entry.value() : entry.value;
-      const color = entry.inactive ? inactiveColor : entry.color;
+      // Do not render entry.value as functions. Always require static string properties.
+      const entryValue = !_.isFunction(entry.value) ? entry.value : null;
+      warn(
```
ckifer, line 176: Do you think a warning or an invariant would be better in this case?

Warnings are still logged in production and aren't in your face enough. 

Invariant will cause an error only in development (in your face) and will be ignored in production builds. 

Mostly curious on thought of what is the better UX

ckifer (commented):
Looks good, just maybe want to have quick conversation around the one comment I made.

src/component/DefaultLegendContent.tsx
chris-mcdonald-dev: I like throwing an error better actually as well for the UX.

Side note, maybe I'm misunderstanding our `warn()` util, but I assumed that's what the `isDev` check was for here (but as a warning instead of an error):
https://github.com/recharts/recharts/blob/21ce248c06ab75893f17b6308deedbdb7a67238e/src/util/LogUtils.ts#L2-L5

I'll switch to using the tiny-invariant dependency instead.

ckifer: Ah yeah, you're not misunderstanding - I just don't have all of the context either since these things existed long before me 😅. Sounds good though

chris-mcdonald-dev: @ckifer Makes sense. 😅
One quick thing I just thought about was backwards compatibility. Throwing an error here would force all existing projects that currently have functions as dataKeys but aren't including the name in a `<Legend/>` to have fatal errors in dev environments. Those scenarios admittedly would likely be a result of dead code. But do we think that very specific, potentially semi-breaking change is worth opting for an error over a warning here?

ckifer: oof, you're right.

Better to have a warning from recharts than the warning they get now from react (functions not valid as a child). 

Next MV this should be changed to invariant, lets keep it a warning for now

ckifer (approved):

src/component/DefaultLegendContent.tsx
chris-mcdonald-dev: Cool, sounds good!

ckifer: This lib has had a lot of unfortunate breaking changes in minor/patch version releases throughout its life. If possible, I'd like to prevent that moving forwards so we'll keep it as a warning in this iteration.

ckifer:
Thank you @chris-mcdonald-dev

Issue #3407, opened 2023-02-27
(Still) Looking for contributors - a Recharts Status Update


ckifer:
Hi all,

Going to pin this as a new issue out of the old one: https://github.com/recharts/recharts/issues/1833

We've figured some things out since then! First of all, thanks to the creator of recharts @arcthur for helping us get the project into the hands of more contributors and maintainers. We want to help recharts to be better, fix bugs, implement features, increase documentation, etc. We couldn't do that with very much velocity without communication, without direction, or without access.

We now have a recharts Slack channel for communication between those who are collaborating or wish to help collaborate on the project: [Join here](https://join.slack.com/t/recharts/shared_invite/zt-1o9t7luva-x9CbZ1_nI2OHXMr6id1T4A) - _**this is a contributors slack channel, please don't use the slack channel for issues**_

Secondly, thanks to those who have joined as interested or invested members of the community - nothing can happen without individuals willing to put in time and effort to manage the community, manage issues, dive into bugs, update documentation, etc. **Huge** shoutouts to the following:
* @nikolasrieble - mainainer
* @Yilun-Sun - triage and contributor
* @proke03 - triage and contributor
* @marcalexiei - contributor
* @lukask-proxora - contributor

...And many more

We've accomplished a few minor released and bug fixes. Notably:
* Security vulnerability fix using `victory-vendor`
* Migration to Jest and RTL
* Addition of Storybook and trying to better documentation
* More...

----

All that said we still need help! All/most of us are contributing to this library because we either like it and have used it in the past, or are currently using it for day jobs. Any contributions, suggestions, bug deep dives, performance profiles, documentation fixes/improvements, etc. are greatly appreciated by the community.

If you want to help us please join the [slack](https://join.slack.com/t/recharts/shared_invite/zt-1o9t7luva-x9CbZ1_nI2OHXMr6id1T4A).

Cheers!

ckifer:
@TheHumanistX sweet! I think someone with some math background is a huge help, I hate math 😂.

I need to tag more issues with "good first issue", but helping with the docs is always good (and will help get you familiar with the library). We're still in progress of moving the website over to storybook. I'll tag and create a few issues today.

TheHumanistX:
@ckifer Just tag me when you have some things ready you think maybe I could help with. I have never contributed to a project before so this may take a second for me to understand and get used to.

ckifer:
[...]

PR #3760, opened 2023-09-15
chore(docs): extends docs for contributors

MateuszTrN:
Adds guidelines to the docs

## Description

as disscused with @nikolasrieble here: https://recharts.slack.com/archives/C042Q5K5UDC/p1694458300038969?thread_ts=1693662193.441469&cid=C042Q5K5UDC

added two bullet points int the docs 

## Related Issue
#3407  

## Motivation and Context

Unifies and standardize  branching names 

## Types of changes

## Checklist:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -40,4 +40,5 @@
 - [ ] My change requires a change to the documentation.
 - [ ] I have updated the documentation accordingly.
 - [ ] I have added tests to cover my changes.
+- [ ] I have added a storybook story to show my changes
 - [ ] All new and existing tests passed.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -40,5 +40,5 @@
 - [ ] My change requires a change to the documentation.
 - [ ] I have updated the documentation accordingly.
 - [ ] I have added tests to cover my changes.
-- [ ] I have added a storybook story to show my changes
+- [ ] I have added a storybook story or extended an existing story to show my changes
 - [ ] All new and existing tests passed.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ _Before_ submitting a pull request, please make sure the following is done…
 
 - Search [GitHub](https://github.com/recharts/recharts/pulls) for an open or closed Pull Request that relates to your submission. You don't want to duplicate effort.
 
-- Fork the repo and create your branch from `master` using the branch naming convention: `(bugfix|feature|chore)/i(#ISSUE_ID)-branch-name`. Example: `bugfix/i(#9999)-fixes-very-important-issue`
+- Fork the repo and create your branch from `master`.
 - If you have added functionality or changed existing functionality, be sure to add a test. Ideally a unit test for helper function, or a test that includes rendering with RTL.
 - If you've changed APIs, make sure that the stories in Storybook are working as expected.
 - Ensure the test suite passes (`npm run test`).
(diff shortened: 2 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -40,4 +40,5 @@
 - [ ] My change requires a change to the documentation.
 - [ ] I have updated the documentation accordingly.
 - [ ] I have added tests to cover my changes.
+- [ ] I have added a storybook story to show my changes
```
nikolasrieble, line 43:
```suggestion
- [ ] I have added a storybook story or extended an existing story to show my changes
```

CONTRIBUTING.md
```diff
@@ -48,7 +48,7 @@ _Before_ submitting a pull request, please make sure the following is done…
 
 - Search [GitHub](https://github.com/recharts/recharts/pulls) for an open or closed Pull Request that relates to your submission. You don't want to duplicate effort.
 
-- Fork the repo and create your branch from `master`.
+- Fork the repo and create your branch from `master` using the branch naming convention: `(bugfix|feature|chore)/i(#ISSUE_ID)-branch-name`. Example: `bugfix/i(#9999)-fixes-very-important-issue`
```
nikolasrieble, line 51: What would you say is the main value from having such a suggestion? 
I am open to it, but want to give room for discussion the pros and cons. 

Alternatively to having a convention for the branch name, it might make sense to instead have a convention for the PR name, as we squash and merge by default, where the PR name becomes the commit message on master.

I personally prefer a rule for the PR title, given the context. 

Either way, if we do care about this, we might want to setup a GHA to validate this rules. 

Curious to hear other opinions and some motivation here.

ckifer: Since we squash and merge I agree @nikolasrieble, you could branch off `master` and PR for all it matters to me 😅.

The only thing I _generally_ follow is conventional commit https://www.conventionalcommits.org/en/v1.0.0/#summary messages as a personal preference.

A general PR name format sounds fine to me, but I don't really expect people to follow it unless right there is some sort of automation. We can update the PR template (which we need to do anyways) to add that

MateuszTrN: Hi,
I feel like I made a fool of myself a bit :D 
**tl;dr; yeah this change has no sense**

My thought process I had two reasons: 
I like to keep branch names clean and conventional, cause I use sourceTree as a git tool, and I like have my branches in folders - which sourcetree does which such convention (separates bugs, chores and features) - and it is helpful when more people work on a repo to have it clean. 
`And it would have been right, if it hadn't been for those meddling forks!`
HOWEVER - I didn't put a thought on it, that the way ppl contribute to this repo is by forking so everyone has very limited amount of branches in their repos and they might not care for any of this, which kills my point - I still can keep my fork as I please. 

But my main reason, that I also  have just learnt is not valid, was in the issue page, right pane. It has this "Development" section - which I am used to have branches (now it has only open PRs, if PR is closed it gets empty again - usually my team use Altassian suite, so I'm used to that it always auto-magically links everything if the issue ID is present anywhere) - but I've just realized that here, issue page does not care for forked repos (which is good, I just fell for force of habits and wanted to have there "on going work" - It's just my first time contributing on github so I didn't think it through).

To sum up - I should write this in the "Motivation" section in the PR - but the point was:
Improve 'Development' section in the right pane of the issue page, which I've learnt, is naive of me to think my suggestion would help.

nikolasrieble: @MateuszTrN I still believe this PR to be valuable, as it allows for us to discuss and align this specific topic. A PR is always a great way to get things going, because it is specific. Thats why I appreciate your contribution here ❤️

MateuszTrN: thanks!
I reverted changes to the branching part. I already applied your suggestion regarding stories.

ckifer (approved):
Thanks!!

PR #3764, opened 2023-09-17
Add few types and tests around generateCategoricalChart

PavelVanecek:
## Description

So I am trying to wrap my head around how generateCategoricalChart works and in the process I added some types and tests. Not all types, just some.

Please correct me if I got something wrong - I'm new to the codebase, I have almost definitely missed something.

## Related Issue

## Motivation and Context

Hopefully make maintenance easier

## How Has This Been Tested?

I didn't make any runtime changes - only types and tests. If the runtime code was not compatible with the types I added `@ts-expect-error` instead.

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -307,6 +307,16 @@ const getTooltipData = (state: CategoricalChartState, chartData: any[], layout: 
   return null;
 };
 
+export type AxisInput = {
+  axes: React.DetailedReactHTMLElement<BaseAxisProps & { ticks?: (string | number)[] }, any>[];
+  graphicalItems: React.DetailedReactHTMLElement<any, any>[];
+  axisType: AxisType;
+  axisIdKey: string;
+  stackGroups: false | StackGroups | null;
+  dataStartIndex: number;
+  dataEndIndex: number;
+};
+
 /**
  * Get the configuration of axis by the options of axis instance
  * @param  {Object} props         Latest props
@@ -321,23 +331,7 @@ const getTooltipData = (state: CategoricalChartState, chartData: any[], layout: 
  */
 export const getAxisMapByAxes = (
   props: CategoricalChartProps,
-  {
-    axes,
-    graphicalItems,
-    axisType,
-    axisIdKey,
-    stackGroups,
-    dataStartIndex,
-    dataEndIndex,
-  }: {
-    axes: React.DetailedReactHTMLElement<BaseAxisProps & { ticks?: (string | number)[] }, any>[];
-    graphicalItems: React.DetailedReactHTMLElement<any, any>[];
-    axisType: AxisType;
-    axisIdKey: string;
-    stackGroups: false | StackGroups | null;
-    dataStartIndex: number;
-    dataEndIndex: number;
-  },
+  { axes, graphicalItems, axisType, axisIdKey, stackGroups, dataStartIndex, dataEndIndex }: AxisInput,
 ): AxisMap => {
   const { layout, children, stackOffset } = props;
   const isCategorical = isCategoricalAxis(layout, axisType);
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -59,6 +59,7 @@ import {
   parseSpecifiedDomain,
   parseDomainOfCategoryAxis,
   getTooltipItem,
+  StackGroups,
 } from '../util/ChartUtils';
 import { ActiveBar } from '../util/BarUtils';
 import { detectReferenceElementsDomain } from '../util/DetectReferenceElementsDomain';
@@ -79,6 +80,7 @@ import {
   adaptEventHandlers,
   GeometrySector,
   AxisDomain,
+  AxisType,
 } from '../util/types';
 import { AccessibilityManager } from './AccessibilityManager';
 
@@ -217,7 +219,7 @@ function isDomainSpecifiedByUser(domain: AxisDomain, allowDataOverflow: boolean,
   return false;
 }
 
-function getDefaultDomainByAxisType(axisType: 'number' | string) {
+function getDefaultDomainByAxisType(axisType: BaseAxisProps['type']): [number, string] | undefined {
   return axisType === 'number' ? [0, 'auto'] : undefined;
 }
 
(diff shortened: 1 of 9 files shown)


ckifer:
These are some awesome changes that we've been meaning to do for a long time, thank you! Type safety was ignored quite a bit when this project was moved from JS to TS.

Worried a little bit about merging this as it has a lot of changes to types and this could potentially break consumers downstream who are using our types. _Most likely_ not, but you never know.

I'll finish reviewing when I can and will try to call out anything that might be risky

test/util/PolarUtils.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { RADIAN, degreeToRadian, getMaxRadius, polarToCartesian, radianToDegree } from '../../src/util/PolarUtils';
```
nikolasrieble, line 1: To reduce the blast radius of this PR I suggest too add all tests separately. Type chances can cause pain for customers, tests never do. 

Thus I would be happy to get the tests merged in first, while paying more attention to the changes of types in the review.

test/util/PolarUtils.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    expect(radianToDegree(rad)).toEqual(deg);
+  });
+});
+
+describe('polarToCartesian', () => {
+  it('should convert zero', () => {
+    expect(polarToCartesian(0, 0, 0, 0)).toEqual({ x: 0, y: 0 });
+  });
+
+  it('should convert 1, Pi/2', () => {
```
nikolasrieble, line 20: Nit: Seems like a cases where parametrised tests could be great.

src/util/ChartUtils.ts
```diff
@@ -847,14 +863,16 @@ export const getStackedData = (data: any, stackItems: any, offsetType: string) =
   return stack(data);
 };
 
+export type StackGroups = any;
```
nikolasrieble, line 866: I do not find that this adds value. What is the intention?

src/util/DataUtils.ts
```diff
@@ -12,7 +12,7 @@ export const mathSign = (value: number) => {
 };
 
 export const isPercent = (value: string | number): value is `${number}%` =>
-  _.isString(value) && value.indexOf('%') === value.length - 1;
+  _.isString(value) && value.length > 1 && value.indexOf('%') === value.length - 1;
```
nikolasrieble, line 15: I do believe that many of these changes will actually fix bugs in a few edge cases, and I believe them to be valuable, but I would prefer to see smaller PRs, each with more focus. 

The focus being generateCategoricalChart means it is almost the whole library.

test/util/ChartUtils.spec.tsx
```diff
@@ -191,7 +192,7 @@ describe('getBandSizeOfAxis', () => {
 });
 
 describe('parseSpecifiedDomain', () => {
-  const domain = [20, 100];
+  const domain: [number, number] = [20, 100];
```
nikolasrieble, line 195: It would be nice to have a type 

```
type Domain = [number, number]
```

nikolasrieble, line 195: This would be a SpecifiedDomain, right? 
```suggestion
  const domain: SpecifiedDomain = [20, 100];
```

src/util/PolarUtils.ts
```diff
@@ -39,7 +50,7 @@ export const getMaxRadius = (
  * @return {Object} Configuration
  */
 export const formatAxisMap = (
-  props: any,
+  props: FormatAxisMapProps,
```
nikolasrieble, line 53: The props are the props of axis itself, right? 

```
    Object.keys(axisObj).forEach(key => {
      axisObj[key] = formatAxisMap(props, axisObj[key], offset, key.replace('Map', ''), chartName);
    });
```suggestion
  props: FormatAxisMapProps,
```
```

File: src/util/PolarUtils.ts [hunk 7]
```diff
@@ -29,6 +29,17 @@ export const getMaxRadius = (
     Math.abs(height - (offset.top || 0) - (offset.bottom || 0)),
   ) / 2;
 
+type FormatAxisMapProps = {
```
nikolasrieble, line 32: I think instead of defining this newly, we could reuse the union of PolarAxisProps and CartesianAxisProps.

src/util/ChartUtils.ts
```diff
@@ -574,16 +588,18 @@ export const getCoordinatesOfGrid = (ticks: Array<TickItem>, min: number, max: n
...
-  },
+  axis:
+    | null
+    | (BaseAxisProps & {
+        duplicateDomain?: any;
+        realScaleType?: 'scaleBand' | 'band' | 'point' | 'linear';
+        scale?: any;
+        axisType?: AxisType;
+        ticks?: any;
+        niceTicks?: any;
```
nikolasrieble, line 599: Wow, I have never seen this before. Will have to investigate where and how this is used.

src/chart/generateCategoricalChart.tsx
```diff
@@ -721,6 +788,10 @@ const calculateOffset = (
   };
 };
 
+type AxisMap = {
+  [k: string]: BaseAxisProps;
+};
```
nikolasrieble, line 793: I wonder whether this is correct. Does the axis map not hold all the props for each axis (i.e. cartesian and polar)?

nikolasrieble (commented):
I am very happy about this PR, yet slightly worried about its size. Chances are that we miss some changes that might break something. 

This does add a lot of value over all, and I would be willing to take the risk, and then fix as we go. 

But I would prefer it even more if we could split this up into separate PRs

(1) Addings tests 
(2) Addings types (if possible multiple smaller ones), starting with the Util files.

ckifer:
+1, smaller PRs will make this easier to review and easier for us to identify issues if we find them post merge

src/util/ChartUtils.ts
PavelVanecek: Documentation - sort of "I don't know what this type is exactly but it's the same as over there". Bringing a bit of information into the sea of `any`.

test/chart/generateCategoricalChart.test.ts
```diff
@@ -193,7 +202,7 @@ describe('generateCategoricalChart', () => {
             yAxisId: 0,
             tickCount: 2,
             type: 'number',
-            padding: { left: 0, right: 0 },
+            padding: { bottom: 0, top: 0 },
```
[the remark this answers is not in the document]
PavelVanecek: XAxisProps and YAxisProps have different paddings (which makes sense). I added a type which forced mock data change which forced the test result change.

test/util/ChartUtils.spec.tsx
PavelVanecek: Quite possible - I still don't fully understand the Domain code.

PavelVanecek:
Thx @ckifer and @nikolasrieble , I will split this up into smaller PRs.

src/util/ChartUtils.ts
nikolasrieble: Fair enough. Could you please add a small comment as to what a StackGroup is?

src/util/PolarUtils.ts
PavelVanecek: On closer look - these appear to be `CategoricalChartProps`.

PavelVanecek:
I'm going to go ahead and decline this PR in favour of other smaller ones.

Issue #3738, opened 2023-09-07
Label Story


d-gottlieb:
Create Label story for Storybook following `Line.stories.tsx`.

Using the [old API documentation](https://recharts.org/en-US/api/Label) as a guide, create the Label story.

Acceptance Criteria:

- All information regarding the Label in the old API docs should be used to create the Story.
- All props and their respective categories should either be taken from or added to the props constants.

Nkiriobasi:
Please, can I take up this?

wanisramdani:
WIll work on this

nikolasrieble:
The missing piece to close this, is to add the informational graphics, as mentioned in the comment:  https://github.com/recharts/recharts/pull/3766#pullrequestreview-1631623542

PR #3766, opened 2023-09-18
Storybook: add Label story #3738

wanisramdani:
## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/API/component/Label.stories.tsx
+++ b/storybook/stories/API/component/Label.stories.tsx
@@ -1,0 +1,108 @@
+import React from 'react';
+import { Args } from '@storybook/react';
+import { ResponsiveContainer, Label, Surface } from '../../../../src';
+
+const LabelProps: Args = {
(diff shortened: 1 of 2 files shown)


storybook/stories/API/component/Label.stories.tsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react';
+import { Args } from '@storybook/react';
```
PavelVanecek, line 2: Hi @wanisramdani , thanks for the PR! For the next one you can also use types from here: https://github.com/recharts/recharts/blob/master/storybook/StorybookArgs.ts

No need to change this PR though - if it works it's all good

ckifer (approved):
LGTM, we should totally add the images on this page https://recharts.org/en-US/api/Label

at some point because of how helpful they are, but we should be good for now.

+1 to the comment about using types, but nbd for this PR

Issue #3305, opened 2023-02-01
Evenly spread axis labels


git-anurag-hub:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Currently, We are trying to evenly distance axis labels so that our UI doesn't look congested. We want to have a easier way to do this as this can have many other implications. As we are heavily based on recharts we are willing to contribute in the feature building if required.

### What does the proposed API look like?
There can be a prop/attributes which can be passed to X-axis component to evenly spread the axis labels.

ckifer:
Hi @git-anurag-hub 

Can you please give an example of what you would like vs what you have now?

git-anurag-hub:
Hey @ckifer, 
So this is the behaviour right now I am getting: 
<img width="792" alt="Screenshot 2023-02-02 at 1 22 47 PM" src="https://user-images.githubusercontent.com/58766532/216264697-24365768-a433-4d75-a509-92ff76b52db9.png">

This is the behaviour I want: 
 <img width="792" alt="Screenshot 2023-02-02 at 1 28 50 PM" src="https://user-images.githubusercontent.com/58766532/216265223-24587b80-817a-447d-8b21-fdc2560a6cbe.png">
 
 Right now what I am doing is initialising the new variable `interval` to calculate what will be the best interval value for my x-axis in the graph & then passing it as a prop `interval` to X-axis.
```
	const interval = data.length > 10 ? Math.floor(data.length / 10) : 0;
        .
        .
        <XAxis
	        dataKey="name"
		stroke="white"
		opacity={0.4}
	        interval={interval}
	/>
``` 

This works in work cases my fails sometimes. What I think can happen is that all the labels are spread uniformly with the graph in the best possible way by default (or by user choice).

ckifer:
Do you mind throwing your example into a codesandbox with a dummy similar dataset? 

I can't promise anything on this since we're trying to clean the repo up a bit currently, but any info helps

git-anurag-hub:
Sure!
You can find the implementation here: https://codesandbox.io/s/dazzling-kowalevski-8p341n

Yilun-Sun:
Hi @git-anurag-hub 
You can consider add a `tickFormatter`:
```Javascript
tickFormatter={value => `${value}                                      `}
```
And you will get
[...]

PR #3768, opened 2023-09-20
CartesianAxis: Improve interval option 'equidistantPreserveStart'

nikolasrieble:
## Description

This PR overhauls the existing interval option `equidistantPreserveStart`

In this variant, we always show the first tick, and starting from the first, we find the smallest N, for which all n-th ticks can be shown without collision. 

A collision is defined as two overlapping ticks, taking into account custom tick formatting, a unit, the fontsize, the angle, et cetera (in alignment with the other existing methods). An overlap is identified if the end of the previous tick overlaps with the start of the next one. 

Downside of this method: 
- When the first tick is very long, hardly any other ticks will be shown. 
- Performance: Instead of iterating over the ticks once, as all other methods do, this method iterates at most M times, M being the total amount of available ticks (i.e. for 1000 datapoints, 1000 times). All other methods are thus more performant. This is a good candidate to discuss in review or to iterate upon. 

## Related Issue
https://github.com/recharts/recharts/issues/3305

## Types of changes

## Checklist:


nikolasrieble:
Extracting https://github.com/recharts/recharts/pull/3769 from this PR to have a prefactoring PR without any behavioural change.

src/cartesian/CartesianAxis.tsx
```diff
@@ -44,7 +44,7 @@ export interface CartesianAxisProps {
   hide?: boolean;
   label?: any;
 
-  minTickGap?: number;
+  minTickGap?: TickGap;
```
nikolasrieble, line 47:
```suggestion
  minTickGap?: number;
```

nikolasrieble:
Screenshot 2023-09-21 at 15 39 22

src/cartesian/getEquidistantTicks.ts
```diff
@@ -0,0 +1,55 @@
...
+  getTickSize: (tick: CartesianTickItem, index: number) => number,
+  ticks: CartesianTickItem[],
+  minTickGap: number,
+): CartesianTickItem[] {
+  const result = (ticks || []).slice();
+
+  const { start: initialStart, end } = boundaries;
+  let index = 0;
+  // Premature optimisation idea 1: Estimate a lower bound, and start from there.
+  // For now, start from every tick
```
nikolasrieble, line 18:
```suggestion
  // Premature optimisation idea 1: Estimate a lower bound, and start from there.
  // For now, start from showing every tick
```

nikolasrieble:
@ckifer What do you think about: 
- Should we deprecate `equidistantPreserveStart`?

ckifer:
I don't see an issue with deprecating it really except the fact that it was very recently introduced. I certainly wouldn't prefer it over the others. 

We'd need to add a deprecation warning and then remove it on 3.0

test/cartesian/getEquidistantTicks.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  test.each([
+    // If all ticks fit, we show all ticks
+    { ticksThatFit: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], resultingTicks: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] },
+    { ticksThatFit: [0, 1, 2, 3, 4, 5, 6, 7, 8], resultingTicks: [0, 2, 4, 6, 8] },
+    { ticksThatFit: [0, 1, 2, 3, 4, 5, 6, 7], resultingTicks: [0, 5] },
+    { ticksThatFit: [0, 1, 2, 3, 4, 5, 6], resultingTicks: [0, 5] },
+    { ticksThatFit: [0, 1, 2, 3, 4, 5], resultingTicks: [0, 5] },
+    { ticksThatFit: [0, 1, 2, 3, 4], resultingTicks: [0] },
```
nikolasrieble, line 26: This section deserves scrutiny. Is this the behaviour we want to achieve?

test/cartesian/getEquidistantTicks.spec.ts
```diff
@@ -0,0 +1,40 @@
+// unit test for getEquidisantTicks
+
+import _ from 'lodash';
```
ckifer, line 3: nit: can we remove the dependency from the test by using a for loop below? 

or else at least import the individual lodash dependency as we're trying to do elsewhere

test/cartesian/getEquidistantTicks.spec.ts
ckifer: is this really true in the browser though?

I feel like all of these should be able to fit? every time?

I kind of think we shouldn't always preserve start and instead show as many as we can while maintaining equidistance

ckifer (approved):
Looks good overall, I love how much the prefactoring helped here

test/cartesian/getEquidistantTicks.spec.ts
nikolasrieble: I construct the ticks in the test such that this is true in the browser. 
But good point, if we were to dynamically allow starting from any tick but still only show equidistant ticks, we could show much more.

test/cartesian/getEquidistantTicks.spec.ts
nikolasrieble: Good catch, thank you.

test/cartesian/getEquidistantTicks.spec.ts
nikolasrieble: Really, we could consider to redo the interface completely. As far as I understand it, the options currently are: 

alwaysShowStart | alwaysShowEnd | collisionHandlingStrategy | currentName
--- | --- | --- | ---- 
true | false | keep left | preserveStart
false | true | keep right | preserveEnd 
true | true | keep left | preserveStartEnd
true | false | compute collision to direct neighbour and then select every nTH | equidistantPreserveStart 
true | false | show every nTH without collision | equidistant

We could have the option to choose `equidistant` without `alwaysShowStart`.

I admittedly find the naming of these strategies not that insightful. We should probably investigate how others do this.

nikolasrieble: The more I think about this, I believe we should simply replace the existing `equidistantPreserveStart` with this implementation. 

This would be a breaking change, as it changes existing behaviour, but the change would only be to the better, namely we would show more ticks than we previously did, and there would yet be no collision.

@ckifer What do you think about this approach? 

I like it, because it does not further complicate the API and it improves the existing behaviour (almost a bug fix)

ckifer: I'm on board - of people complain that their chart looks better then... not sure lol. If anything we could do the same thing with preserveEnd if we wanted no?

nikolasrieble: Done. Thanks for the discussion.

christopherkindl: > I'm on board - of people complain that their chart looks better then... not sure lol. If anything we could do the same thing with preserveEnd if we wanted no?

They will look better, Coltin! :P

christopherkindl:
Thanks so much for this improvement from the entire Tremor team!

test/cartesian/getEquidistantTicks.spec.ts
ckifer: Exactly 🚀. If there are complaints I will say but... they look better 😂

PR #3769, opened 2023-09-20
getTicks: Refactor to simplify and reuse functionality between the implementation of getTicksStart and getTicksEnd

nikolasrieble:
## Description
Refactors getTicks.ts to reuse logic which is shared between getTicksStart and getTicksEnd. The goal of this refatoring is to make it easier to add new methods to compute the ticks, i.e. equidistant ticks. 

Improvements to TickPositioning story:
- Resize (important to test how different interval options lead to ticks disappearing) 
- Add the interval option under the axis as a title

Additionally comments and new types wherever useful. 

## Related Issue
https://github.com/recharts/recharts/issues/3305

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/CartesianAxis.tsx
+++ b/src/cartesian/CartesianAxis.tsx
@@ -25,7 +25,7 @@ export type Orientation = 'top' | 'bottom' | 'left' | 'right';
 export type Unit = string | number;
 /** The formatter function of tick */
 export type TickFormatter = (value: any, index: number) => string;
-
+/** The minimum gap between two adjacent labels. */
 export type TickGap = number;
 
 export interface CartesianAxisProps {
--- a/src/cartesian/CartesianAxis.tsx
+++ b/src/cartesian/CartesianAxis.tsx
@@ -25,8 +25,6 @@ export type Orientation = 'top' | 'bottom' | 'left' | 'right';
 export type Unit = string | number;
 /** The formatter function of tick */
 export type TickFormatter = (value: any, index: number) => string;
-/** The minimum gap between two adjacent labels. */
-export type TickGap = number;
 
 export interface CartesianAxisProps {
   className?: string;
@@ -46,7 +44,7 @@ export interface CartesianAxisProps {
   hide?: boolean;
   label?: any;
 
-  minTickGap?: TickGap;
+  minTickGap?: number;
   ticks?: CartesianTickItem[];
   tickSize?: number;
   tickFormatter?: TickFormatter;
--- a/src/cartesian/CartesianAxis.tsx
+++ b/src/cartesian/CartesianAxis.tsx
@@ -19,14 +19,23 @@ import {
 import { filterProps } from '../util/ReactUtils';
 import { getTicks } from './getTicks';
 
+/** The orientation of the axis in correspondence to the chart */
+export type Orientation = 'top' | 'bottom' | 'left' | 'right';
+/** A unit to be appended to a value */
+export type Unit = string | number;
+/** The formatter function of tick */
+export type TickFormatter = (value: any, index: number) => string;
+
+export type TickGap = number;
+
 export interface CartesianAxisProps {
   className?: string;
   x?: number;
   y?: number;
   width?: number;
   height?: number;
-  unit?: string | number;
-  orientation?: 'top' | 'bottom' | 'left' | 'right';
+  unit?: Unit;
+  orientation?: Orientation;
   // The viewBox of svg
   viewBox?: CartesianViewBox;
   tick?: SVGProps<SVGTextElement> | ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>) | boolean;
@@ -37,11 +46,10 @@ export interface CartesianAxisProps {
   hide?: boolean;
   label?: any;
 
-  minTickGap?: number;
+  minTickGap?: TickGap;
   ticks?: CartesianTickItem[];
   tickSize?: number;
-  /** The formatter function of tick */
-  tickFormatter?: (value: any, index: number) => string;
+  tickFormatter?: TickFormatter;
   ticksGenerator?: (props?: CartesianAxisProps) => CartesianTickItem[];
   interval?: AxisInterval;
   /** Angle in which ticks will be rendered. */
--- a/src/cartesian/getTicks.ts
+++ b/src/cartesian/getTicks.ts
@@ -6,7 +6,7 @@ import { Global } from '../util/Global';
 import {
   doesTickFitInBetweenStartAndEnd,
   getEveryNThTick,
-  getInitialStartAndEnd,
+  getTickBoundaries,
   getNumberIntervalTicks,
   getSizeOfTick,
 } from '../util/TickUtils';
@@ -129,7 +129,7 @@ export function getTicks(props: CartesianAxisProps, fontSize?: string, letterSpa
   };
 
   const sign = ticks.length >= 2 ? mathSign(ticks[1].coordinate - ticks[0].coordinate) : 1;
-  const boundaries = getInitialStartAndEnd(viewBox, sign, sizeKey);
+  const boundaries = getTickBoundaries(viewBox, sign, sizeKey);
 
   if (interval === 'equidistantPreserveStart') {
     candidates = getTicksStart(sign, boundaries, getTickSize, ticks, minTickGap);
--- a/src/cartesian/getTicks.ts
+++ b/src/cartesian/getTicks.ts
@@ -18,7 +18,7 @@ function getTicksEnd(
   boundaries: { start: number; end: number },
   getTickSize: (tick: CartesianTickItem, index: number) => number,
   ticks: CartesianTickItem[],
-  minTickGap: CartesianAxisProps['minTickGap'],
+  minTickGap: number,
 ): CartesianTickItem[] {
   const result = (ticks || []).slice();
   const len = result.length;
@@ -55,7 +55,7 @@ function getTicksStart(
   boundaries: { start: number; end: number },
   getTickSize: (tick: CartesianTickItem, index: number) => number,
   ticks: CartesianTickItem[],
-  minTickGap: CartesianAxisProps['minTickGap'],
+  minTickGap: number,
   preserveEnd?: boolean,
 ): CartesianTickItem[] {
   const result = (ticks || []).slice();
--- a/src/cartesian/getTicks.ts
+++ b/src/cartesian/getTicks.ts
@@ -4,7 +4,7 @@ import { getStringSize } from '../util/DOMUtils';
 import { Props as CartesianAxisProps } from './CartesianAxis';
 import { Global } from '../util/Global';
 import {
-  doesTickFitInBetweenStartAndEnd,
+  isVisible,
   getEveryNThTick,
   getTickBoundaries,
   getNumberIntervalTicks,
@@ -39,7 +39,7 @@ function getTicksEnd(
       result[i] = entry = { ...entry, tickCoord: entry.coordinate };
     }
 
-    const isShow = doesTickFitInBetweenStartAndEnd(sign, entry.tickCoord, size, start, end);
+    const isShow = isVisible(sign, entry.tickCoord, size, start, end);
 
     if (isShow) {
       end = entry.tickCoord - sign * (size / 2 + minTickGap);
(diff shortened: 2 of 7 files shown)


src/cartesian/CartesianAxis.tsx
```diff
@@ -19,14 +19,23 @@ import {
 import { filterProps } from '../util/ReactUtils';
 import { getTicks } from './getTicks';
 
+/** The orientation of the axis in correspondence to the chart */
+export type Orientation = 'top' | 'bottom' | 'left' | 'right';
+/** A unit to be appended to a value */
+export type Unit = string | number;
+/** The formatter function of tick */
+export type TickFormatter = (value: any, index: number) => string;
+
```
nikolasrieble, line 28:
```suggestion
/** The minimum gap between two adjacent labels. */
```

src/cartesian/CartesianAxis.tsx
```diff
@@ -37,11 +46,10 @@ export interface CartesianAxisProps {
   hide?: boolean;
   label?: any;
 
-  minTickGap?: number;
+  minTickGap?: TickGap;
```
MateuszTrN, line 49: why this type is introduced? so far I see it's only alias for number and  when you will hit suggestion prompt in any IDE imho (personal preference) it's easier to wrap your head around when it just hint you that is a number instead some kind of custom type.
even if there are plans to expand it like for example: 
`TickGap = number | string`  or anything like that - I still would like this notation over a custom type, that I don't know and have to see it's definition first to understand it. 
But maybe there is some specific use case that I don't know about and this solves some other issue?

nikolasrieble: No, good point. I thought this would help storybook pick up comments above the Type from a single place, but it turned out to not work 🤦 

Will remove the alias.

src/cartesian/CartesianAxis.tsx
```diff
@@ -19,14 +19,21 @@ import {
 import { filterProps } from '../util/ReactUtils';
 import { getTicks } from './getTicks';
 
+/** The orientation of the axis in correspondence to the chart */
+export type Orientation = 'top' | 'bottom' | 'left' | 'right';
+/** A unit to be appended to a value */
```
ckifer, line 24: nit: a unit could also be a prefix (money)

src/cartesian/getTicks.ts
```diff
@@ -1,81 +1,34 @@
...
-    unit && sizeKey === 'width' ? getStringSize(unit, { fontSize, letterSpacing }) : { width: 0, height: 0 };
+import {
+  doesTickFitInBetweenStartAndEnd,
+  getEveryNThTick,
+  getInitialStartAndEnd,
+  getNumberIntervalTicks,
+  getSizeOfTick,
+} from '../util/TickUtils';
+
+type Sign = 0 | 1 | -1;
```
ckifer, line 14: nit: what is this? from just reading the code I can't really tell

src/cartesian/getTicks.ts
```diff
@@ -86,9 +39,7 @@ function getTicksEnd({
       result[i] = entry = { ...entry, tickCoord: entry.coordinate };
     }
 
-    const isShow =
-      sign * (entry.tickCoord - (sign * size) / 2 - start) >= 0 &&
-      sign * (entry.tickCoord + (sign * size) / 2 - end) <= 0;
+    const isShow = doesTickFitInBetweenStartAndEnd(sign, entry.tickCoord, size, start, end);
```
ckifer, line 42: nit: i know this was here before but can we call this `isVisible` or something a little easier to read?

src/cartesian/getTicks.ts
ckifer: Also now that I looked at `mathSign`, it seems like a really silly function? https://github.com/recharts/recharts/blob/master/src/util/DataUtils.ts#L3C14-L12

src/cartesian/getTicks.ts
```diff
@@ -100,56 +51,29 @@ function getTicksEnd({
...
     const tailGap = sign * (tail.coordinate + (sign * tailSize) / 2 - end);
     result[len - 1] = tail = {
       ...tail,
       tickCoord: tailGap > 0 ? tail.coordinate - tailGap * sign : tail.coordinate,
     };
 
-    const isTailShow =
-      sign * (tail.tickCoord - (sign * tailSize) / 2 - start) >= 0 &&
-      sign * (tail.tickCoord + (sign * tailSize) / 2 - end) <= 0;
+    const isTailShow = doesTickFitInBetweenStartAndEnd(sign, tail.tickCoord, tailSize, start, end);
```
ckifer, line 76: nit: same here, rename?

src/cartesian/getTicks.ts
```diff
@@ -176,9 +96,7 @@ function getTicksStart(
       result[i] = entry = { ...entry, tickCoord: entry.coordinate };
     }
 
-    const isShow =
-      sign * (entry.tickCoord - (sign * size) / 2 - start) >= 0 &&
-      sign * (entry.tickCoord + (sign * size) / 2 - end) <= 0;
+    const isShow = doesTickFitInBetweenStartAndEnd(sign, entry.tickCoord, size, start, end);
```
ckifer, line 99: nit: same here, rename?

src/cartesian/getTicks.ts
```diff
@@ -202,49 +120,27 @@ export function getTicks(props: CartesianAxisProps, fontSize?: string, letterSpa
 
   let candidates: CartesianTickItem[] = [];
 
+  const sizeKey = orientation === 'top' || orientation === 'bottom' ? 'width' : 'height';
+  const unitSize: Size =
+    unit && sizeKey === 'width' ? getStringSize(unit, { fontSize, letterSpacing }) : { width: 0, height: 0 };
+
+  const getTickSize = (content: CartesianTickItem, index: number) => {
+    return getSizeOfTick(tickFormatter, content, index, sizeKey, fontSize, letterSpacing, unitSize, angle);
```
ckifer, line 128: this feels slightly confusing, why wrap a function with a function? Concerned this reduces readability of this file when the goal is the opposite. The rest looks good (and I know this is needed for that), but maybe we can look at this again?

src/util/TickUtils.ts
```diff
@@ -0,0 +1,78 @@
+import _ from 'lodash';
+
+import { TickFormatter } from '../cartesian/CartesianAxis';
+import { getAngledRectangleWidth } from './CartesianUtils';
+import { getStringSize } from './DOMUtils';
+import { getEveryNthWithCondition } from './getEveryNthWithCondition';
+import { Size, CartesianViewBox, CartesianTickItem } from './types';
+
+export function getAngledTickWidth(contentSize: Size, unitSize: Size, angle: number) {
+  const size = { width: contentSize.width + unitSize.width, height: contentSize.height + unitSize.height };
```
ckifer, line 10: sanity check: ideally these won't be NaN or undefined, but calling code needs to ensure that

src/util/TickUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  return getAngledRectangleWidth(size, angle);
+}
+
+export function getInitialStartAndEnd(viewBox: CartesianViewBox, sign: number, sizeKey: string) {
```
ckifer, line 15: nit: name not super clear - getInitialStartAndEnd what? Start and end ticks?

src/util/TickUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+  let start, end;
+
+  const { x, y, width, height } = viewBox;
+  if (sign === 1) {
+    start = sizeKey === 'width' ? x : y;
+    end = sizeKey === 'width' ? x + width : y + height;
+  } else {
+    start = sizeKey === 'width' ? x + width : y + height;
+    end = sizeKey === 'width' ? x : y;
```
ckifer, line 24: nit: can we move the check `sizeKey === 'width'` to a variable?

src/util/TickUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+export function getSizeOfTick(
+  tickFormatter: TickFormatter,
+  entry: CartesianTickItem,
+  index: number,
+  sizeKey: 'width' | 'height',
+  fontSize: string | number,
+  letterSpacing: string | number,
+  unitSize: Size,
+  angle: number,
```
ckifer, line 36: nit: maybe move this input to an object since its getting so busy here for such a small function?

src/util/TickUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+) {
+  const content = _.isFunction(tickFormatter) ? tickFormatter(entry.value, index) : entry.value;
+  // Recharts only supports angles when sizeKey === 'width'
+  return sizeKey === 'width'
+    ? getAngledTickWidth(getStringSize(content, { fontSize, letterSpacing }), unitSize, angle)
+    : getStringSize(content, { fontSize, letterSpacing })[sizeKey];
+}
+
+export function doesTickFitInBetweenStartAndEnd(
```
ckifer, line 45: nit wording: doesTickFitInBetweenStartAndEnd what? does tick fit between start and end tick?

src/util/TickUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+  sign: number,
+  tickPosition: number,
+  size: number,
+  start: number,
+  end: number,
```
ckifer, line 50: nit: lot of input in such a small function again, can maybe use object type and destructure in the function here too?

src/util/types.ts
```diff
@@ -1114,6 +1114,13 @@ export interface BaseAxisProps {
   className?: string;
 }
 
+/** Defines how ticks are places and whether / how tick collisions are handled.
```
ckifer, line 1117: nit nit, typo: 'placed'

ckifer (approved):
LGTM 🚀 

Even though I added lots of comments 99% of it is non-blocking as this is totally the right direction. I think we should try to limit parameters to our functions somehow - after 4 or 5 it gets difficult to keep track of what is what. Maybe moving to object parameters could help that so at least they must be named when calling? idk

src/util/TickUtils.ts
nikolasrieble: Start and end are numbers, which are updated as we traverse the ticks either from the end of the axis towards the start (for the strategy preserveEnd), or from the start to the end.

src/cartesian/getTicks.ts
nikolasrieble: It is called currying and basically partially applies a function. https://javascript.info/currying-partials

The idea is that now I only pass this function down, as opposed to passing down:
- tickFormatter
- sizeKey
- fontSize
- letterSpacing
- unitSize
- angle

nikolasrieble: It ihmo improves readability downstream, where we only call `getStringSize(content, index)``

src/cartesian/CartesianAxis.tsx
nikolasrieble: It theoretically could, but thats not how it is applied in recharts, right? We only append the unit as far as I can see.

src/cartesian/getTicks.ts
nikolasrieble: It is a really silly function 🤦 But I need to time box the refactoring and not explode scope. This totally deserves multiple iterations. 

The sign is an implementation detail for whether we position ticks from left to right or from right to left.

src/util/TickUtils.ts
nikolasrieble: It gets the tick coordinate bounds in between all ticks must fit.

src/util/TickUtils.ts
nikolasrieble: I only moved this function from one file to another. I would prefer to not extend this. 
I do agree with the sentiment though.

src/util/TickUtils.ts
nikolasrieble: Good idea, done.

src/util/TickUtils.ts
nikolasrieble: I decided to rewrite this such that this function no longer exists. This did indeed smell funky. 😶‍🌫️

src/cartesian/CartesianAxis.tsx
ckifer: Yep, just maybe something to support in the future

src/cartesian/getTicks.ts
ckifer: Makes sense

src/cartesian/getTicks.ts
ckifer: Yep makes sense. Sorry was kind of hard to read in the code review format

ckifer:
🚀

Issue #3765, opened 2023-09-17
Consistent interface: Line / Bar / Funnel - activeDot / activeBar / activeShape


andrewangelle:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-ms3vpd?file=/src/index.js)

### Steps to reproduce
Hover over a section of the funnel. Notice activeShape prop is ignored

### What is expected?
I'd expect the customized active shape to render according to the active tooltip the same way that activeDot/Line and activeBar/Bar work respectively.

### What is actually happening?
The activeShape prop is being ignored

| Environment | Info |
|---|---|
| Recharts | v2.8.0 |
| React | 18.2.0 |
| System | macOS Ventura 13.5.2 |
| Browser | Chrome v116.0.5845.179 |

andrewangelle:
I'm working on a fix for this if you want to assign it to me. @ckifer @nikolasrieble

ckifer:
🚀

PR #3772, opened 2023-09-21
Funnel: activeShape should work with Tooltip

andrewangelle:
The existing `activeShape` prop on `<Funnel />` should work with the `<Tooltip />` in the same manner as `<Line />` and `<Bar />` do.

## Description

  - Added another if block to render the graphical item when Tooltip is used without XAxis/YAxis
  - Inside the new if block, render the graphical item and pass in `activeTooltipIndex` which is calculated by matching  `activeItem.payload.value` to `item.props.data[index].value`
  
- #### src/numberAxis/Funnel.tsx
  - Renamed the generically exported type named `Props` to `FunnelProps`
  - Refactored static render method out to FunnelUtils
  - Updated type of `FunnelProps.activeShape` to match the signature of activeLine/activeBar in Line/Bar components
  - Updated the logic for calculating the active index to take into account `props.activeTooltipIndex` in addition to `props.activeIndex`. The former will get passed in from `generateCategoricalChart`
  
  #### Note: 
  `activeShape` and `activeIndex` already exist on Funnel, which means a user could be defining both on their own. My goal was to maintain that behavior while also having `activeShape` be respected by the shape being interacted with by a `<Tooltip />`. See screenshot below for a visual.
  
## Related Issue

#3765 
## Motivation and Context
To provide a consistent interface between Line / Bar / Funnel and their respective props of activeDot / activeBar / activeShape

## How Has This Been Tested?
Unit tests and storybook

Screenshot 2023-09-21 at 8 46 12 AM

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -2126,12 +2126,12 @@ export const generateCategoricalChart = ({
         return null;
       }
       const tooltipEventType = this.getTooltipEventType();
-      const { isTooltipActive, tooltipAxis, activeTooltipIndex, activeLabel, activeItem } = this.state;
+      const { isTooltipActive, tooltipAxis, activeTooltipIndex, activeLabel } = this.state;
       const { children } = this.props;
       const tooltipItem = findChildByType(children, Tooltip);
       const { points, isRange, baseLine } = item.props;
       const { activeDot, hide, activeBar, activeShape } = item.item.props;
-      const hasActive = !hide && isTooltipActive && tooltipItem && (activeDot || activeBar || activeShape);
+      const hasActive = Boolean(!hide && isTooltipActive && tooltipItem && (activeDot || activeBar || activeShape));
       let itemEvents = {};
 
       if (tooltipEventType !== 'axis' && tooltipItem && tooltipItem.props.trigger === 'click') {
@@ -2193,10 +2193,13 @@ export const generateCategoricalChart = ({
            *
            * An example usage case is a FunnelChart
            */
-          const { data: itemData = [] } = graphicalItem.props;
-          const activeData = itemData.find((datum: { value: number }) => datum.value === activeItem.payload.value);
-          const activeIndex = itemData.indexOf(activeData);
-          return [cloneElement(element, { ...item.props, ...itemEvents, activeTooltipIndex: activeIndex }), null, null];
+          const {
+            graphicalItem: { item: xyItem, childIndex },
+          } = this.getItemByXY(this.state.activeCoordinate);
+
+          const elementProps = { ...item.props, ...itemEvents, activeTooltipIndex: childIndex };
+
+          return [cloneElement(xyItem, elementProps), null, null];
         }
       }
 
@@ -2264,8 +2267,7 @@ export const generateCategoricalChart = ({
     }
 
     public getItemByXY(chartXY: { x: number; y: number }) {
-      const { formattedGraphicalItems } = this.state;
-
+      const { formattedGraphicalItems, activeItem } = this.state;
       if (formattedGraphicalItems && formattedGraphicalItems.length) {
         for (let i = 0, len = formattedGraphicalItems.length; i < len; i++) {
           const graphicalItem = formattedGraphicalItems[i];
@@ -2290,6 +2292,33 @@ export const generateCategoricalChart = ({
             if (activeBarItem) {
               return { graphicalItem, payload: activeBarItem };
             }
+          } else if (itemDisplayName === 'Funnel') {
+            /*
+             * To handle possible duplicates in the data set,
+             * match both the data value of the active item to a data value on a graph item,
+             * and match the mouse coordinates of the active item to the coordinates of a trapezoid.
+             * This assumes equal lengths of trapezoids on a funnel to data items.
+             */
+            const activeIndex = item.props.data.findIndex((funnelData: { value: string }, dataIndex: number) => {
+              const valuesMatch = funnelData.value === activeItem.payload.value;
+
+              const indexOfMouseCoordinates = graphicalItem.props.trapezoids.findIndex(
+                (trapezoid: { x: number; y: number }) => {
+                  const xMatches = trapezoid.x === activeItem.labelViewBox.x || trapezoid.x === activeItem.x;
+                  const yMatches = trapezoid.y === activeItem.labelViewBox.y || trapezoid.y === activeItem.y;
+                  return xMatches && yMatches;
+                },
+              );
+
+              const coordinatesMatch = dataIndex === indexOfMouseCoordinates;
+
+              return valuesMatch && coordinatesMatch;
+            });
+
+            return {
+              graphicalItem: { ...graphicalItem, childIndex: activeIndex },
+              payload: graphicalItem.props.data[activeIndex],
+            };
           }
         }
       }
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -2197,7 +2197,7 @@ export const generateCategoricalChart = ({
             graphicalItem: { item: xyItem, childIndex },
           } = this.getItemByXY(this.state.activeCoordinate);
 
-          const elementProps = { ...item.props, ...itemEvents, activeTooltipIndex: childIndex };
+          const elementProps = { ...item.props, ...itemEvents, activeIndex: childIndex };
 
           return [cloneElement(xyItem, elementProps), null, null];
         }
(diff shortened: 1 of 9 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -2152,7 +2153,7 @@ export const generateCategoricalChart = ({
         return typeof tooltipAxis.dataKey === 'function' ? tooltipAxis.dataKey(entry.payload) : null;
       }
 
-      if (hasActive) {
+      if (hasActiveWithAxisDefined) {
```
andrewangelle, line 2156: The calculation is:

```typescript
      const hasActive = !hide && isTooltipActive && tooltipItem && (activeDot || activeBar || activeShape);
      
      const hasActiveWithAxisDefined = hasActive && activeTooltipIndex >= 0;
      
if (hasActiveWithAxisDefined) 
```
Admittedly, this feels off because it's based on `this.state.activeTooltipIndex` being defined which implicitly gets set by the mouse events via `<XAxis/>` and `<YAxis />`. Though, I'm not sure if the problem is with my naming choice or with the way I'm thinking about the problem.

src/chart/generateCategoricalChart.tsx
```diff
@@ -2184,6 +2185,18 @@ export const generateCategoricalChart = ({
...
+         * We hit this block if consumer uses a Tooltip without XAxis and/or YAxis.
+         * In which case, this.state.activeTooltipIndex never gets set
+         * because the mouse events that trigger that value getting set never get trigged without the axis components.
+         *
+         * See FunnelChart story for an example use case
+         */
+        const { data: itemData = [] } = graphicalItem.props;
+        const activeData = itemData.find((datum: { value: number }) => datum.value === activeItem.payload.value);
+        const activeIndex = itemData.indexOf(activeData);
+        return [cloneElement(element, { ...item.props, ...itemEvents, activeTooltipIndex: activeIndex }), null, null];
```
andrewangelle, line 2199: I would have rathered set the `activeIndex` prop. However, in the use case of the Funnel `activeIndex` was pre-existing. So setting that to `activeTooltipIndex` would override the prop value if consumer is already passing in `activeIndex` to Funnel.

src/chart/generateCategoricalChart.tsx
nikolasrieble: I would find it easier to read / understand if it was 
```

const hasActive = !hide && isTooltipActive && tooltipItem && (activeDot || activeBar || activeShape);
      
const hasActiveWithAxisDefined = hasActive && activeTooltipIndex >= 0;
      
if (hasActive) {

  if(activeTooltipIndex >= 0){
     ...
  }
  else {
     ...
  }
  ...
}
```

src/chart/generateCategoricalChart.tsx
```diff
@@ -2184,6 +2185,18 @@ export const generateCategoricalChart = ({
             }),
           ];
         }
+      } else if (hasActive) {
+        /**
+         * We hit this block if consumer uses a Tooltip without XAxis and/or YAxis.
+         * In which case, this.state.activeTooltipIndex never gets set
+         * because the mouse events that trigger that value getting set never get trigged without the axis components.
+         *
+         * See FunnelChart story for an example use case
```
nikolasrieble, line 2194: I suggest to remove this reference, because (a) it is not specific enough (there might be many stories), and (b) we will most certainly forget to remove this if we ever remove the story. 
```suggestion
         * An example usage case would be a FunnelChart
```

src/chart/generateCategoricalChart.tsx
```diff
@@ -2184,6 +2185,18 @@ export const generateCategoricalChart = ({
...
+         */
+        const { data: itemData = [] } = graphicalItem.props;
+        const activeData = itemData.find((datum: { value: number }) => datum.value === activeItem.payload.value);
+        const activeIndex = itemData.indexOf(activeData);
```
nikolasrieble, line 2198: How about duplicates in the data? 🤔

src/numberAxis/Funnel.tsx
```diff
@@ -23,24 +23,25 @@ import {
...
   dataKey: DataKey<any>;
   nameKey?: DataKey<any>;
   data?: any[];
   hide?: boolean;
-  activeShape?: ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>) | TrapezoidProps;
+  activeShape?:
+    | ReactElement<SVGProps<SVGPathElement>>
+    | ((props: FunnelTrapezoidItem) => ReactElement<SVGProps<SVGPathElement>>)
+    | SVGProps<SVGPathElement>
+    | boolean;
```
nikolasrieble, line 44: We could extract a generic

```
type ActiveShape<T> = 
    | ReactElement<SVGProps<SVGPathElement>>
    | ((props: T) => ReactElement<SVGProps<SVGPathElement>>)
    | SVGProps<SVGPathElement>
    | boolean;
```

And reuse it in all places, right?

src/util/FunnelUtils.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return {
+    ...props,
+    ...option,
+    height,
+    x,
+    y,
+  };
+}
+
+export function FunnelTrapezoid({ option, ...props }: { option: FunnelProps['activeShape'] } & FunnelTrapezoidItem) {
```
nikolasrieble, line 27: Just wonderful.

nikolasrieble (commented):
A great PR. Not only did you extend and unify the API, but also did you improve documentation, and refactor in a nice and concise manner. Lovely, thank you for your contribution. 

There are few minor questions I have, but nothing blocking. Curious to hear your perspective, and then lets 🚢

src/chart/generateCategoricalChart.tsx
andrewangelle: Nice catch. I've refactored how I'm calculating the active index, and added a duplicate in the data set for FunnelChart story. Please give my changes a look.

src/numberAxis/Funnel.tsx
andrewangelle: So glad you suggested this! I started to do this previously but stopped. I've added this change and updated the other types to use it.

andrewangelle:
> Should the Tooltip still work, if the activeIndex is set by the user? I wonder whether there should be some hierarchy between the two.
> 
> How else would I as a developer achieve that the activeIndex is set from some other mechanism and can not be overwritten by user action?
> 
> Screenshot 2023-09-22 at 15 29 20

@nikolasrieble 
I think I see what you're saying. So from the user's perspective, the assumption would be that if I'm setting the `activeIndex` then I'm explicitly saying this should always be the active item regardless of other interactions. 

That's a tricky one to gauge.

I see our options as:
- User's value for `activeIndex` takes precedent
  - If I'm setting both `activeItem` and `activeIndex` and rendering a `<Tooltip />` it is possible that the intention is for the tooltip content/payload to be displayed while the first two props are rendering customized content according to some other user state outside of the chart. 
  - Does this suggest that the tooltip should not render the content inside of activeItem under these conditions? It's an easy enough fix. I would just check the existence of the prop before setting an index in the tooltip.
  
- Tooltip's value for `activeIndex` takes precedent, and falls back to user's value in between interactions.

- Extend the api to have separate props for `activeIndex`/`activeItem`/`activeTooltipIndex`/`activeTooltipItem`

src/numberAxis/Funnel.tsx
```diff
@@ -56,9 +54,12 @@ interface FunnelProps extends InternalFunnelProps {
   animationDuration?: AnimationDuration;
   animationEasing?: AnimationTiming;
   id?: string;
+  trapezoids?: FunnelTrapezoidItem[];
+  animationId?: number;
+  activeTooltipIndex?: number;
```
nikolasrieble, line 59: So, if I understand this correctly, we are adding the activeTooltipIndex (and thereby increase complexity of the API) for the reason of not breaking existing implementations. 

The alternative would be to simply set the activeIndex internally, which is already exposed. 

How about only overwriting the activeIndex if it is not set externally, without exposing the activeTooltipIndex? Would that work?

nikolasrieble: I am a bit hesitant to introduce the activeTooltipIndex, because I find it complicates the API further (which is already not trivial). I would prefer if we found a way that could not extend the API, but yet reach feature parity.

nikolasrieble: What I find interesting in this context is that activeIndex is only exposed in a few cases: Funnel, Scatter, RadialBar, Pie. 

The decision we make here will carry weight downstream, as we unify the API further.

andrewangelle: > How about only overwriting the activeIndex if it is not set externally, without exposing the activeTooltipIndex? Would that work?

> I am a bit hesitant to introduce the activeTooltipIndex, because I find it complicates the API further (which is already not trivial). I would prefer if we found a way that could not extend the API, but yet reach feature parity.

Yes, this all makes sense to me. Just pushed up that change.

nikolasrieble:
I tested this on final time, and it works just fine! ❤️ Thank you!

PR #3773, opened 2023-09-21
getLegendProps tests

PavelVanecek:
## Description

Added unit tests for one function.

## Related Issue

## Motivation and Context

Adding tests and types one step at a time.

## How Has This Been Tested?

npm run test

## Types of changes

## Checklist:

### DIFF
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -177,7 +177,7 @@ export const getLegendProps = ({
   legendWidth,
   legendContent,
 }: {
-  children: any;
+  children: ReactNode[];
   formattedGraphicalItems?: Array<FormattedGraphicalItem>;
   legendWidth: number;
   legendContent?: any;
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -180,7 +180,7 @@ export const getLegendProps = ({
   children: ReactNode[];
   formattedGraphicalItems?: Array<FormattedGraphicalItem>;
   legendWidth: number;
-  legendContent?: any;
+  legendContent?: 'children';
 }): LegendProps & { item: ReactElement } => {
   const legendItem = findChildByType(children, Legend);
   if (!legendItem) {
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -17,7 +17,8 @@ import { findEntryInArray, getPercentValue, isNumber, isNumOrStr, mathSign, uniq
 import { filterProps, findAllByType, findChildByType, getDisplayName } from './ReactUtils';
 // TODO: Cause of circular dependency. Needs refactor.
 // import { RadiusAxisProps, AngleAxisProps } from '../polar/types';
-import { AxisType, BaseAxisProps, DataKey, LayoutType, PolarLayoutType, TickItem } from './types';
+import { AxisType, BaseAxisProps, DataKey, LayoutType, LegendType, PolarLayoutType, TickItem } from './types';
+import { Payload as LegendPayload } from '../component/DefaultLegendContent';
 
 export function getValueByDataKey<T>(obj: T, dataKey: DataKey<any>, defaultValue?: any) {
   if (_.isNil(obj) || _.isNil(dataKey)) {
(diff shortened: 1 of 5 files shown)


test/util/ChartUtils.getLegendProps.spec.tsx
```diff
@@ -0,0 +1,494 @@
+import React, { ReactNode } from 'react';
```
ckifer, line 1: a little confused at the file naming convention here. You're just doing this because `ChartUtils` is a giant file?

Tbh I might prefer a folder approach `ChartUtils/getLegendProps.spec.tsx`

ckifer (changes requested):
One small thing, though if there's a good reason to leave as is I'll approve too

test/util/ChartUtils.getLegendProps.spec.tsx
PavelVanecek: This is because I added a mock `jest.mock('../../src/util/ReactUtils');` and that broke the other tests. I would prefer to have functions and utils not read things from a DOM but that's going to take some time until we get there.

Also yes it's a giant file with giant tests.

That said, it should work the same in a folder, I can change that.

ckifer: Lol yep you and me both. 

Thanks!

src/util/ChartUtils.ts
```diff
@@ -165,10 +166,18 @@ export const getMainColorOfGraphicItem = (item: ReactElement) => {
...
-interface FormattedGraphicalItem {
-  props: any;
+export interface FormattedGraphicalItem {
+  props: {
+    sectors?: ReadonlyArray<any>;
+    data?: ReadonlyArray<any>;
+  };
   childIndex: number;
-  item: any;
+  item: ReactElement<{ legendType?: LegendType; hide: boolean; name?: string; dataKey: unknown }>;
```
nikolasrieble, line 175: Love the usage on unknown instead of any - but we could also use the DataKey type here, right? 
from `src/util/types.ts`
``` 
export type DataKey<T> = string | number | ((obj: T) => any);
```

src/util/ChartUtils.ts
```diff
@@ -213,6 +222,7 @@ export const getLegendProps = ({
         type: legendItem.props.iconType || legendType || 'square',
         color: getMainColorOfGraphicItem(item),
         value: name || dataKey,
+        // @ts-expect-error property strokeDasharray is required in Payload but optional in props
```
nikolasrieble, line 225: 😓 Do you think we could provide a reasonable default here if it is not passed? Could we even reuse an existing default?

src/util/ChartUtils.ts
```diff
@@ -165,10 +166,18 @@ export const getMainColorOfGraphicItem = (item: ReactElement) => {
...
+    data?: ReadonlyArray<any>;
+  };
   childIndex: number;
-  item: any;
+  item: ReactElement<{ legendType?: LegendType; hide: boolean; name?: string; dataKey: unknown }>;
+}
+
+interface SectorOrDataEntry {
+  name: any;
+  fill: any;
```
nikolasrieble, line 180: Could we not use any? (But also, the whole PR is already a great improvement, so this is fine, just wondering)

src/util/ChartUtils.ts
nikolasrieble, line 1: Nit: What do you think about moving getLegendProps into a separate file? I personally prefer things to live separately, rather than living in generically named files.

test/util/ChartUtils.getLegendProps.spec.tsx
nikolasrieble: The alternative is to just take getLegendProps out into a separate file.

src/util/ChartUtils.ts
PavelVanecek: I have seen in multiple occasions in this codebase that variable with same name is completely different time so I am hesitating to assume. I will update the type if I track down that it's actually passing the `DataKey`.

src/util/ChartUtils.ts
PavelVanecek: I wanted to avoid runtime changes.

src/util/ChartUtils.ts
PavelVanecek: Yep definitely - I just wanted to keep things limited in one PR.

src/util/ChartUtils.ts
PavelVanecek: So I am a fan of the idea - however can that be a breaking change? I believe the function is not exposed in the official public API and is not documented, but people can still import it directly, right?

It might be paranoid but when moving it I would also re-export it from `ChartUtils.ts` anyway.

PavelVanecek:
> Great work, thank you! This is very valuable, a great investment into further refactoring and stability. ❤️
> 
> I do have a few suggestions and question, but none blocking.

I left individual replies. If you don't mind I would merge this as-is and then follow up in the next PR, is that okay? If @ckifer is happy with the file rename ofc.

src/util/ChartUtils.ts
nikolasrieble: Fair enough 😅

src/util/ChartUtils.ts
nikolasrieble: That sounds fair. These are great considerations! Thanks for sharing!

ckifer:
🚀

PR #3776, opened 2023-09-22
Add types and tests for getDomainOfErrorBars

PavelVanecek:
## Description

Less `any`s, more unit tests.

## Related Issue

## Motivation and Context

Types and tests are always nice

## How Has This Been Tested?

TS, jest

## Types of changes

## Checklist:

### DIFF
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -17,10 +17,19 @@ import { findEntryInArray, getPercentValue, isNumber, isNumOrStr, mathSign, uniq
 import { filterProps, findAllByType, findChildByType, getDisplayName } from './ReactUtils';
 // TODO: Cause of circular dependency. Needs refactor.
 // import { RadiusAxisProps, AngleAxisProps } from '../polar/types';
-import { AxisType, BaseAxisProps, DataKey, LayoutType, LegendType, PolarLayoutType, TickItem } from './types';
+import {
+  AxisType,
+  BaseAxisProps,
+  DataKey,
+  LayoutType,
+  LegendType,
+  PolarLayoutType,
+  SpecificDomain,
+  TickItem,
+} from './types';
 import { Payload as LegendPayload } from '../component/DefaultLegendContent';
 
-export function getValueByDataKey<T>(obj: T, dataKey: DataKey<any>, defaultValue?: any) {
+export function getValueByDataKey<T>(obj: T, dataKey: DataKey<T>, defaultValue?: any) {
   if (_.isNil(obj) || _.isNil(dataKey)) {
     return defaultValue;
   }
@@ -410,7 +419,7 @@ export const appendOffsetOfLegend = (offset: any, items: Array<FormattedGraphica
   return newOffset;
 };
 
-const isErrorBarRelevantForAxis = (layout?: LayoutType, axisType?: AxisType, direction?: 'x' | 'y') => {
+const isErrorBarRelevantForAxis = (layout?: LayoutType, axisType?: AxisType, direction?: 'x' | 'y'): boolean => {
   if (_.isNil(axisType)) {
     return true;
   }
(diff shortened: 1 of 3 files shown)


src/util/types.ts
```diff
@@ -1050,13 +1050,35 @@ export interface GeometrySector {
...
+ * Can be specified in many ways:
+ * - array of numbers
+ * - with special strings like 'dataMin' and 'dataMax'
+ * - with special string math like 'dataMin - 100'
+ * - with keyword 'auto'
+ * - or a function
+ * - array of functions
+ * - or a combination of the above
+ *
+ * See https://recharts.org/en-US/api/XAxis#domain for examples
```
nikolasrieble, line 1065: How about instead adding examples here? 🤔 That would be more stable, i.e. not rely on the specific URL. Given that the website is in a different repository, we would certainly forget to change this comment if we were to change the URL / website.

src/util/types.ts
```diff
@@ -1050,13 +1050,35 @@ export interface GeometrySector {
 export type D3Scale<T> = D3ScaleContinuousNumeric<T, number>;
 
 export type AxisDomainItem = string | number | Function | 'auto' | 'dataMin' | 'dataMax';
-/** The domain of axis */
+/**
+ * The domain of axis.
```
nikolasrieble, line 1054:
```suggestion
 * The domain of axis, always defined by an array of exactly two values, for the min and the max of the axis. 
```

nikolasrieble: The only information that is missing here, is that it must the exactly two values

src/util/types.ts
```diff
@@ -1050,13 +1050,35 @@ export interface GeometrySector {
...
+ */
 export type AxisDomain =
   | string[]
   | number[]
   | [AxisDomainItem, AxisDomainItem]
   | (([dataMin, dataMax]: [number, number], allowDataOverflow: boolean) => [number, number]);
 
+/**
+ * SpecificDomain is an evaluated AxisDomain.
+ * Unlike AxisDomain, it has no variety - it's a tuple of two strings.
```
nikolasrieble, line 1075: Two strings? 🤔 The type below says number. 
Good point though, how about a categorical axis with strings on the xAxis? What does the domain look like in that case?

src/util/types.ts
```diff
@@ -1050,13 +1050,35 @@ export interface GeometrySector {
...
+ * This is after all the keywords and functions were evaluated and what is left is [min, max].
+ *
+ * Know that the min, max values are not guaranteed to be nice numbers - values like -Infinity or NaN are possible.
```
nikolasrieble, line 1078: ❤️

src/util/ChartUtils.ts
```diff
@@ -433,26 +442,26 @@ const isErrorBarRelevantForAxis = (layout?: LayoutType, axisType?: AxisType, dir
...
+  item: ReactElement,
+  dataKey: DataKey<any>,
   layout?: LayoutType,
   axisType?: AxisType,
-) => {
+): SpecificDomain | null => {
   const { children } = item.props;
   const errorBars = findAllByType(children, ErrorBar).filter(errorBarChild =>
     isErrorBarRelevantForAxis(layout, axisType, errorBarChild.props.direction),
   );
```
nikolasrieble, line 454: We could consider refactoring this function, such that errorBars are passed in. Then the tests would also not need to mock this function.

What do you think?

test/util/ChartUtils/getDomainOfErrorBars.spec.tsx
```diff
@@ -0,0 +1,288 @@
...
+        set.add(s);
+      }
+      relevantCases.forEach(verify);
+      returnsNullCases.forEach(verify);
+      const countOfLayoutOptions = 5; // 4 layouts + undefined
+      const countOfAxisOptions = 6; // 5 axes + undefined
+      const countOfDirectionOptions = 3; // 2 directions + undefined
+      const allCombinationsCount = countOfLayoutOptions * countOfAxisOptions * countOfDirectionOptions;
+      expect(set.size).toBe(allCombinationsCount);
+    });
```
nikolasrieble, line 137: This is just wonderful. I was about to start evaluating this myself manually. 😮‍💨

src/util/ChartUtils.ts
```diff
@@ -17,10 +17,19 @@ import { findEntryInArray, getPercentValue, isNumber, isNumOrStr, mathSign, uniq
...
+  LayoutType,
+  LegendType,
+  PolarLayoutType,
+  SpecificDomain,
+  TickItem,
+} from './types';
 import { Payload as LegendPayload } from '../component/DefaultLegendContent';
 
-export function getValueByDataKey<T>(obj: T, dataKey: DataKey<any>, defaultValue?: any) {
+export function getValueByDataKey<T>(obj: T, dataKey: DataKey<T>, defaultValue?: any) {
```
nikolasrieble, line 32: Strictly speaking this could be a breaking change, since it is exported. But I would rather consider this a fix, an object with an unrelated dataKey makes no sense.

nikolasrieble (commented):
This made my day. Wonderful PR. Not only did you add tests, but also another test to automatically verify that all cases are covered. 

I have found only one small inconsistency for a comment, and left a few remarks. I did not find any issue.

src/util/types.ts
PavelVanecek: I am now working on decrypting `getDomainOfDataByKey` and when I'm done with types + tests then I should hopefully have a full set of examples that I can put here.

src/util/types.ts
PavelVanecek: Thanks for the confirmation, I wasn't sure if that's the case judging from types alone.

src/util/types.ts
PavelVanecek: Ah right I don't know what I was thinking.

Also yes I have discovered the existence of categorical axes after I wrote this comment :-X I'll update

src/util/ChartUtils.ts
PavelVanecek: 💯 absolutely. I think some natural refactoring opportunities will pop up as we add more types. Just not in this PR 😬

test/util/ChartUtils/getDomainOfErrorBars.spec.tsx
PavelVanecek: I am contemplating adding a generator and creating these cases on the fly.

src/util/ChartUtils.ts
PavelVanecek: As far as I can see the DataKey is only ever used as `DataKey<any>` anyway. But yes could be breaking.

PavelVanecek:
Prettier is unhappy with a space after comment, I will fix.

src/util/types.ts
PavelVanecek: Hmmm @nikolasrieble the more I look into `getDomainOfDataByKey` the more I'm wondering. What does categorical domain look like? If it's a bunch of strings then all the strings will be part of the domain definition, right? It's only the numerical domain that can simplify the definition to `[min, max]`.

nikolasrieble (approved):
🚀 ✅ Great, thank you. Lets ship

Issue #3253, opened 2023-01-22
Accept 'data-testid' as a prop for every component to write better tests.


akib22:
## Feature Request
When I was migrating tests to jest that I found this issue. To write tests I have to rely on the component's CSS class names and I can't use React Testing Library `screen` utility functions which are recommended. Sometimes React Testing Library's `container.querySelector` don't work as expected then have to rely on the `document` object. And somehow the name of the CSS class changes test will fail, which is not true. If components accept `data-testid` then writing tests will be easier and don't have to rely on CSS class names. So I proposed, accept 'data-testid' as a prop for every component to write better tests.

## Current behavior
```ts
it('Render 1 path in Cross', () => {
    const { container } = render(
      <Surface width={400} height={400}>
        <Cross top={50} left={50} x={150} y={180} width={200} height={200} stroke="#000" fill="none" />
      </Surface>,
    );

    // accessing components using CSS class name
    expect(container.querySelectorAll('.recharts-cross')).toHaveLength(1);
});
```

## Expected behavior
```ts
it('Render 1 path in Cross', () => {
    const { container } = render(
      <Surface width={400} height={400}>
       // passing testid to the component
        <Cross data-testid="myTestId" top={50} left={50} x={150} y={180} width={200} height={200} stroke="#000" fill="none" />
      </Surface>,
    );

   // accessing element using testid
    expect(screen.getAllByTestId('myTestId').toHaveLength(1);
});
```

ckifer:
Hi @akib22 - thanks for your contributions!

This is definitely an issue (finding the components that you need while using RTL), but I'm not 100% sold on if this is the correct approach to solving it. I think maybe the lower level components (Layer, Surface, Cross, Polygon, etc.) could accept a `data-testid`, but I'm not sure the higher level components (Line, Bar, BarChart, LineChart, etc.) should. Especially because getting by test id is technically an [escape hatch in RTLs patterns](https://testing-library.com/docs/queries/bytestid). 

What happens for instance when a consumer of recharts wants to test a `Sector` but their RTL preset for test ids happens to a different attribute (such as `data-test`). This will then prompt that consumer to make an issue here that we should support their ID as a prop as well (which we can't do). 
 
The best thing I can think of which improves accessibility and testing at the same time is to add `<title>` tags to our svg elements and use `getByTitle`

ckifer:
What we could do it accept all data-* props and spread them into the base element of that component

mrskiro:
What is the current status?
[...]

PR #3783, opened 2023-09-23
Vitest Configuation Update

branberry:
Updated tests to pass using Vitest. This PR depends on a [change in the react-smooth](https://github.com/recharts/react-smooth/pull/81) library to deal with the unhandled exceptions we receive in the tests.

https://github.com/vitest-dev/vitest/issues/3117

Additional changes:
- updated the test command to not watch by default. This is to prevent pushes from being stuck due to running the test in watch mode for the push husky checks.
- turned off the no-extraneous-dependencies error so that we no longer receive errors when importing `vitest`.
- have Vitest ignore `node_modules` and `react-smooth` tests.


.eslintrc
```diff
@@ -51,7 +50,7 @@
     "react/no-unused-prop-types": "off",
     "class-methods-use-this": "error",
     "import/no-extraneous-dependencies": [
-      "error",
+      "off",
```
branberry, line 53: Added this to prevent errors when importing `vitest`. I believe we can configure `vitest` to work globally like `jest`, but didn't spend too much time trying to figure out how haha

test/component/Label.spec.tsx
```diff
@@ -12,6 +13,13 @@ const data = [
   { name: 'Page F', uv: 189, pv: 4800, amt: 2400 },
 ];
 
+beforeEach(() => {
+  vi.useFakeTimers();
+});
```
branberry, line 18: this should be remoed

test/component/Label.spec.tsx
```diff
@@ -143,17 +151,15 @@ describe('<Label />', () => {
...
       </LineChart>,
     );
-    // run all timers so animation finishes and labels render
-    jest.runAllTimers();
 
-    expect(container.querySelectorAll('.recharts-line .recharts-line-curve').length).toEqual(1);
-    expect(screen.getByText(/400/i)).toBeInTheDocument();
+    setTimeout(() => {
+      expect(container.querySelectorAll('.recharts-line .recharts-line-curve').length).toEqual(1);
+      expect(screen.getByText(/400/i)).toBeInTheDocument();
```
branberry, line 162: Fake timers weren't working, but using a regular timeout seemed to resolve the issue.

test/component/ResponsiveContainer.spec.tsx
```diff
@@ -230,16 +230,17 @@ describe('<ResponsiveContainer />', () => {
...
+      <ResponsiveContainer style={{ color: 'red', backgroundColor: '#FF00FF' }} data-testid="container">
         <div data-testid="inside" />
       </ResponsiveContainer>,
     );
-
-    expect(container.querySelector('.recharts-responsive-container')).toHaveStyle({
-      color: 'red',
-      backgroundColor: '#FF00FF',
-    });
+    expect(screen.getByTestId('container')).toHaveStyle('background-color: rgb(255, 0, 255)');
```
branberry, line 242: Using the data test ID to more easily get the right container

test/component/ResponsiveContainer.spec.tsx
```diff
@@ -252,8 +253,8 @@ describe('<ResponsiveContainer />', () => {
     expect(container.querySelector('.recharts-responsive-container')).toHaveStyle({
       width: '100px',
       height: '100px',
-      backgroundColor: 'red',
-      color: 'red',
+      'background-color': 'rgb(255,0,0)',
+      color: 'rgb(255,0,0)',
```
branberry, line 257: Looks like color enum values like `'red'` get coerced into the rgb format, so updated the test to reflect that. Not sure if this is problematic.

test/setupTests.ts
```diff
@@ -1,3 +1,7 @@
 import '@testing-library/jest-dom';
 
 global.requestAnimationFrame = fn => window.setTimeout(fn, 0);
+Object.defineProperty(window, 'requestAnimationFrame', {
```
branberry, line 4: This can be removed

.eslintrc
PavelVanecek: How come eslint still errors on this when `vitest` is indeed declared as a devDependency?

(I'm not a vitest nor eslint configuration expert, it's a genuine question)

branberry: Hey Pavel! I believe it's a rule to prevent the import of dev dependencies that causes this error. I think this rule exists because dev dependencies are typically stripped in production, but since we only use them in tests, it shouldn't be an issue (I think!)

test/chart/FunnelChart.spec.tsx
```diff
@@ -59,7 +59,8 @@ describe('<FunnelChart />', () => {
 
     expect(container.querySelectorAll('.recharts-funnel-trapezoid')).toHaveLength(5);
     // all trapezoids are visible
-    expect(container.querySelectorAll('.recharts-trapezoid')).toHaveLength(5);
+
```
branberry, line 62: Realized this is not correct. Need to look at this one again.

test/component/Label.spec.tsx
branberry: I spoke too soon :/ I incorrectly thought this fixed the tests, but using an actual timer meant the expect was never called as the test was no longer in scope. There are some weird issues with timers and whatnot with Vitest that I am doing my best to figure out, but for now, my boneheaded approach of using real timers isn't an option 🫠

.eslintrc
ckifer: FYI config for global vitest is pretty easy
https://vitest.dev/config/#globals
```
// tsconfig.json
{
  "compilerOptions": {
    "types": ["vitest/globals"]
  }
}
```

```
// vite.config.ts
import { defineConfig } from 'vitest/config'

export default defineConfig({
  test: {
    globals: true,
  },
})
```

As for the lint rule, I don't really get why it doesn't work. `devDependencies: false` should never error when devDeps are imported

src/component/ResponsiveContainer.tsx
```diff
@@ -27,6 +27,7 @@ export interface Props {
     width: number;
     height: number;
   };
+  'data-testid'?: string;
```
ckifer, line 30: What if someones test key is different than `data-testid`? could we allow all data-* attributes? 

 http://www.typescriptlang.org/docs/handbook/jsx.html#attribute-type-checking
"Note: If an attribute name is not a valid JS identifier (like a data-* attribute), it is not considered to be an error"

So data-* attributes are implicitly supported by TS, we can remove the explicit addition.

we could have a function called `getBaseProps` or similar that looks for props that match data-* and spreads them on the components.

test/component/ResponsiveContainer.spec.tsx
ckifer: should be fine

ckifer (changes requested):
since we already have globals enabled, lets add them to the tsconfig as well.

I'm concerned about the addition of the data-testid prop specifically, wondering if we can allow them implicitly instead.
See this as another project that does this https://github.com/cloudscape-design/components/blob/f4a2bed320f1fcbf8f6eacc41af5ec9a92096a0b/src/internal/base-component/index.ts#L23-L36

.eslintrc
branberry: Ah, it looks like we have `devDependencies: true` which might be why it's giving us an error then

branberry:
@ckifer I added a function like you mentioned to grab the `data-` props specifically, and spread that into the `div` for the responsive container component.  Thanks for that! 

Unfortunately, I was unable to setup the `vi` namespace globally. I followed the documentation, but I am still getting errors saying it can't find the namespace. Let me know if you catch anything I missed. 

Thanks again for the review!

src/component/ResponsiveContainer.tsx
```diff
@@ -83,7 +82,16 @@ export const ResponsiveContainer = forwardRef(
...
+    const getDataProps = (props: Partial<Props>) => {
+      const dataProps: Partial<Record<keyof Props, Props[keyof Props]>> = {};
+      Object.keys(rest).forEach((prop: keyof Props) => {
+        if (prop.match(/^data-/)) {
+          dataProps[prop] = props[prop];
+        }
+      });
 
+      return dataProps;
+    };
```
ckifer, line 94: nice! we can probably extract this elsewhere though. It will probably get used again and it isn't specific to this file. Can add a test for it elsewhere as well. Maybe one of the util files?

ckifer (changes requested):
Just one more thing I promise! Thank youuuu

src/component/ResponsiveContainer.tsx
branberry: I think that's a good idea! Will do.

nikolasrieble:
I merged master into the base branches, please also rebase or merge the vitest branch into this branch.

ckifer:
Yeah looks like we need to delete the old jest snapshots and gen new ones with npm test -- -u

nikolasrieble:
This PR already solves quite a few issues in the vitest branch. I suggest we merge this, and then start another to solve the remaining problems. We do not need to solve everything right here right now, simply because the base branch is already failing 😅 

Wdyt? @ckifer

src/component/ResponsiveContainer.tsx
```diff
@@ -83,7 +82,16 @@ export const ResponsiveContainer = forwardRef(
         containerHeight: containerRef.current.clientHeight,
       };
     }, []);
+    const getDataProps = (props: Partial<Props>) => {
```
nikolasrieble, line 85: This does seem unrelated. Is it? 🤔

ckifer: Yes and no - there was no great way to get the responsive container reference without adding a testid (although className would've worked, id would've worked).

We have already proposed doing something like this elsewhere https://github.com/recharts/recharts/issues/3253#issuecomment-1423448329 and I think this is a great way of allowing users to specify data-* props they want on base elements without it messing with our implementation or being super specific.

I did ask this to be refactored to be elsewhere, maybe ideally we make the tests pass with `className` and then in a separate PR we add this helper function in a util file and add this ability back to responsive container

ckifer: @branberry what do you think?

branberry: @ckifer If you'd like, I could move this function out into a separate file as a part of another PR so that we aren't blocking this one. Does that work for you?

ckifer: I'd rather remove this from this PR and add it back in it's own PR entirely rather than hiding the initial addition here

nikolasrieble: I wonder whether we should even land this separately on master.

ckifer: Snapshots will succeed when we get this into master - I think our CI is running on master always as compared to on the target branch

PR #3784, opened 2023-09-24
Add tests and types for getDomainOfDataByKey

PavelVanecek:
## Description

Tests and types only.

I don't think that all of the behaviour of `getDomainOfDataByKey` function is intentional :-X it appears to me like a product of a developer that doesn't have the habit of writing unit tests. I'm not certain if it's worth fixing some of these and what the release strategy should be.

Some of the behaviour is obviously wrong (sorting strings) and deserves a fix, on the other hand perhaps we should wait and see how it interacts with rest of the system?

## Related Issue

## Motivation and Context

Observing how the library works

## How Has This Been Tested?

Jest, TS

## Types of changes

## Checklist:

### DIFF
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -26,6 +26,7 @@ import {
   PolarLayoutType,
   NumberDomain,
   TickItem,
+  CategoricalDomain,
 } from './types';
 import { Payload as LegendPayload } from '../component/DefaultLegendContent';
 
@@ -45,25 +46,31 @@ export function getValueByDataKey<T>(obj: T, dataKey: DataKey<T>, defaultValue?:
   return defaultValue;
 }
 /**
- * Get domain of data by key
+ * Get domain of data by key.
  * @param  {Array}   data      The data displayed in the chart
  * @param  {String}  key       The unique key of a group of data
  * @param  {String}  type      The type of axis
  * @param  {Boolean} filterNil Whether or not filter nil values
  * @return {Array} Domain of data
  */
-export function getDomainOfDataByKey<T>(data: Array<T>, key: string, type: string, filterNil?: boolean) {
-  const flattenData = _.flatMap(data, entry => getValueByDataKey(entry, key));
+export function getDomainOfDataByKey<T>(
+  data: Array<T>,
+  key: DataKey<T>,
+  type: BaseAxisProps['type'],
+  filterNil?: boolean,
+): NumberDomain | CategoricalDomain {
+  const flattenData: unknown[] = _.flatMap(data, (entry: T): unknown => getValueByDataKey(entry, key));
 
   if (type === 'number') {
-    const domain = flattenData.filter(entry => isNumber(entry) || parseFloat(entry));
+    // @ts-expect-error parseFloat type only accepts strings
+    const domain: number[] = flattenData.filter(entry => isNumber(entry) || parseFloat(entry));
 
     return domain.length ? [_.min(domain), _.max(domain)] : [Infinity, -Infinity];
   }
 
   const validateData = filterNil ? flattenData.filter(entry => !_.isNil(entry)) : flattenData;
 
-  // 支持Date类型的x轴
+  // Supports x-axis of Date type
   return validateData.map(entry => (isNumOrStr(entry) || entry instanceof Date ? entry : ''));
 }
 
@@ -512,12 +519,12 @@ export const parseErrorBarsOfAxis = (
  */
 export const getDomainOfItemsWithSameAxis = (
   data: any[],
-  items: any[],
-  type: string,
+  items: ReactElement[],
+  type: BaseAxisProps['type'],
   layout?: LayoutType,
   filterNil?: boolean,
 ) => {
-  const domains = items.map(item => {
+  const domains: (NumberDomain | CategoricalDomain | null)[] = items.map(item => {
     const { dataKey } = item.props;
 
     if (type === 'number' && dataKey) {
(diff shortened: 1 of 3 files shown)


src/util/ChartUtils.ts
```diff
@@ -45,25 +46,31 @@ export function getValueByDataKey<T>(obj: T, dataKey: DataKey<T>, defaultValue?:
...
+  key: DataKey<T>,
+  type: BaseAxisProps['type'],
+  filterNil?: boolean,
+): NumberDomain | CategoricalDomain {
+  const flattenData: unknown[] = _.flatMap(data, (entry: T): unknown => getValueByDataKey(entry, key));
 
   if (type === 'number') {
-    const domain = flattenData.filter(entry => isNumber(entry) || parseFloat(entry));
+    // @ts-expect-error parseFloat type only accepts strings
+    const domain: number[] = flattenData.filter(entry => isNumber(entry) || parseFloat(entry));
```
nikolasrieble, line 66: Nit: Naming
This is not the domain yet, it is only the parsedData, right?

nikolasrieble, line 66:
```suggestion
    const parsedData: number[] = flattenData.filter(entry => isNumber(entry) || parseFloat(entry));
```

src/util/ChartUtils.ts
```diff
@@ -45,25 +46,31 @@ export function getValueByDataKey<T>(obj: T, dataKey: DataKey<T>, defaultValue?:
...
 
     return domain.length ? [_.min(domain), _.max(domain)] : [Infinity, -Infinity];
```
nikolasrieble, line 68: Nit: We could extract [Infinity, -Infinity] as a constant, DEFAULT_DOMAIN

Random side note: 
This line has an interesting edge case, i.e. when the data contains only a single data point. Then the resulting domain is [x, x] which results in a very weird axis and chart.

src/util/types.ts
```diff
@@ -1079,6 +1084,8 @@ export type AxisDomain =
  */
 export type NumberDomain = [min: number, max: number];
 
+export type CategoricalDomain = (number | string | Date)[];
```
nikolasrieble, line 1087: I assume more specifically: 

```suggestion
type CategoricalDomainValue = number | string | Date
export type CategoricalDomain = [CategoricalDomainValue, CategoricalDomainValue];
```

This would capture the fact that it must be exactly two values, right?

test/util/ChartUtils/getDomainOfDataByKey.spec.tsx
```diff
@@ -0,0 +1,169 @@
...
+      const data = [{ x: 7 }];
+      const result = getDomainOfDataByKey(data, 'x', 'number');
+      expect(result).toEqual([7, 7]);
+    });
+
+    it('should ignore NaN', () => {
+      const data = [{ x: NaN }];
+      const result = getDomainOfDataByKey(data, 'x', 'number');
+      expect(result).toEqual([Infinity, -Infinity]);
+    });
```
nikolasrieble, line 95:
```suggestion
    it('should ignore NaN', () => {
      const data = [{ x: NaN }];
      const result = getDomainOfDataByKey(data, 'x', 'number');
      expect(result).toEqual([Infinity, -Infinity]);
    });
    
    it('creates a collapsed domain for a single data point', () => {
      const data = [{ x: 1 }];
      const result = getDomainOfDataByKey(data, 'x', 'number');
      expect(result).toEqual([1, 1]);
    });
```

nikolasrieble: What do you think about adding this case?

test/util/ChartUtils/getDomainOfDataByKey.spec.tsx
```diff
@@ -0,0 +1,169 @@
+import { getDomainOfDataByKey } from '../../../src/util/ChartUtils';
+
+describe('getDomainOfDataByKey', () => {
```
nikolasrieble, line 3: Just to iterate: This is a very high value contribution, thanks a ton. Improving our overall confidence to change things is one of the most important things that does not itself map to an issue or specific feature request.

nikolasrieble (approved):
High value contribution. ❤️ Increasing our confidence to change things going forward. Thanks a ton. 

I have a few remarks to consider, but nothing blocking.

src/util/ChartUtils.ts
PavelVanecek: Maybe, I didn't add the variable, only the @ts- comment

src/util/types.ts
PavelVanecek: I am not convinced it's always two values. If an axis is `type = category` then its domain should be a list of all categories, right? When given `['fruits', 'vegetables', 'dairy', 'meat']`, recharts will keep all the categories. Unlike numerical axis of course.

The code certainly allows multiple values.

test/util/ChartUtils/getDomainOfDataByKey.spec.tsx
```diff
@@ -0,0 +1,169 @@
...
+      expect(resultWithTrue).toEqual(resultWithFalse);
+    });
+
+    it('should ignore 0 if it is specified as a string', () => {
+      const data = [{ x: '0' }, { x: '100' }, { x: '200' }];
+      const result = getDomainOfDataByKey(data, 'x', 'number');
+      expect(result).toEqual(['100', '200']);
+    });
+
+    it('should use the same number as both min and max if there is only one data point', () => {
```
PavelVanecek, line 85: Indeed @nikolasrieble here's a test for the single value case. I don't understand the usefulness of such domain but I'm still discovering what all this stuff is for.

src/util/types.ts
nikolasrieble: Right, I totally forgot about the PointScale, which simply has values in the domain. Sorry.

PR #3785, opened 2023-09-24
Expand tests for getValueByDataKey function

PavelVanecek:
## Description

There were already tests present so I expanded on some edge cases.

## Related Issue

## Motivation and Context

Testsss

## How Has This Been Tested?

Tests!

## Types of changes

- [X] Tesssstsss

## Checklist:

### DIFF
--- a/test/util/ChartUtils.spec.tsx
+++ b/test/util/ChartUtils.spec.tsx
@@ -242,16 +242,94 @@ describe('parseScale', () => {
 });
 
 describe('getValueByDataKey', () => {
-  const data = { a: 1, b: 2, c: 3 };
+  const data: Record<string, unknown> = {
+    a: 1,
+    b: 2,
+    c: 3,
+    1: 4,
+    u: undefined,
+    n: null,
+    nested: { obj: { children: 7 } },
+    arr: [{ x: 6 }, { y: 5 }],
+    nan: NaN,
+  };
 
-  it('of function', () => {
+  it('should call dataKey if it is a function', () => {
     const fn: DataKey<typeof data> = entry => entry.a;
-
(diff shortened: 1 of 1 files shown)


test/util/ChartUtils.spec.tsx
```diff
@@ -242,16 +242,94 @@ describe('parseScale', () => {
 });
 
 describe('getValueByDataKey', () => {
-  const data = { a: 1, b: 2, c: 3 };
+  const data: Record<string, unknown> = {
```
nikolasrieble, line 245: Nice, I think this type could be a nice one to generally use for a row of data.

PavelVanecek: Eventually I would like to find a solution that relies on generics instead of unknowns - but for now this will do

PavelVanecek: Also I should really be using `PropertyKey` instead, and adding a test of what happens when DataKey returns a symbol!

nikolasrieble: A side note: I used Symbols in a few workaround to add an index to the data before passing it to recharts. I then use the index to identify an active datapoint from a mouseOver, or a Tooltip. 

So for my use case this would have to be 

```
Record<string | Symbol, unknown>
```

nikolasrieble: Unrelated to this PR, but generally relevant: 

Should we possibly do this in the library itself? The whole getActiveIndex is quite expensive (iterating over the whole dataset every time) and could be done much simpler if we just add an index with a Symbol to the dataset as a preprocessing step in the Chart.

nikolasrieble (approved):
❤️

PR #3803, opened 2023-09-29
RadialBar: activeShape should work with Tooltip

andrewangelle:
The existing activeShape prop on `<RadialBar />` should work with the `<Tooltip />` in the same manner as `<Line />`, `<Bar />`, `<Funnel/>` do.
## Description

  - Added another if block to render the graphical item when `activeShape` is truthy.
Inside the new if block, render the graphical item and pass in `activeIndex` which defaults to user defined `activeIndex` and falls back to `activeTooltipIndex` when undefined.

- #### src/polar/RadialBar.tsx
  - Renamed the generically exported type named Props to RadialBarProps
  - Refactored static render method out to RadialBarUtils
  - Updated type of `RadialBar.activeShape` to match the signature of activeLine/activeBar/activeShape in Line/Bar/Funnel components

## Related Issue
#3792 

## Motivation and Context

To provide a consistent interface between Line / Bar / Funnel / RadialBar and their respective props of activeDot / activeBar / activeShape

## How Has This Been Tested?
Unit tests and storybook

Screenshot 2023-09-29 at 5 23 19 PM

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -2151,6 +2151,12 @@ export const generateCategoricalChart = ({
             return [graphicalItem, this.renderActiveBar({ item, childIndex: activeTooltipIndex })];
           }
 
+          if (activeShape) {
+            const activeIndex =
+              element.props.activeIndex !== undefined ? element.props.activeIndex : activeTooltipIndex;
+            return [cloneElement(element, { ...item.props, ...itemEvents, activeIndex }), null, null];
+          }
+
           if (!_.isNil(activePoint)) {
             return [
               graphicalItem,
@@ -2172,8 +2178,8 @@ export const generateCategoricalChart = ({
            * An example usage case is a FunnelChart
            */
           const {
-            graphicalItem: { item: xyItem, childIndex },
-          } = this.getItemByXY(this.state.activeCoordinate);
+            graphicalItem: { item: xyItem = element, childIndex },
+          } = this.getItemByXY(this.state.activeCoordinate) ?? { graphicalItem };
 
           const elementProps = { ...item.props, ...itemEvents, activeIndex: childIndex };
 
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1,11 +1,10 @@
-import React, { Component, cloneElement, isValidElement, createElement, ComponentProps, ReactElement } from 'react';
+import React, { Component, cloneElement, isValidElement, createElement, ReactElement } from 'react';
 import classNames from 'classnames';
 import _, { isArray, isBoolean, isNil } from 'lodash';
 import invariant from 'tiny-invariant';
 import { getTicks } from '../cartesian/getTicks';
 import { Surface } from '../container/Surface';
 import { Layer } from '../container/Layer';
-import { Bar } from '../cartesian/Bar';
 import { Tooltip } from '../component/Tooltip';
 import { Legend } from '../component/Legend';
 import { Curve } from '../shape/Curve';
@@ -52,7 +51,6 @@ import {
   parseDomainOfCategoryAxis,
   getTooltipItem,
 } from '../util/ChartUtils';
-import { ActiveBar } from '../util/BarUtils';
 import { detectReferenceElementsDomain } from '../util/DetectReferenceElementsDomain';
 import { inRangeOfSector, polarToCartesian } from '../util/PolarUtils';
 import { shallowEqual } from '../util/ShallowEqual';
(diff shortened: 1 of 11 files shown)


src/polar/RadialBar.tsx
```diff
@@ -39,18 +41,17 @@ type RadialBarDataItem = SectorProps & {
   background?: SectorProps;
 };
 
-type RadialBarShape = ReactElement | ((props: Props) => ReactNode);
-type RadialBarBackground = ReactElement | ((props: Props) => ReactNode) | SectorProps | boolean;
+type RadialBarBackground = ActiveShape<SectorProps>;
```
nikolasrieble, line 44: Nice.

src/polar/RadialBar.tsx
```diff
@@ -366,8 +356,8 @@ export class RadialBar extends PureComponent<Props, State> {
         return null;
       }
 
-      const props = {
-        cornerRadius,
+      const props: RadialBarSectorProps = {
+        cornerRadius: typeof cornerRadius === 'string' ? parseInt(cornerRadius, 10) : cornerRadius,
```
nikolasrieble, line 360: Nit: We could extract `parseCornerRadius`

src/chart/generateCategoricalChart.tsx
```diff
@@ -2151,6 +2151,12 @@ export const generateCategoricalChart = ({
             return [graphicalItem, this.renderActiveBar({ item, childIndex: activeTooltipIndex })];
           }
 
+          if (activeShape) {
+            const activeIndex =
+              element.props.activeIndex !== undefined ? element.props.activeIndex : activeTooltipIndex;
+            return [cloneElement(element, { ...item.props, ...itemEvents, activeIndex }), null, null];
```
nikolasrieble, line 2157: Just wondering, do you see a reason why activeBar should be treated separately? Could we not treat all activeShapes the same? 

Or is there some specific behaviour the I am not aware of?

src/util/RadialBarUtils.tsx
```diff
@@ -0,0 +1,49 @@
...
+  if (isValidElement(option)) {
+    sectorShape = cloneElement(option, props);
+  } else if (_.isFunction(option)) {
+    sectorShape = option(props);
+  } else if (_.isPlainObject(option) && !_.isBoolean(option)) {
+    const elementProps = typeGuardSectorProps(option, props);
+    sectorShape = <Sector {...elementProps} />;
+  } else {
+    sectorShape = <Sector {...props} />;
+  }
```
nikolasrieble, line 42: I have seen this logic before in various variants. Should we possibly extract this as a helper and reuse? 

Screenshot 2023-10-01 at 12 14 46

src/util/RadialBarUtils.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  if (props.isActive) {
+    return <Layer className="recharts-active-shape">{sectorShape}</Layer>;
```
nikolasrieble, line 45: Not sure about best practice, but we are already using this twice so far, we could extract it as a constant. 

What is the use case of this className? Testing? Enabling the user to further adapt the style? 

If the reason is testing, should we then not be using data-testid attribute instead?

storybook/stories/API/chart/RadialBarChart.stories.tsx
```diff
@@ -1,30 +1,55 @@
...
   argTypes: {
     ...CategoricalChartProps,
+    activeIndex: {
+      description:
+        'The index of the individual shapes that make up the Funnel to be marked as active, and render props.activeShape as a result',
+      table: {
+        type: {
+          summary: 'number | undefined',
+        },
+        defaultValue: undefined,
```
nikolasrieble, line 19: Do we need to define these here? Do they not get auto-filled from the code? 🤔

nikolasrieble (approved):
Overall great addition! I have a few comments and questions but nothing blocking. Thank you for your (repeated ❤️ ) contribution

src/util/RadialBarUtils.tsx
andrewangelle: > What is the use case of this className? Testing? Enabling the user to further adapt the style?

I believe if the user provides `activeShape` an object of attributes, or as a boolean, then they could use this classname to target styling.

storybook/stories/API/chart/RadialBarChart.stories.tsx
andrewangelle: Should be fine. I removed these.

src/util/RadialBarUtils.tsx
andrewangelle: Addressed here: https://github.com/recharts/recharts/pull/3803/commits/d28ad574c21165a70e11697e70a9e753a7d063c3

Let me know if you have any feedback on that refactor.

src/chart/generateCategoricalChart.tsx
andrewangelle: Addressed here: https://github.com/recharts/recharts/pull/3803/commits/80bd132bfecaee13a84d09b661a86a59c425a0eb

nikolasrieble:
Thank you @andrewangelle! I tested it again and found no issue.

PR #3807, opened 2023-10-03
Add AxisMap return type to few functions in generateCategoricalChart

PavelVanecek:
The type was already there - but only used in State. I reused the same as return type of the `getAxis` functions

## Description

types only

## Related Issue

## Motivation and Context

TS

## How Has This Been Tested?

TS

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -70,6 +70,7 @@ import {
   TickItem,
   adaptEventHandlers,
   GeometrySector,
+  AxisType,
 } from '../util/types';
 import { AccessibilityManager } from './AccessibilityManager';
 import { isDomainSpecifiedByUser } from '../util/isDomainSpecifiedByUser';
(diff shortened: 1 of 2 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -287,13 +288,29 @@ const getTooltipData = (state: CategoricalChartState, chartData: any[], layout:
...
+  {
+    axes,
+    graphicalItems,
+    axisType,
+    axisIdKey,
+    stackGroups,
+    dataStartIndex,
+    dataEndIndex,
+  }: {
+    axes: any;
```
nikolasrieble, line 300: It seems we can narrow this further from usage, right?

nikolasrieble (approved):
Small steps, quick to review, thank you.

src/chart/generateCategoricalChart.tsx
PavelVanecek: Indeed we can, I will continue doing that in followup PRs.

PR #3809, opened 2023-10-03
Add generateCategoricalChart types: string properties, and events

PavelVanecek:
## Description

Types!

## Related Issue

## Motivation and Context

I want to refactor this whole thing but I need typescript to work with me

## How Has This Been Tested?

TS, Jest

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1003,13 +1003,16 @@ export const generateCategoricalChart = ({
   return class CategoricalChartWrapper extends Component<CategoricalChartProps, CategoricalChartState> {
     static displayName = chartName;
 
-    uniqueChartId: any;
+    uniqueChartId: string;
 
-    clipPathId: any;
+    clipPathId: string;
 
+    /**
+     * legendInstance appears never used - can we remove it?
+     */
     legendInstance: any;
 
-    deferId: any;
+    deferId: ReturnType<typeof defer> | null;
 
     accessibilityManager = new AccessibilityManager();
 
@@ -1025,7 +1028,7 @@ export const generateCategoricalChart = ({
       ...defaultProps,
     };
 
-    container?: any;
+    container?: HTMLElement;
 
     constructor(props: CategoricalChartProps) {
       super(props);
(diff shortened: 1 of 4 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -1415,6 +1420,7 @@ export const generateCategoricalChart = ({
 
     clearDeferId = () => {
       if (!isNil(this.deferId) && deferClear) {
+        // @ts-expect-error types miss the relationship between deferClear and defer
```
nikolasrieble, line 1423: What do you suggest we do with these statements? 
Should we create a follow up issue? 

I am a bit sceptical of the value of introducing types if that requires us to also `@ts-expect-error`.

nikolasrieble: Do you see a way how you could still fix these in this PR?

nikolasrieble (commented):
Great small PR in the right direction. Curious to hear your thoughts on how to proceed with these types that require `@ts-expect-error`

src/chart/generateCategoricalChart.tsx
PavelVanecek: These two are forming a pair and if we want the pair to resolve to the same generic then they need to be in the same generic container - like an object or a Class. I can fix this one.

PavelVanecek:
So about the `@ts-expect-error` in general. In my opinion it is not amazing to have that but it is 100% better than `any` because:

- It shows, explicitly, with a comment, that the types have a problem. Developers are alerted to the presence of the problem, can read what it is, and can apply extra caution.
- All the _other_ currently existing places in code, where our function is used, are now strongly typed. `ts-expect-error` is only applied to one single line.
- If a problem is fixed and the `ts-expect-error` is no longer necessary, typescript will highlight that as an error. On the other hand, `any` will spread silently even when it's no longer needed.
- And most important, all the future uses of code are going to be strongly typed.

So you will see me sometimes replacing `any` with `@ts-expect-error`. I tend to do that more often when the required refactor is too large (out of scope of single PR), or when it touches code we want to get rid of anyway (element cloning).

src/util/deferer.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Will execute callback fn asynchronously.
+ * It will detect the appropriate function to use.
+ *
+ * Named after the famous Swiss tennis player, Roger Deferer.
+ *
+ * @param {Function} callback will be executed asynchronously, with no arguments
+ * @returns {Function} a cancel function.
+ */
+export function deferer(callback: () => void): CancelFunction {
```
nikolasrieble, line 12: Nice, thank you.

nikolasrieble (approved):
🚀

Issue #3793, opened 2023-09-26
Consistent interface: Pie - activeShape


andrewangelle:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-4thsnj?file=/src/index.js)

### Steps to reproduce
Hover over any of the sections. Notice activeShape prop is ignored

### What is expected?
I'd expect the customized active shape to render according to the active tooltip the same way that activeDot/activeBar/activeShape work with Line/Bar/Funnel

### What is actually happening?
The activeShape prop is being ignored

| Environment | Info |
|---|---|
| Recharts | v2.8.0 |
| React | 18.2.0 |
| System | macOS Ventura 13.5.2 |
| Browser | Chrome v116.0.5845.179 |

andrewangelle:
Assign me please!

ckifer:
Thanks!

PR #3818, opened 2023-10-05
Pie: activeShape should work with Tooltip

andrewangelle:
The existing `activeShape` prop on `<Pie />` should work with the `<Tooltip />` in the same manner as `<Line />` and `<Bar />` do.

## Description

  - added a block to `getItemByXY` to set the find and set `activeIndex` when Tooltip is used with Pie
  
- #### src/polar/Pie.tsx
  - Refactored to use `ActiveShape` type for `props.activeShape` and `props.InactiveShape`
  
  - Made `propTransformer` optional.
  
- #### test/chart/PieChart.spec.tsx
  
- #### test/polar/Pie.spec.tsx
  - Updated some types in the existing tests around activeShape with activeIndex
  
## Related Issue
#3793 

## Motivation and Context

To provide a consistent interface between Line / Bar / Funnel / RadialBar / Pie and their respective props of activeDot / activeBar / activeShape

## How Has This Been Tested?

Unit tests and storybook.

Screenshot 2023-10-05 at 9 12 37 AM

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -73,6 +73,7 @@ import {
 import { AccessibilityManager } from './AccessibilityManager';
 import { isDomainSpecifiedByUser } from '../util/isDomainSpecifiedByUser';
 import { deferer, CancelFunction } from '../util/deferer';
+import { getActiveShapeIndexForTooltip } from '../util/ActiveShapeUtils';
 
 export type GraphicalItem<Props = Record<string, any>> = ReactElement<
   Props,
@@ -2146,11 +2147,13 @@ export const generateCategoricalChart = ({
           }
         } else {
           /**
-           * We hit this block if consumer uses a Tooltip without XAxis and/or YAxis.
-           * In which case, this.state.activeTooltipIndex never gets set
-           * because the mouse events that trigger that value getting set never get trigged without the axis components.
+           * We hit this block if consumer uses a Tooltip and this.state.activeTooltipIndex never gets set
            *
-           * An example usage case is a FunnelChart
+           * One example is usage of Tooltip without XAxis and/or YAxis.
+           * The mouse events that normally trigger this.state.activeTooltipIndex getting set
+           * never get trigged without the axis components.
+           *
+           * An example usage case is a FunnelChart/Pie/Scatter
            */
           const {
             graphicalItem: { item: xyItem = element, childIndex },
@@ -2251,27 +2254,12 @@ export const generateCategoricalChart = ({
             if (activeBarItem) {
               return { graphicalItem, payload: activeBarItem };
             }
-          } else if (itemDisplayName === 'Funnel') {
-            /*
-             * To handle possible duplicates in the data set,
-             * match both the data value of the active item to a data value on a graph item,
-             * and match the mouse coordinates of the active item to the coordinates of a trapezoid.
-             * This assumes equal lengths of trapezoids on a funnel to data items.
-             */
-            const activeIndex = item.props.data.findIndex((funnelData: { value: string }, dataIndex: number) => {
-              const valuesMatch = funnelData.value === activeItem.payload.value;
-
-              const indexOfMouseCoordinates = graphicalItem.props.trapezoids.findIndex(
-                (trapezoid: { x: number; y: number }) => {
-                  const xMatches = trapezoid.x === activeItem.labelViewBox.x || trapezoid.x === activeItem.x;
-                  const yMatches = trapezoid.y === activeItem.labelViewBox.y || trapezoid.y === activeItem.y;
-                  return xMatches && yMatches;
-                },
-              );
-
-              const coordinatesMatch = dataIndex === indexOfMouseCoordinates;
-
-              return valuesMatch && coordinatesMatch;
+          } else if (itemDisplayName === 'Funnel' || itemDisplayName === 'Pie') {
+            const activeIndex = getActiveShapeIndexForTooltip({
+              itemDisplayName,
+              graphicalItem,
+              activeTooltipItem: activeItem,
+              itemData: item.props.data,
             });
 
             const childIndex = item.props.activeIndex === undefined ? activeIndex : item.props.activeIndex;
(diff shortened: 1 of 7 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -2280,6 +2280,19 @@ export const generateCategoricalChart = ({
               graphicalItem: { ...graphicalItem, childIndex },
               payload: graphicalItem.props.data[activeIndex],
             };
+          } else if (itemDisplayName === 'Pie') {
+            const tooltipPayload = activeItem.tooltipPayload?.[0].payload.payload;
+
+            const activePieIndex = item.props.data.findIndex((datum: typeof tooltipPayload) => {
+              return _.isEqual(tooltipPayload, datum);
+            });
```
nikolasrieble, line 2288: I believe the same holds true as above, this could also have duplicates, right? 
I wonder whether we can further refactor this logic a bit and possibly shared the computation of the activeIndex?

nikolasrieble (commented):
Pretty nice! 🚀 I am only wondering about duplicates and refactoring to reduce duplication in generateCategoricalChart.

src/chart/generateCategoricalChart.tsx
andrewangelle: Yes. I initially thought this wasn't possible, but I was incorrect. I've added a duplicate data item to the story. I've also refactored the logic to handle getting the active index and accounting for duplicates for multiple graph components.

src/util/ActiveShapeUtils.tsx
```diff
@@ -63,3 +78,140 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
...
+          const xMatches = shapeData.x === activeTooltipItem?.labelViewBox?.x || shapeData.x === activeTooltipItem.x;
+          const yMatches = shapeData.y === activeTooltipItem?.labelViewBox?.y || shapeData.y === activeTooltipItem.y;
+          return xMatches && yMatches;
+        }
+
+        if (isPie(itemDisplayName, shapeData) && isPie(itemDisplayName, activeTooltipItem)) {
+          const startAngleMatches = shapeData.endAngle === activeTooltipItem.endAngle;
+          const endAngleMatches = shapeData.startAngle === activeTooltipItem.startAngle;
+          return startAngleMatches && endAngleMatches;
+        }
```
nikolasrieble, line 189: Could we instead pass a comparison function? That could simplify this component a little.

nikolasrieble: Or, we could simply pass a list of keys for comparison. 
For the Pie: ['endAngle', 'startAngle'], for the scatter: ['x', 'y', 'z'], etc. 

Just wondering whether we can simplify this function to not know about Pies, Scatters, etc.

src/util/ActiveShapeUtils.tsx
```diff
@@ -63,3 +78,140 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
...
+  itemDisplayName,
+  activeTooltipItem,
+  graphicalItem,
+  itemData,
+}: GetActiveShapeIndexForTooltip): number {
+  const shapeKey = getShapeDataKey(itemDisplayName);
+
+  const tooltipPayload: Partial<ShapeData<typeof itemDisplayName>> = isScatter(itemDisplayName, activeTooltipItem)
+    ? activeTooltipItem.payload
+    : activeTooltipItem.tooltipPayload?.[0].payload.payload;
```
ckifer, line 172: nit: this optional chaining should probably keep going to prevent bad access 

`activeTooltipItem.tooltipPayload?.[0]?.payload?.payload`

src/util/ActiveShapeUtils.tsx
```diff
@@ -63,3 +78,140 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
...
+      mouseCoordinateMatches[mouseCoordinateMatches.length - 1],
+    );
+
+    const coordinatesMatch = dataIndex === indexOfMouseCoordinates;
+
+    return valuesMatch && coordinatesMatch;
+  });
+
+  // get the last index in case of multiple matches
+  const activeIndex = itemData.indexOf(activeItemMatches[activeItemMatches.length - 1]);
```
ckifer, line 214: what if length is 0?

ckifer (changes requested):
I have some concerns with the reliance on displayName. I fear this approach ties us to it more (even though the previous implementation does exactly the same). Perhaps we could brainstorm some ideas on how this could be mitigated in the future.

nothing to be done there for now, but I think my nits/other comments at least can be addressed

Overall great improvement

src/polar/Pie.tsx
```diff
@@ -471,28 +472,16 @@ export class Pie extends PureComponent<Props, State> {
...
     const { activeShape, blendStroke, inactiveShape: inactiveShapeProp } = this.props;
     return sectors.map((entry, i) => {
+      const isActive = this.isActiveIndex(i);
       const inactiveShape = inactiveShapeProp && this.hasActiveIndex() ? inactiveShapeProp : null;
-      const sectorOptions = this.isActiveIndex(i) ? activeShape : inactiveShape;
+      const sectorOptions = isActive ? activeShape : inactiveShape;
       const sectorProps = {
         ...entry,
         stroke: blendStroke ? entry.fill : entry.stroke,
+        tabIndex: -1,
```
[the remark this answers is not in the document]
andrewangelle: I was trying to keep feature parity with the old implementation which had this hardcoded here:
https://github.com/recharts/recharts/pull/3818/files#diff-df78eb0233df1557f58b751b9c4d31b9479393942600664c0d7cf3ac3ef0a1b8L482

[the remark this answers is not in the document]
ckifer: ah gotcha, thank you. Hard to see in the diff

src/util/ActiveShapeUtils.tsx
andrewangelle: We'd get `activeIndex` as `-1`, i.e. no match. Which i think is what we'd want for a fallback, correct?

ckifer: when length is zero this happens no?
`activeItemMatches[0 - 1]` === `activeItemMatches[-1]` === out of bounds exception

ckifer: Oh wow javascript continues to surprise me, I guess that just returns undefined huh 🙃

ckifer: All good, ty

src/util/ActiveShapeUtils.tsx
```diff
@@ -63,3 +78,140 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
...
+  return itemDisplayName === 'Funnel';
+}
+
+function isPie(itemDisplayName: ItemDisplayName, _item: unknown): _item is PieItem {
+  return itemDisplayName === 'Pie';
+}
+
+function isScatter(itemDisplayName: ItemDisplayName, _item: unknown): _item is ScatterItem {
+  return itemDisplayName === 'Scatter';
+}
```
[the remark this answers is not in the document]
andrewangelle: I can pass in `graphicalItem` instead of `itemDisplayName` and check the existence of the particular shape key(trapezoids, sector, points) inside of `graphicalItem.props`.

```typescript
function isFunnel(graphicalItem: GraphicalItem, _item: unknown): _item is FunnelItem {
  return 'trapezoids' in graphicalItem.props
}
```

src/util/ActiveShapeUtils.tsx
```diff
@@ -63,3 +78,140 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
...
+function getShapeDataKey(itemDisplayName: ItemDisplayName): GraphicalItemShapeKey {
+  let shapeKey: GraphicalItemShapeKey;
+
+  if (itemDisplayName === 'Funnel') {
+    shapeKey = 'trapezoids';
+  } else if (itemDisplayName === 'Pie') {
+    shapeKey = 'sectors';
+  } else if (itemDisplayName === 'Scatter') {
+    shapeKey = 'points';
+  }
```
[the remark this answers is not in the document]
andrewangelle: If i keep the function, but use the type guards to assign this value after the change described in my other comment then it should remove the need for `itemDisplayName` anywhere in this flow.

ckifer (approved):
Thank you! 🚀

src/util/ActiveShapeUtils.tsx
andrewangelle: List of keys won't work because I need to do an `||` comparison a few lines above this for funnel.

I can use the type guards with the incoming `graphicalItem` and `activeItem` to get the different comparisons without this function needing to know anything about the component type.

ckifer (approved):

Issue #3684, opened 2023-08-05
label doesn't work in Reference Data


MrOxMasTer:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://github.com/recharts/recharts/issues/new?assignees=&labels=&projects=&template=----please-use-recharts-issue-helper---.md](https://github.com/recharts/recharts/issues/new?assignees=&labels=&projects=&template=----please-use-recharts-issue-helper---.md)

### Steps to reproduce
не нужно

### What is expected?
that the signature will appear near the point

### What is actually happening?
Error

| Environment | Info |
|---|---|
| Recharts | v2.7.2 |
| React | "react": "^18.2.0", |
| System | Windows 11 |
| Browser | Chrome |

![image](https://github.com/recharts/recharts/assets/59291123/65f267e6-cea0-49c1-96f3-42ee3b63984a)
![image](https://github.com/recharts/recharts/assets/59291123/8e9c29c3-0488-436b-bdf9-6f81ddd79f55)

ckifer:
Interesting, `Label` is a valid child of reference dot. For now you can use the `label` prop

https://recharts.org/en-US/api/ReferenceDot#label

ckifer:
Could you reproduce using a sandbox? Seems to be working 
https://codesandbox.io/s/simple-bar-chart-forked-4c7s77?file=/src/App.tsx

MrOxMasTer:
> Interesting, `Label` is a valid child of reference dot. For now you can use the `label` prop
> 
> https://recharts.org/en-US/api/ReferenceDot#label

my label doesn't work at all. Not with the help of props, not with the help of children

ckifer:
Can you reproduce? I can't reproduce the issue so I'm not sure how to help

MrOxMasTer:
> Can you reproduce? I can't reproduce the issue so I'm not sure how to help

I do not know how to show you better. Maybe there is some kind of library conflict going on? I can throw you the code or throw off the list of libraries. Right now I'll try to test it in codebox

MrOxMasTer:
I do not know why, but everything works for me in codebox too https://codesandbox.io/s/simple-bar-chart-forked-rdyg7c?file=/src/App.tsx

MrOxMasTer:
my list of libraries:

"dependencies": {
        "@radix-ui/react-progress": "^1.0.3",
[...]

Issue #3752, opened 2023-09-11
Use lodash individual packages imports instead of the entire thing


ckifer:
## Do you want to request a feature or report a bug?
Bundle Size Improvement

We currently have `lodash` as a dependency. `lodash` is known to be large and not have great tree-shaking.
https://lodash.com/per-method-packages

Instead of relying on all of `lodash`, how about we use the lodash individual packages such as `lodash.groupBy`, etc.

## Acceptance Criteria 

1. Remove all `import _ from lodash`
2. Identify what functions are used in the project 
3. import all individual functions directly like `import isFunction from 'lodash/isFunction'`
4. Replace all instances of `_` with the equivalent function from individual packages
5. Run tests and confirm behavior where needed

hamzaM98:
Hello, Can I take this?

hmza98:
Hello, Can i take this?

nikolasrieble:
It is not clear whether this effort would result in a large gain as to https://www.blazemeter.com/blog/import-lodash-libraries

> The smallest bundle size could also be reached by using the babel-plugin-lodash together with lodash-webpack-plugin for cherry-picking only the used functions.

nikolasrieble:
It seems that there are a few caveats though: https://www.npmjs.com/package/lodash-webpack-plugin

> DISCLAIMER: Using this plugin without enabling the proper feature sets may cause lodash functions to behave in unexpected ways. Methods may appear to work, however they might return incorrect results.

ckifer:
It should amount to roughly the same gain. 

The issue with the plugin is that consumers attempt to use it and leave out functions that we use internally causing breakage for them. If we only include the packages we use then no one has to do anything special to get bundle size reduction

nikolasrieble:
I am looking into it, and will start with deprecating lodash for native functions where possible / reasonable. 

See https://www.npmjs.com/package/lodash.isarray for example

ckifer:
Sounds good, would much rather not have lodash at all except for specific and complex things (groupBy is something I always use in my projects for example)

nikolasrieble:
Reading more into this, it seems importing lodash from different packages is discouraged. https://lodash.com/per-method-packages

Instead, the suggest way to go, is to import specific methods by path, as in 
```
import capitalize from 'lodash/capitalize';
```

That can nicely be done in a single PR, since it merely is a change of import.

ckifer:
Oh, RIP 😅. I should have researched more. 

Cool! Much easier then

[...]

PR #3820, opened 2023-10-06
remove reduceCSSCalc

HHongSeungWoo:
## Description

I have drafted a PR for removing the "reduce-css-calc" library used in recharts.

## Related Issue

#3684 

## Motivation and Context

It seems that making this change would result in benefits such as removing library dependencies and reducing bundle size.

## Types of changes

## Checklist:


HHongSeungWoo:
If this change looks okay, I will continue with the work.

ckifer:
Replacing this library would be great!

I'm a bit concerned that the implementations are not the same. We probably shouldn't be creating temporary elements to get css calculation values... I'm not sure this approach is the right direction. 

@nikolasrieble  thoughts?

HHongSeungWoo:
@ckifer 
In my opinion, the this implementation has the following issues

1. number + number calculations do not work correctly.
2. library existing returns values for invalid expressions, but in this implementation, return value for an invalid expression is an empty string.
3. new element always created to obtain the computed css

for the first issue, capHeight, lineHeight used in the current Text implementation are not exposed externally.
second issue, invalid expressions do not seem to be a problem since they do not work correctly even in the dy
third issue, there might be ways to reuse objects or use the CSSStyleSheet API

HHongSeungWoo:
I understand the need for caution when making changes to such widely used libraries. If you have any concerns, even the slightest ones, I'm willing to close the PR. :)

src/component/Text.tsx
```diff
@@ -190,6 +189,12 @@ const getWordsByLines = ({ width, scaleToFit, children, style, breakAll, maxLine
 
 const DEFAULT_FILL = '#808080';
 
+const reduceCSSCalc = (calc: string) => {
+  const tempEle = document.createElement('div');
+  tempEle.style.width = calc;
+  return tempEle.style.width.slice(5, -1).trim();
+};
```
nikolasrieble, line 196: Adding this without any unit tests is not sufficient.

nikolasrieble: Where will this behave different than `import reduceCSSCalc from 'reduce-css-calc';`
It will certainly not be identical, I assume - given the code size difference.

src/component/Text.tsx
```diff
@@ -190,6 +189,12 @@ const getWordsByLines = ({ width, scaleToFit, children, style, breakAll, maxLine
 
 const DEFAULT_FILL = '#808080';
 
+const reduceCSSCalc = (calc: string) => {
+  const tempEle = document.createElement('div');
```
nikolasrieble, line 193: In how far is this temporary, if you create it on every call?

nikolasrieble (commented):
I am not confident this is the right way. If we want to continue here, the next step should certainly be unit tests.

src/component/Text.tsx
HHongSeungWoo: > Adding this without any unit tests is not sufficient.

I'd be happy to add testing if it's needed. However, since it relies on browser functionality, a browser testing library like Cypress is required.

HHongSeungWoo: > Where will this behave different than `import reduceCSSCalc from 'reduce-css-calc';` It will certainly not be identical, I assume - given the code size difference.

This code is referenced from https://drafts.csswg.org/css-values-3/#calc-syntax.
The differences I am aware of between reduce-css-calc and this code are in the calculation of unitless numbers and the return for invalid expressions.

src/component/Text.tsx
HHongSeungWoo: > In how far is this temporary, if you create it on every call?

Can place the tempEle outside of the function scope to avoid creating it in every call

src/component/Text.tsx
nikolasrieble: Right, so how would (a) unitless numbers and (b) invalid expressions change the product behaviour here?

HHongSeungWoo: image


I have made some modifications and also written a simple test code.
I excluded cases where the result of `reduce-css-calc` is `calc(~~)` because the dy attribute doesn't work correctly in those cases. 
Since invalid expressions are being ignored, i can expect the same behavior as before

ckifer:
I don't really think we want to add another suite of tests to the project without it being a separate endeavor from this PR. We can do this kind of testing in storybook with interaction tests if we really want without adding Cypress (as much as I love Cypress). 

I still think there is a better solution for removing this library, maybe with a different approach altogether

src/component/Text.tsx
nikolasrieble: Wow, thanks for looking into performance. This alone is reason to look into this more.

HHongSeungWoo:
I wanted to remove the library without any breaking changes. I understand it might look a bit odd. Personally, I also think it would be better to change capHeight and lineHeight to numbers and calculate them directly.

ckifer:
Yeah we certainly don't want any breaking changes with a refactor like this. 

If we can do that while retaining behavior that would be the direction I'd rather go in

HHongSeungWoo:
I thought this method was a way to remove it while maintaining functionality. Bringing in another library or writing new code to remove it seemed like a futile effort to me. My focus is on reducing the bundle size.

ckifer:
Sure, makes sense. 

My concern is reliance on the DOM - we've had a lot of issues with text measurement using the DOM in [this function](https://github.com/recharts/recharts/blob/1ea54d868cf43600a3d000326a7877a60d90a890/src/util/DOMUtils.ts#L76). 

It can't be unit tested, it creates shadow elements, and it may or may not create issues for server side rendering

ckifer:
This one would be good to take to reduce bundle size https://github.com/recharts/recharts/issues/3752
Hasn't been any progress for a while so I can unassign

ckifer:
If we can do that, add tests, and prove functionality is the same I'm all for it

test/util/ReduceCssCalcPrototype.spec.ts
HHongSeungWoo, line 1: @ckifer @nikolasrieble Please review it

test/util/ReduceCssCalcPOC.spec.ts
```diff
@@ -0,0 +1,135 @@
+import reduceCSSCalc from 'reduce-css-calc';
```
ckifer, line 1: this all feels like we're introducing so much complexity for us to manage for very little gain - package size will probably remain similar with these additions

test/util/ReduceCssCalcPrototype.spec.ts
```diff
@@ -0,0 +1,315 @@
+import reduceCSSCalc from 'reduce-css-calc';
```
ckifer, line 1: this file looks a lot cleaner, I like it more than the above. 

But still at the end of the day how much bundle size could this possibly save?

ckifer (commented):
Curious about the benchmarks and bundle size comparisons

the tradeoffs of us owning this don't really seem that great to me...

test/util/ReduceCssCalcPOC.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+import { log } from 'console';
+
+const conversionRates: Record<string, number> = {
+  cm: 96 / 2.54,
+  mm: 96 / 25.4,
+  pt: 96 / 72,
+  pc: 96 / 6,
+  in: 96,
+  Q: 96 / (2.54 * 40),
+  px: 1,
```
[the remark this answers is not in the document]
HHongSeungWoo: https://developer.mozilla.org/en-US/docs/Web/CSS/length

test/util/ReduceCssCalcPrototype.spec.ts
HHongSeungWoo: I'll cleanup the commits, check bundle size and performance

HHongSeungWoo:
@ckifer 
bundle size(my project)

### old
스크린샷 2023-10-14 11-46-06

### new
스크린샷 2023-10-14 11-32-36

new reduceCSSCalc is 5.4x faster then `reduce-css-calc`.

ckifer:
pulled this down and it looks great, need to move the dep to dev deps and we're good

ckifer (approved):

PavelVanecek:
Now that this is merged we can certainly delete the test that's comparing the two functions? It's failing anyway, quietly, hidden behind a try/catch+log.

nikolasrieble:
Thank you for sharing @jounii! This is helpful context. Thank you for unlocking this use case, @HHongSeungWoo

PR #3821, opened 2023-10-06
Scatter: activeShape should work with Tooltip

andrewangelle:
The existing `activeShape` prop on `<Scatter />` should work with the `<Tooltip />` in the same manner as `<Line />` and `<Bar />` do.

## Description

  - added a block to `getItemByXY` to set the find and set `activeIndex` when Tooltip is used with Scatter
  
- #### src/cartesian/Scatter.tsx
  - Refactored to use `ActiveShape` type for `props.activeShape` and `props.shape`
  
  - Made `propTransformer` optional.
  - Added case to render a `<Symbols />` component
  
- #### test/chart/ScatterChart.spec.tsx
  
## Related Issue

#3794 
## Motivation and Context

To provide a consistent interface between Line / Bar / Funnel / RadialBar / Pie / Scatter and their respective props of activeDot / activeBar / activeShape

## How Has This Been Tested?

Unit tests and storybook

Screenshot 2023-10-06 at 10 03 07 AM

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Scatter.tsx
+++ b/src/cartesian/Scatter.tsx
@@ -11,14 +11,12 @@ import { findAllByType, filterProps } from '../util/ReactUtils';
 import { Global } from '../util/Global';
 import { ZAxis, Props as ZAxisProps } from './ZAxis';
 import { Curve, Props as CurveProps, CurveType } from '../shape/Curve';
-import { Symbols, Props as SymbolsProps } from '../shape/Symbols';
 import { ErrorBar } from './ErrorBar';
 import { Cell } from '../component/Cell';
 import { uniqueId, interpolateNumber, getLinearRegression } from '../util/DataUtils';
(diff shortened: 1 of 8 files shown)


nikolasrieble:
I will want to test this in a bit more detail, I remember that the Tooltip interaction for the Scatter was rather different to all other components (especially for handling multiple scatter points with the same xAxis value)

nikolasrieble (approved):
Great! I could not find any issue here. Thank you for your contribution! This is amazing.

ckifer:
Does this not have a ton of overlap with https://github.com/recharts/recharts/pull/3818?

Could we process one at a time so we can view the diff without the overlap?

ckifer (changes requested):
Same comments as on https://github.com/recharts/recharts/pull/3818, would maybe like to get one merged before the other

andrewangelle:
@ckifer No problem. I'll close for now and re-open when the other gets merged.

Issue #3822, opened 2023-10-07
Create performance tests


PavelVanecek:
- [X] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
We know there are performance issues but we lack visibility. We need three things to improve the situation:
1. Sample data with known performance issues. There are some reports in github issues, should be a good start
2. Tools and docs for manual testing and discovery. This should be mostly covered by browser devtools and https://github.com/recharts/recharts/issues/3437
3. Automated deterministic performance test suite.

### What does the proposed API look like?
We should have a suite of performance tests that are:

- deterministic - means running the same test twice will produce same results
- the more automated the better
- if they integrate well with GitHub and can run on pull requests, would be a bonus

PavelVanecek:
So we had to rollback reassure so I will keep this open.

PR #3829, opened 2023-10-08
Add reassure perf tester

PavelVanecek:
Update contributing docs and add one sample test.
Before I go ahead and add more tests, is this workflow acceptable? Reassure does allow CI integration but let's start with dev machine for now and see how well it works for us.

## Related Issue

https://github.com/recharts/recharts/issues/3822

## Motivation and Context

## How Has This Been Tested?

`npm run test:perf`

image

image

## Types of changes

## Checklist:


test/chart/AreaChart.perf-test.tsx
```diff
@@ -0,0 +1,38 @@
...
+      pv: i * 200,
+      amt: i * 300,
+    });
+  }
+  return data;
+}
+
+describe('AreaChart', () => {
+  test('Simple test', async () => {
+    const data = createDataArray();
```
nikolasrieble, line 27: I wonder whether it would reduce the run duration if we were to create the data only one time when loading the file - as opposed to every time the test is running?

nikolasrieble (approved):
I read the docs, and think this is a great tool. It certainly seems more used and better maintained than the storybook addon.

nikolasrieble:
@ckifer Would love your opinion here before merging.

ckifer (approved):
Looks good to me, my only hesitation is we have a lot of tools and this adds another. Tools are great but I don't want this to be a think that is installed, added to the docs, but is never used. 

We'll also need to watch our CI run time and how much this adds

test/chart/AreaChart.perf-test.tsx
PavelVanecek: I added some measurements around my test like this:

```
describe('AreaChart', () => {
  test('Simple test', async () => {
    console.time('createDataArray');
    const data = createDataArray();
    console.timeEnd('createDataArray');
    console.time('measurePerformance');
    await measurePerformance(
      <AreaChart width={100} height={50} data={data}>
        <Area type="monotone" dataKey="uv" stroke="#ff7300" fill="#ff7300" />
        <Tooltip />
        <XAxis />
        <YAxis />
        <Brush />
      </AreaChart>,
    );
    console.timeEnd('measurePerformance');
  });
});
```

The result is:
- Total run time: 15.115 seconds
- Data generation: 1 millisecond
- `measurePerformance` call: 3643 milliseconds
- "Other" (I suspect jest and reassure and npm): 12 seconds

... so: no, I don't think it's worth it optimizing the `createDataArray` call.

image

image

nikolasrieble: Thank you for looking into it. I was just curious, appreciate your answer.

PR #3838, opened 2023-10-10
Add tests and types for getStackGroupsByAxisId

PavelVanecek:
## Description

I want to have props and state in generateCategoricalChart typed before I attempt the proof of concept

## Related Issue

## Motivation and Context

Confidence

## How Has This Been Tested?

TS, Jest

## Types of changes

## Checklist:

### DIFF
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -976,14 +976,31 @@ export const getStackedData = (
   return stack(data);
 };
 
+type AxisId = string;
+type StackId = string | number | symbol;
+
+export type ParentStackGroup = {
+  hasStack: boolean;
+  stackGroups: Record<StackId, ChildStackGroup>;
+};
+
+export type ChildStackGroup = {
+  numericAxisId: string;
+  cateAxisId: string;
+  items: Array<ReactElement>;
+  stackedData?: ReadonlyArray<Series<Record<string, unknown>, string>>;
+};
+
+export type AxisStackGroups = Record<AxisId, ParentStackGroup>;
+
 export const getStackGroupsByAxisId = (
-  data: any,
-  _items: Array<any>,
+  data: ReadonlyArray<Record<string, unknown>> | undefined,
+  _items: Array<ReactElement>,
   numericAxisId: string,
   cateAxisId: string,
-  offsetType: any,
+  offsetType: StackOffsetType,
   reverseStackOrder: boolean,
-) => {
+): AxisStackGroups => {
   if (!data) {
     return null;
   }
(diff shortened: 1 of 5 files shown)


src/util/ChartUtils.ts
```diff
@@ -976,33 +976,52 @@ export const getStackedData = (
...
+  items: Array<ReactElement>;
+  stackedData?: ReadonlyArray<Series<Record<string, unknown>, string>>;
+};
+
+export type AxisStackGroups = Record<AxisId, ParentStackGroup>;
+
 export const getStackGroupsByAxisId = (
-  data: any,
-  _items: Array<any>,
+  data: ReadonlyArray<Record<string, unknown>> | undefined,
```
ckifer, line 997: this any is so dangerous 😓. What if `data` is `null`?

Eventually we should make `data` generic yeah?

src/util/ChartUtils.ts
```diff
@@ -976,33 +976,52 @@ export const getStackedData = (
...
+  const stackGroups: Record<AxisId, ParentStackGroup> = items.reduce((result, item) => {
     const { stackId, hide } = item.props;
 
     if (hide) {
       return result;
     }
 
-    const axisId = item.props[numericAxisId];
-    const parentGroup = result[axisId] || { hasStack: false, stackGroups: {} };
+    const axisId: AxisId = item.props[numericAxisId];
```
ckifer, line 1020: nit nit: does `AxisId` give any more clarity than if we just declared this as `string`. Imo I see `string` and I know its a string, I see a type alias of `AxisId` and I'm thinking its something more complex

ckifer (commented):
Looks awesome! Two tiny comments/questions

src/util/ChartUtils.ts
PavelVanecek: This area withs with maps of maps. I wanted to give names to the keys even though they are both `string`.

Before:
```
const obj: Record<string, Array>
const data: Record<string, Array>
```

After:
```
const obj: Record<AxisId, Array>
const data: Record<StackId, Array>
```

ckifer: Gotcha, reasoning makes sense in this specific case

ckifer (approved):

src/util/ChartUtils.ts
PavelVanecek: The whole library is full of missing null-checks because at the moment, the way typescript is configured it behaves like Java, i.e. each and every type is nullable. We should definitely turn on https://www.typescriptlang.org/tsconfig#strictNullChecks. When I tried, the whole repository lights up with hundreds of errors so I chose to postpone that until later.

ckifer: Gotcha

PR #3839, opened 2023-10-10
Scatter: activeShape should work with Tooltip

andrewangelle:
The existing `activeShape` prop on `<Scatter />` should work with the `<Tooltip />` in the same manner as `<Line />` and `<Bar />` do.

## Description

  - Removed a line in `getItemByXY` referencing `itemDisplayName`
  
- #### src/cartesian/Scatter.tsx
  - Refactored to use `ActiveShape` type for `props.activeShape` and `props.shape`
  
  - Added case to render a `<Symbols />` component
  
- #### test/chart/ScatterChart.spec.tsx
  
## Related Issue

#3794 
## Motivation and Context

To provide a consistent interface between Line / Bar / Funnel / RadialBar / Pie / Scatter and their respective props of activeDot / activeBar / activeShape

## How Has This Been Tested?

Unit tests and storybook

Screenshot 2023-10-06 at 10 03 07 AM

## Types of changes

## Checklist:

### DIFF
--- a/src/util/ActiveShapeUtils.tsx
+++ b/src/util/ActiveShapeUtils.tsx
@@ -5,6 +5,7 @@ import { Trapezoid } from '../shape/Trapezoid';
 import { Sector } from '../shape/Sector';
 import { Layer } from '../container/Layer';
 import { GraphicalItem } from '../chart/generateCategoricalChart';
+import { Symbols, SymbolsProps } from '../shape/Symbols';
 
 /**
  * This is an abstraction for rendering a user defined prop for a customized shape in several forms.
@@ -14,7 +15,7 @@ import { GraphicalItem } from '../chart/generateCategoricalChart';
  *  - a render prop(inline function that returns jsx)
  *  - a react element
  */
-type ShapeType = 'trapezoid' | 'rectangle' | 'sector';
+type ShapeType = 'trapezoid' | 'rectangle' | 'sector' | 'symbols';
 
 export type ShapeProps<OptionType, ExtraProps, ShapePropsType> = {
   shapeType: ShapeType;
@@ -45,6 +46,8 @@ function ShapeSelector<ShapePropsType>({
       return <Trapezoid {...elementProps} />;
     case 'sector':
       return <Sector {...elementProps} />;
+    case 'symbols':
+      return <Symbols {...(elementProps as unknown as SymbolsProps)} />;
     default:
       return null;
   }
@@ -84,8 +87,6 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
  * This is an abstraction to handle identifying the active index from a tooltip mouse interaction
  */
 type GraphicalItemShapeKey = 'trapezoids' | 'sectors' | 'points';
-
-export type ItemDisplayName = 'Funnel' | 'Pie' | 'Scatter';
 
 type FunnelItem = {
   x: number;
--- a/src/util/ActiveShapeUtils.tsx
+++ b/src/util/ActiveShapeUtils.tsx
@@ -9,11 +9,16 @@ import { Symbols, SymbolsProps } from '../shape/Symbols';
 
 /**
  * This is an abstraction for rendering a user defined prop for a customized shape in several forms.
- * It will handle taking in:
+ *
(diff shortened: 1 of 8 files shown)


src/util/ActiveShapeUtils.tsx
```diff
@@ -45,6 +46,8 @@ function ShapeSelector<ShapePropsType>({
       return <Trapezoid {...elementProps} />;
     case 'sector':
       return <Sector {...elementProps} />;
+    case 'symbols':
+      return <Symbols {...(elementProps as unknown as SymbolsProps)} />;
```
ckifer, line 50: wish we didn't have to cast here

ckifer (approved):
LGTM, nice addition + some simplification!

src/util/ActiveShapeUtils.tsx
```diff
@@ -14,7 +15,7 @@ import { GraphicalItem } from '../chart/generateCategoricalChart';
  *  - a render prop(inline function that returns jsx)
  *  - a react element
  */
-type ShapeType = 'trapezoid' | 'rectangle' | 'sector';
+type ShapeType = 'trapezoid' | 'rectangle' | 'sector' | 'symbols';
```
PavelVanecek, line 18: Is it possible to elaborate in the JSDoc above what each of these values mean?

I'm reading _"It will handle taking in: *  - an object of svg properties *  - a boolean *  - a render prop(inline function that returns jsx) *  - a react element"_ but then the type itself is a union of strings? What's the relationship?

src/util/ActiveShapeUtils.tsx
PavelVanecek: +1

src/util/ScatterUtils.tsx
PavelVanecek, line 1: I wonder if perhaps a `src/components/ScatterSymbol.tsx` would have been a better place for this file?

PavelVanecek:
I have some small questions but overall looks good

src/util/ScatterUtils.tsx
ckifer: That may add the component to the public api which I don't think we want to do. We may want to think about the private/public separation mechanisms sooner than later

src/util/ActiveShapeUtils.tsx
andrewangelle: Agreed. I fell back to the cast after initially not being able to resolve the type error.

This is error:
Screenshot 2023-10-12 at 7 47 37 AM

It goes away if I add a guard:

```typescript
function isSymbolsProps(shapeType: ShapeType, _elementProps: unknown): _elementProps is SymbolsProps {
  return shapeType === 'symbols';
}
```

And then wrap the render in an if statement. It will always be true because we're already inside the `case 'symbols':` block.
```typescript
function ShapeSelector<ShapePropsType>({
  shapeType,
  elementProps,
}: {
  shapeType: ShapeType;
  elementProps: ShapePropsType;
}) {
  switch (shapeType) {
    ...
    case 'symbols':
      if (isSymbolsProps(shapeType, elementProps)) {
        return <Symbols {...elementProps} />;
      }
      break;
  }
}
```

andrewangelle: @ckifer @PavelVanecek  I've commited the above, but I don't mind reverting if either of you think this is not an optimal solution to the initial type error.

src/util/ActiveShapeUtils.tsx
andrewangelle: I see how this comment is misleading. My intention was to explain purpose of the entire section of code from lines 18 - 85 in the initial commit. 

I've updated the comment to be clearer. Feel free to leave further feedback. 
```javascript
/**
 * This is an abstraction for rendering a user defined prop for a customized shape in several forms.
 *
 * <Shape /> is the root and will handle taking in:
 *  - an object of svg properties
 *  - a boolean
 *  - a render prop(inline function that returns jsx)
 *  - a react element
 *
 * <ShapeSelector /> is a subcomponent of <Shape /> and used to match a component
 * to the value of props.shapeType that is passed to the root.
 *
 */
```

ckifer (approved):

PR #3841, opened 2023-10-10
tooltip internal payload type

SharonGrossman:
## Description

adding ```TPayload``` type to the Tooltip generic argument types

## Motivation and Context

I would like to replace the ```payload: any;``` type with ```payload: TPayload``` type in order to control and achieve better type safety additional data when using a Custom Tooltip
## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.

### DIFF
--- a/src/component/DefaultTooltipContent.tsx
+++ b/src/component/DefaultTooltipContent.tsx
@@ -13,45 +13,46 @@ function defaultFormatter<TValue extends ValueType>(value: TValue) {
 export type TooltipType = 'none';
 export type ValueType = number | string | Array<number | string>;
 export type NameType = number | string;
-export type Formatter<TValue extends ValueType, TName extends NameType> = (
+export type PayloadType = number | string | Array<number | string> | { [key: string]: any };
+export type Formatter<TValue extends ValueType, TName extends NameType, TPayload extends PayloadType> = (
   value: TValue,
   name: TName,
-  item: Payload<TValue, TName>,
+  item: Payload<TValue, TName, TPayload>,
   index: number,
-  payload: Array<Payload<TValue, TName>>,
+  payload: Array<Payload<TValue, TName, TPayload>>,
 ) => [React.ReactNode, TName] | React.ReactNode;
 
-export interface Payload<TValue extends ValueType, TName extends NameType> {
+export interface Payload<TValue extends ValueType, TName extends NameType, TPayload extends PayloadType> {
   type?: TooltipType;
   color?: string;
-  formatter?: Formatter<TValue, TName>;
+  formatter?: Formatter<TValue, TName, TPayload>;
   name?: TName;
   value?: TValue;
   unit?: ReactNode;
   dataKey?: string | number;
-  payload?: any;
+  payload?: TPayload;
   chartType?: string;
   stroke?: string;
   strokeDasharray?: string | number;
   strokeWidth?: number | string;
 }
 
-export interface Props<TValue extends ValueType, TName extends NameType> {
+export interface Props<TValue extends ValueType, TName extends NameType, TPayload extends PayloadType> {
   separator?: string;
   wrapperClassName?: string;
   labelClassName?: string;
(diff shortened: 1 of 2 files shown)


src/component/DefaultTooltipContent.tsx
```diff
@@ -13,45 +13,46 @@ function defaultFormatter<TValue extends ValueType>(value: TValue) {
 export type TooltipType = 'none';
 export type ValueType = number | string | Array<number | string>;
 export type NameType = number | string;
-export type Formatter<TValue extends ValueType, TName extends NameType> = (
+export type PayloadType = number | string | Array<number | string> | { [key: string]: any };
```
ckifer, line 16: nit: instead of `{ [key: string]: any }` can we use `Record<string, any>`?

ckifer (changes requested):
We need to make sure that this isn't a breaking change else we'll have to wait for a major release. Is there any way you can prove that this works in all cases?

SharonGrossman:
> We need to make sure that this isn't a breaking change else we'll have to wait for a major release. Is there any way you can prove that this works in all cases?

I'm not sure a simple unit test can catch this, do you have anything I can do in mind?

PavelVanecek:
I'm always for a better type safety.

This can be a breaking change - for projects who went ahead and imported undocumented types and used them.

PavelVanecek:
Also yes @SharonGrossman having a unit test would be amazing; it won't have any `expect` calls in it but it will have a bunch of variables showcasing that defining a variable this way works, and that way works, and on the other hand this won't work because `// @ts-expect-error explanation of the error`.

ckifer:
Thanks for the options @PavelVanecek! Since we can't really gauge impact of this I would like to try our best not to break consumers if we can. For trust reasons with the consumer if nothing else. For 2 how likely is payload to be a number? Wouldn't it be more likely an object? 

3 seems the most viable to me although a bit annoying until the next MV. 

+1 on the unit test suggestions.

SharonGrossman:
Thanks for the comments guys, I think 3 is the most reasonable & safe solution, if we agree I can get to doing that.
About the unit test, I can try to do one test without assertion, although it's a little weird IMO

ckifer:
Thanks for being flexible! 

I say go for it - one thing to note is just to make sure the naming is okay. When we make the major version jump we could either 
1. rename back to the old names and let the consumer break with instructions to fix
2. use the new names forever and remove the old

If we do 1 make sure the names are clear they are temporary, if we do 2 make sure they are something clean we can use going into the future. No preference on which.

SharonGrossman:
I'm up for renaming when major is released, but the names can be like you suggested can work

ckifer:
Cool, first option it is then. Feel free to make the new type whatever you want as long as it makes sense of course. Maybe we can add a jsdoc comment above it explaining what it is as well as make an issue for the major release follow up

ckifer:
Add the default to the PR and lets test it quite a bit. I will pull it down, do a local release, and try to see if there are any common scenarios where TS might break. If not, we can ship - if people break and report it we can give an easy way to mitigate or expand the type to accommodate if it makes sense

SharonGrossman:
Okay, I created a default type, let me know if you think there's anything to do beside test it 🙏

ckifer:
Cool, if anyone on the thread wants to pull this down and try to break it feel free

ckifer:
@SharonGrossman mind fixing the merge conflicts on this? After that I'll pull it down and try to break it

ckifer:
Pulled it down, I think this works fine because the default includes `Record<string, any>` - if it didn't this wouldn't be viable. Just merge master and we should be good here

ckifer:
I tried to fix merge conflicts on this but couldn't in a reasonable amount of time :/

ckifer:
bump on fixing the merge conflict here

ckifer:
Going to close this as stale for now

PR #3843, opened 2023-10-11
Calculate offset types

PavelVanecek:
## Description

On a quest to make recharts better!

## Related Issue

## Motivation and Context

Testssss

## How Has This Been Tested?

TS, Jest

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -656,11 +656,11 @@ const getAxisNameByLayout = (layout: LayoutType) => {
 
 /**
  * Calculate the offset of main part in the svg element
- * @param  {Object} props          Latest props
- * graphicalItems The instances of item
- * xAxisMap       The configuration of x-axis
- * yAxisMap       The configuration of y-axis
- * @param  {Object} prevLegendBBox          the boundary box of legend
+ * @param  {Object} params.props          Latest props
+ * @param  {Array}  params.graphicalItems The instances of item
+ * @param  {Object} params.xAxisMap       The configuration of x-axis
+ * @param  {Object} params.yAxisMap       The configuration of y-axis
+ * @param  {Object} prevLegendBBox        The boundary box of legend
  * @return {Object} The offset of main part in the svg element
  */
 const calculateOffset = (
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -75,6 +75,8 @@ import { AccessibilityManager } from './AccessibilityManager';
 import { isDomainSpecifiedByUser } from '../util/isDomainSpecifiedByUser';
 import { deferer, CancelFunction } from '../util/deferer';
 import { getActiveShapeIndexForTooltip, isFunnel, isPie, isScatter } from '../util/ActiveShapeUtils';
+import { Props as YAxisProps } from '../cartesian/YAxis';
+import { Props as XAxisProps } from '../cartesian/XAxis';
 
 export type GraphicalItem<Props = Record<string, any>> = ReactElement<
   Props,
@@ -662,16 +664,26 @@ const getAxisNameByLayout = (layout: LayoutType) => {
  * @return {Object} The offset of main part in the svg element
  */
 const calculateOffset = (
-  { props, graphicalItems, xAxisMap = {} as BaseAxisProps, yAxisMap = {} as BaseAxisProps }: any,
-  prevLegendBBox?: any,
-) => {
+  {
+    props,
+    graphicalItems,
+    xAxisMap = {},
+    yAxisMap = {},
+  }: {
+    props: CategoricalChartProps;
+    graphicalItems: Array<ReactElement>;
+    xAxisMap?: { [axisId: string]: XAxisProps };
+    yAxisMap?: { [axisId: string]: YAxisProps };
+  },
+  prevLegendBBox?: DOMRect | null,
+): ChartOffset => {
   const { width, height, children } = props;
   const margin = props.margin || {};
   const brushItem = findChildByType(children, Brush);
   const legendItem = findChildByType(children, Legend);
 
   const offsetH = Object.keys(yAxisMap).reduce(
-    (result: any, id: any) => {
+    (result, id) => {
       const entry = yAxisMap[id];
       const { orientation } = entry;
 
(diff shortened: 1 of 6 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -662,16 +664,26 @@ const getAxisNameByLayout = (layout: LayoutType) => {
  * @return {Object} The offset of main part in the svg element
```
ckifer, line 664: Can we correct the JSDoc too since its there?

src/chart/generateCategoricalChart.tsx
```diff
@@ -662,16 +664,26 @@ const getAxisNameByLayout = (layout: LayoutType) => {
...
+  {
+    props,
+    graphicalItems,
+    xAxisMap = {},
+    yAxisMap = {},
+  }: {
+    props: CategoricalChartProps;
+    graphicalItems: Array<ReactElement>;
+    xAxisMap?: { [axisId: string]: XAxisProps };
+    yAxisMap?: { [axisId: string]: YAxisProps };
```
ckifer, line 676: nit: could we use `Record<string, X/YAxisProps>` instead?

src/chart/generateCategoricalChart.tsx
```diff
@@ -707,6 +719,7 @@ const calculateOffset = (
   }
 
   if (legendItem && prevLegendBBox) {
+    // @ts-expect-error margin is optional in props but required in appendOffsetOfLegend
     offset = appendOffsetOfLegend(offset, graphicalItems, props, prevLegendBBox);
```
ckifer, line 723: what happens when margin isn't provided 😅? wouldn't this have caused runtime errors in the `appendOffsetOfLegend` calculations below?

src/util/ChartUtils.ts
```diff
@@ -437,27 +380,41 @@ export const getBarPosition = ({
   return result;
 };
 
-export const appendOffsetOfLegend = (offset: any, items: Array<FormattedGraphicalItem>, props: any, legendBox: any) => {
+export const appendOffsetOfLegend = (
+  offset: ChartOffset,
+  _unused: unknown,
```
ckifer, line 385: if `items` was unused could we not just remove it?

src/util/ChartUtils.ts
```diff
@@ -437,27 +380,41 @@ export const getBarPosition = ({
...
   if (legendProps) {
-    const box = legendBox || {};
+    const { width: boxWidth, height: boxHeight } = legendBox || {};
     const { align, verticalAlign, layout } = legendProps;
 
-    if ((layout === 'vertical' || (layout === 'horizontal' && verticalAlign === 'middle')) && isNumber(offset[align])) {
-      newOffset = { ...offset, [align]: newOffset[align] + (box.width || 0) };
+    if (
+      (layout === 'vertical' || (layout === 'horizontal' && verticalAlign === 'middle')) &&
+      align !== 'center' &&
```
ckifer, line 402: is this a new condition? why?

src/util/getLegendProps.ts
```diff
@@ -0,0 +1,66 @@
+import { ReactNode, ReactElement } from 'react';
+import { Legend, Props as LegendProps } from '../component/Legend';
+import { FormattedGraphicalItem, getMainColorOfGraphicItem } from './ChartUtils';
+import { findChildByType } from './ReactUtils';
+import { Payload as LegendPayload } from '../component/DefaultLegendContent';
+
+interface SectorOrDataEntry {
+  name: any;
+  fill: any;
```
ckifer, line 9: these can't be typed?

ckifer (commented):
Love the improvements! Just some questions

src/chart/generateCategoricalChart.tsx
PavelVanecek: I chose the indexer because Record does not allow me to give a name to the key. Is there another difference between Record and object I am not aware of?

src/chart/generateCategoricalChart.tsx
PavelVanecek: I think the whole chart just won't render without errors. Not sure where that's coming from. Definitely suspicious.

src/util/ChartUtils.ts
PavelVanecek: That could be a breaking change.

src/util/ChartUtils.ts
PavelVanecek: To satisfy typescript - without this condition, TS was flagging (correctly) that `offset` is never going to have property `center`.

src/util/getLegendProps.ts
PavelVanecek: Yes but in some other PR

src/chart/generateCategoricalChart.tsx
ckifer: No it's the same, I just personally prefer it for brevity but your reason makes sense

src/chart/generateCategoricalChart.tsx
ckifer: Hmm - I think there is default margin? Maybe it's optional but defaulted

src/util/ChartUtils.ts
ckifer: How? I guess I generally wouldn't expect anyone externally to be using this function and if they are it certainly isn't documented. Would say semver doesn't really apply to helper functions (doesn't help that we export literally everything but still)

src/util/ChartUtils.ts
ckifer: Gotcha, cool. As long as the logic shakes out to be the same

src/util/ChartUtils.ts
PavelVanecek: Perhaps the next major change in recharts should be to stop exporting things and have a clear distinction between private and public APIs

ckifer: Keep or remove for this one I guess doesn't matter to me, I would argue if someone breaks it was their risk to accept (and in this case the fix is easy). 

But yes completely agree, that helps us with maintainability and even allows us to export some non-component functions that may be helpful and document them.

src/chart/generateCategoricalChart.tsx
PavelVanecek: Updated

PR #3844, opened 2023-10-11
Change type of graphicalItems from any to ReadonlyArray<ReactElement>

PavelVanecek:
## Description

Types!

## Related Issue

## Motivation and Context

Types

## How Has This Been Tested?

TS, Jest

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -146,7 +146,7 @@ export class Bar extends PureComponent<Props, State> {
     offset,
   }: {
     props: Props;
-    item: Bar;
+    item: ReactElement;
     barPosition: any;
     bandSize: number;
     xAxis: InternalBarProps['xAxis'];
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -152,9 +152,9 @@ const getActiveCoordinate = (
 const getDisplayedData = (
   data: any[],
   { graphicalItems, dataStartIndex, dataEndIndex }: CategoricalChartState,
-  item?: any,
+  item?: ReactElement,
 ): any[] => {
-  const itemsData = (graphicalItems || []).reduce((result: any, child: any) => {
+  const itemsData = (graphicalItems || []).reduce((result, child) => {
     const itemData = child.props.data;
 
     if (itemData && itemData.length) {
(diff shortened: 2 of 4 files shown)


src/cartesian/Bar.tsx
```diff
@@ -146,7 +146,7 @@ export class Bar extends PureComponent<Props, State> {
     offset,
   }: {
     props: Props;
-    item: Bar;
```
ckifer, line 149: was `Bar` referencing the class in this case? :o. Is that not still more accurate than a generic ReactElement?

src/chart/generateCategoricalChart.tsx
```diff
@@ -874,6 +874,7 @@ export const generateCategoricalChart = ({
           (axisMap && axisMap[id]) || entry.axisType === 'zAxis',
           `Specifying a(n) ${entry.axisType}Id requires a corresponding ${
             entry.axisType
+            // @ts-expect-error we should stop reading data from ReactElements
```
ckifer, line 877: 💯% agree, the data has to come from somewhere though. Eventually, we refactor this to get the data from... well data instead of elements 😅. In the meantime the invariant is still useful with the expect error

ckifer (commented):
Looks good, just one question on the class thing

src/cartesian/Bar.tsx
ckifer: Could we do `ReactElement<BarProps>` to give the needed specificity?

PavelVanecek: So two answers, one short term (this PR) one long term (recharts future).

Short term: Yes, using `Bar` is more accurate. But it is only accurate on the receiving end! In the sending end there is no guarantee. `generateCategoricalChart` sends ReactElement. So this was a false positive, going against Barbara Liskov principle.

Long term: In my opinion, architecture where a component requires a _rendered instance of itself_ to look up in the global state before it decides if it can render - is not a good architecture. This whole function should be different and this argument should not exist.

ckifer: Agreed with the long term for sure, one of the major issues with why generateCategoricalChart is how it is. 

Unfortunate that short term we can't have some sort of guarantee of what we're receiving, but I guess that's just the cards we've been deal with this little mess. 

Thanks

ckifer (approved):

PR #3848, opened 2023-10-12
Add stackGroups and zAxisMap to CategoricalChartState, fix types

PavelVanecek:
## Description

Types!

## Related Issue

## Motivation and Context

One less `any`

## How Has This Been Tested?

TS, Jest

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -859,6 +859,23 @@ export interface CategoricalChartProps {
   tabIndex?: number;
 }
 
+type AxisObj = {
+  xAxis?: BaseAxisProps;
+  xAxisTicks?: Array<TickItem>;
+
+  yAxis?: BaseAxisProps;
+  yAxisTicks?: Array<TickItem>;
+
+  zAxis?: BaseAxisProps;
+  zAxisTicks?: Array<TickItem>;
+
+  angleAxis?: BaseAxisProps;
+  angleAxisTicks?: Array<TickItem>;
+
+  radiusAxis?: BaseAxisProps;
+  radiusAxisTicks?: Array<TickItem>;
+};
+
 export const generateCategoricalChart = ({
   chartName,
   GraphicalChild,
@@ -884,23 +901,6 @@ export const generateCategoricalChart = ({
       const numericAxisId = item.props[`${numericAxisName}Id`];
       // axisId of the categorical axis
       const cateAxisId = item.props[`${cateAxisName}Id`];
-
-      type AxisObj = {
-        xAxis?: BaseAxisProps;
-        xAxisTicks?: Array<TickItem>;
-
-        yAxis?: BaseAxisProps;
-        yAxisTicks?: Array<TickItem>;
-
-        zAxis?: BaseAxisProps;
-        zAxisTicks?: Array<TickItem>;
-
-        angleAxis?: BaseAxisProps;
-        angleAxisTicks?: Array<TickItem>;
-
-        radiusAxis?: BaseAxisProps;
-        radiusAxisTicks?: Array<TickItem>;
-      };
 
       const axisObjInitialValue: AxisObj = {};
 
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -51,6 +51,7 @@ import {
   parseDomainOfCategoryAxis,
   getTooltipItem,
   BarPosition,
+  AxisStackGroups,
 } from '../util/ChartUtils';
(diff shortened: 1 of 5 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -859,11 +884,31 @@ export const generateCategoricalChart = ({
...
+
+        zAxis?: BaseAxisProps;
+        zAxisTicks?: Array<TickItem>;
+
+        angleAxis?: BaseAxisProps;
+        angleAxisTicks?: Array<TickItem>;
+
+        radiusAxis?: BaseAxisProps;
+        radiusAxisTicks?: Array<TickItem>;
+      };
```
nikolasrieble, line 903: Any reason to define this type inside of the function? Not against it, just wondering. 
If this object is something that is passed around, we might reuse it elsewhere. 🤔

src/util/ChartUtils.ts
```diff
@@ -977,20 +977,22 @@ export const getStackedData = (
...
-type StackId = string | number | symbol;
+export type StackId = string | number | symbol;
 
 export type ParentStackGroup = {
   hasStack: boolean;
   stackGroups: Record<StackId, ChildStackGroup>;
 };
 
-export type ChildStackGroup = {
+export type GenericChildStackGroup<T> = {
```
nikolasrieble, line 987: Nice.

test/util/ChartUtils/getStackedDataOfItem.spec.ts
```diff
@@ -42,31 +56,49 @@ describe('getStackedDataOfItem', () => {
...
     const group = Symbol('my mock group');
-    const stackGroups = { a: { stackedData: [group], items: [item] } };
+    const stackGroups: Record<StackId, GenericChildStackGroup<typeof group>> = {
+      a: {
+        stackedData: [group],
+        items: [item],
+        numericAxisId: '',
+        cateAxisId: '',
+      },
+    };
```
nikolasrieble, line 81: How about extracting a helper function to generate the stackGroup? 

```
generateMockStackGroupFor(group, item){
    return {
      a: {
        stackedData: [group],
        items: [item],
        numericAxisId: '',
        cateAxisId: '',
      },
    };
}
```

test/util/DataUtils.spec.ts
```diff
@@ -72,6 +72,15 @@ describe('is functions', () => {
         expect(isNumber(value)).toBe(result);
       });
     });
+
+    it('should allow type refinement', () => {
```
nikolasrieble, line 76: This seems like a test of Typescript behaviour itself. Did you intend to commit this? 🤔 If yes, why?

nikolasrieble (changes requested):
A few remarks, but only one question really: 

Why the additional test in DataUtils.spec.ts?

src/chart/generateCategoricalChart.tsx
ckifer: yeah +1, generally types aren't defined in functions even if they aren't reused

test/util/DataUtils.spec.ts
ckifer: I guess its a test that `isNumber` ends up allowing type refinement yeah? But also yeah +1 I don't really see the value - if we know it can do type refinement it never won't be able to

src/chart/generateCategoricalChart.tsx
PavelVanecek: 🤷 I was just going for the smallest scope required. I can move it.

test/util/DataUtils.spec.ts
PavelVanecek: So I wrote this test for two reasons:

1. I was struggling with a type error in different file. That error might have been caused by the missing refinement. I wasn't certain if it does behave that way so I wrote a test. And then I didn't delete the test when I was done debugging.
2. I wanted to have an example for the next time someone asks how to write unit tests for types - I remember it came up in the other PR.

That said, I can delete it if we feel like it's taking up too much space.

test/util/ChartUtils/getStackedDataOfItem.spec.ts
PavelVanecek: Added

src/chart/generateCategoricalChart.tsx
PavelVanecek: Moved

test/util/DataUtils.spec.ts
ckifer: eh I think its fine but maybe we can add a comment that explains it since it is slightly confusing at a glance

ckifer (approved):
🚀

Issue #3823, opened 2023-10-07
Proof of Concept of plan 5a


PavelVanecek:
- [X] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
In https://github.com/recharts/recharts/discussions/3717 we have discussed two approaches of refactoring recharts. This is the tracker to create a proof of concept branch where we can observe the code changes and impact on performance and API.

### What does the proposed API look like?
- Stop relying on `displayName`.
- Instead of parsing children array and reading props, have each component push their existence (and their props) to a shared config.
- `generateCategoricalChart` will still have to render each component but it won't have to include dozens of if layout == 'horizontal' because that information would have been pushed by the components.
- This looks like a "Push model"

PavelVanecek:
We have decided to go forward with 5b

PR #3850, opened 2023-10-12
Adding types to renderCursor

PavelVanecek:
Bonus: extracted getRadialCursorPoints function so that I can have it typed differently from the rest: it returns different object. And added unit test.

## Related Issue

## Motivation and Context

I think I can now make sense of the Tooltip rendering and proceed with https://github.com/recharts/recharts/issues/3823

## How Has This Been Tested?

TS, Jest

## Types of changes

## Checklist:

### DIFF
--- a/src/util/types.ts
+++ b/src/util/types.ts
@@ -1266,11 +1266,13 @@ export const adaptEventsOfChild = (
   return out;
 };
 
+export type TooltipEventType = 'axis' | 'item';
+
 export interface CategoricalChartOptions {
   chartName?: string;
   GraphicalChild?: any;
-  defaultTooltipEventType?: string;
-  validateTooltipEventTypes?: string[];
+  defaultTooltipEventType?: TooltipEventType;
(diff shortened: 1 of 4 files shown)


src/util/types.ts
```diff
@@ -1266,11 +1266,13 @@ export const adaptEventsOfChild = (
   return out;
 };
 
+export type TooltipEventType = 'axis' | 'item';
```
nikolasrieble, line 1269: Nice! I wonder whether we can make this an actual option to the user instead of having it be hard defined per chart type. 

For example in the case of the multi line chart, it is a standard use case to only want to show a single dots Tooltip (as opposed to recharts default behaviour of showing all values per xAxis value). 

If I understand correctly this is the switch between the two behaviours.

nikolasrieble: Totally unrelated to this PR, just wanted to share some more context knowledge.

nikolasrieble (approved):
Love it, thank you for yet another small and targeted improvement.

ckifer (approved):
🚀

PR #3852, opened 2023-10-13
DO NOT MERGE: Request for comments: Plan 5a Proof of concept

PavelVanecek:
So this doesn't pass tests, I don't think it actually renders the Tooltip at all. All the problems are fixable, I am just wondering what do you think about the general direction before I sink the time into fixing them.

In the meantime I would like to use the Pull Request as a vehicle for comments and discussion.

Please focus on:
- Overall architecture and data flow
- Breaking API changes
- Any changes we want to bring to master independent of the refactor to minimize the PR size and reduce risk?

Please do not yet pay attention to:
- Variable naming
- Component split
- File names
- `any`

## Description

As described in https://github.com/recharts/recharts/discussions/3717:
- The `Tooltip` component is now purely a marker. It renders nothing. It only cleans its props, adds some defaults, and pushes everything into a React.Context
- `generateCategoricalChart` no longer needs to search for the `Tooltip` Element in DOM because it has all its data in the context. It gets some precomputed data and global state and passes it down to a new component:
- `TooltipRenderer` This is grouping all the small individual renders that used to be split in three places before:
   1. `Tooltip.render`
   2. `generateCategoricalChart.renderCursor`
   3. `generateCategoricalChart.renderTooltip`

## Related Issue

https://github.com/recharts/recharts/issues/3823

## Motivation and Context

## How Has This Been Tested?

I tested it won't work, please don't merge it yet.

## Types of changes

## Checklist:

- [X] Only discuss, no merge

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1,4 +1,4 @@
-import React, { Component, cloneElement, isValidElement, createElement, ReactElement } from 'react';
+import React, { Component, cloneElement, isValidElement, ReactElement } from 'react';
 import isNil from 'lodash/isNil';
 import isFunction from 'lodash/isFunction';
 import range from 'lodash/range';
@@ -12,17 +12,13 @@ import clsx from 'clsx';
 // eslint-disable-next-line no-restricted-imports
 import type { DebouncedFunc } from 'lodash';
 import invariant from 'tiny-invariant';
-import { getRadialCursorPoints } from '../util/cursor/getRadialCursorPoints';
 import { getTicks } from '../cartesian/getTicks';
 import { Surface } from '../container/Surface';
 import { Layer } from '../container/Layer';
 import { Tooltip } from '../component/Tooltip';
 import { Legend } from '../component/Legend';
-import { Curve } from '../shape/Curve';
-import { Cross } from '../shape/Cross';
-import { Sector } from '../shape/Sector';
 import { Dot } from '../shape/Dot';
-import { isInRectangle, Rectangle } from '../shape/Rectangle';
+import { isInRectangle } from '../shape/Rectangle';
 
 import {
   filterProps,
@@ -89,8 +85,8 @@ import { isDomainSpecifiedByUser } from '../util/isDomainSpecifiedByUser';
 import { getActiveShapeIndexForTooltip, isFunnel, isPie, isScatter } from '../util/ActiveShapeUtils';
 import { Props as YAxisProps } from '../cartesian/YAxis';
 import { Props as XAxisProps } from '../cartesian/XAxis';
-import { getCursorPoints } from '../util/cursor/getCursorPoints';
-import { getCursorRectangle } from '../util/cursor/getCursorRectangle';
+import { TooltipRenderer } from '../renderer/TooltipRenderer';
+import { ChartContextContainer } from '../context/chartContext';
 
 export interface MousePointer {
   pageX: number;
@@ -112,6 +108,22 @@ const ORIENT_MAP = {
 const FULL_WIDTH_AND_HEIGHT = { width: '100%', height: '100%' };
 
 const originCoordinate: Coordinate = { x: 0, y: 0 };
+
+/**
+ * This function exists as a temporary workaround.
+ *
+ * Why? generateCategoricalChart does not render `{children}` directly;
+ * instead it passes them through `renderByOrder` function which reads their handlers.
+ *
+ * So, this is a handler that does nothing.
+ * Once we get rid of `renderByOrder` and switch to JSX only, we can get rid of this handler too.
+ *
+ * @param {JSX} element as is in JSX
+ * @returns {JSX} the same element
+ */
+function renderAsIs(element: React.ReactElement): React.ReactElement {
+  return element;
+}
 
 const calculateTooltipPos = (rangeObj: any, layout: LayoutType): any => {
   if (layout === 'horizontal') {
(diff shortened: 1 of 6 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -2312,7 +2250,6 @@ export const generateCategoricalChart = ({
         Scatter: { handler: this.renderGraphicChild },
         Pie: { handler: this.renderGraphicChild },
         Funnel: { handler: this.renderGraphicChild },
-        Tooltip: { handler: this.renderCursor, once: true },
```
PavelVanecek, line 2315: Tooltip no longer uses the cloning map!

src/component/Tooltip.tsx
```diff
@@ -90,6 +53,33 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Def
...
+      useTranslate3d: props.useTranslate3d,
+      // @ts-expect-error TODO
+      allowEscapeViewBox: props.allowEscapeViewBox,
+      // @ts-expect-error TODO
+      reverseDirection: props.reverseDirection,
+      offsetTopLeft: props.offset,
+      // @ts-expect-error TODO
+      position: props.position,
+    };
+    setContext(newContext);
```
PavelVanecek, line 77: The public `Tooltip` will now only gather and cleanup the public props and push to context - nothing more.

src/context/chartContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Payload } from '../component/DefaultTooltipContent';
+
+export type TooltipContentType = ReactElement | ((props: any) => ReactNode);
+
+export type AllowInDimension = {
+  x: boolean;
+  y: boolean;
+};
+
+export type ChartContextType = {
```
PavelVanecek, line 13: The context might grow quite a bit, we might consider splitting to one context per component.

src/renderer/TooltipRenderer.tsx
PavelVanecek, line 1: Here is the actual Tooltip + Cursor rendering. Reads from props (shared global state) and context (public props).

PavelVanecek:
The more I work on the codebase the less I am convinced we can have a clean split between 5a ("Push") and 5b ("Pull"). It might end up as a combination of both, either way.

src/renderer/TooltipRenderer.tsx
```diff
@@ -0,0 +1,383 @@
...
+
+  return isValidElement(cursor) ? cloneElement(cursor, cursorProps) : createElement(cursorComp, cursorProps);
+}
+
+type UniqueFunc<TValue extends ValueType, TName extends NameType> = (entry: Payload<TValue, TName>) => unknown;
+type UniqueOption<TValue extends ValueType, TName extends NameType> = boolean | UniqueFunc<TValue, TName>;
+
+function defaultUniqBy<TValue extends ValueType, TName extends NameType>(entry: Payload<TValue, TName>) {
+  return entry.dataKey;
+}
```
ckifer, line 112: this is defined in the file above too?

ckifer (commented):
I like the direction! We'll have to be careful about how unwieldy that context gets, it could quite literally get as bad as generateCategoricalChart itself lol

src/chart/generateCategoricalChart.tsx
```diff
@@ -1967,27 +1906,23 @@ export const generateCategoricalChart = ({
...
-      return cloneElement(tooltipItem, {
-        viewBox: { ...offset, x: offset.left, y: offset.top },
-        active: isTooltipActive,
-        label: activeLabel,
-        payload: isTooltipActive ? activePayload : [],
-        coordinate: activeCoordinate,
-      });
+      const { layout } = this.props;
+      return (
+        <TooltipRenderer
```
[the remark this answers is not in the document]
PavelVanecek: Yes that's the idea. The published ones are "markers", private ones render.

src/renderer/TooltipRenderer.tsx
PavelVanecek: I don't actually intend to merge this the way it is now. All defaults resolving should be happening before moving data to context. So the private renderer can only work with one shape of props.

ckifer:
I think tests are basically passing besides snaps?

PavelVanecek:
There's still couple of tests red. As I'm going through them I'm
discovering small problems. I want to get all tests green to make sure I
didn't miss any roadblock, and then I move on to 5b.

ckifer:
Sounds good!

PavelVanecek:
I can't get the latest changes to work. In the name of performance optimization, this PR introduced mutable variables outside of state: https://github.com/HHongSeungWoo/recharts/commit/830bd0040cd0554f0b252c1e2e3bb8c457668ae8 and I can't figure out if it's meant to be that way and I should follow that or if that's an oversight.

HHongSeungWoo:
Not using the state was an intentional change. Currently, the tooltip is being rerendered by outer components, so there is no need to trigger state change. but you can change it if necessary

PavelVanecek:
I don't have enough insight to tell if it's necessary or not but I can tell it looks like an oversight. I recommend to add unit tests that show that it is intentional and what unnecessary re-renders does it prevent. Otherwise you can have someone come and go "This is not The React Way(tm)" and change it.

PavelVanecek:
And a comment too 😬

src/renderer/TooltipRenderer.tsx
nikolasrieble: Combining the Tooltip and the Cursor generally seems like a great refactoring. Do you see a way to get this in independent on the conceptual change?

PavelVanecek: Yeah so without the conceptual change the component of `TooltipWithCursor` will accept merged props from both. There might be slight complications since they are both rendered at different times (not sure why?). I will look into that.

PavelVanecek:
Sibling PR: https://github.com/recharts/recharts/pull/3965

PavelVanecek:
Declining in favour of https://github.com/recharts/recharts/pull/3965

src/renderer/TooltipRenderer.tsx
PavelVanecek: https://github.com/recharts/recharts/pull/3994

Issue #3640, opened 2023-06-23
Automatic bar width is set to zero when there is only one data point


Nicolapps:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/tiny-bar-chart-forked-gppdsy?file=/src/App.tsx)

### Steps to reproduce
* Create a `BarChart` with a `XAxis` of type `number` and a `domain`
* Use a data set that has only one data point

### What is expected?
The bar has a width that makes it visible.

### What is actually happening?
The bar has a zero width by default, which makes it invisible.

| Environment | Info |
|---|---|
| Recharts | v2.7.2 |
| React | 18.2.0 |
| System | macOS Ventura 13.4 |
| Browser | Safari 16.5, Firefox 114.0.2, Chrome 114.0.5735.133 |

Nicolapps:
It is worth noting that if your chart is affected by this behavior, it is possible to set manually `barSize` on `Bar` in the case where there is only one data point (and leave it `undefined` otherwise to keep using the automatic width).

ckifer:
This behavior is the same in previous versions right? (I think it is)

Nicolapps:
Yes, I have the same behavior with earlier versions.

graup:
This is still broken and the `barSize` workaround doesn't work for me. Does anyone know any other workarounds?

graup:
I created another example showing how this works (or doesn't work) with date axes: https://codesandbox.io/p/sandbox/tiny-bar-chart-forked-vk9fyv?file=%2Fsrc%2FApp.tsx

Here is what I expect:
![Screenshot 2024-03-19 at 11 41 19](https://github.com/recharts/recharts/assets/898549/2bb99284-5222-4267-9ec5-74491f7d4f20)
![expected](https://github.com/recharts/recharts/assets/898549/a9474dc3-38ef-44ef-b0e3-267a70aaeba8)

I don't believe the PR achieves what's in the second image, instead making the bar always 1/3 the width of the chart. How can we have one bar that's the size of one tick?

graup:
After much digging, I have concluded that there is no easy & correct fix for this.

[...]

PR #3853, opened 2023-10-13
Fix: bar width when there is only one data point

souzolliveira:
## Description

- Adding condition to return a bar size when there is only one data point;
- Creating test for it.

## Related Issue

- Bug #3640 

## Motivation and Context

- I choose this bug to solve for a graduation issue and because it seems easy.

## How Has This Been Tested?

- I created a test to count one rectangle for one data point.

## Screenshots:

image

## Types of changes

## Checklist:


test/chart/BarChart.spec.tsx
```diff
@@ -14,6 +14,19 @@ describe('<BarChart />', () => {
...
+    const { container } = render(
+      <BarChart width={800} height={200} data={[{ x: 1, y: 200 }]}>
+        <XAxis type="number" dataKey="x" />
+        <YAxis />
+        <Tooltip />
+        <Bar dataKey="y" />
+      </BarChart>,
+    );
+
+    expect(container.querySelectorAll('.recharts-rectangle')).toHaveLength(1);
```
ckifer, line 27: can you confirm that this test doesn't pass before your change? If the rectangle gets rendered with a 0 width this would still pass

src/util/ChartUtils.ts
```diff
@@ -1266,18 +1266,28 @@ export const getBandSizeOfAxis = (
     }
   }
 
-  if (axis && ticks && ticks.length >= 2) {
-    const orderedTicks = _.sortBy(ticks, o => o.coordinate);
-    let bandSize = Infinity;
+  // @ts-expect-error we need to rethink width type
```
ckifer, line 1269: Instead of doing this maybe we re-think the type sent to this function. `BaseAxisProps` does not include `width` but `XAxis` and YAxis` props do. 

What if this type became a union of `XAxisProps` | `YAxisProps` | `ZAxisProps` and type narrowing can be done from there. That way we can add this with correct TS defs.

ckifer (changes requested):
Thanks for contributing! A few questions

src/util/ChartUtils.ts
souzolliveira: Fixed

test/chart/BarChart.spec.tsx
souzolliveira: Really, this condition can be satisfied in this case if the rectangle renders with a 0 width. I will verify.

souzolliveira: The .recharts-rectangle isn't rendered when its width is zero, then, this test fails without my changes.

src/util/ChartUtils.ts
```diff
@@ -1251,11 +1252,7 @@ export const parseSpecifiedDomain = (specifiedDomain: any, dataDomain: any, allo
  * @param  {Boolean} isBar if items in axis are bars
  * @return {Number} Size
  */
-export const getBandSizeOfAxis = (
-  axis?: BaseAxisProps,
-  ticks?: Array<TickItem>,
-  isBar?: boolean,
-): number | undefined => {
+export const getBandSizeOfAxis = (axis?: XAxisProps, ticks?: Array<TickItem>, isBar?: boolean): number | undefined => {
```
ckifer, line 1255: now the type isn't accurate because this function isn't called for only `XAxis`

We need a union type + type narrowing

So
```ts
// defined above or elsewhere
type CartesianAxisProps = XAxisProps | YAxisProps | ZAxisProps


// in function
 if (axis && 'width' in axis && axis.width && ticks) {
```

ckifer (changes requested):

src/util/ChartUtils.ts
souzolliveira: I applied the fixes. Can you verify if it is correct now? Pls

src/util/ChartUtils.ts
```diff
@@ -1251,8 +1254,9 @@ export const parseSpecifiedDomain = (specifiedDomain: any, dataDomain: any, allo
  * @param  {Boolean} isBar if items in axis are bars
  * @return {Number} Size
  */
+type CartesianAxisProps = XAxisProps | YAxisProps | ZAxisProps;
```
ckifer, line 1257: the jsdoc belongs to the function not the type, please move the type above the jsdoc comment

ckifer (changes requested):
looks good besides one tiny thing

src/util/ChartUtils.ts
```diff
@@ -32,6 +32,11 @@ import {
   ChartOffset,
 } from './types';
 import { getLegendProps } from './getLegendProps';
+import { Props as XAxisProps } from '../cartesian/XAxis';
+import { Props as YAxisProps } from '../cartesian/YAxis';
+import { Props as ZAxisProps } from '../cartesian/ZAxis';
+
+type CartesianAxisProps = XAxisProps | YAxisProps | ZAxisProps;
```
ckifer, line 39: 😅 generally I'd put this below the re-export below but all good

ckifer (approved):
LGTM

This is an improvement but tbh numerical axis bar charts do not make much sense (unless the numbers are treated as categories). This is probably why this didn't work in the first place.

@nikolasrieble @PavelVanecek 
Any thoughts about this? If this produces an ugly chart (it does, see uncentered ticks) wouldn't it be better to throw an invariant when someone attempts to use two numerical axes for a bar chart? I see no reason not to ship this to improve current behavior, but maybe this is something we discourage in the next MV

ckifer:
@souzolliveira thanks, going to wait for some other reviews

ckifer:
also build is failing

src/util/ChartUtils.ts
souzolliveira: At the line 43?

souzolliveira:
@ckifer I could increase the x-axis width to centralize the unique bar in the axis if you tell me the path to do this.
Did my changes break the tests that failed in the build?

nikolasrieble:
> @ckifer I could increase the x-axis width to centralize the unique bar in the axis if you tell me the path to do this. Did my changes break the tests that failed in the build?

Yes, this specifically 

```
 <BarChart /> › Render a bar › renders a bar if size is limited
```

PavelVanecek (approved):
In my opinion, as a library we should allow use-cases that don't make sense to us. Maybe there's a valid chart that needs numerical axis and bars together, who knows.

ckifer:
Agreed in principle, but the flexibility does add a lot of pressure on the side of the maintainers. 

We support this but is it expected to look a certain way or be functional? I guess it might be useful depending on the case and customizations applied. I wish the original creators had limited flexibility a bit (even though it's one of the things people like about the lib now) 😅

Doesn't matter for now though, just trying to think aloud a bit. Thanks for the input!

ckifer:
@souzolliveira looks like test is still failing

TheCleric:
To write in favor of having bar charts with numerical X and Y axes, here are some ones I am using for a project:

https://bferentz.vercel.app/

In a lot of charts on that page, I am comparing a numeric value (potential points scored in a game) against another numeric value (the % chance of those points being score).

ckifer:
@TheCleric nice, thanks for the data point (and cool site!)

At small scale things look relatively non-functional but these types of examples help challenge (at least my) perception of what people are able to do with the lib. Kudos for creativity and we'll make sure number support doesn't go away

ckifer:
@souzolliveira any chance you could merge the latest and fix this test?

souzolliveira:
> @souzolliveira any chance you could merge the latest and fix this test?

Yes, sorry for the delay, I will update my branch and correct the tests.

ckifer:
No worries at all

test/chart/BarChart.spec.tsx
```diff
@@ -19,7 +19,7 @@ describe('<BarChart />', () => {
   ];
 
   type CustomLabelProps = Partial<{ x: number; y: number; index: number }>;
-
+
```
nikolasrieble, line 22:
```suggestion
```

src/util/ChartUtils.ts
```diff
@@ -1283,18 +1288,26 @@ export const getBandSizeOfAxis = (
...
+      let bandSize = Infinity;
+      bandSize = axis.width / 3;
+      return bandSize === Infinity ? 0 : bandSize;
+    }
 
-    for (let i = 1, len = orderedTicks.length; i < len; i++) {
-      const cur = orderedTicks[i];
-      const prev = orderedTicks[i - 1];
+    if (ticks.length >= 2) {
+      const orderedTicks = _.sortBy(ticks, o => o.coordinate);
```
nikolasrieble, line 1299:
```suggestion
      const orderedTicks = sortBy(ticks, o => o.coordinate);
```

nikolasrieble:
Trying to land this, I merged master via UI. Testing this locally, tests are failing 🤔 
Strangely the CI was not triggered by actions in the UI.

nikolasrieble (changes requested):
Lets make sure to have all tests pass before this lands.

ckifer:
Going to close this as stale for now, potentially @graup will take it over

PR #3865, opened 2023-10-15
Lodash: Only use where needed - isNil

nikolasrieble:
## Description

https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L11973

No need to import lodash for checking whether something `== null`
The only reason could be readability, so I instead added a helper function isNil. 

## Related Issue
https://github.com/recharts/recharts/issues/3752

## Types of changes

## Checklist:

### DIFF
--- a/src/util/isNil.ts
+++ b/src/util/isNil.ts
@@ -1,0 +1,3 @@
+export function isNil(value: unknown) {
+  return value == null;
+}
--- a/src/util/isNil.ts
+++ b/src/util/isNil.ts
@@ -1,3 +1,3 @@
-export function isNil(value: unknown) {
+export function isNil(value: unknown): value is null | undefined {
   return value == null;
 }
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -5,6 +5,7 @@ import React, { PureComponent, ReactElement, SVGProps } from 'react';
 import classNames from 'classnames';
 import Animate from 'react-smooth';
 import _ from 'lodash';
+import { isNil } from '../util/isNil';
 import { Curve, CurveType, Point as CurvePoint } from '../shape/Curve';
 import { Dot, Props as DotProps } from '../shape/Dot';
 import { Layer } from '../container/Layer';
@@ -207,7 +208,7 @@ export class Area extends PureComponent<Props, State> {
         }
       }
 
-      const isBreakPoint = _.isNil(value[1]) || (hasStack && _.isNil(originalValue));
+      const isBreakPoint = isNil(value[1]) || (hasStack && isNil(originalValue));
       if (layout === 'horizontal') {
         return {
           x: getCateCoordinateOfLine({ axis: xAxis, ticks: xAxisTicks, bandSize, entry, index }),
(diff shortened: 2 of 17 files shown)


src/util/isNil.ts
```diff
@@ -0,0 +1,3 @@
+export function isNil(value: unknown) {
+  return value == null;
+}
```
ckifer, line 3: so TypeScript can be weird with this. Does this know that `== null` can be undefined or null? Maybe a better way to check this would be to see if it type narrows for something that can be both undefined and null

ckifer (commented):
Changes all look good, one question

src/util/isNil.ts
nikolasrieble: We could just make it a type guard as well.

src/util/isNil.ts
```diff
@@ -0,0 +1,3 @@
+export function isNil(value: unknown) {
```
nikolasrieble, line 1:
```suggestion
export function isNil(value: unknown): value is null | undefined {
```

src/util/isNil.ts
ckifer: Cool, that works for me

PavelVanecek: If it was me I would add unit and type tests too - first to show that indeed yes yes double equal does check for undefined. There will always be some people who are not sure about that bit. And second test to show that this function allows use inside an `Array.prototype.filter` call; it will need the typeguard to work that way.

PavelVanecek (approved):

src/util/isNil.ts
nikolasrieble: Thank you Pavel, I will do that. That is exactly why I asked you for PR review as well. ❤️ Appreciate it.

ckifer: Sometimes when a function that validates a type (this helper) is within the scope of another function that does something (filter) TypeScript isn't able type narrow correctly.

Here is this function as an example https://www.typescriptlang.org/play?ts=4.7.4#code/FAjGHsDsGcBcAICW0ByiA28C88AUA3AQ3QFcBTALnhMgGtJwB3SASmwD54BvYeeAJzKwS-SPCKky2HJBLp0AbmABfMFDjxC-flVxx+iSAHN4AH2qQAJmQBmhspbPxZ8lgG0AutnhuA5L4AaZzl0DyVgAHoI+Dt0WDJBR1BCMTgMTAAjKRdMcH4LaztIB3gyfDIxWAALcBIjKvhGKVAqslBaGLykWGAIGARY+MTvLX4AOkGEgmIObl4BIRExAEJkNHRp9BYlZRYgA

ckifer:
@nikolasrieble looks like tiny tiny formatting issue breaking the build

also the type narrowing doesn't work in the case of filtering (or anything else) because of TS

src/util/isNil.ts
PavelVanecek: So from a helper like this I have two expectations:

1. the runtime behaviour - you have that covered well in your tests @nikolasrieble 
2. and the compile time behaviour - that's missing from tests. @ckifer 's example is one of them, I would add another:

https://www.typescriptlang.org/play?ts=5.2.2#code/MYewdgzgLgBAlhAcnANjAvDAFANwIYoCuApgFwyFgDWYIA7mAJTn5HHwQxiEpoA+FMABNiAMzhhiQjAD4YAbwBQMGACdiUQqrAxWJDJm68A3IoC+ixaMrAoccDFFgszGNFUSA5jAFG08tQ0tHQByEJgLRVBIWFZydy8fLh40TCcXSwB6TJgoAAsENzyQHmkAI3Z81XpEvB1iVWrVKPBoeDBWOCF4qA8wb0xWU0U4UWwAQgRkFFwCRkYFZRhs3ILOCGLSrhBYKvol6Lb1cUlut17EwYJTCyA

PavelVanecek: To fix the array filter example we would need another helper: https://www.typescriptlang.org/play?ts=5.2.2#code/GYVwdgxgLglg9mABDAzgEQKbBmDATRAHgBUA+ACgDcBDAGxAwC5FjEAfRcPLHfdxMCFq0AlMxr0MyFC0QBvAFCJEAJwxQQKpBIaIAhAF4BQ2gG4FAXwUKICFFETUVK5uXsqcAc35ceuAhyCwiIA2gC6iEYhAOTRADTGwmHmCgD0qYgA7lIQ1Ehq2LiOSE4q1ACeAnCZNnYO2LRQGGp4zO5e4ZGOzgB0DU0q5KiYhfgiCkA

nikolasrieble:
Instead of continuing here, I prefer to land https://github.com/recharts/recharts/pull/3902

PR #3868, opened 2023-10-16
Remove unit test that was comparing new/old reduce-css-calc

PavelVanecek:
Why? It's not testing anything - it throws, and the error is not reported as a failing test but instead as a log message in the console.

Instead I have replaced it with a `.not.toThrow()` call which at least tests that our new implementation doesn't throw. The old one throws so I removed it completely.

## Related Issue

https://github.com/recharts/recharts/pull/3820

## Motivation and Context

image

## How Has This Been Tested?

Jest

## Types of changes

## Checklist:

### DIFF
--- a/test/util/ReduceCssCalcPrototype.spec.ts
+++ b/test/util/ReduceCssCalcPrototype.spec.ts
@@ -1,5 +1,3 @@
-import oldReduceCSSCalc from 'reduce-css-calc';
-import { log } from 'console';
 import { reduceCSSCalc, safeEvaluateExpression } from '../../src/util/ReduceCSSCalc';
 
 describe('number calculate', () => {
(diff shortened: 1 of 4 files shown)


test/util/ReduceCssCalcPrototype.spec.ts
```diff
@@ -101,16 +99,7 @@ describe('reduce-css-calc', () => {
...
           `calc(${capHeight} / -${lineHeight})`,
         ].forEach(calcString => {
-          try {
-            const newCalc = reduceCSSCalc(calcString);
-            const prevCalc = oldReduceCSSCalc(calcString);
-
-            if (!(prevCalc.includes('calc') || prevCalc.includes(' '))) {
-              expect(prevCalc).toEqual(newCalc);
-            }
-          } catch (e) {
```
PavelVanecek, line 111: Test failures from `expect` are reported as an exception - so this `catch` turns this test into an always green one. Don't catch exceptions in tests.

ckifer: Some of the tests were added to explicitly show that the old reduce css calc threw errors in those cases, but agreed should be maybe been a separate block with expect to throw

Issue #846, opened 2017-08-23
[Bug] No animation when changing dataKey if data is not recreated


howardya:
## Do you want to request a feature or report a bug?
Bug

## What is the current behavior?
Data provided to the charts will only be animated if it is recreated on every render. 

For instance, this is perfect http://jsfiddle.net/qbmg1567/581/ 

However, if the data is constant and not recreated on every render, then the animation will not work. For instance, http://jsfiddle.net/qbmg1567/582/ this one will not animate the changes when we press the button. 

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/qbmg1567/).
http://jsfiddle.net/qbmg1567/581/ 
http://jsfiddle.net/qbmg1567/582/

## What is the expected behavior?
Expect both http://jsfiddle.net/qbmg1567/581/ and http://jsfiddle.net/qbmg1567/582/ to have animation

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
NA

951759534:
I also met the same problem

Rahulkr204:
@howardya did you find any solution ?

tomtobac:
Facing the same issue too

brakon:
What I found is the animation is not triggered if your component rerenders several times in a row. This can happen if the render method is being called even if the data you are passing doesnt change.

To fix this, I added a shouldComponentUpdate to the component contaning the chart, and had it only render when the data did indeed change.

acontreras89:
I'm not sure why this is closed, I'm still facing this issue as of today.

AnthonyCrowcroft:
also still experiencing this as well. Were there any good work arounds.

arabold:
For everybody coming here: I found a workaround for this issue. The hint is actually in the title of this bug report: Simply recreate the data when the `dataKey` changes. For example, if `data` is your chart data, use something like this:

```ts
  const chartData = useMemo(() => data.slice(), [dataKey, data])
  return (
    <BarChart data={chartData}> ... </BarChart>
  )
```

Performance should be acceptable for most scenarios but it would be great if this could be fixed in the library itself instead of requiring the above workaround.

muba00:
I had the same problem, I solved it by adding **<BarChart key={Math.random()} .....**

max-crcl:
[...]

PR #3870, opened 2023-10-16
Add `dataKey` prop on `CategoricalChartProps`

charpeni:
## Description

Introducing a `dataKey` prop on `CategoricalChartProps`, so we could rely on it to fire whether the chart should animate between two states. `dataKey` is an explicit prop to controller whether the `dataKey` changes or not and if we should animate to accommodate for the possibility that `data` doesn't change, but we toggle between `dataKey`.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/846.

## Motivation and Context

As described in #846—an issue reported six years ago—where changing the `dataKey` on the chart's elements doesn't trigger an animation between the two states. Instead, it just happens right away.

The reason why this is happening is behind the [`getDerivedStateFromProps`](https://react.dev/reference/react/Component#static-getderivedstatefromprops) lifecycle function:

https://github.com/recharts/recharts/blob/9b6046b141be493f64291909fe8439c5fae715de/src/chart/generateCategoricalChart.tsx#L1207-L1214

It means that the animation between two states will only happen if at least one of the listed props here changes: `data`, `width`, `height`, `layout`, `stackOffset`, `margin`, and `children`. One way to assert this is that changing `stackOffset`, which uses a primitive type, simultaneously changing the `dataKey` used on the chart's elements will trigger the animation between two states as expected. Unfortunately, that behavior isn't explicit at all.

### Current behavior as of `2.9.0`

Behavior: It just happens.

🔗  CodeSandbox: https://codesandbox.io/s/recharts-current-3sskmy

https://github.com/recharts/recharts/assets/7189823/1ea667da-2d0f-4893-9cf2-da6bec41ac88

### Using `dataKey` as `key` prop

React supports a `key` prop to explicitly tell React to re-render a component based on the provided `key`. We can leverage this by passing `dataKey`. Unfortunately, it isn't working as expected (or it does, actually). It will have the same effect if we also define a `key` prop on the chart's elements (e.g., `<Line>`).

Behavior: Re-renders the full chart like its first render.

🔗 CodeSandbox: https://codesandbox.io/s/recharts-datakey-as-key-mr4dk5

https://github.com/recharts/recharts/assets/7189823/678b7255-92f1-4541-aacd-b4433f9a65c8

### Proposed solution: supports `dataKey` on the chart

Behavior: Animates between two states

🔗 CodeSandbox: https://codesandbox.io/s/recharts-fix-proposal-hkyzn9

https://github.com/recharts/recharts/assets/7189823/a1b13600-e437-4911-95c1-2da098b07cf6

Provided in the previous section.

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -72,6 +72,7 @@ import {
   adaptEventHandlers,
   GeometrySector,
   AxisType,
+  DataKey,
 } from '../util/types';
 import { AccessibilityManager } from './AccessibilityManager';
 import { isDomainSpecifiedByUser } from '../util/isDomainSpecifiedByUser';
@@ -821,6 +822,7 @@ export interface CategoricalChartState {
 
   legendBBox?: DOMRect | null;
 
+  prevDataKey?: DataKey<any>;
   prevData?: any[];
   prevWidth?: number;
   prevHeight?: number;
@@ -840,6 +842,7 @@ export interface CategoricalChartProps {
   compact?: boolean;
   width?: number;
   height?: number;
+  dataKey?: DataKey<any>;
   data?: any[];
   layout?: LayoutType;
   stackOffset?: StackOffsetType;
@@ -1180,7 +1183,7 @@ export const generateCategoricalChart = ({
       nextProps: CategoricalChartProps,
       prevState: CategoricalChartState,
     ): CategoricalChartState => {
-      const { data, children, width, height, layout, stackOffset, margin } = nextProps;
+      const { dataKey, data, children, width, height, layout, stackOffset, margin } = nextProps;
 
       if (_.isNil(prevState.updateId)) {
         const defaultState = createDefaultState(nextProps);
@@ -1197,6 +1200,7 @@ export const generateCategoricalChart = ({
             prevState,
           ),
 
+          prevDataKey: dataKey,
           prevData: data,
           prevWidth: width,
           prevHeight: height,
@@ -1207,6 +1211,7 @@ export const generateCategoricalChart = ({
         };
       }
       if (
+        dataKey !== prevState.prevDataKey ||
         data !== prevState.prevData ||
         width !== prevState.prevWidth ||
         height !== prevState.prevHeight ||
@@ -1249,6 +1254,7 @@ export const generateCategoricalChart = ({
             },
             prevState,
           ),
+          prevDataKey: dataKey,
           prevData: data,
           prevWidth: width,
           prevHeight: height,


src/chart/generateCategoricalChart.tsx
```diff
@@ -821,6 +822,7 @@ export interface CategoricalChartState {
 
   legendBBox?: DOMRect | null;
 
+  prevDataKey?: DataKey<any>;
```
nikolasrieble, line 825: Do we need to use any? 🤔

charpeni: I had the same though, I really dislike using `any`, and when I looked for `DataKey` usages, they were all using it like this. 😅

ckifer: yeah they're all like this right now unfortunately

That being said this is a great fix, we can address the DataKey issue with some of our refactorings that have been happening across the lib

ckifer (approved):
🚀

charpeni:
I wanted to make sure checks were all green (they are) and add the proper documentation. But the documentation lives within another directory, so turning it ready for review.

ckifer (commented):
I guess another thing to consider is what this does to UX - this allows the issue to be fixed but now I have to specify dataKey in two places. Maybe we allow the user to specify dataKey here and it just works in all of the children too if they don't specify there?

But.. this is backwards compatible so its fine. A tiny duplication on the consumer side for a nice improvement

charpeni:
> but now I have to specify dataKey in two places.

It is worth mentioning that the top-level `dataKey` is totally optional and is only used as a way to "tag" that the `dataKey` changed, and we should be running the transition animation between datasets. It will have the same effect as `data={[...data]}`, but has the convenience of being more explicit about the change: changing `dataKey` will trigger the animation.

The proposed solution was to memoize the data based on `data` and `dataKey`, but the memorization had nothing to do with `dataKey` and therefore, the rules of hooks should be screaming at you through ESLint. That's what I meant by introducing `dataKey` as an explicit solution and preventing it from being removed by inexperienced readers.

For use cases where you don't change the `dataKey` on inner elements, the top-level one isn't required and should be avoided.

ckifer:
Cool!

You can add the docs changes to the storybook as we're using that as the new SOT for the docs, that is within this repo.

charpeni:
I just pushed two commits to add documentation for `dataKey`, but also to add examples related to toggling between `dataKey`s:

Screenshot 2023-10-17 at 8 49 59 AM

ckifer (approved):
Thank you!

BjornGjorv:
@ckifer were this fix included in the 2.9.1? I still have issues, but it may be caused by something else. 🤷‍♂️

charpeni:
Not included in `2.9.1`: https://github.com/recharts/recharts/compare/v2.9.0...v2.9.1.

ckifer:
Sorry it was not - I was only including severe issues from 2.9 in 2.9.1. 
And since this is technically a new addition (even though it allows a workaround for a bug) it doesn't quite fit the semver definition of a patch release and I'm trying to be more strict about that. 

Will be out soon enough though!

BjornGjorv:
> Sorry it was not - I was only including severe issues from 2.9 in 2.9.1. And since this is technically a new addition (even though it allows a workaround for a bug) it doesn't quite fit the semver definition of a patch release and I'm trying to be more strict about that.
> 
> Will be out soon enough though!

Allright, looking forward to it!

PR #3880, opened 2023-10-18
Lodash: Only use where needed - isFunction

nikolasrieble:
## Description
No need to use lodash for isFunction. 

## Types of changes

## Checklist:


src/util/isFunction.ts
```diff
@@ -0,0 +1,3 @@
+export function isFunction(value: unknown): value is Function {
+  return typeof value === 'function';
```
PavelVanecek, line 2: So I looked into this recently and discovered that lodash is doing something way more complicated than`typeof`: https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L3063

Are we okay with replacing it with `typeof` check? I saw some Safari bugs mentioned in there saying that `typeof () => {}` returns `Object`. But I didn't go in the detail.

HHongSeungWoo: https://github.com/HHongSeungWoo/recharts/blob/0c1aa93cb4ea2de06a31b8efce6a1a27f446af7d/src/util/DataUtils.ts#L19 
How about doing it this way?

ckifer:
@nikolasrieble what would you  like to do with this and the other PR?

src/util/isFunction.ts
nikolasrieble: As https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L3063 indeed seems to handle more edge cases, I believe it is safest to just still use lodash.isfunction

Issue #3861, opened 2023-10-15
ResponsiveContainer: Refactor and simplify


nikolasrieble:
Coming from https://github.com/recharts/recharts/issues/172
Our usage is bloated and does not align with the recommended usage.

- [ ] Upgrade react-resize-detector
- [ ] Refactor to align usage with best practice as documented in the readme (i.e. use the https://www.npmjs.com/package/react-resize-detector#1-react-hook)

This issues is part investigation, since it might be that we can not refactor to best practice for backwards compatibility / edge cases. 

Goals further are: 
- Add / extend tests 
- Extract helper functions
- Reduce useCallback, useEffect usage to required minimum https://react.dev/learn/you-might-not-need-an-effect

nikolasrieble:
@HHongSeungWoo Would this also interest you?

nikolasrieble:
Great work, @HHongSeungWoo! Closing this.

PR #3881, opened 2023-10-19
Update Typescript version

HHongSeungWoo:
## Description

I am currently investigating issue #3861 and experimenting with a simple implementation using ResizeObserver. 
The current version of TypeScript being used does not support ResizeObserver.

## Related Issue

#3861

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "terser-webpack-plugin": "^5.1.1",
     "ts-jest": "^27.1.5",
     "ts-loader": "^8.0.12",
-    "typescript": "^4.4.4",
+    "typescript": "4.4.4",
     "update-browserslist-db": "^1.0.10",
     "webpack": "^5.65.0",
     "webpack-bundle-analyzer": "^4.4.0",
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "terser-webpack-plugin": "^5.1.1",
     "ts-jest": "^27.1.5",
     "ts-loader": "^8.0.12",
-    "typescript": "4.1.3",
+    "typescript": "^4.9.5",
     "update-browserslist-db": "^1.0.10",
     "webpack": "^5.65.0",
     "webpack-bundle-analyzer": "^4.4.0",
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "terser-webpack-plugin": "^5.1.1",
     "ts-jest": "^27.1.5",
     "ts-loader": "^8.0.12",
-    "typescript": "^4.9.5",
+    "typescript": "^4.4.4",
     "update-browserslist-db": "^1.0.10",
     "webpack": "^5.65.0",
     "webpack-bundle-analyzer": "^4.4.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -152,7 +152,7 @@
     "terser-webpack-plugin": "^5.1.1",
     "ts-jest": "^27.1.5",
     "ts-loader": "^8.0.12",
-    "typescript": "4.1.3",
+    "typescript": "^4.9.5",
```
PavelVanecek, line 155: Why not 5+?

ckifer: That would be a breaking change no?, we can't upgrade without a major version release

HHongSeungWoo: I actually don't mind version 5 and above, it's just that build test was bothersome :cry:

ckifer: TS 5 has too many TS 5 things such as the satisfies operator which would break any consumer who isn't on 5 (a lot of people)

ckifer: Actually that is 4.9...

Not sure, but 95% sure if we ever used that and our consumers were on a lesser version they would break

ckifer: I would actually advocate that we leave the TS version alone else we break backwards compatibility. We either wait until v3 or we do something like proposed in https://github.com/microsoft/TypeScript/issues/22605 (which is a lot of work but possible)

HHongSeungWoo: I have reviewed all the type definitions for the 4.9.5 update, and as far as I can tell, there doesn't seem to be any significant issues that I missed.
Even if there were any, they would likely be minor and understandable for users trying to keep up with our version

PavelVanecek: Hold on, we ship the source code to npm? With typescript as a required dependency? That doesn't sound right. That would mean that projects using JS (no TS) are not able to use recharts at all? If that's true then we should be fixing that instead of bumping TS version

HHongSeungWoo: We can use it without using TS.

ckifer: We ship the type defs to typescript consumers which means if we use a feature incompatible with the consumers TS version they will not be able to build their code without upgrading. Something like satisfies still exists in the d.ts files unless I'm misunderstanding how TS works

ckifer: https://stackoverflow.com/a/51670675

I agree with the answer in the SOF, but we can't really guard well against some PR in recharts using a new TS feature in the next minor release that will break consumers because type signatures changed. Satisfies is a good example (or something else becoming more strict). It only matters in the public API only sure but unfortunately current state of things has no separation

HHongSeungWoo: How about adding a build test then?

HHongSeungWoo: It seems sufficient to test only the success of the 4.1 TypeScript build and remove it in the 3.0 release

HHongSeungWoo: Sure, that sounds good to me

package.json
```diff
@@ -152,7 +152,7 @@
     "terser-webpack-plugin": "^5.1.1",
     "ts-jest": "^27.1.5",
     "ts-loader": "^8.0.12",
-    "typescript": "4.1.3",
+    "typescript": "^4.4.4",
```
ckifer, line 155: version is still 4.9.5 in the lockfile, you'll need to remove the `^`, install, and then add it back to make sure the lockfile updates

ckifer (changes requested):

package.json
HHongSeungWoo: fixed it

ckifer (approved):

Issue #3358, opened 2023-02-11
AreaChart leaves white space with intermittent values and connectNulls prop looks awful


Ian-Bright:
I am having issues with the stacked area chart. There are values in the data that don't exist across the entire time range and they are causing holes to appear in the chart:

<img width="689" alt="Screen Shot 2023-02-11 at 11 47 34 AM" src="https://user-images.githubusercontent.com/66142236/218275853-e9d7b198-5a1f-4f04-80d2-31f7dbb52870.png">

<img width="700" alt="Screen Shot 2023-02-11 at 11 47 26 AM" src="https://user-images.githubusercontent.com/66142236/218275856-8829bbc1-b7a6-4b24-9635-a893d202a905.png">

I tried using the connectNulls prop but the rendering looks horrendous:

<img width="686" alt="Screen Shot 2023-02-11 at 11 51 55 AM" src="https://user-images.githubusercontent.com/66142236/218275892-f71fe479-7613-4973-82a5-ed8d9434e8e9.png">

<img width="689" alt="Screen Shot 2023-02-11 at 11 52 02 AM" src="https://user-images.githubusercontent.com/66142236/218275895-2b8df211-e8fa-4875-8120-0686422842e3.png">

Any clue how this can be corrected?

nikolasrieble:
The approach that first comes to mind is to preprocess the data and fill in the gaps with whatever logic you see fit before passing it to recharts. 

Independently, this seems to be a bug with the behaviour of the stacked AreaChart and connectNulls.

It would greatly help us, if you could provide a toy dataset (i.e. as few data points as possible) and some code to reproduce this issue.

ckifer:
+1 to the above, a codesandbox would be awesome

nikolasrieble:
Released a fix with https://github.com/recharts/recharts/releases/tag/v2.10

Issue #2176, opened 2020-06-22
Trailing Line on AreaChart if values aren't defined


CharlesMcMillan:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://jsfiddle.net/mutualslump/Lwvymzrx/4/](https://jsfiddle.net/mutualslump/Lwvymzrx/4/)

### Steps to reproduce
Fill in an AreaChart with array values for the datakey (lower/upper bounds)...leave some of these datakeys null/undefined in the data array

### What is expected?
A hard cutoff of the area chart, so that it doesn't try to draw a line from the null values to the non null values

### What is actually happening?
When you look at the linked repro (https://jsfiddle.net/mutualslump/Lwvymzrx/4/), you'll see that recharts is drawing a line from 'Page D' (which does not have any areacharts datakey to 'Page E' (which does have that datakey)

| Environment | Info |
|---|---|
| Recharts | v2.0.0-beta.6 |
| React | 15.6.2 |
| System | MacOS |
| Browser | Chrome 83.0 |

andriyloza:
I have the same issue with AreaChart and try to draw line instead of triangles in case null/undefined value

hbermani:
Facing the same issue here.  connectNulls={true} didn't resolve it for me. 

In the end I just used a hacky way to not display the stroke and just display the filled area, I provided stroke="00B1EB" value instead of "#00B1EB" so it doesn't draw it and only draw the area.

seiyab:
Here I reproduced in CodeSandbox.
https://codesandbox.io/s/serene-lichterman-g0s9t6?file=/src/Repro.tsx

ckifer:
Issue happened between these two releases https://github.com/recharts/recharts/compare/v2.0.0-beta.3...v2.0.0-beta.4

culprit looks to be https://github.com/recharts/recharts/compare/v2.0.0-beta.3...v2.0.0-beta.4#diff-1496ed5301e625b1e310643256ed6b2af380dd919af777f25d65beecac8f7b22R195-R206

When going to fix though things get confusing fast. What looks to be an easy fix (account for connectNulls) seems to be that `connectNulls` never really worked as expected and axis.scale(value[1]) (which returns a value) was giving the previous behavior. Will keep digging.

ckifer:
[...]

Issue #1601, opened 2018-12-28
How to disable Stacked Area Chart Draw 0 when data is undefined


JQuu:
I recently noticed when using StackedId in Area chart, it always draw 0 if there is no data for one x-axis, this is possible in a normal area chart [Non-Stacked Area Char works fine - JSFiddle](https://jsfiddle.net/6yoqhvfk/). 

However when using stackedArea chart it will always bring areas to 0.
[Stacked Area Chart with no data - JSFiddle](https://jsfiddle.net/n9xjcurh/1/)

Is there a way to make stacked chart behavior consistent with non-stacked cart?

dd1mk:
Hi, are there any updates on this feature?

PR #3882, opened 2023-10-19
Remove duplicate types In Area chart

HHongSeungWoo:
## Description

~~In the stack area chart, when `connectNulls` is set to true and there's a null value, I have adjusted it so that the graph now smoothly connects without any breaks~~

I removed duplicate types and did some code refactoring.

## Related Issue

~~#2176, #1601, maybe #3358~~

## Types of changes

## Checklist:


storybook/stories/Examples/AreaChart.stories.tsx
```diff
@@ -407,3 +407,81 @@ export const AreaChartFillByValue = {
...
+      { month: '2015.08', a: 3490, b: 4300, c: 2100 },
+      { month: '2015.09' },
+    ];
+
+    const toPercent = (decimal: number, fixed = 0) => {
+      return `${(decimal * 100).toFixed(fixed)}%`;
+    };
+
+    return (
+      <AreaChart width={600} height={400} data={data} margin={{ top: 10, right: 30, left: 0, bottom: 0 }}>
```
ckifer, line 478: the y axis on the story gets cut off so I think we need to increase left margin here

ckifer (commented):
This is a good fix overall, I know its been an issue for a while.

There is some interesting behavior when only a few areas in the stack have `connectNulls`

With a regular area chart you get a straight line connecting the values where there is null in between such as this story:
image

With a stacked chart you don't get that 
image
The value (yellow here), goes down to either 0 or the lowest point on the stack. 0 is fine if the null are on the start of end of the data and there is nothing to connect to but `connectNulls` is "Whether to connect a graph area across null points." which would draw more like this I think 
image

does `connectNulls` mean go to 0? @nikolasrieble thoughts?

nikolasrieble:
> does `connectNulls` mean go to 0? @nikolasrieble thoughts?

This is a great point, Coltin! The API should behave consistently across components. In this case I would expect to connect between the two points. It should only fall back 0 if not possible (i.e. first or last value)

HHongSeungWoo:
If the `connectNulls` feature is intended to remove noise from the chart, then i agree.

`null null(replace 0) 1 2 null(connect) 3 4 null(replace 0)` Is this the way want it to be processed?

HHongSeungWoo:
We think this needs to be fixed too, right?

image

ckifer:
That's how I would think it would work based off of my current assumptions. 

Maybe we can take a peek at what other charting libs look like, it's kind of hard to know what would be preferable/standard

HHongSeungWoo:
If that's the case, I will revert everything in this PR except for the code cleanup.
It seems that consensus is a priority in addressing this issue

ckifer:
Sounds good, lets revisit this.

One thing is reverting to the old behavior (which is better than current and "fixes" the bugs linked), another is changing the behavior entirely. Maybe we use the old behavior (+connect nulls) for now and later adjust based on what we think/know connectNulls should do

ckifer:
Take this example https://www.highcharts.com/demo/highcharts/area-stacked
and start replacing the data with nulls - without `connectNulls` I think they get behavior correct i.e.
image


With connect nulls they behave the same way that we would have if we merged this fix (go to 0 for all nullish values) 
image

In regular area they connect the line (as `connectNulls` would work for a `Line`) 
image

After looking at that I think how we had it was correct

Issue #3816, opened 2023-10-05
Having multiple charts on one page causes accessibility issues


bourke:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://hs4v2f.csb.app/](https://hs4v2f.csb.app/)

### Steps to reproduce
Load the page and test using an accessibility tester (e.g., https://chrome.google.com/webstore/detail/ibm-equal-access-accessib/lkcagbfjnkomcinoddgooolagloogehp/related?utm_source=ext_sidebar&hl=en-GB). 

### What is expected?
The page should pass accessibility tests for WCAG checkpoint 2.4.1 regarding accessible names for ARIA landmarks. 

### What is actually happening?
Because the `recharts-wrapper` `<div>` has a hardcoded `role="region"` without `aria-label`, when multiple charts exist on a page, the ARIA landmarks (`region` is considered a landmark) aren't identified and can't be navigated to quickly using screen readers. A similar phenomenon occurs with tooltip wrappers; they have their roles defined as `dialog` without any identifiers.

| Environment | Info |
|---|---|
| Recharts | v2.8.0 |
| React | 18.2.0 |
| System | macOS 13.6 |
| Browser | Chrome 117.0.5938.132 (arm64) |

---
I realise this issue could be considered a feature request rather than a bug report, but since the code does include ARIA roles without the necessary identifiers, I'm reporting it as a bug. I'll be happy to submit a PR if this isn't on your internal roadmap.

nikolasrieble:
@bourke Thank you for creating this issue. We would love if you could pick this up yourself.

julianna-langston:
It looks like some of the charts have `role="img"`, and some of them have `role="region"`. Would it be easier to replace the sankey plots with `role="img"`, or just to drop the role entirely, rather than trying to juggle aria-label's?

ckifer:
I think in other issues we've discussed that img isn't even the correct role to use on charts anymore right? We should be using the newer spec "graphic-*"?

julianna-langston:
What the "correct" role is open to debate, and many accessibility experts can have reasonable disagreements on it. 

Let me take you through my though process.

When I'm writing code for any reason to go on the web, I add `role` to elements for the sole purpose of communicating something to a screen reader user. Let's go through a couple of options, their pros/cons, and how they would be read by screen readers.

[...]

PR #3884, opened 2023-10-20
Add aria-label attribute (and associated props) for wrapper divs

bourke:
## Description

This PR adds the following functionality to Recharts:

**For charts**:
- Adds the `aria-label` attribute required by the ARIA role of `region` defined on the wrapper `<div>` for `generateCategoricalChart`, `Sankey`, and `TreeMap`;
- Allows developers to supply a value for the `aria-label` attribute using the `ariaLabel` prop;
- Generates a fallback value if `ariaLabel` is `undefined`.

**For tooltips**:
- The same as for charts, except the `aria-label` is required by the ARIA role of `dialog`.

## Related Issue

#3816 

## Motivation and Context

Accessibility rules dictate that, on any given page, multiple elements with the same ARIA role need to have unique identifiers. The `aria-label` attribute is the most straightforward way to supply these unique identifiers.

## How Has This Been Tested?

I’ve tested these changes against the existing Recharts test suite as well as our project’s accessibility test suite, which applies WCAG rules to a set of automated browser tests using Playwright.

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/Sankey.tsx
+++ b/src/chart/Sankey.tsx
@@ -10,6 +10,7 @@ import { Tooltip } from '../component/Tooltip';
 import { Rectangle, Props as RectangleProps } from '../shape/Rectangle';
 import { shallowEqual } from '../util/ShallowEqual';
 import { filterSvgElements, validateWidthHeight, findChildByType, filterProps } from '../util/ReactUtils';
+import { uniqueId } from '../util/DataUtils';
 import { getValueByDataKey } from '../util/ChartUtils';
 import { Margin, DataKey, SankeyLink, SankeyNode } from '../util/types';
 
@@ -385,6 +386,8 @@ interface SankeyProps {
   onMouseLeave?: any;
 
   sort?: boolean;
+
+  wrapperAriaLabel?: string;
 }
 
 type Props = SVGProps<SVGElement> & SankeyProps;
@@ -428,6 +431,13 @@ export class Sankey extends PureComponent<Props, State> {
     nodes: [] as SankeyNode[],
     links: [] as SankeyLink[],
   };
+
+  private wrapperAriaLabel: string;
+
+  constructor(props: Props) {
+    super(props);
+    this.wrapperAriaLabel = props.wrapperAriaLabel ?? uniqueId('Recharts Sankey Chart ');
+  }
 
   static getDerivedStateFromProps(nextProps: Props, prevState: State): State {
     const { data, width, height, margin, iterations, nodeWidth, nodePadding, sort } = nextProps;
@@ -705,6 +715,7 @@ export class Sankey extends PureComponent<Props, State> {
         className={classNames('recharts-wrapper', className)}
         style={{ ...style, position: 'relative', cursor: 'default', width, height }}
         role="region"
+        aria-label={this.wrapperAriaLabel}
       >
         <Surface {...attrs} width={width} height={height}>
           {filterSvgElements(children)}
--- a/src/chart/Sankey.tsx
+++ b/src/chart/Sankey.tsx
@@ -432,7 +432,7 @@ export class Sankey extends PureComponent<Props, State> {
     links: [] as SankeyLink[],
   };
 
-  private wrapperAriaLabel: string;
+  wrapperAriaLabel: string;
 
   constructor(props: Props) {
     super(props);
--- a/storybook/stories/API/chart/Sankey.stories.tsx
+++ b/storybook/stories/API/chart/Sankey.stories.tsx
@@ -18,6 +18,7 @@ export default {
     margin,
     data,
     sort: { description: 'Whether to sort the data or not' },
+    wrapperAriaLabel: { description: 'Optional aria-label applied to the wrapper <div> element of the chart' },
   },
   component: Sankey,
 };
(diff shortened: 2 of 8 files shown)


src/chart/Sankey.tsx
```diff
@@ -429,6 +432,13 @@ export class Sankey extends PureComponent<Props, State> {
     links: [] as SankeyLink[],
   };
 
+  wrapperAriaLabel: string;
+
+  constructor(props: Props) {
+    super(props);
+    this.wrapperAriaLabel = props.wrapperAriaLabel ?? uniqueId('Recharts Sankey Chart ');
```
ckifer, line 439: why do we need a constructor here (and all others in PR with same approach)? Its probably more simple and just as efficient to reference props directly

Why is there a space in all of the uniqueId calls? I also don't think these need to be unique?

storybook/stories/API/chart/Sankey.stories.tsx
```diff
@@ -18,6 +18,7 @@ export default {
     margin,
     data,
     sort: { description: 'Whether to sort the data or not' },
+    wrapperAriaLabel: { description: 'Optional aria-label applied to the wrapper <div> element of the chart' },
```
ckifer, line 21: why not just just use native `aria-label` and pass it down instead of adding a new prop?

ckifer (changes requested):
Some questions. I also think we can do this in less code - we don't need any of the added constructors

src/chart/Sankey.tsx
bourke: They do need to be unique; that's the motivation for the PR. 

> Region
>
> A [region](https://w3c.github.io/aria/#region) landmark is a perceivable section of the page containing content > that is sufficiently important for users to be able to navigate to the section.
> 
> - A region landmark must have a label.
> - If a page includes more than one region landmark, each should have a unique label (see [Step 3](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/#aria_lh_step3) above).
> - The region landmark can be used identify content that named landmarks do not appropriately describe.

(from [https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/#aria_lh_region](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/#aria_lh_region))

The space is there so that screen readers can pronounce the name.

storybook/stories/API/chart/Sankey.stories.tsx
bourke: That's a fair point. I admit it's a matter of opinion, but in this case I think having a name that explicitly tells the user what's going to happen with the prop is better. 

What's more, using the native HTML attributes could lead to confusion in another way. If I didn't know better, I'd assume that any "extra" attributes applied to `<Sankey>` would end up on the `<svg>` element, not the wrapper.

storybook/stories/API/props/ChartProps.ts
```diff
@@ -340,4 +340,16 @@ Also see https://d3js.org/d3-shape/stack#stack-offsets
       category: 'General',
     },
   },
+  wrapperLabel: {
+    description: `An optional value to be used as the aria-label for the wrapper element. When no value is provided
+      the aria-label will be set to the value of the \`title\` prop; if present; otherwise
```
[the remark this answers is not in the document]
bourke: Good point; I'll fix this.

src/chart/Sankey.tsx
nikolasrieble: I would like to understand this better - how does the space help screen readers? 
Without a comment to explain this in the code, I am afraid the next developer might remove the space.

src/chart/Sankey.tsx
```diff
@@ -705,6 +715,7 @@ export class Sankey extends PureComponent<Props, State> {
         className={classNames('recharts-wrapper', className)}
         style={{ ...style, position: 'relative', cursor: 'default', width, height }}
         role="region"
+        aria-label={this.wrapperAriaLabel}
       >
         <Surface {...attrs} width={width} height={height}>
```
nikolasrieble, line 720: So, we can pass props directly to the Surface, such as ariaLabel but we can now also pass an ariaLabel to the wrapping div? 

How would those interact? Would they be redundant to another? Should users prefer one way over the other?

src/chart/Sankey.tsx
nikolasrieble: Should we possibly extract a helper function with an intention revealing name, such as:

```
// Some useful reference
function addTrailingSpaceForScreenReaders(label: string){
   return label+' '
}
```

ckifer: +1, that space will get removed in a heartbeat by someone who thinks it was an accident. Would be better if we can wrap it somehow

src/chart/Sankey.tsx
ckifer: +1

storybook/stories/API/chart/Sankey.stories.tsx
ckifer: Hmm yeah I guess makes sense. We really, before making changes like these need to figure out broader accessibility so there isn't opportunity for this confusion in the first place. Imo this should just be normal `ariaLabel` and the user should be able to assume with confidence that it goes to the correct place regardless.

ckifer:
Going to close this as stale for now, please feel free to update and we can reopen

Issue #3832, opened 2023-10-09
Support for custom-shaped Funnel components


codebreaker-v2-0:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
I'm currently working on a project where the design of the Funnel Chart is in the shape of a perfect funnel which has varying bar sizes proportionate to the number of customers at that given stage. At present, the Funnel component doesn't support custom shapes for Funnel chart. And also the customisation is also very limited. I'm just curious to know why the customisation is kept very limited in the case of Funnel Chart as compared to Bar/Line/Pie charts

### What does the proposed API look like?
Providing additional prop like **shape={//customShape}** 
```
<FunnelChart >
    <Funnel data={data} dataKey="value" shape={//customShape} >
        <LabelList dataKey="name" fill="#000" />
    </Funnel>
    <Tooltip />
</FunnelChart>
```

ckifer:
Probably just because funnel was added later than the other components and shape was never added as a feature. 

If one doesn't already exist we can use this as a feature request

nikolasrieble:
@andrewangelle As you have already solved a big of challenges around (active)Shapes, and API alignment - would you consider picking this up as well?

andrewangelle:
@nikolasrieble of course! assign me.

nikolasrieble:
Released with https://github.com/recharts/recharts/releases/tag/v2.10

PR #3886, opened 2023-10-20
feat: Funnel support custom shape prop

andrewangelle:
## Description

Add support for shape prop in Funnel for user to provide a customized shape.

## Related Issue

#3832 
## Motivation and Context

Unified api for shape/activeShape support between Funnel/Line/Bar/RadialBar/Scatter

## How Has This Been Tested?
unit tests and storybook

## Types of changes

## Checklist:


src/util/ActiveShapeUtils.tsx
```diff
@@ -27,7 +27,7 @@ export type ShapeProps<OptionType, ExtraProps, ShapePropsType> = {
   option: OptionType;
   isActive: boolean;
   activeClassName?: string;
-  propTransformer?: (option: OptionType, props: ExtraProps) => ShapePropsType;
+  propTransformer?: (option: OptionType, props: unknown) => ShapePropsType;
```
ckifer, line 30: what was `ExtraProps` and why is `unknown` better?

src/util/ActiveShapeUtils.tsx
```diff
@@ -74,15 +82,17 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
   ...props
 }: ShapeProps<OptionType, ExtraProps, ShapePropsType>) {
   let shape: React.JSX.Element;
+  // eslint-disable-next-line one-var
+  let nextProps: ShapePropsType;
 
   if (isValidElement(option)) {
-    shape = cloneElement(option, props);
+    nextProps = propTransformer(option.props as OptionType, props);
```
ckifer, line 89: any way we don't have to cast here?

ckifer (changes requested):
Would like to see if we can remove that cast

src/util/ActiveShapeUtils.tsx
```diff
@@ -74,15 +82,17 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
...
+    shape = cloneElement(option, nextProps);
   } else if (_.isFunction(option)) {
     shape = option(props);
   } else if (_.isPlainObject(option) && !_.isBoolean(option)) {
-    const shapeProps = props as unknown as ExtraProps;
```
andrewangelle, line 83: @ckifer 
> what was `ExtraProps` and why is `unknown` better?

Allows to remove a cast here.

src/util/ActiveShapeUtils.tsx
andrewangelle: looks like there is. 😃

src/util/ActiveShapeUtils.tsx
```diff
@@ -74,15 +82,17 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
...
 }: ShapeProps<OptionType, ExtraProps, ShapePropsType>) {
   let shape: React.JSX.Element;
+  // eslint-disable-next-line one-var
+  let nextProps: ShapePropsType;
 
-  if (isValidElement(option)) {
-    shape = cloneElement(option, props);
+  if (isValidElement<OptionType>(option)) {
+    nextProps = propTransformer(option.props, props);
+    shape = cloneElement<OptionType>(option, nextProps);
```
ckifer, line 90: even though we remove `as` this is still casting.

Its probably fine since we're doing all of the checks we do here but still

storybook/stories/API/chart/FunnelChart.stories.tsx
```diff
@@ -17,6 +17,16 @@ export default {
...
+    shape: {
+      description: 'The customized shape to be rendered.',
+      table: {
+        type: {
+          summary: 'Function | boolean | ReactElement | object',
+        },
+        defaultValue: undefined,
+        category: 'General',
+      },
+    },
```
nikolasrieble, line 29: I believe this is documented in multiple places. Should we extract a shared description, and import it? 

That would enforce standardised user facing communication for this property.

test/chart/FunnelChart.spec.tsx
```diff
@@ -229,4 +229,55 @@ describe('<FunnelChart />', () => {
     const activeShape = container.querySelectorAll('.recharts-active-shape');
     expect(activeShape).toHaveLength(0);
   });
+
+  test('renders customized shape when shape is set to be a function', () => {
```
nikolasrieble, line 233: Just a curious idea: 

How about instead of duplicating these tests per component, why not create a single test file, for shape, and test all supported components at once with a parametrised test?

nikolasrieble: That would further enforce standardised behaviour and reduce copy pasted code /test cases

andrewangelle: I did this for the Funnel cases, can you take a look and confirm I'm on the right path. If so, I'll convert the others.

src/util/ActiveShapeUtils.tsx
andrewangelle: Made another update check again, please.

ckifer (approved):
LGTM!

nikolasrieble (approved):
Thank you for cleaning up the tests! Appreciate it!

PR #3890, opened 2023-10-21
Testing: Migrate from jest to vitest

nikolasrieble:
## Description
In https://github.com/recharts/recharts/pull/3705 we verified the vitest would suffice our requirements, and yes, it does. 

As we need to migrate to vitest for 3.0 release, but do not need the 3.0 release to migrate to vitest, lets do it now. 

Thank you @branberry @akamfoad @PavelVanecek @ckifer for your support! 

On an unrelated note: 1.0 release is upcoming for vitest https://github.com/vitest-dev/vitest/releases

TODO:
- [ ] create follow up tickets as defined in https://github.com/recharts/recharts/pull/3705

## Types of changes

## Checklist:


vitest.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from 'vitest/config';
+import react from '@vitejs/plugin-react';
+
+// https://vitejs.dev/config/
+// eslint-disable-next-line import/no-default-export
+export default defineConfig({
+  plugins: [react()],
+  test: {
+    environment: 'jsdom',
+    globals: true,
```
nikolasrieble, line 10: Does this really work? If yes, why do we need to import `vi`?

ckifer: Works in every other project I've used vite in, doesn't work in our current config for some reason though.

akamfoad: I think so, tests are passing even though some of the tests didn't have vi imported explicitly.

Personally I prefer to import things directly and leave globals away as much as possible.

But this has implications as mentioned here:

https://vitest.dev/guide/migration.html#globals-as-a-default

nikolasrieble: Oh I am certain it does work generally, I think something is just not right in this PR 😅

akamfoad: Oh my bad😂

I don't have a computer with me rn, but can check what is going on tomorrow.

tsconfig.json
```diff
@@ -1,7 +1,6 @@
 {
   "compilerOptions": {
     "jsx": "react",
-    "module": "commonjs",
     "noImplicitAny": true,
```
nikolasrieble, line 4:
```suggestion
    "module": "commonjs",
    "noImplicitAny": true,
```

nikolasrieble:
> As we need to migrate to vitest for 3.0 release, but do not need the 3.0 release to migrate to vitest, lets do it now.

This might have been premature 😅 
After all we did upgrade the node versions in the vitest branch

ckifer:
>  After all we did upgrade the node versions of CI in the vitest branch

Lol I have not a clue what all has been done in this branch. While thats fine and all lets make sure we know what we're merging when we merge it

akamfoad:
@nikolasrieble The issue is not vitest, actually the tests are running perfectly, locally, the issue here is the CI I think

BTW Node 14 is deprecated, so we should kick it out of or CI and use last maintenance instead, wdyt?
https://nodejs.dev/en/about/releases/

image

ckifer:
We use multiple node versions to try to ensure little as possible breaking changes on those versions. 14 can be removed, but while 16 is deprecated too we should keep it around for at least a little while imo.

Also there shouldn't be an issue with using `ci` as its doing a clean dependency install...

HHongSeungWoo:
It is due to the version in the lock.json file. The default npm version of 14 is 6, which is not compatible. To resolve the issue, you need to use a different package manager or install npm in the CI

akamfoad:
@ckifer I'd target Maintenance, Active (LTS) and Current versions only, 16 is about to get EOLed as well, but it is still active as of now, so we can keep it in the CI for a couple of months until next active changes, we then kick it out as 14.

ckifer:
Ah cool nice find. Yep that makes sense to me @akamfoad  we might add 20 when we remove 16

ckifer (approved):
LGTM, lets do it

PR #3891, opened 2023-10-22
Refactoring mouse event

HHongSeungWoo:
## Description

I cleaned up the code and optimized performance.

## Types of changes

## Checklist:


src/chart/generateCategoricalChart.tsx
```diff
@@ -1499,28 +1473,41 @@ export const generateCategoricalChart = ({
...
+      }
+    };
+
+    accessibilityMouseHandler = (e: MousePointer) => {
+      const mouse = this.getMouseInfo(e);
+      const nextState: CategoricalChartState = mouse ? { ...mouse, isTooltipActive: true } : { isTooltipActive: false };
+      this.setState(nextState);
+      this.triggerSyncEvent(nextState);
+
+      // I don't understand why it is receiving keyboard or focus events and triggering mouse events
```
HHongSeungWoo, line 1509: @ckifer Can I know the reason?

ckifer: Not sure what you mean. This is for accessibility see https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-accessibility--docs

Click that chart then use the arrow keys to navigate the tooltip

src/chart/generateCategoricalChart.tsx
```diff
@@ -1322,21 +1298,23 @@ export const generateCategoricalChart = ({
...
       if (!this.container) {
         return null;
       }
 
       const containerOffset = getOffset(this.container);
-      const e = calculateChartCoordinate(event, containerOffset);
+      const e = {
+        chartX: Math.round(event.pageX - containerOffset.left),
+        chartY: Math.round(event.pageY - containerOffset.top),
+      };
```
ckifer, line 1313: why remove this? These are the same and the function is much easier to test in isolation than this

src/chart/generateCategoricalChart.tsx
```diff
@@ -174,14 +174,10 @@ const getDisplayedData = (
...
     return result;
   }, []);
-  if (itemsData && itemsData.length > 0) {
+  if (itemsData.length > 0) {
     return itemsData;
   }
 
-  if (item && item.props && item.props.data && item.props.data.length > 0) {
-    return item.props.data;
-  }
```
ckifer, line 183: why can this be deleted?

src/chart/generateCategoricalChart.tsx
```diff
@@ -313,7 +309,7 @@ export const getAxisMapByAxes = (
   const isCategorical = isCategoricalAxis(layout, axisType);
 
   // Eliminate duplicated axes
-  const axisMap: AxisMap = axes.reduce((result: AxisMap, child: ReactElement): AxisMap => {
+  return axes.reduce((result: AxisMap, child: ReactElement): AxisMap => {
```
ckifer, line 312: nit nit, non-blocking: the return here and below slightly decreases readability (at least for me)

src/chart/generateCategoricalChart.tsx
```diff
@@ -1094,9 +1088,9 @@ export const generateCategoricalChart = ({
   return class CategoricalChartWrapper extends Component<CategoricalChartProps, CategoricalChartState> {
     static displayName = chartName;
 
-    clipPathId: string;
+    readonly eventEmitterSymbol: Symbol = Symbol('rechartsEventEmitter');
```
ckifer, line 1091: what does this do?

src/chart/generateCategoricalChart.tsx
```diff
@@ -1283,35 +1274,20 @@ export const generateCategoricalChart = ({
...
-      this.cancelThrottledTriggerAfterMouseMove();
-    }
-
-    cancelThrottledTriggerAfterMouseMove() {
-      if (typeof (this.triggeredAfterMouseMove as any).cancel === 'function') {
-        (this.triggeredAfterMouseMove as any).cancel();
-      }
+      this.removeListener();
+      (
+        this.triggeredAfterMouseMove as DebouncedFunc<typeof CategoricalChartWrapper.prototype.triggeredAfterMouseMove>
```
ckifer, line 1283: could this be typed where its defined instead of casted here?

src/chart/generateCategoricalChart.tsx
```diff
@@ -1138,7 +1129,7 @@ export const generateCategoricalChart = ({
           top: this.props.margin.top ?? 0,
         },
         coordinateList: this.state.tooltipTicks,
-        mouseHandlerCallback: this.handleMouseMove,
+        mouseHandlerCallback: this.accessibilityMouseHandler,
```
ckifer, line 1132: I don't think this is accurate - mouseHandlerCallback should still be handleMouseMove

src/util/DataUtils.ts
```diff
@@ -11,6 +11,10 @@ export const mathSign = (value: number) => {
   return -1;
 };
 
+// use legacy isFinite only if there is a problem (aka IE)
+// eslint-disable-next-line no-restricted-globals
+export const isFinit = Number.isFinite ? Number.isFinite : isFinite;
```
ckifer, line 16: this is probably fine to remove entirely now

ckifer (commented):
some comments and questions, thanks!

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: I was curious about why onMouseMove is triggered for virtual events, not real mouse events.(probably because there seems to be no way to trigger changes in tooltips...)  
Similarly, other events seem to function the same way... 
These unpredictable synthetic events with names similar to default handlers in recharts make it difficult to use.

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: item is an element of graphicalItems, and from the code alone, it's not clear how it is being used. If passing item is necessary, it can be passed as part of graphicalItems, so I removed it.

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: I personally dislike unnecessary assignments. lol

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: I added this to prevent calling applySyncEvent when syscMethod is not a function.

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: Since triggeredAfterMouseMove is throttled, it doesn't seem appropriate here.

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: Okay

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: Personally, I don't think this function needs to be tested, and at least for now, as it is not reused, I think it's not a bad idea to do this for readability

src/util/DataUtils.ts
HHongSeungWoo: Great!

HHongSeungWoo:
What do you think about using lightweight stores like Jotai for refactoring, separate from this issue?

src/chart/generateCategoricalChart.tsx
ckifer: Agreed but unfortunately not much else is clear by code alone either 😅. We don't have the luxury of making changes like this without knowing the complete set of implications, especially when refactoring. 

Not ideal, but it's what we've been dealt

src/chart/generateCategoricalChart.tsx
ckifer: Eh it's preference. I only call it out cause this is a large block of code as the return

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: I believe I've checked everything overall, but I can revert it if you want. it doesn't seem that important.

The chart and each of its elements have the data props, but there are too many inconsistencies and ambiguities in the handling process. 😢😢😢

ckifer: Agree, it's hard to work with which is why we're trying to make these changes. Thanks for the effort, it's super appreciated. 

Let me pull it down later/tomorrow and test

src/chart/generateCategoricalChart.tsx
ckifer: Hm yeah this case is fine I guess - don't want to do too many things inline or without tests else we'll end up worse than we are now with generateCategoricalChart, but this instance should be fine

src/chart/generateCategoricalChart.tsx
```diff
@@ -1119,17 +1115,14 @@ export const generateCategoricalChart = ({
 
       this.clipPathId = `${props.id ?? uniqueId('recharts')}-clip`;
 
-      if (props.throttleDelay) {
-        this.triggeredAfterMouseMove = _.throttle(this.triggeredAfterMouseMove, props.throttleDelay);
-      }
+      // trigger 60fps
+      this.throttleTriggeredAfterMouseMove = _.throttle(this.triggeredAfterMouseMove, props.throttleDelay ?? 1000 / 60);
```
ckifer, line 1119: why are we doing this if it isn't provided by the user? Does the 16ms make a difference? (this is a behavior breaking change even if unnoticeable, probably fine though)

ckifer:
> What do you think about using lightweight stores like Jotai for refactoring, separate from this issue? 

See https://github.com/recharts/recharts/pull/3852. We are trying to do without any external dependency for now, using context

src/chart/generateCategoricalChart.tsx
ckifer: I'm not the original implementer of any of this so I have no idea. Totally agree that things are unpredictable and confusing, we just have to be very careful while untangling things.. Trying to prevent swaths of breaking changes to build more trust and confidence with consumers (previously things were randomly merged and released).

But yeah... I don't understand either 😅

HHongSeungWoo:
> > 이 문제와 별도로 리팩토링을 위해 Jotai와 같은 경량 스토어를 사용하는 것에 대해 어떻게 생각하시나요?
> 
> [#3852를](https://github.com/recharts/recharts/pull/3852) 참조하세요 . 지금은 컨텍스트를 사용하여 외부 종속성 없이 수행하려고 노력하고 있습니다.

Sure, I'll take a look. However, I'm concerned about re-renders caused by using the context.

src/chart/generateCategoricalChart.tsx
```diff
@@ -1634,7 +1602,7 @@ export const generateCategoricalChart = ({
 
     handleTouchMove = (e: React.TouchEvent) => {
       if (e.changedTouches != null && e.changedTouches.length > 0) {
-        this.handleMouseMove(e.changedTouches[0]);
+        this.throttleTriggeredAfterMouseMove(e.changedTouches[0]);
```
ckifer, line 1605: aren't these two functions slightly different? Missing e.persist now

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: It was added for performance reasons. As you mentioned, it might not be noticeable to the average users, but it will significantly improve performance for users with large datasets.

HHongSeungWoo: Even for the average user, other parts of the application can be impacted by performance issues, so having a default value might not be a bad idea (for now).

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: same!

HHongSeungWoo: Come to think of it, there's no 'persist' for touch events, right? It was never there originally

ckifer: I think it doesn't matter the type of event - all synthetic events need e.persist (react < 17 at least)

https://legacy.reactjs.org/docs/legacy-event-pooling.html

src/chart/generateCategoricalChart.tsx
ckifer: Can you remove this comment here for now? Else reword it to be more generic? (Don't use "I" since once this is merged no one will know who "you" are 😅)

src/chart/generateCategoricalChart.tsx
HHongSeungWoo: I know. But we are not passing events here

src/chart/generateCategoricalChart.tsx
ckifer: pulled it down, and yeah! this is totally fine to remove. I think the GitHub UI just makes it difficult to see how everything is connected when reviewing. Nice cleanup!

ckifer:
@HHongSeungWoo can you fix merge conflicts? Will get in after

HHongSeungWoo:
fixed it!

ckifer (approved):

PR #3892, opened 2023-10-22
Fix #2862

### LINKED ISSUE #2862: Scatter chart performance issues
Handler returned by `combineEventHandlers` [here](https://github.com/recharts/recharts/blob/86d4d563b8fa213373c56646367185e4adefbbee/src/chart/generateCategoricalChart.tsx#L1941) is an anonymous function which changes on every render. Because of this all Scatter elements rerender on every props or state change causing performance issues in big charts.
I have a scatter chart with 18 lines and hundreds of data points in my app and toggling tooltip in such chart takes 0.5 second.

Demo [here](https://codesandbox.io/s/scatter-chart-of-three-dimensions-forked-z2hiud?file=/src/App.tsx)
Hover points to show tooltip and see the console output. On every tooltip state change Scatter is rerendered, but it shouldn't normally happen.
_(It is impossible to track rendering of the Scatter itself, for demo purposes it is done using custom shape)_ 

For instance Line charts don't need such handlers and as a result are much more performant.

HHongSeungWoo:
## Description

If the reference of the combining function is the same, it has been modified to use caching.

## Related Issue

#2862

## Types of changes

## Checklist:


src/util/ChartUtils.ts
```diff
@@ -668,31 +668,31 @@ export const getTicksOfAxis = (
...
  * combine the handlers
  * @param  {Function} defaultHandler Internal private handler
- * @param  {Function} parentHandler  Handler function specified in parent component
- * @param  {Function} childHandler   Handler function specified in child component
+ * @param  {Function} childHandler Handler function specified in child component
  * @return {Function}                The combined handler
  */
-export const combineEventHandlers = (defaultHandler: Function, parentHandler: Function, childHandler: Function) => {
-  let customizedHandler: Function;
+const handlerWeakMap = new WeakMap();
```
ckifer, line 674: I like this change but how has this been tested? Can we add unit tests?

ckifer (changes requested):
Would like some more tests here if possible

src/util/ChartUtils.ts
HHongSeungWoo: I'll try to add it as time allows.

ckifer (approved):

ckifer:
Want to get some other eyes on this, but looks good to me!

PR #3893, opened 2023-10-22
Fix #2589

### LINKED ISSUE #2589: Performance Issues with Cartesian Axis (getStringSize)

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/magical-kilby-r8npx?file=/src/App.js)

### Steps to reproduce
Start a performance profile in chrome dev tools and you can see that there is a force reflow for every axis tick, which in the example is 5.

### What is expected?
the recharts render method would not utilize getStringSize because of its performance implications.

As sample size increases and several charts are rendered on the same page the above reflow completely cripples the application performance.

| Environment | Info |
|---|---|
| Recharts | v2.0.8 |
| React | 16.8.1 |
| System | OSX and Windows |
| Browser | Chrome, Safari, and Firefox |

---
this issue was submitted previously https://github.com/recharts/recharts/issues/1465 and closed.

HHongSeungWoo:
## Description

While not a fundamental solution, it will better utilize caching.

## Related Issue

#2589

## Types of changes

## Checklist:

### DIFF
--- a/src/util/DOMUtils.ts
+++ b/src/util/DOMUtils.ts
@@ -73,14 +73,23 @@ export const getStyleString = (style: CSSProperties) =>
     '',
   );
 
+function removeInvalidKeys(obj: Record<string, any>) {
+  const copyObj = { ...obj };
+  Object.keys(copyObj).forEach(key => {
+    if (!copyObj[key]) {
+      delete copyObj[key];
+    }
+  });
+  return copyObj;
+}
+
 export const getStringSize = (text: string | number, style: CSSProperties = {}): Size => {
   if (text === undefined || text === null || Global.isSsr) {
     return { width: 0, height: 0 };
   }
 
-  const str = `${text}`;
-  const styleString = getStyleString(style);
-  const cacheKey = `${str}-${styleString}`;
+  const copyStyle = removeInvalidKeys(style);
+  const cacheKey = JSON.stringify({ text, copyStyle });
 
   if (stringCache.widthCache[cacheKey]) {
     return stringCache.widthCache[cacheKey];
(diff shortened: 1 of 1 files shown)


src/util/DOMUtils.ts
```diff
@@ -96,13 +105,10 @@ export const getStringSize = (text: string | number, style: CSSProperties = {}):
...
-    const measurementSpanStyle: Record<string, any> = { ...SPAN_STYLE, ...style };
-    Object.keys(measurementSpanStyle).map(styleKey => {
-      (measurementSpan.style as Record<string, any>)[styleKey] = measurementSpanStyle[styleKey];
-      return styleKey;
-    });
+    const measurementSpanStyle: Record<string, any> = { ...SPAN_STYLE, ...copyStyle };
+    Object.assign(measurementSpan.style, measurementSpanStyle);
 
-    measurementSpan.textContent = str;
+    measurementSpan.textContent = `${text}`;
```
ckifer, line 111: The template literal isn't used so can probably remove in favor of just text (though ik this was here before)

src/util/DOMUtils.ts
```diff
@@ -73,14 +73,23 @@ export const getStyleString = (style: CSSProperties) =>
     '',
   );
 
+function removeInvalidKeys(obj: Record<string, any>) {
+  const copyObj = { ...obj };
+  Object.keys(copyObj).forEach(key => {
+    if (!copyObj[key]) {
```
ckifer, line 79: I'm confused, why would this ever not be true if you just called object.keys? Only if someone passed in. A falsy value?

ckifer (changes requested):

src/util/DOMUtils.ts
HHongSeungWoo: need type casting if you don't convert it into a string.

src/util/DOMUtils.ts
HHongSeungWoo: In functions like getTicks, we are passing undefined or an empty string to font-size.

src/util/DOMUtils.ts
ckifer: It is a string right? Would `String(text)` be more appropriate?

HHongSeungWoo: That doesn't seem like a good idea in several ways. There's no need to box primitive types into objects, and it's performance-wise not optimal

ckifer: Hm, I don't like the fact that it's there 😅. But only because it wouldn't need to be there if the type was correct. Guess it's fine

ckifer (approved):

src/util/DOMUtils.ts
nikolasrieble: Then I wonder, should we even call getStringSize in those cases?

HHongSeungWoo: It's an issue stemming from not knowing the calculated font size before the DOM is rendered. If the component waits for rendering in the DOM and calculates it all at once, the problem should be resolved. With the ongoing refactorings, it might naturally get resolved, don't you think?

Issue #2329, opened 2020-11-25
Event object for onTouchMove is null


AlexCat:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/heuristic-feistel-nrkg1?file=/src/App.js)

### Steps to reproduce
1. Open on mobile device
2. Try to fire onTouchMove event (swipe)
3. Event object for onTouchMove will be null

### What is expected?
1. Event object contains relevant data

### What is actually happening?
1.Event object for onTouchMove is null

| Environment | Info |
|---|---|
| Recharts | v1.8.5 |
| React | 17 |
| System | codesandbox.io |
| Browser | Google Chrome Version 87.0.4280.66 |

xile611:
Sorry for not handling your issue in time, please try the latest version. If the problem persists, please open a new issue according to the issue guide.

nikolasrieble:
Released fix with https://github.com/recharts/recharts/releases/tag/v2.10

PR #3894, opened 2023-10-22
Change in the value of CategoricalChartFunc

HHongSeungWoo:
## Description

We cannot predict null as a type, but the value can be null.
If you are not going to change the type definition of CategoricalChartFunc, it would be better to return an empty object.

## Related Issue

#2329

## Types of changes

## Checklist:


ckifer:
I guess this is fine - it doesn't really mean much in the grand scheme of that issue (because they still don't get the data they want), but all good. Now the type is more accurate

ckifer (approved):

PR #3895, opened 2023-10-22
fix: target supported node versions in CI

akamfoad:
## Description

Target only supported node versions, this includes 3 versions that are currently one of
- Maintained
- LTS
- Next
## Types of changes

## Checklist:


HHongSeungWoo:
To run the test, you also need to downgrade the version of @testing-library/user-event.

HHongSeungWoo:
recommmend ^13.5.0

akamfoad:
@HHongSeungWoo What would be the reason for that?

HHongSeungWoo:
Oh, the build succeeded. I had to create a new package-lock file because the build was failing due to the new syntax in TS. Seems like there were some dependency issues.

akamfoad:
It was in fact due to dependency issue, I'm guessing it was from classnames, I'd mitigate that in a different PR, I have seen the warning for @types/classnames quite a few times now, looks like classnames is packing their types now.

Here is the fail in CI: https://github.com/recharts/recharts/actions/runs/6599091572/job/17927744175 

Thanks for the suggestion.

akamfoad:
@PavelVanecek I removed the perf file since it was causing TypeChecking issues, it is not a big file and we have it in the history of the repo and I'm sure you have it locally too, we can add it back when we're confident that the performance tests work.

ckifer (approved):

PavelVanecek:
@akamfoad yep that's fine, that's what git is for. We can always retrieve it from there later.

Issue #3899, opened 2023-10-23
Risky code in getStrokeDasharray leads to infinite loop when strokeDasharray is '0'


vjohansen:
I was using strokeDasharray='0' to disable dashed lines (maybe incorrect) but it leads to an infinite loop in getStrokeDasharray .

I suggest the code is changed so infinite loops do not happen even on incorrect user input. From a code style perspective it should be obvious that an infinite loop does not happen (maybe : if (i>10) {console.warn .. ; break}) . 

![image](https://github.com/recharts/recharts/assets/634765/671094b3-019e-4baa-ac3b-3fd6f486681a)

vjohansen:
This may explain why setting isAnimationActive={false} in #1812 fixes hangs.

There seem to be a relation with strokeDasharray and isAnimationActive 

"Line chart strokeDasharray does not apply if isAnimationActive is false from the start and set to true in componentDidMount" #2373

ckifer:
Thanks for the report, will get this fixed. A lot of this is an inherited legacy so I never would've known that line existed or these things are related. 

Agreed that code is incredibly risky and we shouldn't trust human input ever (especially in loops) 😅

ckifer:
Should be resolved by #3904 - will be fixed in next release. Thanks!

ckifer:
released in 2.9.1

PR #3904, opened 2023-10-25
fix: infinite loop when Line strokeDasharray is 0

ckifer:
## Description

Fix infinite loop case in `Line` where `strokeDasharray="0"` causes an infinite loop

## Related Issue

https://github.com/recharts/recharts/issues/3899

## Motivation and Context

This is a code smell and a bug

## How Has This Been Tested?

- Ran storybook with strokeDasharray as 0 and reproduced, made changes and see that infinite loop no longer happens (and animation still works as expected
- Add unit test

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Line.tsx
+++ b/src/cartesian/Line.tsx
@@ -225,7 +225,6 @@ export class Line extends PureComponent<Props, State> {
     const restLength = totalLength - length;
 
     let remainLines: number[] = [];
-    // prevent infinite loop by checking length https://github.com/recharts/recharts/issues/3899
     for (let i = 0, sum = 0; i < lines.length; sum += lines[i], ++i) {
       if (sum + lines[i] > remainLength) {
         remainLines = [...lines.slice(0, i), remainLength - sum];
--- a/src/cartesian/Line.tsx
+++ b/src/cartesian/Line.tsx
@@ -208,12 +208,16 @@ export class Line extends PureComponent<Props, State> {
     }
   }
 
+  generateSimpleStrokeDasharray = (totalLength: number, length: number): string => {
+    return `${length}px ${totalLength - length}px`;
+  };
+
   getStrokeDasharray = (length: number, totalLength: number, lines: number[]) => {
     const lineLength = lines.reduce((pre, next) => pre + next);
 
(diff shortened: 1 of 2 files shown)


src/cartesian/Line.tsx
```diff
@@ -211,12 +211,18 @@ export class Line extends PureComponent<Props, State> {
   getStrokeDasharray = (length: number, totalLength: number, lines: number[]) => {
     const lineLength = lines.reduce((pre, next) => pre + next);
 
+    // if lineLength is 0 return the default when no strokeDasharray is provided
+    if (!lineLength) {
+      return `${length}px ${totalLength - length}px`;
```
ckifer, line 216: See lines below for where this comes from and why I returned this specifically 
https://github.com/recharts/recharts/pull/3904/files#diff-acabe8d166eb869ed5f168003bedf80f5bad4827ce727453b57733b3c0cf1974R439-R444

nikolasrieble: This works, but it would be nicer to encapsulate in a helper function to communicate that both cases should be the same, right? 

```
function generateSimpleLine(totallength, length){
    return `${length}px ${totalLength - length}px`
}
```

ckifer: Sure, easy change

src/cartesian/Line.tsx
```diff
@@ -211,12 +211,18 @@ export class Line extends PureComponent<Props, State> {
...
+    }
+
     const count = Math.floor(length / lineLength);
     const remainLength = length % lineLength;
     const restLength = totalLength - length;
 
-    let remainLines = [];
-    for (let i = 0, sum = 0; ; sum += lines[i], ++i) {
+    let remainLines: number[] = [];
+    // prevent infinite loop by checking length https://github.com/recharts/recharts/issues/3899
```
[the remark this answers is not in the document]
ckifer: eh yeah. sometimes we link issues where things have slipped by in the past and adding a comment adds some awareness. This one should be pretty obvious by looking at it since it was a for-loop without a break condition 😓

ckifer:
Changes made

akamfoad (approved):

Issue #3907, opened 2023-10-26
CartesianGrid - Can no longer render CartesianGrid alone


ckifer:
### Discussed in https://github.com/recharts/recharts/discussions/3906

<div type='discussions-op-text'>

<sup>Originally posted by **sahillihas2021** October 26, 2023</sup>
I am displaying a multilinechart (Having multiple Y-Axis), it's working perfectly fine in my local, but when I am deploying it to the internal dev server, it is giving an error for the same code on : ERROR

Cannot read properties of null (reading 'ticks')
TypeError: Cannot read properties of null (reading 'ticks')
at CartesianGrid.render (https://react.qa.he.az.devmfi.net/static/js/bundle.js:197529:66)
at finishClassComponent (https://react.qa.he.az.devmfi.net/static/js/bundle.js:167192:35)
at updateClassComponent (https://react.qa.he.az.devmfi.net/static/js/bundle.js:167149:28)
at beginWork (https://react.qa.he.az.devmfi.net/static/js/bundle.js:168775:20)
at HTMLUnknownElement.callCallback (https://react.qa.he.az.devmfi.net/static/js/bundle.js:153766:18)
at Object.invokeGuardedCallbackDev (https://react.qa.he.az.devmfi.net/static/js/bundle.js:153810:20)
at invokeGuardedCallback (https://react.qa.he.az.devmfi.net/static/js/bundle.js:153867:35)
at beginWork$1 (https://react.qa.he.az.devmfi.net/static/js/bundle.js:173741:11)
at performUnitOfWork (https://react.qa.he.az.devmfi.net/static/js/bundle.js:172988:16)
at workLoopSync (https://react.qa.he.az.devmfi.net/static/js/bundle.js:172911:9)

Due to: [Refer the Code below)
```jsx
yAxisComponents.push(
<YAxis
key={yAxisId}
yAxisId={yAxisId}
orientation={orientation}
axisLine={{ display: 'none' }}
tickLine={{ display: 'none' }}
tick={{
fill: tickColor,
fontSize: 16,
fontWeight: 'bold'
}}
/>
);
```

Code:
```jsx
import React, { useState, useEffect } from 'react';
import { LineChart, Line, XAxis, YAxis, CartesianGrid, Tooltip } from 'recharts';
import PropTypes from 'prop-types';
import Loading from '../../Loading/Loading';
function MultipleYAxesLineChart({ data, CustomXAxisTick, loading }) {
  MultipleYAxesLineChart.propTypes = {
[...]

PR #3912, opened 2023-10-27
fix: CartesianGrid throws error when xAxis or yAxis doesn't exist

akamfoad:
## Description

- add: LineChart example toggling children
    This example showcases a LineChart with ability to toggle content of the chart like Line, Tooltip and etc.

    This example introduced to reproduce a bug reported on #3906 where keeping only CartesianGrid rendered would throw an error since the xAxis and yAxis aren't present (not rendered)

- fix: optional xAxis & yAxis access without checks
    xAxis and yAxis are optional and CartesianGrid shouldn't use them if they don't exist.

## Related Issue

Fixes: #3906 #3907

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/CartesianGrid.tsx
+++ b/src/cartesian/CartesianGrid.tsx
@@ -307,10 +307,12 @@ export class CartesianGrid extends PureComponent<Props> {
 
       horizontalPoints = horizontalCoordinatesGenerator(
         {
-          yAxis: {
-            ...yAxis,
-            ticks: isHorizontalValues ? horizontalValues : yAxis.ticks,
-          },
+          yAxis: yAxis
+            ? {
+                ...yAxis,
+                ticks: isHorizontalValues ? horizontalValues : yAxis.ticks,
+              }
+            : undefined,
           width: chartWidth,
           height: chartHeight,
           offset,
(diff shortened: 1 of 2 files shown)


ckifer (approved):
Thank you @akamfoad !

src/cartesian/CartesianGrid.tsx
```diff
@@ -325,10 +327,12 @@ export class CartesianGrid extends PureComponent<Props> {
...
         {
-          xAxis: {
-            ...xAxis,
-            ticks: isVerticalValues ? verticalValues : xAxis.ticks,
-          },
+          xAxis: xAxis
+            ? {
+                ...xAxis,
+                ticks: isVerticalValues ? verticalValues : xAxis.ticks,
+              }
```
nikolasrieble, line 334: Interesting, could we have caught this with better types?

ckifer: I've learned that you can't really trust our types in terms of undefined/null. In this case we should always have a y axis but because of how the components are composed you can't guarantee the user provided one. 

One way would be to always have a default axis used for calculation but not show it

Issue #3911, opened 2023-10-27
[BarChart] Broken types for ActiveShape


klassm:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://github.com/klassm/recharts-bar-chart-repro](https://github.com/klassm/recharts-bar-chart-repro)

### Steps to reproduce
- Downgrade to recharts 2.8.0 -> compiles
- Upgrade to recharts 2.9.9 -> broken with multiple issues

### What is expected?
- The name property can be used without casting (as string)
- The IData property, so the interface for the data, can be used within the Shape component (which was possible beforehand and still is, it is just prohibited by typing)

### What is actually happening?
For the name:
```
Type  BarProps  is not assignable to type  Omit<SVGProps<SVGPathElement>, "radius"> 
Types of property  name  are incompatible.
Type  string | number | undefined  is not assignable to type  string | undefined
```
For the rest of the properties - they are just not available. Looking at the code, the `ActiveShape` interface just always uses `BarProps` and ignores the type of the data - which is also why typescript does not allow to have it.

| Environment | Info |
|---|---|
| Recharts | v2.9.0 |
| React | 18.2.28 |
| System | MacOS |
| Browser | Chrome |

akamfoad:
This change have been introduced by #3803, the signature of **shape** and others changed

```diff
-  shape?: ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>);
+  shape?: ActiveShape<BarProps, SVGPathElement>;
```

So before the props where of type `any` I think, that is why you could do that, I understand this might be a breaking change, but you could work around it with render prop and make the `(prop: any) => <CustomeShape />`

ckifer:
@andrewangelle FYI

klassm:
Thanks for the response. Yes, casting it to any works - so there are workarounds; it is just - they shouldn't be necessary (especially casting to any -.-)

akamfoad:
@klassm 

Indeed, sorry if that landed poorly, I meant so you're not blocked until we fix this issue.

We respect the semver semantics and we had no idea this breaking change was introduced.

- Thanks for reporting, hopefully we'll release a minor/patch fix to revert this back to the pre-breaking-change interface. Though I cannot promise about ETA since we all work on recharts on our own time, and contributions are always welcome.

Thanks again for the bug report.

ckifer:
+1 to the above
 
[...]

PR #3915, opened 2023-10-27
fix(ActiveShape): Fix breaking change in ActiveShape type 

andrewangelle:
Fix ActiveShape type that causes  breaking change in Bar shape prop

## Description

Extend the definition of `ActiveShape` to allow consumer to define the type of props argument on their own render prop being passed to `shape`.

## Related Issue
#3911 

## How Has This Been Tested?
unit tests

## Types of changes

## Checklist:

### DIFF
--- a/src/util/types.ts
+++ b/src/util/types.ts
@@ -20,6 +20,7 @@ import {
   TransitionEvent,
   UIEvent,
   WheelEvent,
+  JSX,
 } from 'react';
 import _ from 'lodash';
 import { ScaleContinuousNumeric as D3ScaleContinuousNumeric } from 'victory-vendor/d3-scale';
@@ -1320,5 +1321,6 @@ export type Size = { width: number; height: number };
 export type ActiveShape<PropsType = Record<string, any>, ElementType = SVGElement> =
   | ReactElement<SVGProps<ElementType>>
   | ((props: PropsType) => ReactElement<SVGProps<ElementType>>)
+  | ((props: unknown) => JSX.Element)
   | SVGProps<ElementType>
   | boolean;
--- a/test/chart/BarChart.spec.tsx
+++ b/test/chart/BarChart.spec.tsx
@@ -2,10 +2,16 @@ import { fireEvent, render } from '@testing-library/react';
 import React from 'react';
 
 import { vi } from 'vitest';
-import { Bar, BarChart, Rectangle, Tooltip, XAxis, YAxis } from '../../src';
+import { Bar, BarChart, Rectangle, RectangleProps, Tooltip, XAxis, YAxis } from '../../src';
+
+type DataType = {
+  name: string;
+  uv: number;
+  pv: number;
+};
 
 describe('<BarChart />', () => {
-  const data = [
+  const data: DataType[] = [
     { name: 'food', uv: 400, pv: 2400 },
     { name: 'cosmetic', uv: 300, pv: 4567 },
     { name: 'storage', uv: 300, pv: 1398 },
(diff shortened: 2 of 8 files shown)


test/chart/BarChart.spec.tsx
```diff
@@ -265,7 +271,7 @@ describe('<BarChart />', () => {
   });
 
   test('Render customized shapem when shape is set to be a function', () => {
-    const renderShape = (props: any) => {
+    const renderShape = (props: RectangleProps & DataType) => {
```
andrewangelle, line 274: This is very similar to the implementation provided by the user in the reproduction linked in the issue. 

This also would throw a type error prior to the change in this PR.

src/util/types.ts
```diff
@@ -1320,5 +1321,6 @@ export type Size = { width: number; height: number };
 export type ActiveShape<PropsType = Record<string, any>, ElementType = SVGElement> =
   | ReactElement<SVGProps<ElementType>>
   | ((props: PropsType) => ReactElement<SVGProps<ElementType>>)
+  | ((props: unknown) => JSX.Element)
```
andrewangelle, line 1324: This is the fix

ckifer:
So now if someone has `activeShape={props => <Trapezoid {...props} />}` they are most likely going to get the implicit `any` errors that just appeared in our CI no?

ckifer (approved):
Definitely better either way and fixes the reported bug, might cause some other tsc errors for folks, but those should be easy for them to fix. Merging this for now

andrewangelle:
> So now if someone has `activeShape={props => <Trapezoid {...props} />}` without an explicit type they are most likely going to get the implicit `any` errors that just appeared in our CI no?

Yes. But that was the behavior prior to `2.9.0`, so we're likely fixing more folks than we're breaking. For the record, I'd like to remove this in the next major version. Assuming our types are accurate, I think we should be throwing a type error if the user is assigning a random type to a parameter that we internally define.

ckifer:
Agreed but we should take a closer look to ensure the type doesn't leave anything out before that

ckifer:
Hmm, If I cherry pick this commit from `v2.9.0` tag I get a bunch of compilation failures
```
No overload matches this call.
  Overload 1 of 2, '(props: Props | Readonly<Props>): Bar', gave the following error.
    Type '(props: RectangleProps & DataType) => React.JSX.Element' is not assignable to type 'ActiveShape<BarProps, SVGPathElement> | undefined'.
      Type '(props: RectangleProps & DataType) => React.JSX.Element' is not assignable to type '(props: unknown) => Element'.
        Types of parameters 'props' and 'props' are incompatible.
          Type 'unknown' is not assignable to type 'Omit<SVGProps<SVGPathElement>, "radius"> & RectangleProps & DataType'.
            Type 'unknown' is not assignable to type 'Omit<SVGProps<SVGPathElement>, "radius">'.
  Overload 2 of 2, '(props: Props, context: any): Bar', gave the following error.
    Type '(props: RectangleProps & DataType) => React.JSX.Element' is not assignable to type 'ActiveShape<BarProps, SVGPathElement> | undefined'.
```

Any idea?

ckifer:
@andrewangelle https://github.com/ckifer/recharts/blob/patch/2.9.1/test/chart/BarChart.spec.tsx#L294

I pushed my patch branch up.. I can't figure out why that type isn't working
The difference I see is 

`(property) shape?: ActiveShape<BarProps, SVGPathElement>` (working) vs `(property) shape?: ActiveShape<BarProps, SVGPathElement> | undefined` (not working), but no idea why

andrewangelle:
@ckifer I can't track down why master has no errors, but cherry-picked into 2.9.0 does.

It does seem like `| ((props: any) => JSX.Element)` instead of `| ((props: unknown) => JSX.Element)` removes the type errors. Do you want me to PR that?

ckifer:
Hmm, no worries I can just change it and publish that for now. I don't have plans to merge 2.9.1 back into master or do another patch release if I can help it. There's been too much other stuff merged since. Thank you!

PR #3918, opened 2023-10-29
Fixed the initial mount re-rendering bug in ResponsiveContainer

HHongSeungWoo:
## Types of changes

## Checklist:

### DIFF
--- a/src/component/ResponsiveContainer.tsx
+++ b/src/component/ResponsiveContainer.tsx
@@ -12,6 +12,7 @@ import React, {
   useEffect,
   useMemo,
   CSSProperties,
+  useCallback,
 } from 'react';
 import { throttle } from 'lodash';
 import { isPercent } from '../util/DataUtils';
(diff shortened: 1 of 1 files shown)


src/component/ResponsiveContainer.tsx
```diff
@@ -77,10 +78,22 @@ export const ResponsiveContainer = forwardRef<HTMLDivElement, Props>(
       containerHeight: initialDimension.height,
     });
 
+    const setContainerSize = useCallback((newWidth: number, newHeight: number) => {
+      setSizes(prevState => {
+        const roundedWidth = Math.round(newWidth);
+        const roundedHeight = Math.round(newHeight);
+        if (prevState.containerWidth === roundedWidth && prevState.containerHeight === roundedHeight) {
```
ckifer, line 85: I guess I only see a slight chance of this happening no? And if the previous state is the same as the new state what's wrong with just not checking this and returning the new state?

ckifer (commented):
Just one question on the callback

src/component/ResponsiveContainer.tsx
nikolasrieble: If this is not done, the state would always change, right?
In line 89, we always create a new object.

ckifer: Wouldn't it most likely always change anyways? This only helps if I resize the screen and move it to the exact height and width pixels it was before. Though I might be mistaken - it definitely doesn't hurt to be here, more curious than anything

HHongSeungWoo: In browsers that support sub-pixels, values may differ due to precision issues. 
Therefore, I compared the values, and if they were the same, I returned the previous state as is to prevent re-rendering

ckifer: Gotcha thanks. It's a mini optimization imo but again, doesn't hurt to be here

ckifer (approved):

Issue #3922, opened 2023-10-31
Bad type on activeDot onClick


cplprince:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-sz6kpd?file=/src/index.tsx)

### Steps to reproduce
Your onClick fuction on activeDot on Line is badly typed. If you use the type which is written in the onClick, then there's not a payload attribute on 2. position of the onClick.

https://pasteboard.co/DAYWcO1i0xa7.png

https://pasteboard.co/BYM0L0t2XByp.png

### What is expected?
Have at least :any type on the second position for the payload.

### What is actually happening?
We have a TS error 

| Environment | Info |
|---|---|
| Recharts | v2.9.1 |
| React | 17.0.2 |
| System | Windows 11 |
| Browser | Version 118.0.5993.118 (Official Build) (64-bit) |

ckifer:
Did this just happen in 2.9.1 or was it like this previously?

ckifer:
@andrewangelle 😅. Sorry to keep tagging you but fyi. Not sure if it's my latest change for 2.9.1 yet or not

cplprince:
we were on the 2.3.2 before. The payload type was missing but at least it was usable.

cplprince:
If you want, I can check if it works differently on different version?

cplprince:
btw we are using "typescript": "^4.6.2", if it makes any difference 🙂

ckifer:
>  I hate the types 

Can you explain why you hate the types 😅? The goal is to make them better, more accurate - not worse.

andrewangelle:
> Sorry to keep tagging you but fyi. Not sure if it's my latest change for 2.9.1 yet or not

@ckifer Please feel free to continue to do so. I have a fix. I'll have it up today.

ckifer:
@andrewangelle 🙌 appreciate you!

ckifer:
Fixed in 2.9.2, resolving this

cplprince:
@ckifer look better but payload.payload does not exist, would it be possible to have it in such a way that Required<DotProps>['onClick']<T> where the T is a generic for the payload.payload? if non is provided then the payload.payload is any?

andrewangelle:
[...]

PR #3923, opened 2023-10-31
fix(Line): onClick for activeDot is incorrectly typed

andrewangelle:
fix(Line): onClick for activeDot is incorrectly typed

## Description

Have `activeDot` include the type for `DotProps` which fixes the incorrect types for the event handlers that are added in [line 2036 in generateCategoricalChart](https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.tsx#L2036
) which give a different payload to the callback mentioned in the issue.

## Related Issue

#3922 

## Motivation and Context

Fixes incorrect types that introduced a breaking change 

## How Has This Been Tested?
Manually. It's a purely typescript issue. Screenshot below.

Screenshot 2023-10-31 at 4 22 50 PM

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Line.tsx
+++ b/src/cartesian/Line.tsx
@@ -65,7 +65,7 @@ interface LineProps extends InternalLineProps {
   hide?: boolean;
 
   // whether have dot in line
-  activeDot?: ActiveShape<DotProps>;
+  activeDot?: ActiveShape<DotProps> | DotProps;
   dot?: LineDot;
 
   onAnimationStart?: () => void;


andrewangelle:
@ckifer I pulled down your v2.9.1 patch branch and cherry picked this commit into it locally to make sure there were no surprises there. I was able to run the tests/build/linter all successfully on my local with this commit.

nikolasrieble (approved):
🚀 Thank you. Can you please double check whether this issue also exists with other ActiveShape usage cases?

andrewangelle:
@nikolasrieble  I console logged the events coming in from the onClick for the others and they are receiving a normal event payload that you'd expect from an svg element event handler which is consistent with the current type definition for those.

It'll take me a bit longer to track down if this is how they were typed prior to v2.9.0. 

This one for `activeDot` is different than the rest because the chart wrapper adds a different event handler and renders the active dots separately. The other use cases of `activeBar/activeShape` are handled via the same rendering method in the chart wrapper and don't have this same overwriting of the event.

ckifer (approved):

ckifer:
Maybe a tiny bit late for this but `Area` also has an `activeDot` prop. That one however, does not use the new ActiveShape type

https://github.com/recharts/recharts/blob/master/src/cartesian/Area.tsx#L66

andrewangelle:
For the components touched in the `ActiveShape` changes, the types for the event handlers all look to be equivalent in the latest version versus 2.8.0. Though admittedly, it's hard to anticipate every which way users could have been using that prop to anticipate other ways to test that haven't already been covered. I'll continue to think about this and see if I can find any inconsistencies that might have been missed.

ckifer:
Yep agreed! Thank you. 

It's completely impossible to be honest - it's going to be tough to narrow types in minor version releases. Generally external facing type changes are breaking changes unfortunately

Issue #3934, opened 2023-11-05
Tooltip: Show both cursor and tooltip after moving out of the chart


nikolasrieble:
This issue is an extension of https://github.com/recharts/recharts/issues/855. 

Goal: Implement a lockable Tooltip (by click) with a custom cursor. 

Expected behaviour: When the Tooltip is locked and the mouse is moved out of the chart area, both Tooltip and Cursor remain where they are and are visible. 

Current behaviour: When the Tooltip is locked and the mouse is moved out of the chart area, the Tooltip is still visible but the custom cursor disappears. 

For the sake of this behaviour, we can ignore the locking context. 
In this reproduction, the cursor should also remain visible https://codesandbox.io/s/simple-bar-chart-always-show-tooltip-ekoiu2?file=/src/App.tsx

TODO: 
- [ ] Investigate workaround 
- [x] Investigate API extension in recharts

nikolasrieble:
Released with https://github.com/recharts/recharts/releases/tag/v2.10

Issue #855, opened 2017-08-30
Show Tooltip on load and keep visible


hellofantastic:
## Do you want to request a feature or report a bug?
Feature?
## What is the current behavior?
Tooltips only show on hover of a Bar 

## What is the expected behavior?
Possible setting so Tooltips permanently showing

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Not Applicable?

Is it possible to show tooltips onload and keep them visible near their associated bars

myangsc:
same question here, any one has any suggestion?

aesopwolf:
In theory you should be able to pass `active` and `payload` props to make this happen, but it appears the Tooltip isn't receiving those properly:

![image](https://user-images.githubusercontent.com/6847633/30460569-79c55a2c-996c-11e7-80fb-861e94e52ffa.png)

aesopwolf:
This should probably be filed as a bug report^. @xile611

myangsc:
Can change the logic of the isTooltipActive, activeTooltipIndex inside generateCategoricalChart.js to make it work

aesopwolf:
Similar problem over here https://github.com/recharts/recharts/issues/790

uiruson:
Any update of this fix?

christinajensen:
+1 :)

EpicNem:
+1

NaelFR:
+1 😁

mikkelhartung:
+1

mattspeller:
+1

Johnsonok:
+1

dutgcom:
+1

schevall:
+1

ThanhPhat1080:
+1

massile:
Any updates ?

austere-rm:
+1

johnfiorilla-refundlabs:
+1

SkryGun:
+1

schevall:
another one, for that matter =0 => +1

BrunoQuaresma:
+1

darcot:
+1

QuynhNguyen:
+1

CibiAananth:
+1

itaiMona:
+1

cle-skov-dk:
+1

Raymond427:
+1

dentov13:
+1

vtereshyn:
nice :) 👍

[...]

Issue #2967, opened 2022-09-14
How can I dismiss tooltip if it triggered by click?


EmmmaG:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Currently if tooltip triggered by click, it can't be dismissed. I think users are expecting a feature to dismiss tooltip by clicking away out of the chart/

### What does the proposed API look like?
it should be included when user use ``` trigger='click' ```

ckifer:
Hi @EmmmaG 

As of the latest version of recharts the escape key should work to dismiss the tooltip.

Works here when changing trigger to click
https://codesandbox.io/s/simple-line-chart-forked-3pn9vn?file=/src/App.tsx

Agreed though, the tooltip should dismiss when clicking out of the chart. Labeling this as a feature request.

ArcherP-Calibrate:
I also need this answer on this because clicking escape is not a intuitive method of closing the component

PR #3935, opened 2023-11-05
Fix: Respect the user configuration of Tootip - active

nikolasrieble:
## Description
The Tooltip has a property named `active`. It can be set by the user. 
Setting it from the outside has no effect, since it is overwritten by internal logic, when copying the element. Instead, we should respect the users configuration. 

Interestingly, this fix makes this workaround redundant https://github.com/recharts/recharts/issues/855

Additionally, the cursor (itself a Tooltip prop), should also respect the Tooltip and be shown. 

This change is:
- breaking 
- fixes an old bug (i.e. active never worked) 
- solves a user need 
- enables further customisation (my need: I want to have a lockable Tooltip and Cursor, which I can hack together for the Tooltip, but not for the cursor.) 

I see no use cases of a customer this could break, since it merely changes the behaviour when `active` is set - something that previously had no effect at all. 

## Related Issue
https://github.com/recharts/recharts/issues/3934

## Types of changes

## Checklist:


nikolasrieble:
The missing piece here is that when pressing ESC, the cursor no longer disappears. 

We could: 
- Move isDismissed on the Tooltip from state to props, and mark it as internal for the documentation, access it when rendering the cursor (which is a function that takes in the Tooltip)
- ?

Any ideas @PavelVanecek ?

ckifer:
Wouldn't the user be able to control isDismissed with `isActive`? If they set is active to false it should(?) disappear?

PavelVanecek:
So is this a breaking change? Depends who you ask.
Based on the documentation, no:
> [active]
> If set true, the tooltip is displayed. If set false, the tooltip is hidden, usually calculated internally.

However, based on my experience with how people use software, it's 100% a breaking change. If this was at work where we have paying customers, I would 100% insist on breaking change protocol. But we're not at work, we're opensource library, I think we can go with "technically".

I don't understand what broke the Escape functionality. That was inside the Tooltip component, right? And it only worked if `active` is not set. Is that different now?

If the Escape closing no longer works when `active=true` then that's something that we document and ship. So maybe the documentation, instead of the mysterious "usually calculated internally" should say:

- If active is `true`, then Tooltip is always displayed
- If active is `false`, then Tooltip is never displayed
- If active is `undefined`, then Recharts will control when the Tooltip displays. This includes mouse and keyboard controls.

ckifer:
Dismiss on escape was something that was explicitly added as functionality from what I remember - but only without the active property in mind. If it still works with the internal check then I think this doesn't break anything. 

Does it break it without the active property set?

If it does I think it's a breaking change and we'll see users create issues about it.

ckifer:
But if you look at issues like https://github.com/recharts/recharts/issues/2967 we should probably have another, better way to dismiss anyways

nikolasrieble:
All great input, thank you. 

Next steps: 
(1) Clearly document the behaviour before this change. 
- Storybook, Play function, possibly RTL tests 
(2) Clearly document the behaviour after this change. 

This PR first and foremost fixes the active props on the Tooltip. It does **not** break dismissing, but it is not a full fix. Specifically: When setting active from the outside, then the dismissing only dismisses the Tooltip, but not the cursor. 

I believe it is generally not a good idea to mingle (a) user facing props with (b) internally set props. Specifically, we should never override props set by users with internal logic (exactly what is fixed here).

ckifer:
Got it, thanks for the clarification. We should probably make dismiss on click outside of the chart one of the follow ups - issue has been around for far too long

src/chart/generateCategoricalChart.tsx
```diff
@@ -1751,11 +1751,13 @@ export const generateCategoricalChart = ({
       const { isTooltipActive, activeCoordinate, activePayload, offset, activeTooltipIndex, tooltipAxisBandSize } =
         this.state;
       const tooltipEventType = this.getTooltipEventType();
+      // Now the cursor is a part of the Tooltip itself, and it should be shown (by default) when the Tooltip is active.
```
nikolasrieble, line 1754:
```suggestion
      // The cursor is a part of the Tooltip, and it should be shown (by default) when the Tooltip is active.
```

src/chart/generateCategoricalChart.tsx
```diff
@@ -1950,11 +1952,16 @@ export const generateCategoricalChart = ({
 
       const { isTooltipActive, activeCoordinate, activePayload, activeLabel, offset } = this.state;
 
+      // The user can set isActive directly on the Tooltip,
+      // and we must respect the users override to enable customisation.
+      // The Tooltip is active if the user has set isActive, or if the tooltip is active due to a mouse event.
```
nikolasrieble:
```suggestion
      // The user can set isActive on the Tooltip,
      // and we respect the user to enable customisation.
      // The Tooltip is active if the user has set isActive, or if the tooltip is active due to a mouse event.
```

storybook/stories/Examples/Tooltip.stories.tsx
```diff
@@ -52,8 +101,29 @@ export const LockedByClick = {
...
+      fireEvent.mouseMove(chart, { clientX: 400, clientY: 100 });
+      expect(screen.getByText('590')).toBeVisible();
+
+      // When the mouse leaves the chart, the Tooltip still stays, but the cursor does not.
+      fireEvent.mouseOut(chart);
+
+      // The Tooltip is still visible, although the mouse is out of the chart.
+      expect(screen.getByText('590')).toBeVisible();
+    }, 0);
+  },
```
nikolasrieble, line 126:
```suggestion
```

ckifer:
Looks like this breaks the lock by click interaction test?

nikolasrieble:
> Looks like this breaks the lock by click interaction test?

It does, but funnily no such test exists. I added a single play function to one story - where the does not fail. 
Strangely this play function is applied to all stories in the same file - where it fails. 
I did not yet find out why. 

🤔 Can reproduce locally though when running the scripts. 
But when starting storybook, and I manually run the script, it all works fine.

ckifer:
Oh I totally thought there was a test for that. Weird..

src/component/Tooltip.tsx
```diff
@@ -38,7 +38,12 @@ function renderContent<TValue extends ValueType, TName extends NameType>(
 }
 
 export type TooltipProps<TValue extends ValueType, TName extends NameType> = ToltipContentProps<TValue, TName> & {
-  active?: boolean;
+  /**
+   * If true, then Tooltip is always displayed, once an activeIndex is set by mouse over, or programmatically.
+   * If false, then Tooltip is never displayed.
+   * If active is undefined, Recharts will control when the Tooltip displays. This includes mouse and keyboard controls.
+   */
```
nikolasrieble, line 45: This nicely shows up in the Storybook ArgsTable. As a reminder: We would like to have such comments above every single prop that is user facing, because then it is documented in the code and in storybook identically.

nikolasrieble:
This change strictly speaking is breaking - specifically if a user has set active to false or true - both which previously would simply have been ignored, but now is functional. 

Given that (a) this was already documented to be functional (wrongly) and (b) there is high customer demand for this to work https://github.com/recharts/recharts/issues/855, I believe we should merge this as a bug fix.

ckifer (approved):
Sweet - love the tests.

If we get any reports of breaking changes from this we just explain and give the simple solution of removing the isActive prop

Issue #3359, opened 2023-02-13
Animation of Line of Line chart in responsive container breaks after window resize with hide flag


agatsko-bachoo:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/area-chart-in-responsive-repro-2023-forked-31qhd5?file=/src/App.tsx)

### Steps to reproduce
1. Resize window to be smaller or bigger
2. Click on legend item once to hide it
3. Click on legend second time to show it

### What is expected?
Animation to be the same as initial render before resize

### What is actually happening?
The animation of the new line is broken, sometimes animating from bottom, sometimes with small pieces at the end 

| Environment | Info |
|---|---|
| Recharts | v2.4.1 |
| React | 17.0 |
| System | Windows |
| Browser | Chrome |

HHongSeungWoo:
@ckifer Is this behavior a bug, or is it intended?
If it's the same animation as the initial rendering, it would look like this.

[animation.webm](https://github.com/recharts/recharts/assets/34399997/16a5d111-dd32-4384-b8b2-6d40de7bcd72)

ckifer:
Hmm @HHongSeungWoo it seems I can't play that video format. 

Imo if a the something is hidden or shown shouldn't re-animate at all. But if it does it should do the whole animation from as it does on initial render - or at least the animation for the thing that was shown

HHongSeungWoo:
Hmm... I can identify the bug in CodeSandbox, but I can't reproduce it locally. For now, I'll skip this issue

ckifer:
I can reproduce it locally with the exact code in the sandbox

ckifer:
and it only shows itself when the chart is over a certain height it seems

ckifer:
<img width="1168" alt="image" src="https://github.com/recharts/recharts/assets/25180830/5430abbf-876e-4b33-99e6-c63b09970fa5">
In the animate function on Line log the current and total strokeDasharray, then try changing strokeDasharray manually

ckifer:
`getTotalLength` returns a too small of a number in certain circumstances here https://github.com/recharts/recharts/blob/27a47c1f3c5f1864854bf6e58811ada7cb1756a0/src/cartesian/Line.tsx#L183

ckifer:
```ts
[...]

PR #3946, opened 2023-11-09
fix: Line never checks for new totalLength on component update

ckifer:
## Description

`getTotalLength` only gets called on initial render in Line. When hiding and showing `Line` components in a specific chart the total length remains the same as it was before, when in fact it is now different (chart adjusts based on available space). In some cases this leads to the charts `strokeDasharray` not being correct and sometimes partially showing during animation. See videos

## Related Issue

fixes https://github.com/recharts/recharts/issues/3359

## Motivation and Context

Fix bug

## How Has This Been Tested?

See videos 

- add storybook entry for toggling `Line` components via `Legend` click

#### Before

https://github.com/recharts/recharts/assets/25180830/55e4baa3-d10e-4d44-bfa1-0703a6d75cae

#### After

https://github.com/recharts/recharts/assets/25180830/e81226ff-f12b-40ec-b79f-30c33afe41ec

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Line.tsx
+++ b/src/cartesian/Line.tsx
@@ -184,6 +184,17 @@ export class Line extends PureComponent<Props, State> {
     this.setState({ totalLength });
   }
 
+  componentDidUpdate(): void {
+    if (!this.props.isAnimationActive) {
+      return;
+    }
+
+    const totalLength = this.getTotalLength();
+    if (totalLength != null && totalLength !== this.state.totalLength) {
+      this.setState({ totalLength });
+    }
+  }
+
   static getDerivedStateFromProps(nextProps: Props, prevState: State): State {
     if (nextProps.animationId !== prevState.prevAnimationId) {
       return {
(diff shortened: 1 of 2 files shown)


ckifer:
@HHongSeungWoo

src/cartesian/Line.tsx
```diff
@@ -184,6 +184,17 @@ export class Line extends PureComponent<Props, State> {
     this.setState({ totalLength });
   }
 
+  componentDidUpdate(): void {
+    if (!this.props.isAnimationActive) {
+      return;
+    }
+
+    const totalLength = this.getTotalLength();
+    if (totalLength != null && totalLength !== this.state.totalLength) {
```
HHongSeungWoo, line 193: This part feels redundant. Since the null case does not exist and it's a primitive type, it seems fine to use `this.setState({ totalLength: this.getTotalLength() });`

akamfoad (approved):

PavelVanecek (approved):

src/cartesian/Line.tsx
HHongSeungWoo: The tooltip inherits PureComponent, it already performs shallow comparisons.
Primitive type references always remain the same, so it won't rerender, but not calling it explicitly also seems good

PR #3950, opened 2023-11-10
2 a11y violations for <Brush /> fixed

enriquetamames-cpi:
## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Brush.tsx
+++ b/src/cartesian/Brush.tsx
@@ -31,6 +31,8 @@ interface InternalBrushProps {
 interface BrushProps extends InternalBrushProps {
   className?: string;
 
+  ariaLabel?: string;
+
   height: number;
   travellerWidth?: number;
   traveller?: BrushTravellerType;
@@ -127,6 +129,7 @@ export class Brush extends PureComponent<Props, State> {
     padding: { top: 1, right: 1, bottom: 1, left: 1 },
     leaveTimeOut: 1000,
     alwaysShowText: false,
+    ariaLabel: 'chart brush',
   };
 
   constructor(props: Props) {
@@ -487,7 +490,7 @@ export class Brush extends PureComponent<Props, State> {
   }
 
   renderTravellerLayer(travellerX: number, id: BrushTravellerId) {
-    const { y, travellerWidth, height, traveller } = this.props;
+    const { y, travellerWidth, height, traveller, ariaLabel } = this.props;
     const x = Math.max(travellerX, this.props.x);
     const travellerProps = {
       ...filterProps(this.props),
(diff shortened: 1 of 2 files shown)


ckifer:
thanks for the contribution!

Imo we should fix these in a more global way... This addresses two small issues but there is a ton that remain and will probably solved with a different approach in the future.

Take a look at this PR https://github.com/recharts/recharts/pull/3884

enriquetamames-cpi:
@ckifer Absolutely there are tons of accessibility issues but these  solves at least 2 comign from `< Brush />`

src/cartesian/Brush.tsx
```diff
@@ -497,10 +500,14 @@ export class Brush extends PureComponent<Props, State> {
       height,
     };
 
+    const ariaValueNow = `ID: ${id}, Min value: ${data[startIndex].name}, Max value: ${data[endIndex].name}`;
```
ckifer, line 503: can you give an example of what this ends up being? Does this work for every type of chart that can possibly use brush? Any possibility of index access errors?

ckifer:
Looks like build is failing

enriquetamames-cpi:
`aria-valuenow` can only be a number, so I changed that so the build passes

src/cartesian/Brush.tsx
```diff
@@ -127,6 +129,7 @@ export class Brush extends PureComponent<Props, State> {
     padding: { top: 1, right: 1, bottom: 1, left: 1 },
     leaveTimeOut: 1000,
     alwaysShowText: false,
+    ariaLabel: 'chart brush',
```
ckifer, line 132: isn't this now misleading? The users `ariaLabel` won't have full control on what actually ends up in the label. It will be put into the "ID" of the `ariaLabelBrush`. We should have what is in `ariaLabelBrush` be the default value and allow the user to override the entire thing, rather than just a portion of it.

ckifer (changes requested):

storybook/stories/API/cartesian/Brush.stories.tsx
```diff
@@ -6,6 +6,10 @@ import { Brush, ResponsiveContainer, ComposedChart, Line } from '../../../../src
 import { getStoryArgsFromArgsTypesObject } from '../props/utils';
 
 const GeneralProps: Args = {
+  ariaLabel: {
+    description: 'The ARIA label of the brush.',
+    table: { type: { summary: 'string' }, category: 'General' },
```
ckifer, line 11: nit: we could add the default value here

ckifer (approved):
This is a good start I think. Issues will now start to arise that other components don't have first class label support but that is probably a good thing. Thanks

PR #3951, opened 2023-11-11
Split Tooltip into TooltipBoundingBox

PavelVanecek:
## Description

I think this is an opportunity to have two clearly defined components with split responsibility. One doing the position (BoundingBox) and one doing the content (Tooltip).

## How Has This Been Tested?

unit tests

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Tooltip.tsx
+++ b/src/component/Tooltip.tsx
@@ -2,15 +2,19 @@
  * @fileOverview Tooltip
  */
 import React, { PureComponent, CSSProperties, ReactNode, ReactElement, SVGProps } from 'react';
-import { translateStyle } from 'react-smooth';
-import { DefaultTooltipContent, ValueType, NameType, Payload, Props as DefaultProps } from './DefaultTooltipContent';
+import {
+  DefaultTooltipContent,
+  ValueType,
+  NameType,
+  Payload,
+  Props as ToltipContentProps,
+} from './DefaultTooltipContent';
+import { TooltipBoundingBox } from './TooltipBoundingBox';
 
 import { Global } from '../util/Global';
 import { UniqueOption, getUniqPayload } from '../util/payload/getUniqPayload';
-import { AllowInDimension, AnimationDuration, AnimationTiming, Coordinate } from '../util/types';
-import { getTooltipTranslate } from '../util/tooltip/translate';
+import { AllowInDimension, AnimationDuration, AnimationTiming, CartesianViewBox, Coordinate } from '../util/types';
 
-const EPS = 1;
 export type ContentType<TValue extends ValueType, TName extends NameType> =
   | ReactElement
   | ((props: TooltipProps<TValue, TName>) => ReactNode);
@@ -22,7 +26,7 @@ function defaultUniqBy<TValue extends ValueType, TName extends NameType>(entry: 
 function renderContent<TValue extends ValueType, TName extends NameType>(
   content: ContentType<TValue, TName>,
   props: TooltipProps<TValue, TName>,
-) {
+): ReactNode {
   if (React.isValidElement(content)) {
     return React.cloneElement(content, props);
   }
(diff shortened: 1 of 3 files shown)


src/component/Tooltip.tsx
```diff
@@ -33,30 +37,25 @@ function renderContent<TValue extends ValueType, TName extends NameType>(
   return <DefaultTooltipContent {...props} />;
 }
 
-export type TooltipProps<TValue extends ValueType, TName extends NameType> = DefaultProps<TValue, TName> & {
+export type TooltipProps<TValue extends ValueType, TName extends NameType> = ToltipContentProps<TValue, TName> & {
```
PavelVanecek, line 40: Changes to props:

1. "Sort lines alphabetically"
2. `viewBox` uses (identical) imported type instead of inline.

src/component/Tooltip.tsx
```diff
@@ -67,92 +66,25 @@ export class Tooltip<TValue extends ValueType, TName extends NameType> extends P
   static defaultProps = {
     active: false,
     allowEscapeViewBox: { x: false, y: false },
-    reverseDirection: { x: false, y: false },
-    offset: 10,
-    viewBox: { x: 0, y: 0, height: 0, width: 0 },
+    animationDuration: 400,
```
PavelVanecek, line 69: "Sort lines alphabetically"

src/component/TooltipBoundingBox.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  updateBBox() {
+    if (this.wrapperNode && this.wrapperNode.getBoundingClientRect) {
+      const box = this.wrapperNode.getBoundingClientRect();
+
+      if (
+        Math.abs(box.width - this.lastBoundingBox.width) > EPS ||
+        Math.abs(box.height - this.lastBoundingBox.height) > EPS
+      ) {
+        this.lastBoundingBox.width = box.width;
```
PavelVanecek, line 51: I think this instance assignment is a potential bug but I copy pasted it without a change.

PavelVanecek:
This will also make my PoC branches easier - there's tons of conflicts.

src/component/TooltipBoundingBox.tsx
nikolasrieble: If you think this is a bug, please do elaborate, and create a follow up issue. We can tackle this independently. I appreciate this you only copied it in this PR.

src/component/TooltipBoundingBox.tsx
```diff
@@ -0,0 +1,159 @@
...
+  viewBox: CartesianViewBox;
+  wrapperStyle: CSSProperties;
+};
+
+type State = {
+  dismissed: boolean;
+  dismissedAtCoordinate: Coordinate;
+};
+
+const EPS = 1;
```
nikolasrieble, line 28: Nit: If you know what EPS stands for, I would sure love to replace this with a proper name.

nikolasrieble (approved):
Nice cleanup, I like it. It further simplifies the Tooltip and marginally increases test coverage. 

Well done. ❤️ 🚀

src/component/TooltipBoundingBox.tsx
ckifer: I think @HHongSeungWoo recently changed modified this section?

PavelVanecek: @HHongSeungWoo confirmed in another thread that this is intentional: https://github.com/recharts/recharts/pull/3852#issuecomment-1806724245

I do not have any evidence that this is a bug. It's only an antipattern that can lead to bugs - this thing here is why `useState` and `useRef` exists.

PavelVanecek: (Reading `defaultProps` and cloning elements are also antipatterns 😬 )

src/component/TooltipBoundingBox.tsx
PavelVanecek: "Epsilon"

> denotes the smallest quantity like a term which is taken as the zero in some limit

src/component/TooltipBoundingBox.tsx
ckifer: This should probably be a function component then no? 😅

src/component/TooltipBoundingBox.tsx
HHongSeungWoo: This is no longer necessary. It seems to have been added due to the structure of obtaining the tooltip's viewbox, which could have caused an infinite loop

src/component/TooltipBoundingBox.tsx
HHongSeungWoo: While we have many anti-patterns, I think this one is similar to useMemo.
Due to `getTooltipTranslate`, this seems to be the best option for now.

Issue #1465, opened 2018-09-04
CartesianAxis on charts with large datasets causes performance issues with the default interval


lightsinthesky:
## Do you want to request a feature or report a bug?

Bug

## What is the current behavior?

With a large dataset (~10000 items) rendering on a simple line chart,  adding an XAxis with the default interval (preserveEnd) results in a chart taking several seconds to render (depending on cpu). Drilling into the perf you can see that getTicksEnd is the blocking function (more specifically getStringSize which is forcing reflow)

<img width="715" alt="screen shot 2018-09-04 at 2 47 39 pm" src="https://user-images.githubusercontent.com/8852073/45048148-7f564580-b051-11e8-8618-10ff447cdd06.png">

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: http://jsfiddle.net/ndLhnegs/).

goto : https://jsfiddle.net/dbwnv9fp/15/ and "run" then add interval={1000} to the XAxis and notice the speed increase.

## What is the expected behavior?

adding an xaxis should not increase the render time by 4-5x for large datasets.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

recharts 1.1.0 - chrome latest (osx and windows).

weiyuxuan:
Try adding a specific key on the XAxis component, which can prevent XAxis to reRender.

m3h0w:
@lightsinthesky thank you so much for reporting this - I've been struggling so much with the performance I almost changed the library but maybe with setting the interval somehow based on the viewport I can hack around this issue.

@xile611 why is this issue closed?

Eric24:
> Try adding a specific key on the XAxis component, which can prevent XAxis to reRender.

If you mean setting the `dataKey` property, that is something we always do. I has no effect. The root issue is the getStringSize() function.

weiyuxuan:
> > Try adding a specific key on the XAxis component, which can prevent XAxis to reRender.
> 
> If you mean setting the `dataKey` property, that is something we always do. I has no effect. The root issue is the getStringSize() function.

I mean the `key` property  instead of the `dataKey` property.

Eric24:
The `key` property? I don't see that in the documentation. What does the `key` property do?

PR #3953, opened 2023-11-12
Reduce calls to `getStringSize` when calculating visible ticks (#2589)

bernardobelchior:
## Description

Reduce the number of times `getStringSize` is called when calculating which ticks are rendered.
This optimization is applied to `getTicksEnd, `getTicksStart` and `getEquidistantTicks`, but the latter is unlikely to benefit from it since the algorithm does not iterate over the candidate ticks sequentially. 

I have created [this repository](https://github.com/bernardobelchior/recharts-repro), based on the JSFiddle available in [this issue](https://github.com/recharts/recharts/issues/1465#issue-356913907) to test the changes. It generates a chart with 10000 random data points.

Based on Recharts' master branch (pointing to commit `8cdfab7facc4d1d107c11a1fc3d4a26e716dddca` at the time of writing), the Chrome DevTools' Performance chart is as follows:

image

You can see that 95% time is taken by `getStringSize()` and descendants, which account for 3.2 seconds on my M1 Macbook Pro from 2020. In fact, I counted and `getStringSize()` is being called around 40k times to render the chart.

Since we know that `getStringSize` is expensive, whatever we can do to reduce the number of times it's called will likely yield a performance improvement. As such, this PR is comprised of two changes:
1. Call `getTickSize` in `getTicksEnd`, `getTicksStart` and `getEquidistantTicks` only when absolutely necessary;
1. Update the `isVisible` calculation to check if the tick won't be visible even before we take size into account: this way we can skip calculating the string size if we already know that the tick won't be visible.

After this change, `getStringSize` still takes up most of the time (88%), but the absolute value is now much smaller (1.8 seconds):

image

`getStringSize` is now called around 18k times, which means there's still a lot of room for future improvements.

## Related Issue

#2589 
#1465

## Motivation and Context

Explained above

## How Has This Been Tested?

Added some unit tests. There were none, so it's hard to see if there's no regression, but I ran the new tests against the `isVisible` function before and after these changes and the results were the same.

If there's any edge case you'd like me to test, please let me know 😄 

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/getEquidistantTicks.ts
+++ b/src/cartesian/getEquidistantTicks.ts
@@ -31,11 +31,19 @@ export function getEquidistantTicks(
     }
 
     // Check if the element collides with the next element
-    const size = getTickSize(entry, index);
+    const i = index;
+    let size: number | undefined;
+    const getSize = () => {
+      if (size === undefined) {
+        size = getTickSize(entry, i);
+      }
+
+      return size;
+    };
 
     const tickCoord = entry.coordinate;
     // We will always show the first tick.
-    const isShow = index === 0 || isVisible(sign, tickCoord, () => size, start, end);
+    const isShow = index === 0 || isVisible(sign, tickCoord, getSize, start, end);
 
     if (!isShow) {
       // Start all over with a larger stepsize
@@ -46,7 +54,7 @@ export function getEquidistantTicks(
 
     if (isShow) {
       // If it can be shown, update the start
-      start = tickCoord + sign * (size / 2 + minTickGap);
+      start = tickCoord + sign * (getSize() / 2 + minTickGap);
       index += stepsize;
     }
   }
--- a/src/cartesian/getEquidistantTicks.ts
+++ b/src/cartesian/getEquidistantTicks.ts
@@ -35,7 +35,7 @@ export function getEquidistantTicks(
 
     const tickCoord = entry.coordinate;
     // We will always show the first tick.
-    const isShow = index === 0 || isVisible(sign, tickCoord, size, start, end);
+    const isShow = index === 0 || isVisible(sign, tickCoord, () => size, start, end);
 
     if (!isShow) {
       // Start all over with a larger stepsize
--- a/src/cartesian/getTicks.ts
+++ b/src/cartesian/getTicks.ts
@@ -24,9 +24,11 @@ function getTicksEnd(
 
   for (let i = len - 1; i >= 0; i--) {
     let entry = result[i];
-    const size = getTickSize(entry, i);
+    let size: number | undefined;
+    const getSize = () => size ?? getTickSize(entry, i);
+
     if (i === len - 1) {
-      const gap = sign * (entry.coordinate + (sign * size) / 2 - end);
+      const gap = sign * (entry.coordinate + (sign * getSize()) / 2 - end);
       result[i] = entry = {
         ...entry,
         tickCoord: gap > 0 ? entry.coordinate - gap * sign : entry.coordinate,
@@ -35,10 +37,10 @@ function getTicksEnd(
       result[i] = entry = { ...entry, tickCoord: entry.coordinate };
     }
 
-    const isShow = isVisible(sign, entry.tickCoord, size, start, end);
+    const isShow = isVisible(sign, entry.tickCoord, getSize, start, end);
 
     if (isShow) {
-      end = entry.tickCoord - sign * (size / 2 + minTickGap);
+      end = entry.tickCoord - sign * (getSize() / 2 + minTickGap);
       result[i] = { ...entry, isShow: true };
     }
   }
@@ -69,7 +71,7 @@ function getTicksStart(
       tickCoord: tailGap > 0 ? tail.coordinate - tailGap * sign : tail.coordinate,
     };
 
-    const isTailShow = isVisible(sign, tail.tickCoord, tailSize, start, end);
+    const isTailShow = isVisible(sign, tail.tickCoord, () => tailSize, start, end);
 
     if (isTailShow) {
       end = tail.tickCoord - sign * (tailSize / 2 + minTickGap);
@@ -80,10 +82,11 @@ function getTicksStart(
   const count = preserveEnd ? len - 1 : len;
   for (let i = 0; i < count; i++) {
     let entry = result[i];
-    const size = getTickSize(entry, i);
+    let size: number | undefined;
+    const getSize = () => size ?? getTickSize(entry, i);
 
     if (i === 0) {
-      const gap = sign * (entry.coordinate - (sign * size) / 2 - start);
+      const gap = sign * (entry.coordinate - (sign * getSize()) / 2 - start);
       result[i] = entry = {
         ...entry,
         tickCoord: gap < 0 ? entry.coordinate - gap * sign : entry.coordinate,
@@ -92,10 +95,10 @@ function getTicksStart(
       result[i] = entry = { ...entry, tickCoord: entry.coordinate };
     }
 
-    const isShow = isVisible(sign, entry.tickCoord, size, start, end);
+    const isShow = isVisible(sign, entry.tickCoord, getSize, start, end);
 
     if (isShow) {
-      start = entry.tickCoord + sign * (size / 2 + minTickGap);
+      start = entry.tickCoord + sign * (getSize() / 2 + minTickGap);
       result[i] = { ...entry, isShow: true };
     }
   }
--- a/src/cartesian/getTicks.ts
+++ b/src/cartesian/getTicks.ts
@@ -25,7 +25,13 @@ function getTicksEnd(
   for (let i = len - 1; i >= 0; i--) {
     let entry = result[i];
     let size: number | undefined;
-    const getSize = () => size ?? getTickSize(entry, i);
+    const getSize = () => {
+      if (size === undefined) {
+        size = getTickSize(entry, i);
+      }
+
+      return size;
+    };
 
     if (i === len - 1) {
       const gap = sign * (entry.coordinate + (sign * getSize()) / 2 - end);
@@ -83,7 +89,13 @@ function getTicksStart(
   for (let i = 0; i < count; i++) {
     let entry = result[i];
     let size: number | undefined;
-    const getSize = () => size ?? getTickSize(entry, i);
+    const getSize = () => {
+      if (size === undefined) {
+        size = getTickSize(entry, i);
+      }
+
+      return size;
+    };
 
     if (i === 0) {
       const gap = sign * (entry.coordinate - (sign * getSize()) / 2 - start);
--- a/src/util/TickUtils.ts
+++ b/src/util/TickUtils.ts
@@ -30,6 +30,8 @@ export function isVisible(
   start: number,
   end: number,
 ): boolean {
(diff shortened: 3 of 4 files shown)


ckifer:
Thanks for the analysis!

Yeah.. unfortunately the long term solution here requires quite a bit more work with a broader swath of changes. Some of the refactoring we're doing should enable us to reduce these calls quite a bit more I think!

bernardobelchior:
Thanks you, @ckifer! 
I have a couple question regarding testing:
1. Does Recharts have any automated visual tests? Or should I go through Storybook stories to check whether there were any regressions? 
2. Do you think there's any additional unit test I should add in this PR? We had no tests for `isVisible`, so we might want to add some more tests there

Thanks!

ckifer:
1. We run visual regression with chromatic - it isn't perfect but it will cover some things for sure  
2. I haven't looked at the PR yet but will let you know

bernardobelchior:
Also, @ckifer it seems like `getStringSize`'s cache has a limit of 2k values and it completely clears it when it's over that. 

There's several improvements we can think about:
1. Allow users to customize the number of entries in the cache: it'll affect memory usage but might be worth the tradeoff? The downside is that users shouldn't need to think about this when using a charting library
2. Evict older entries from the cache instead of clearing it altogether;

nikolasrieble:
> Also, @ckifer it seems like `getStringSize`'s cache has a limit of 2k values and it completely clears it when it's over that.
> 
> There's several improvements we can think about:
> 
> 1. Allow users to customize the number of entries in the cache: it'll affect memory usage but might be worth the tradeoff? The downside is that users shouldn't need to think about this when using a charting library
> 2. Evict older entries from the cache instead of clearing it altogether;


re: 1

Given that recharts is known specifically for its customisability, this could make sense. Really though the customer would not necessarily unterstand how the big the cache should be, because the amount of calls are a recharts implementation detail. 

What would be the options? Rather, the cache size should be a function of the amount of ticks.
In which cases are we calling it? 
For all ticks on all axes, right? We could have a cache option where the `cacheSize = axisCount * dataLength`. Would that make sense? 

Not sure it would be a great experience to offer the user to set a number. 

re: 2 
Seems a good idea to try out. 
I wonder if it helps though, simply because for example, for the tick positioning, we simply measure every tick. If we have 1001 data points, but the cache is of size 1000, then it makes no difference whether you evict all, or the latest one. In both cases you would never have the string in question in the cache. 

That being said, this is tackling an important point, thanks a lot for your help here! ❤️

src/util/TickUtils.ts
```diff
@@ -23,7 +23,19 @@ export function getTickBoundaries(viewBox: CartesianViewBox, sign: number, sizeK
...
+export function isVisible(
+  sign: number,
+  tickPosition: number,
+  getSize: () => number,
+  start: number,
+  end: number,
+): boolean {
+  if (sign * tickPosition < sign * start || sign * tickPosition > sign * end) {
+    return false;
+  }
```
nikolasrieble, line 35: A very nice idea ❤️ 
Could you please add a small comment here for the motivation. It will help the next person working on this piece of the code.

src/cartesian/getTicks.ts
```diff
@@ -24,9 +24,11 @@ function getTicksEnd(
 
   for (let i = len - 1; i >= 0; i--) {
     let entry = result[i];
-    const size = getTickSize(entry, i);
+    let size: number | undefined;
+    const getSize = () => size ?? getTickSize(entry, i);
```
nikolasrieble, line 28: This, I do not understand. When would this ever return size? When would size not be undefined?

src/cartesian/getTicks.ts
```diff
@@ -69,7 +71,7 @@ function getTicksStart(
       tickCoord: tailGap > 0 ? tail.coordinate - tailGap * sign : tail.coordinate,
     };
 
-    const isTailShow = isVisible(sign, tail.tickCoord, tailSize, start, end);
+    const isTailShow = isVisible(sign, tail.tickCoord, () => tailSize, start, end);
```
nikolasrieble, line 74: Does this even work? Is getTickSize not already evaluated in line 67? 

```
const tailSize = getTickSize(tail, len - 1);
```

Would it not make more sense to define: 
```
const getTailSize = () => getTickSize(tail, len - 1);
const isTailShow = isVisible(sign, tail.tickCoord, getTailSize, start, end);
```

If your solution works, then there is something I am missing. What did I fail to understand? 🤔

nikolasrieble (commented):
I have a few question about some implementation details, but the idea is both simple as well as clear. Love the improvement. Lets figure out the details and land this. 

Thank you for this great improvement.

src/cartesian/getTicks.ts
bernardobelchior: In this case it doesn't really matter. 

In line 68 we're using the `tailSize`, so we need to compute it anyway. I don't think we can optimize it here.

src/util/TickUtils.ts
bernardobelchior: Sure! Let me know if the comment I added works 👍

src/cartesian/getTicks.ts
bernardobelchior: Sorry, I made a mistake here. It was a leftover from some refactors I made.
Thanks, updated 👍

src/cartesian/getEquidistantTicks.ts
```diff
@@ -35,7 +35,7 @@ export function getEquidistantTicks(
 
     const tickCoord = entry.coordinate;
     // We will always show the first tick.
-    const isShow = index === 0 || isVisible(sign, tickCoord, size, start, end);
+    const isShow = index === 0 || isVisible(sign, tickCoord, () => size, start, end);
```
nikolasrieble, line 38: Is the tick size not already computed in line 34? 
If I understand correctly, this should be 

```
    // Check if the element collides with the next element
    const getSize = () => getTickSize(entry, index);

    const tickCoord = entry.coordinate;
    // We will always show the first tick.
    const isShow = index === 0 || isVisible(sign, tickCoord, getSize, start, end);
```

Now we would not call getTickSize directly.

src/cartesian/getTicks.ts
```diff
@@ -24,9 +24,17 @@ function getTicksEnd(
...
     let entry = result[i];
-    const size = getTickSize(entry, i);
+    let size: number | undefined;
+    const getSize = () => {
+      if (size === undefined) {
+        size = getTickSize(entry, i);
+      }
+
+      return size;
+    };
```
nikolasrieble, line 34:
```suggestion
    const getSize = () => getTickSize(entry, i);
```

My question remains: In which case would size already be set? Why is this not just

bernardobelchior:
> > Also, @ckifer it seems like `getStringSize`'s cache has a limit of 2k values and it completely clears it when it's over that.
> > There's several improvements we can think about:
> > 
> > 1. Allow users to customize the number of entries in the cache: it'll affect memory usage but might be worth the tradeoff? The downside is that users shouldn't need to think about this when using a charting library
> > 2. Evict older entries from the cache instead of clearing it altogether;
> 
> re: 1
> 
> Given that recharts is known specifically for its customisability, this could make sense. Really though the customer would not necessarily unterstand how the big the cache should be, because the amount of calls are a recharts implementation detail.
> 
> What would be the options? Rather, the cache size should be a function of the amount of ticks. In which cases are we calling it? For all ticks on all axes, right? We could have a cache option where the `cacheSize = axisCount * dataLength`. Would that make sense?

That makes sense if charts are static. However, if we add 1 more data point (or change a data point), we may evict the whole cache because we'd be over the cache size. This might work, though, if we implement cache eviction by the oldest key. 
In that case, in case a chart re-renders because of new data points, we'd only evict a subset of the cache, which would reduce cache misses.
 
> Not sure it would be a great experience to offer the user to set a number.

Yeah, I'm not sure either 😅 

> re: 2 Seems a good idea to try out. I wonder if it helps though, simply because for example, for the tick positioning, we simply measure every tick. If we have 1001 data points, but the cache is of size 1000, then it makes no difference whether you evict all, or the latest one. In both cases you would never have the string in question in the cache.

That's right, but you would still hit the cache for all the potential ticks that didn't change. If we evict the whole cache, we know we'll have 100% cache misses; if we evict the oldest key, there's a chance we might have some cache hits. 

Also, another problem is that the cache is global for Recharts. If we have two charts on the same page and they both have 2k potential ticks, they'll share the same cache and cause cache misses. It might make sense for the cache to be scoped to the chart. 

> That being said, this is tackling an important point, thanks a lot for your help here! ❤️

Thanks for the kind words 😄

nikolasrieble:
All great points! Cache being shared across charts is a good reason to be able to set it separately. Then all my suggestions also do not make sense. 😓 

Also, your point with non static charts makes a lot of sense. Would you like to create a follow up issue, and implement cache eviction in the FIFO strategy?

src/cartesian/getEquidistantTicks.ts
bernardobelchior: Doing so triggers the following lint error:

image

But I think I managed to find a workaround. Pushing a new commit 😄

bernardobelchior:
> Also, your point with non static charts makes a lot of sense. Would you like to create a follow up issue, and implement cache eviction in the FIFO strategy?

[Created an issue](https://github.com/recharts/recharts/issues/3955), but I don't want to commit to implementing since I don't know when I'll have some free time to do that.

src/cartesian/getTicks.ts
nikolasrieble: It does, I was not aware. Would it make sense to add this as a comment?

src/cartesian/getEquidistantTicks.ts
```diff
@@ -31,11 +31,19 @@ export function getEquidistantTicks(
...
+        size = getTickSize(entry, i);
+      }
+
+      return size;
+    };
 
     const tickCoord = entry.coordinate;
     // We will always show the first tick.
-    const isShow = index === 0 || isVisible(sign, tickCoord, size, start, end);
+    const isShow = index === 0 || isVisible(sign, tickCoord, getSize, start, end);
```
nikolasrieble, line 46: We are now redefining this in three places. I wonder: Could we instead extract this combination of getSize with isVisible into a new helper function?

nikolasrieble (approved):
I could not find any issue while manually testing in storybook. This is a great first step to improvement. Lets ship! 🚀

src/cartesian/getTicks.ts
bernardobelchior: I think we can see that from the code; it's pretty local to the lines 89-115. In which line do you think a comment would fit?

src/cartesian/getEquidistantTicks.ts
bernardobelchior: I don't think we could because there might be other places where `getSize` is used. 
For example, in `getTicksEnd` we use it in lines 37, 46 and 49. It isn't just used for `isVisible`, so it's hard to abstract that

src/cartesian/getTicks.ts
nikolasrieble: Totally fair. Not a requirement.

ckifer:
Didn't have a chance to look at this but great conversation here 🚀🚀🚀

Issue #3937, opened 2023-11-06
Radar chart with a PolarAngleAxis with type "number" does not properly align points with the ticks


tran-simon:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-896dvj?file=/src/index.js)

<img width="347" alt="image" src="https://github.com/recharts/recharts/assets/8755930/9ab10a4b-75c9-4c46-b416-490e9c0718ce">

### Steps to reproduce
Have a Radar Chart that uses a PolarAngleAxis with type="number". The data is an array of objects containing the angle and the value

### What is expected?
Points on the chart should be aligned with the axis ticks at the correct angle

### What is actually happening?
The whole polygon seems to be rotate about 45 degrees counterclockwise

| Environment | Info |
|---|---|
| Recharts | v2.9.2 |
| React | 18.2.0 |
| System | codesandbox |
| Browser | Chrome |

ckifer:
Going to close this in favor of https://github.com/recharts/recharts/issues/2969 since its pre-existing. Example will still be linked from there

ckifer:
Should be fixed in next minor

Issue #2969, opened 2022-09-15
RadarChart for type="number" data rendering wrongly


vileanco:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/hardcore-danny-odzium)

### Steps to reproduce
See CodeSandbox.

```
import {
  PolarAngleAxis,
  PolarGrid,
  PolarRadiusAxis,
  Radar,
  RadarChart
} from "recharts";

const data = [
  { angle: 0, r: 1 },
  { angle: 90, r: 2 },
  { angle: 180, r: 3 },
  { angle: 270, r: 4 }
];

export default function App() {
  return (
    <RadarChart data={data} width={360} height={360}>
      <PolarGrid gridType="circle" />

      <PolarRadiusAxis type="number" dataKey="r" />

      <PolarAngleAxis
        dataKey="angle"
        axisLineType="circle"
        type="number"
        domain={[0, 360]}
      />

      <Radar
        type="number"
        name="r"
        dataKey="r"
        fillOpacity={0}
        stroke={"#000"}
      />
    </RadarChart>
  );
}
```

### What is expected?
The drawn graph should match the data.

### What is actually happening?
The angle degrees do not match the graph. Eg. angle 90 is drawn with radius 3 even though it should be drawn with radius 2.

| Environment | Info |
|---|---|
| Recharts | v2.1.6 |
| React | 18.2.0 |
| System | macOS Monterey 12.3.1 |
| Browser | 105.0.5195.102 |

---
Trying to use RadarChart for non-categorical data for polar coordinate graphs.

tran-simon:
I have the same issue https://github.com/recharts/recharts/issues/3937

It seems like the whole graph is rotated counter clockwise

akamfoad:
Hi @vileanco @tran-simon 

Looks like removing `type="number"` on `PolarAngleAxis` fixes the issue, here I forked and applied this change, the Chart looks correct now, let us know if this fixes your issue

https://codesandbox.io/s/solitary-wave-3jqdwv?file=/src/App.js:401-482

ckifer:
+1 - no need to specify number type on these charts. Ideally you wouldn't be able to, but until then you can just remove it

tran-simon:
[...]

PR #3960, opened 2023-11-14
fix(RadarChart): angleAxis type='number' dose not render correctly

HHongSeungWoo:
## Description

I took a brief look at the RadarChart, so I'm not sure if this is the good solution, but it seems to address the reported issue

## Related Issue

#2969
#3937

## Types of changes

- [?] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/src/polar/Radar.tsx
+++ b/src/polar/Radar.tsx
@@ -110,11 +110,12 @@ export class Radar extends PureComponent<Props, State> {
     const { cx, cy } = angleAxis;
     let isRange = false;
     const points: RadarPoint[] = [];
+    const angleBandSize = angleAxis.type !== 'number' ? bandSize ?? 0 : 0;
 
     displayedData.forEach((entry, i) => {
       const name = getValueByDataKey(entry, angleAxis.dataKey, i);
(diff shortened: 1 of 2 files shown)


src/polar/Radar.tsx
```diff
@@ -110,11 +110,12 @@ export class Radar extends PureComponent<Props, State> {
     const { cx, cy } = angleAxis;
     let isRange = false;
     const points: RadarPoint[] = [];
+    const angleBandSize = angleAxis.type !== 'number' ? bandSize ?? 0 : 0;
```
ckifer, line 113: what path does this go through when axis type is number
https://github.com/recharts/recharts/blob/master/src/util/ChartUtils.ts#L1271

Is there anything we can do in that function to guard against this?

HHongSeungWoo: I have modified the Radar to minimize side effects,
It seems appropriate to make the changes at this line

ckifer (approved):
Pulled it down, this fixes the bug without changing current functionality otherwise 🚢

Issue #2948, opened 2022-08-23
"0" Displayed in PieChart Tooltip After Resizing


joncursi:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/eloquent-sound-njrnz8?file=/src/App.js)

### Steps to reproduce
1. Construct a pie chart that only has 1 item in the data object.
2. Observe that the tooltip appears to render normally on initial hover.
3. Resize the browser viewport
4. Observe that the tooltip now renders a "0" inside of it on all subsequent hovers

### What is expected?
No "0" rendered inside the tooltip after resizing

### What is actually happening?
A "0" is rendered inside the tooltip after resizing

| Environment | Info |
|---|---|
| Recharts | v2.1.6 |
| React | 18.2.0 |
| System | macOS |
| Browser | Chrome latest |

---
I have included a link to a minimal reproduction on codesandbox. Once you load it up, all you need to do is resize the viewport to observe the "0" appearing inside the tooltip.

joncursi:
Demonstration:

![2022-08-23 18 20 51](https://user-images.githubusercontent.com/4974609/186276420-cf3b067a-d5e0-4eb8-8981-92c7d0b1db03.gif)

PR #3967, opened 2023-11-16
fix(tooltip): The getTooltipData in the pie chart is returning incorrect data

HHongSeungWoo:
## Description

The behavior of always returning 0 when there is only one tick resulted in `calculateActiveTickIndex` returning inconsistent values.

## Related Issue

#2948

## Types of changes

## Checklist:


src/util/ChartUtils.ts
```diff
@@ -100,9 +100,8 @@ export const calculateActiveTickIndex = (
   let index = -1;
   const len = ticks?.length ?? 0;
 
-  // if there are 1 or less ticks ticks then the active tick is at index 0
-  if (len <= 1) {
-    return 0;
+  if (len === 0) {
+    return index;
```
ckifer, line 104: this will return `-1` now yeah? How is that better than 0?

HHongSeungWoo: I removed what seemed to be unnecessary code. 
I pushed the code again because there seemed to be an issue with the previous modifications

src/util/ChartUtils.ts
```diff
@@ -98,20 +98,15 @@ export const calculateActiveTickIndex = (
...
 ) => {
   let index = -1;
-  const len = ticks?.length ?? 0;
-
-  // if there are 1 or less ticks ticks then the active tick is at index 0
-  if (len <= 1) {
-    return 0;
-  }
 
   if (axis && axis.axisType === 'angleAxis' && Math.abs(Math.abs(axis.range[1] - axis.range[0]) - 360) <= 1e-6) {
```
HHongSeungWoo, line 102: It for the angleAxis, so i isolated it to make it easier to split the code later on

ckifer (changes requested):
Logic looks good to me from what I can tell - I won't really feel confident in this change without tests though. At least some simple cases for each path if they aren't already there

HHongSeungWoo:
I'll take a closer look after work :smile:

HHongSeungWoo:
Oh God, is this what hell is like? Other issues keep popping up :sob:

ckifer:
Lol take your time - no need to rush on it. Like I said, the logic looks good, but there is so much reliance on tiny things in the code that we can't have any confidence when we make changes like this. We'll get there...

HHongSeungWoo:
@ckifer Is getTooltipData necessary for the pie chart? It seems useless and it’s causing a bug.

https://github.com/recharts/recharts/blob/b6bffe4eea6e168692ddfd4d2133980fe732ea7e/src/chart/generateCategoricalChart.tsx#L264

ckifer:
Not sure @HHongSeungWoo - I didn't write this originally we're in the same place about knowing what will break what.

Does it use getToolTipData? If it does, what happens when it is removed? What happened when different settings are tried on Pie?

Removal may fix a bug, it also may cause more bugs. We need to care it doesn't.

ckifer:
This issue https://github.com/recharts/recharts/issues/3971 is from the latest version, wonder if it's an easier fix

HHongSeungWoo:
I will remove getToolTipData from the pie chart and commit it with some examples for review.

i'll fix that issue right now

HHongSeungWoo:
@ckifer In the pie chart, there is an issue with the `calculateActiveTickIndex` not finding the correct activeIndex.
While I want to address the problem, it is expected to breaking changes.
For now, I have resolved #2948 and made adjustments to avoid returning incorrect data.

src/chart/generateCategoricalChart.tsx
```diff
@@ -255,13 +255,25 @@ const getTooltipContent = (
...
  */
-const getTooltipData = (state: CategoricalChartState, chartData: any[], layout: LayoutType, rangeObj?: any) => {
+const getTooltipData = (
+  chartName: string,
+  state: CategoricalChartState,
+  chartData: any[],
+  layout: LayoutType,
+  rangeObj?: any,
+): object => {
+  // PieChart is temporarily returning null because the 'calculateActiveTickIndex' is returning an incorrect activeIndex
```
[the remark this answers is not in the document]
HHongSeungWoo: One concern with this modification is that it no longer returns payload and index in the mouse handler of PieChart. While this fixes a part that often returns incorrect data and causes bugs, it could be a breaking change for users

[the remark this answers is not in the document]
ckifer: Thats certainly a breaking change that will be noticed, I don't think we'll be able to fix it for now if thats the only solution

Issue #3981, opened 2023-11-20
Tooltip rendering crashes when resizing the chart "Cannot read properties of undefined (reading 'payload')"


tran-simon:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/new-voice-tgydrx?file=/src/App.js)

### Steps to reproduce
1. Hover the chart to show the tooltip
2. Press the button
3. Resize the window (to show/hide the scrollbar)
4. Repeat 3 times

The app crashes sometimes with "Cannot read properties of undefined (reading 'payload')" when the window is resized.

The crash is not consistent, so you should reload and try multiple times to reproduce

### What is expected?
The tooltip should appear correctly

### What is actually happening?
The whole app crashes (null pointer exception)

| Environment | Info |
|---|---|
| Recharts | v2.10.1 |
| React | 18.2.0 |
| System | codesandbox |
| Browser | Chrome |

---
This is due to a mistake in `getItemByXY` in `generateCategoricalChart.tsx`. 
https://github.com/recharts/recharts/blob/0d5326a8cdcad34a3b9d7c0644abbb645ddffe87/src/chart/generateCategoricalChart.tsx#L2250-L2252

https://github.com/recharts/recharts/blob/7fb227dae542c3d3093506e6d80a2c2c366f9a26/src/util/ActiveShapeUtils.tsx#L142-L152

`isFunnel`, `isPie` and `isScatter` do not check if `activeItem` is defined, so it might cast it as a defined `FunnelItem`/`PieItem`/`ScatterItem` even if `activeItem` is undefined

ckifer:
Thanks for the fix. I can't reproduce at all, I wonder why its so random

tran-simon:
@ckifer yeah it's really hard to reproduce, I'll keep looking for a way to reproduce it consistently

But I still think that part of the code is smelly, `activeItem` is typed as optional but the `isFunnel`/`is...` functions cast it as defined without checking if it actually is

ckifer:
makes sense

ckifer:
fixed merged and released in 2.10.2 - please comment back if this persists

PR #3982, opened 2023-11-20
Fix tooltip rendering crash when activeItem is undefined

tran-simon:
## Description

Sometimes, the tooltip rendering would crash when the window has been resized

> TypeError: Cannot read properties of undefined (reading 'payload')

This is due to a mistake in `getItemByXY` in `generateCategoricalChart.tsx`. 
https://github.com/recharts/recharts/blob/0d5326a8cdcad34a3b9d7c0644abbb645ddffe87/src/chart/generateCategoricalChart.tsx#L2250-L2252

https://github.com/recharts/recharts/blob/7fb227dae542c3d3093506e6d80a2c2c366f9a26/src/util/ActiveShapeUtils.tsx#L142-L152

`isFunnel`, `isPie` and `isScatter` do not check if `activeItem` is defined, so it might cast it as a defined `FunnelItem`/`PieItem`/`ScatterItem` even if `activeItem` is undefined

I've added a nullcheck for activeItem

## Related Issue

https://github.com/recharts/recharts/issues/3981

## Motivation and Context

This fixes a crash

## How Has This Been Tested?

I manually checked that the crash doesn't happen anymore

image

## Types of changes

## Checklist:

### DIFF
--- a/src/util/ActiveShapeUtils.tsx
+++ b/src/util/ActiveShapeUtils.tsx
@@ -140,15 +140,15 @@ type GetActiveShapeIndexForTooltip = {
 };
 
 export function isFunnel(graphicalItem: GraphicalItem, _item: unknown): _item is FunnelItem {
-  return 'trapezoids' in graphicalItem.props;
+  return _item != null && 'trapezoids' in graphicalItem.props;
 }
 
 export function isPie(graphicalItem: GraphicalItem, _item: unknown): _item is PieItem {
-  return 'sectors' in graphicalItem.props;
+  return _item != null && 'sectors' in graphicalItem.props;
 }
 
 export function isScatter(graphicalItem: GraphicalItem, _item: unknown): _item is ScatterItem {
-  return 'points' in graphicalItem.props;
+  return _item != null && 'points' in graphicalItem.props;
 }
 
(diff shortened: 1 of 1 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -2247,9 +2247,10 @@ export const generateCategoricalChart = ({
               return { graphicalItem, payload: activeBarItem };
             }
           } else if (
-            isFunnel(graphicalItem, activeItem) ||
-            isPie(graphicalItem, activeItem) ||
-            isScatter(graphicalItem, activeItem)
+            activeItem != null &&
+            (isFunnel(graphicalItem, activeItem) ||
+              isPie(graphicalItem, activeItem) ||
+              isScatter(graphicalItem, activeItem))
```
ckifer, line 2253: instead of doing it this way could we add the check in each `is..` function?

That way this is never an issue elsewhere that those typeguards are used.

so for example
```ts
 export function isScatter(graphicalItem: GraphicalItem, _item: unknown): _item is ScatterItem { 
   return Boolean(_item) && 'points' in graphicalItem.props; 
 } 
```

ckifer (changes requested):

Issue #3987, opened 2023-11-21
Tooltips aren't showing up if the top-level data prop is an empty array


charpeni:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link

[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/998w87?file=%2Fsrc%2FApp.tsx%3A21%2C1-22%2C1)

### Steps to reproduce

A regression was introduced in [`2.9.3`](https://github.com/recharts/recharts/compare/v2.9.2...v2.9.3), where tooltips aren't showing up if the top-level `data` is an empty array.

See the reproduction link above.

The following works perfectly (meaning that tooltips are showing up) on `2.9.2`, but stopped working starting from `2.9.3`:

```tsx
<ComposedChart data={[]}>
  [...]
  <Tooltip />
  <Bar data={pageData} dataKey="uv" />
</ComposedChart>
```

### What is expected?

[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/recharts-tooltip-issue-repro-2-9-0-np2qcc?file=%2Fsrc%2FApp.tsx%3A21%2C1-22%2C1)

<img width="634" alt="Screenshot 2023-11-21 at 1 53 07 PM" src="https://github.com/recharts/recharts/assets/7189823/3ac235c0-c812-49cd-a324-0257bf843bcb">

I would expect tooltips to appear even though the top-level `data` is empty, like it used to be the case previously.

### What is actually happening?

See the reproduction link above.

<img width="630" alt="Screenshot 2023-11-21 at 1 53 41 PM" src="https://github.com/recharts/recharts/assets/7189823/aa19e870-7ad5-45a1-a5d3-44bfb099c103">

When upgrading from `2.9.2` to `2.9.3` (or newer), tooltips aren't showing up anymore.

I investigated a little bit, and the reason is that `payload` contained within the tooltip is now an empty array and doesn't contain other data keys.

This is only happening with an empty array:

```tsx
// Tooltips aren't showing up
<ComposedChart data={[]} />

// OK
<ComposedChart />
<ComposedChart data={['keep me']} />
```

As a workaround, we could omit `data` instead of passing an empty array, but in that case, it will be leading to another issue: https://github.com/recharts/recharts/issues/3985.

[...]

PR #3988, opened 2023-11-22
fix: dataEndIndex can be -1

HHongSeungWoo:
## Description

As dataEndIndex is -1 in an empty data array, it causes issues with #3987.
It doesn't seem to be intentional

## Related Issue

#3987

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -647,11 +647,20 @@ const tooltipTicksGenerator = (axisMap: AxisMap) => {
 const createDefaultState = (props: CategoricalChartProps): CategoricalChartState => {
   const { children, defaultShowTooltip } = props;
   const brushItem = findChildByType(children, Brush);
-  const startIndex = (brushItem && brushItem.props && brushItem.props.startIndex) || 0;
-  const endIndex =
-    brushItem?.props?.endIndex !== undefined
-      ? brushItem?.props?.endIndex
-      : (props.data && props.data.length !== 0 && props.data.length - 1) || 0;
+  let startIndex = 0;
+  let endIndex = 0;
+  if (props.data && props.data.length !== 0) {
+    endIndex = props.data.length - 1;
+  }
+
+  if (brushItem && brushItem.props) {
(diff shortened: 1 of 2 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -649,7 +649,9 @@ const createDefaultState = (props: CategoricalChartProps): CategoricalChartState
   const brushItem = findChildByType(children, Brush);
   const startIndex = (brushItem && brushItem.props && brushItem.props.startIndex) || 0;
   const endIndex =
-    brushItem?.props?.endIndex !== undefined ? brushItem?.props?.endIndex : (props.data && props.data.length - 1) || 0;
+    brushItem?.props?.endIndex !== undefined
+      ? brushItem?.props?.endIndex
+      : (props.data && props.data.length !== 0 && props.data.length - 1) || 0;
```
ckifer, line 654: This approach makes this section of code look really confusing imo. If we moved the check and the resulting value outside of the ternary this becomes so much easier to read

ckifer (changes requested):

nikolasrieble (commented):
Could we add a test to this change? I see how a unit test can be difficult, but a RTL or storybook test should be possible.

HHongSeungWoo:
I will also add a test

test/chart/generateCategoricalChart.spec.tsx
```diff
@@ -230,4 +231,27 @@ describe('generateCategoricalChart', () => {
       );
     });
   });
+
+  describe('createDefaultState', () => {
+    it('Should have the correct dataIndex', () => {
```
nikolasrieble, line 236: Why did you combine these four separate test cases into a single test? Is there any interaction that I am missing? If not, please split it up into 4 tests.

HHongSeungWoo: it was just ctrl+c,v :sweat_smile:

nikolasrieble (approved):
Thank you for adding the tests! ❤️ 🚀

ckifer (approved):

Issue #3990, opened 2023-11-23
Regression: horizontal bar chart with tooltip


ilteoood:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/recharts-issue-template-forked-9mffgj?file=/src/index.js)

### Steps to reproduce
With the cursor, hover on one of the bars
The background for that bar will turn to grey

### What is expected?
The same behavior of version 2.9.3: the tooltip is rendered when you hover the bar, without any additional background

### What is actually happening?
The background of a bar turns to grey

| Environment | Info |
|---|---|
| Recharts | v2.10.1 |
| React | 18.2.0 |
| System | OS X 13.6.2 |
| Browser | Chrome |

makedatanotlore:
Can confirm we are seeing the same errant behavior on a custom Tooltip component, showing a grey background where before there was none.

Likely related to this, we can no longer hover individual bars in our stacked bar chart.

It feels strange that a breaking change such as this was intended, especially when it wasn't part of a new major. It must be a bug?

ckifer:
Definitely not intentional. Probably a result of trying to fix a different bug. 

@HHongSeungWoo?

HHongSeungWoo:
It seems there was a small logical error during the process of removing lodash. Sorry about that! 😅

ckifer:
fixed in 2.10.2

nbran:
Seeing same issue with vertical bars in 2.12.0

In the clip below, as soon as the `active` prop for `<Tooltip />` is set to `true` all that's rendered is a gray background taking the full width of the X-axis position.

[Screencast from 02-15-2024 06:19:42 PM.webm](https://github.com/recharts/recharts/assets/25080013/500f8a5a-d92c-47ad-8376-414b3ad9a712)

ckifer:
that would be the cursor, could you try setting `cursor={false}` on your tooltip? If that doesn't fix the issue could you provide a codesandbox?

ckifer:
Still seems like it shouldn't be happening. @HHongSeungWoo do you mind taking a look at some point?

HHongSeungWoo:
@ckifer It seems normal for the cursor to be displayed when the TooltipEventType is 'axis'. 
Do you want this behavior to be modified?

[...]

PR #3992, opened 2023-11-24
fix: type check

HHongSeungWoo:
## Description

I should have checked if it's a boolean type, but I was converting the value to a boolean

## Related Issue

#3990 

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1305,7 +1305,7 @@ export const generateCategoricalChart = ({
     getTooltipEventType() {
       const tooltipItem = findChildByType(this.props.children, Tooltip);
 
-      if (tooltipItem && Boolean(tooltipItem.props.shared)) {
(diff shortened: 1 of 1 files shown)


ckifer:
This was using isBoolean previously?

Any other places in code that do this?

HHongSeungWoo:
> This was using isBoolean previously?
> 
> Any other places in code that do this?

yes.

I can't check everything, but it's not in `generateCategoricalChart.tsx` at least

ckifer:
Thanks 👍

nikolasrieble:
Would it make sense to add a test to verify this does not break again?

trainoasis:
I think this might have broken the `ReferenceArea` - it is now without grayish color as well, even when you define stroke or color prop. Can you check pls?

HHongSeungWoo:
Could you please create an issue with a reproducible example??

trainoasis:
@HHongSeungWoo investigated a bit more & seems this isn't related to your changes at all! It was a very hidden change on our side. Sorry for false alarm! 🙌🏼

PR #3996, opened 2023-11-25
Add lint-test script

PavelVanecek:
## Description

I have noticed that there's a `lint` and `lint-storybook` command in the pre-push hook, but there is no `lint-test`. So I added one and fixed some of the errors.

Some of the other errors are conflicting with my other pull request so my next steps will be:

1. Merge everything I have open at the moment
2. Fix all the lint-test violations
3. Add lint-test to pre-push hook
4. Figure out how to do a typescript check in pre-push too

## Related Issue

Couldn't find one

## Motivation and Context

We want code style and type safety in the `test` folder too

## How Has This Been Tested?

npm run test

## Types of changes

## Checklist:

### DIFF
--- a/test/cartesian/YAxis.spec.tsx
+++ b/test/cartesian/YAxis.spec.tsx
@@ -41,14 +41,14 @@ describe('<YAxis />', () => {
     expect(ticks[1].getAttribute('y')).toBe('297.5');
   });
 
-  const cases1: [number, AxisDomain, string][] = [
+  const casesThatShowTicks: [number, AxisDomain, string][] = [
     // [ticksLength, domain, textContentOfTickElement]
     [5, [0, 10000], '10000'],
     [4, [0, 'dataMax'], '9800'],
     [4, [0, 'dataMax - 100'], '9800'],
   ];
 
-  test.each(cases1)('Should render %s ticks when domain={%s}', (length, domain, textContent) => {
+  test.each(casesThatShowTicks)('Should render %s ticks when domain={%s}', (length, domain, textContent) => {
     render(
       <AreaChart width={600} height={400} data={data}>
         <YAxis type="number" stroke="#ff7300" domain={domain} />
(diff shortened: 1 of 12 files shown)


test/cartesian/YAxis.spec.tsx
```diff
@@ -40,12 +41,14 @@ describe('<YAxis />', () => {
     expect(ticks[1].getAttribute('y')).toBe('297.5');
   });
 
-  each([
+  const cases1: [number, AxisDomain, string][] = [
```
nikolasrieble, line 44: It would be great to use a better name here.

test/cartesian/YAxis.spec.tsx
```diff
@@ -72,20 +75,19 @@ describe('<YAxis />', () => {
...
-          <YAxis stroke="#ff7300" domain={domain} />
-          <Area dataKey="noExist" stroke="#ff7300" fill="#ff7300" />
-        </AreaChart>,
-      );
-      const ticks = document.querySelectorAll('text');
-
-      expect(ticks).toHaveLength(0);
-    },
-  );
+  const cases2: [AxisDomain][] = [[[0, 'dataMax + 100']], [[0, 'dataMax - 100']], [['auto', 'auto']]];
```
nikolasrieble, line 78: If we do not want to use good names, we could also just inline the object as well.

nikolasrieble (approved):
Great cleanup, thank you!

test/cartesian/YAxis.spec.tsx
PavelVanecek: Done!

test/cartesian/YAxis.spec.tsx
PavelVanecek: If I inline then I cannot have explicit type definition. I could use a sea of `as const` and rely on typescript to infer what I need which I didn't want to do. Or I could use `as` which I don't want since it's just `any` in a different coat. Or I could use `any` which I don't want.

I found a better name instead!

Issue #4004, opened 2023-11-30
Pie chart with animation causes 'non-unique keys' react-dom warning


lennartfranke:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/recharts-issue-template-forked-lhq77z?file=%2Fsrc%2Findex.js)

### Steps to reproduce
Open codesandbox and open console, see react non-unique key warnings.

### What is expected?
All generated sector elements have a unique key in any case.

### What is actually happening?
If `isAnimationActive` is set to `true` all sectors in the pie chart are initially rendered with the same start and end angle and therefore have the same key prop.

| Environment | Info |
|---|---|
| Recharts | v2.10.2 |
| React | react 18.2.0 |
| System | MacOS 14.1.1 |
| Browser | Chrome 119.0.6045.159 |

ckifer:
@imagine Life

ckifer:
@imagineLife it looks like there are more instances of this. Can you go through the demo and look for all instances? They don't show on the storybook for some reason...

`npm run demo` -> ComposedChart - produces another duplicate set of keys for tooltip

ckifer:
same with legend item

imagineLife:
Surely 👍 
- Can this convo be "migrated" to [this issue](https://github.com/recharts/recharts/issues/3273)? there are perhaps _many_ keys that need updating that aren't "throwing" lint errors during development :/ I can post more based on some of my findings either here or the other issue

ckifer:
Summarized over there

ckifer:
This should be resolved in the latest (see sandbox https://codesandbox.io/s/pie-chart-of-straight-angle-forked-rvsnly)

We'll try to be more careful when making key changes!

HHongSeungWoo:
Unfortunately, it seems that `Bar` also need to be reverted

imagineLife:
> Unfortunately, it seems that `Bar` also need to be reverted
Hi @HHongSeungWoo 👋 Do you have &/or can you put together an example showing the key error, like a code sandbox? :/ The `bar` tests all pass and the `demo` bar charts all render without key errors, I'd love to see what props and/or features cause the key issue! (_and likely reset the recent key-change back to index-based until more thorough testing is done on cases like this!_)

[...]

PR #4005, opened 2023-11-30
chore: fix pie sector key value

imagineLife:
## Description

Updates the `key` value for pie sectors

## Related Issue
[4004](https://github.com/recharts/recharts/issues/4004)

## Motivation and Context
Fix a duplicate-key error

## How Has This Been Tested?
- [x] new passing test based on codesandbox example in original issue

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -71,6 +71,7 @@ export type PieSectorDataItem = SectorProps & {
   tooltipPosition?: Coordinate;
   value?: number;
   paddingAngle?: number;
+  dataKey?: string;
 };
 
 interface PieProps extends PieDef {
@@ -486,7 +487,6 @@ export class Pie extends PureComponent<Props, State> {
         stroke: blendStroke ? entry.fill : entry.stroke,
         tabIndex: -1,
       };
-
       return (
         <Layer
           ref={(ref: HTMLElement) => {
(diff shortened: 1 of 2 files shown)


src/polar/Pie.tsx
```diff
@@ -497,7 +497,7 @@ export class Pie extends PureComponent<Props, State> {
           tabIndex={-1}
           className="recharts-pie-sector"
           {...adaptEventsOfChild(this.props, entry, i)}
-          key={`sector-${entry?.startAngle}=${entry?.endAngle}`}
+          key={`sector-${entry?.startAngle}-${entry?.endAngle}-${entry.midAngle}`}
```
ckifer, line 500: does the guarantee uniqueness for each sector?

imagineLife: 👋 
- it looks like `midAngle` is the most and/or _only_ easily-identifiable unique field of each `entry` object (_each "slice" of the pie_)
- no duplicate-key lint err is present in any of the tests, including the new test, duplicate of the sandbox example which _did_ reproduce the key error in the test prior to adding this `midAngle` data piece to the key

ckifer: 🚀 - good enough for a mitigation for now, lets monitor issues and see if this is brought up again after I merge and patch. Thank you for jumping in!

imagineLife:
I'll try the demos & make PR(s) accordingly 🤔 still learning the repo a bit I suppose!

src/polar/Pie.tsx
VictorHugoLeme: Hi! Sorry for excavating this issue, but it is still present.
I've been having the warning below (logged as an error) when I have 2 or more cells with value equal to '0'.
`Warning: Encountered two children with the same key, 'sector-360-360-360'.`

ckifer: I can't reproduce https://codesandbox.io/p/sandbox/two-simple-pie-chart-forked-25xllj

You probably need to update recharts @VictorHugoLeme

Issue #3273, opened 2023-01-24
Remove react/no-array-index-key comments


marcalexiei:
Follow to [discussion](https://github.com/recharts/recharts/pull/3270#discussion_r1084889331) in #3270

For every `eslint-disable-next-line react/no-array-index-key` we should find a valid string or number instead of using an index

BuddhikaGeasman:
@ckifer If this is not resolved yet. Any chance I could take it?

ckifer:
@BuddhikaGeasman there was a recent report of an issue in the latest comments here https://github.com/recharts/recharts/issues/4004 

If you can find the error mentioned there that would be more important than removing these lint ignores. I actually think removing these ignores is impossible without re-thinking some things. Any time we use data to try and address these there are duplicate key errors

DrJShoff:
This error also appears in scatter chart when dots are drawn on top of each other. This causes duplications of dots if the chart is resized or the data updated as the chart can't keep track of all dots as they have the same key.

imagineLife:
Perhaps components that render chartable items should have _(more) required unique props_.  

### Uniquely Identify Items
One recurring error in duplicate keys is revealed when 2 chartable items (pie slices, dots on a scatter, bars, could be more) have the same cartable "data": x values, y values, etc.

**One idea** to tackle this seems to be _set more required props_ for renderable elements: something like a "name", or "id". There does not seem to be enough required props that an be leveraged for a unique key when iterating over data to render.    


**Another idea** could be to add a required prop called uniqueId (or key or something) where developers who use recharges must bring-their-own key to recharts. This approach would perhaps be a bit more “breaking”. Recharges, itself, would “pass along” the key provided by the developer to the charting rendering. The benefit here would be that recharges, itself, stops “caring” about the key the way it is currently trying to.

ckifer:
Imo we should only put this burden on the consumer if they are iterating through things themselves. If we iterate, we need to be the ones who ensure keys are unique. 

> set more required props for renderable elements

The consumer shouldn't have to have more data in their data in order to render a chart without errors

> developers who use recharts must bring-their-own key to recharts

This doesn't work because that key will be the same for every element of every component giving us the same problem we currently have

Not really sure of a great way - we can hash the whole object but that has the same chances of collision. We can generate a UUID based of the object but same thing, we can use a UUID but that negates the purpose...

PR #4009, opened 2023-12-02
Chore: adjust pie sector key value

imagineLife:
## Description
Updates the `sector` (pie "slice") rendering logic to render `null` at the "beginning" of an animation.  
This allows for more explicit key values of sectors once the sectors have starting+ending angles.    
The `no-array-index-key` seems to _only be appearing_ in pie chart sectors when (1) **animation is enabled** (2) **the first rendering** of each slice where each slice has _"nothing" to render that can uniquely identify each sector_. Once the slices "have size" in the animation, the `no-array-index-key` stops appearing.

## Related Issue
[3273](https://github.com/recharts/recharts/issues/3273)

## Motivation and Context
set more explicit key value in a pie chart.

## How Has This Been Tested?
- [x] pie demos work without key error: Pie, PieChart, ResponsiveContainer, all "manual" testing
- [x] unit tests are updated & pass

## Types of changes
- [x] technical detail

## Checklist:

- [x] I have added tests to cover my changes
  - assert 1-slice pie renders with animation
  - update 6-slice pie to include animation & `v` attribute seen in `demo` data code
- [x] All new and existing tests passed

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -479,6 +479,7 @@ export class Pie extends PureComponent<Props, State> {
   renderSectorsStatically(sectors: PieSectorDataItem[]) {
     const { activeShape, blendStroke, inactiveShape: inactiveShapeProp } = this.props;
     return sectors.map((entry, i) => {
+      if (entry?.startAngle === 0 && entry?.endAngle === 0 && sectors.length !== 1) return null;
       const isActive = this.isActiveIndex(i);
       const inactiveShape = inactiveShapeProp && this.hasActiveIndex() ? inactiveShapeProp : null;
       const sectorOptions = isActive ? activeShape : inactiveShape;
@@ -497,8 +498,7 @@ export class Pie extends PureComponent<Props, State> {
           tabIndex={-1}
           className="recharts-pie-sector"
           {...adaptEventsOfChild(this.props, entry, i)}
-          // eslint-disable-next-line react/no-array-index-key
-          key={`sector-${i}`}
+          key={`sector-${entry?.startAngle}-${entry?.endAngle}-${entry.midAngle}`}
         >
           <Shape option={sectorOptions} isActive={isActive} shapeType="sector" {...sectorProps} />
         </Layer>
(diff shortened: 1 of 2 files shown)


src/polar/Pie.tsx
```diff
@@ -479,6 +479,7 @@ export class Pie extends PureComponent<Props, State> {
   renderSectorsStatically(sectors: PieSectorDataItem[]) {
     const { activeShape, blendStroke, inactiveShape: inactiveShapeProp } = this.props;
     return sectors.map((entry, i) => {
+      if (entry?.startAngle === 0 && entry?.endAngle === 0 && sectors.length !== 1) return null;
```
ckifer, line 482: why `sectors.length !== 1`? Is 1 when animation is starting?

Are the angle values always going to be 0 or could they be undefined? If they're undefined this will go through

ckifer (commented):
some questions

src/polar/Pie.tsx
imagineLife: Hello 👋 
> why sectors.length !== 1? Is 1 when animation is starting?   

when there is 1 sector, there is 1 pie "slice". With 1 pie sector/slice, there will not be a duplicate key error, as there will not be multiple sectors/slices.  
The animation starting is represented by the sector(s) startAngle & endAngle both being `0`.  

> Are the angle values always going to be 0 or could they be undefined? If they're undefined this will go through  

The angle values are never `undefined` as far as I can tell 🤔 The angles change, during an animation, from 0 to the angle of the pie slice, "tick-by-tick".   

The duplicate-key error appears in lets say an animated pie with 3 slices:
- at the beginning of the animation, all 3 slices have the same startAngle of 0
- at the beginning of the animation, all 3 slices have the same endAngle of 0
- the angle(s) of each separate slice are the (only?!) uniquely-identifying attributes that can be used to associate each item with a unique key
- Due to the animation of each slice starting with each slice angle value at `0`, the beginning of the animation throws the duplicate key error. Once the animation gets past `0` values for each slice, the duplicate key error goes away during the rest of the animation.

ckifer: Thanks for the detailed explanation 🚀

ckifer:
Let's merge and monitor again, looks good to me though

Issue #4011, opened 2023-12-05
RadialBar background={{ className }} ignored


nik-webdevelop:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/rechart-radial-bar-classname-bug-cl3ll5?file=%2Fsrc%2FApp.tsx)

### Steps to reproduce
Check if RadialBar gets 'custom-class' in background's class list

### What is expected?
background={{ className }} to be merged with background svg element's classes

### What is actually happening?
background={{ className }} ignored 

| Environment | Info |
|---|---|
| Recharts | v2.10.3 |
| React | react 18.2.0 |
| System | Win 10 x64 |
| Browser | Google Chrome Version 119.0.6045.160 (Official Build) (64-bit) |

ckifer:
Hey @nik-webdevelop 

This should be pretty easy to fix - would you like to contribute?

https://github.com/recharts/recharts/blob/master/src/polar/RadialBar.tsx#L370

Just take `backgroundProps.className` and append it to that string using `clsx` there

```jsx
clsx('recharts-radial-bar-background-sector', backgroundProps.className)
```

ckifer:
This should be fixed on next release 

Going to close for now

nik-webdevelop:
@ckifer Sorry I forgot about this issue. Thank you very much for the fix!

PR #4027, opened 2023-12-21
allow className to be passed to Radial Bar background

ckifer:
## Description

* Fix issue where className is ignored if sending for RadialBar `background` prop
* Fix some additional issues in tests which were giving me trouble (wrap some items in `svg`)

Note:

The pie test was failing because of animation, it was flaky though. I'm running it on a pretty beefy box so I think the tests finished before the default animation time could.

The others don't fail but they log annoying messages in the console and that could potentially produce the wrong behavior. SVG elements without their container aren't usually able to be used for much

## Related Issue

#4011 

## Motivation and Context

Fix a small fix

## How Has This Been Tested?

See added unit test

N/A

## Types of changes

## Checklist:

### DIFF
--- a/test/util/FunnelUtils.spec.tsx
+++ b/test/util/FunnelUtils.spec.tsx
@@ -25,19 +25,25 @@ describe('funnelUtils', () => {
     expect(res).toEqual(mockRes);
   });
   it('<FunnelTrapezoid /> returns a trapezoid shape with no options', () => {
-    const { container } = render(<FunnelTrapezoid option={{}} {...mockProps} isActive={Boolean(true)} />);
+    const { container } = render(
+      <svg width={100} height={100}>
+        <FunnelTrapezoid option={{}} {...mockProps} isActive={Boolean(true)} />
+      </svg>,
+    );
     expect(container.querySelector('g')).toHaveClass('recharts-active-shape');
   });
   it('<FunnelTrapezoid /> returns a trapezoid shape with x & y from options', () => {
     const { container } = render(
-      <FunnelTrapezoid
-        option={{ x: 10, y: 10 }}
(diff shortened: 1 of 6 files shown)


test/util/FunnelUtils.spec.tsx
```diff
@@ -25,19 +25,25 @@ describe('funnelUtils', () => {
     expect(res).toEqual(mockRes);
   });
   it('<FunnelTrapezoid /> returns a trapezoid shape with no options', () => {
-    const { container } = render(<FunnelTrapezoid option={{}} {...mockProps} isActive={Boolean(true)} />);
+    const { container } = render(
+      <svg width={100} height={100}>
```
nikolasrieble, line 29: Interesting. Did these fail before? If not, why change them? If yes, how come we did not notice? 
I believe this is worth comment more on in the PR summary. Please share some more details, if only for knowledge sharing.

nikolasrieble (approved):
🚢

test/util/FunnelUtils.spec.tsx
ckifer: The pie test was failing because of animation, it was flaky though. I'm running it on a pretty beefy box so I think the tests finished before the default animation time could. 

The others don't fail but they log annoying messages in the console and that could potentially produce the wrong behavior. SVG elements without their container aren't usually able to be used for much

ckifer: Generally I'd keep the two types of changes separate, but this is mostly just laziness :p. If the bug fix was larger/more impactful I would separate things out. There are other test "errors" that have popped up since the last time I worked on things. Seems I should take a a look at those (mostly key related)

PR #4031, opened 2023-12-24
Referenceline get end points tests

PavelVanecek:
## Motivation and Context

I am trying to apply plan 5b on the ReferenceLine, and I found this `getEndPoints` function and realized I do not feel comfortable making changes in there so I wrote some tests and improved some of the types and made it a separate PR.

## How Has This Been Tested?

git push

## Types of changes

## Checklist:

### DIFF
--- a/test/util/ReactUtils.spec.tsx
+++ b/test/util/ReactUtils.spec.tsx
@@ -18,30 +18,36 @@ import { adaptEventHandlers, adaptEventsOfChild } from '../../src/util/types';
 
 describe('ReactUtils untest tests', () => {
   describe('filterProps', () => {
-    test('should call filterProps wtesth any boolean and return a null result', () => {
-      expect(filterProps(true)).toBeNull();
-      expect(filterProps(false)).toBeNull();
+    test.each([true, false, 125, null, undefined])('should return null when called with %s', input => {
+      expect(filterProps(input)).toBeNull();
     });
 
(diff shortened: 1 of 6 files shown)


test/util/ReactUtils.spec.tsx
```diff
@@ -18,32 +18,38 @@ import { adaptEventHandlers, adaptEventsOfChild } from '../../src/util/types';
...
 
-    test('should expect viewBox on type "svg"', () => {
+    test('should return viewBox string on type "svg"', () => {
       expect(filterProps({ test: '1234', helloWorld: 1234, viewBox: '0 0 0 0' }, false, 'svg')).toEqual({
         viewBox: '0 0 0 0',
       });
     });
 
+    test('should return viewBox object on type "svg"', () => {
+      // I think this is a bug because SVG wants a string viewBox and this will let the object through
```
ckifer, line 43: the only component that actually would require this to be external facing is `Surface` I think. Surface takes care of things here https://github.com/recharts/recharts/blob/master/src/container/Surface.tsx#L38

Any other component that uses an SVG directly would also have to do that though

ckifer (approved):
These tests are awesome, thank you!

test/util/ReactUtils.spec.tsx
ckifer: Since Surface references viewBox directly we could probably always filter viewBox out even for SVG

Issue #710, opened 2017-05-27
Zoom and pan


serializable:
This is a feature request

It would be nice if the charts had a zoom and pan feature (per axis).
If there is a way to achieve this, please let me know. Otherwise, can you think of a "creative way" to add this in? Or if you give me a few pointers I might fork and try to add it in

Thanks

al3x-huang:
Also interested in this

oliverwatkins:
also interested

appsmehta:
Also Interested in this

angulo:
Interested too

matepaiva:
+1

matepaiva:
Oh, now I see that Brush does the job. So I dont think this feature is really necessary.

swapbuk:
@matepaiva What is Brush? Is it component in rechart or configuration in component ?

xile611:
Check [this demo](http://recharts.org/#/en-US/examples/BrushBarChart).

hueitan:
If you are still interested in a smoothly way to **zoom** the graph, here's the solution

> update the Axis padding style with the animation

jsfiddle : http://jsfiddle.net/udrq0h2d/
screen gif demo:
![recharts-zoom-in-out-example](https://user-images.githubusercontent.com/2560096/32661383-0dbba4ca-c627-11e7-8345-5083a37b7e99.gif)

golestanirad:
> Oh, now I see that Brush does the job. So I dont think this feature is really necessary.

what is that?

matepaiva:
I don't actually remember, but I think this is it: http://recharts.org/en-US/api/Brush

gmilette:
this feature is still necessary, in case the user wants to pan the chart without using the brush component

leonardolessa:
I think pan and brush are completely different features, is there any way to achieve the result of zooming Y axis?

jtiagodev:
Have you considered the following example?
http://recharts.org/en-US/examples/HighlightAndZoomLineChart

howdyhyber:
Also interested in this feature, its hassle to use brush and its much easier to just pan or zoom

Jarrio:
+1

sseira:
Would love to be able to zoom into a specific area in the scatter plot

aniketmore311:
did anyone figure out zooming and panning?  I can use the brush to pan the chart but how do I implement zoom? I don't think domains work on the category chart.

robertgutu:
[...]

PR #4036, opened 2024-01-03
BrushY: A dummy vertical Brush for scaling/panning the YAxis

kerguler:
The changes clone and transform Brush as a vertical slider, but do not cast a function to it.

## Description

The new BrushY keeps track of the startIndex and endIndex within a user-defined number of steps. BrushY can be linked to the YAxis to set an appropriate domain for vertical scrolling/zooming.

## Related Issue

Zoom and pan - Issue #710 : https://github.com/recharts/recharts/issues/710#top

## Motivation and Context

Brush performs horizontal scrolling/zooming by filtering the source data as needed. BrushY removes the filtering step and provides a dummy vertical slider to scale/pan the YAxis.

## How Has This Been Tested?

A demo is included in the demo/component folder to show how the link between BrushY and YAxis can be established by the user. The demo uses the useState hook to update the YAxis limits. In addition, a test is included in the test/cartesian folder (similar to the test for Brush) to confirm that the component is properly rendered.

To enable rendering the component and remove the data (domain) filtering step, changes are also done in generateCategoricalChart.tsx.

## Screenshot:

Screenshot 2024-01-03 at 08 04 19

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x ] My code follows the code style of this project.
- [ x ] My change requires a change to the documentation.
- [ x ] I have added tests to cover my changes.
- [ x ] All new and existing tests passed.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -41,7 +41,7 @@
     "react/jsx-filename-extension": ["error", { "extensions": [".ts", ".tsx", ".js", ".jsx"] }],
 
     "no-var": "error",
-    "react/jsx-no-bind": "off",
+    "react/jsx-no-bind": "error",
 
     "no-unused-vars": "off",
     "one-var": ["error", { "uninitialized": "always" }],
@@ -76,8 +76,7 @@
     "import/prefer-default-export": "off",
     "import/no-named-as-default": "error",
     "import/no-cycle": "error",
-    "import/no-default-export": "off",
-    "import/no-unresolved": "off",
+    "import/no-default-export": "error",
     "import/no-import-module-exports": "error",
     "import/no-relative-packages": "error",
     "no-use-before-define": "off",
@@ -85,7 +84,14 @@
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/explicit-module-boundary-types": "off",
     "@typescript-eslint/ban-types": "off",
-    "@typescript-eslint/no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      {
+        "args": "after-used",
+        "ignoreRestSiblings": true,
+        "argsIgnorePattern": "^_"
+      }
+    ],
     "no-restricted-imports": [
       "error",
       {
--- a/.eslintrc
+++ b/.eslintrc
@@ -41,7 +41,7 @@
     "react/jsx-filename-extension": ["error", { "extensions": [".ts", ".tsx", ".js", ".jsx"] }],
 
     "no-var": "error",
-    "react/jsx-no-bind": "error",
+    "react/jsx-no-bind": "off",
 
     "no-unused-vars": "off",
     "one-var": ["error", { "uninitialized": "always" }],
--- a/.eslintrc
+++ b/.eslintrc
@@ -76,7 +76,8 @@
     "import/prefer-default-export": "off",
     "import/no-named-as-default": "error",
     "import/no-cycle": "error",
-    "import/no-default-export": "error",
+    "import/no-default-export": "off",
+    "import/no-unresolved": "off",
     "import/no-import-module-exports": "error",
     "import/no-relative-packages": "error",
     "no-use-before-define": "off",
@@ -84,14 +85,7 @@
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/explicit-module-boundary-types": "off",
     "@typescript-eslint/ban-types": "off",
-    "@typescript-eslint/no-unused-vars": [
-      "error",
-      {
-        "args": "after-used",
-        "ignoreRestSiblings": true,
-        "argsIgnorePattern": "^_"
-      }
-    ],
+    "@typescript-eslint/no-unused-vars": "off",
     "no-restricted-imports": [
       "error",
       {
--- a/demo/component/BrushYDemo.tsx
+++ b/demo/component/BrushYDemo.tsx
@@ -1,99 +1,51 @@
-import React, { Component } from 'react';
-import { Surface, BrushY } from 'recharts';
+import React, { Component, useState } from 'react';
+import { LineChart, Line, Brush, BrushY, XAxis, YAxis } from 'recharts';
 
-export default class Demo extends Component<any, any> {
-  static displayName = 'BrushYDemo';
+const data = [
+  { day: '1', value: 10 },
+  { day: '2', value: 20 },
+  { day: '3', value: 10 },
+  { day: '4', value: 30 },
+  { day: '5', value: 50 },
+  { day: '6', value: 10 },
+  { day: '7', value: 30 },
+  { day: '8', value: 20 },
+  { day: '9', value: 10 },
+  { day: '10', value: 70 },
+  { day: '11', value: 40 },
+  { day: '12', value: 20 },
+  { day: '13', value: 10 },
+  { day: '14', value: 10 },
+];
 
-  data = [
-    '2015-10-01',
-    '2015-10-02',
-    '2015-10-03',
-    '2015-10-04',
-    '2015-10-05',
-    '2015-10-06',
-    '2015-10-07',
-    '2015-10-08',
-    '2015-10-09',
-    '2015-10-10',
-    '2015-10-11',
-    '2015-10-12',
-    '2015-10-13',
-    '2015-10-14',
-    '2015-10-15',
-    '2015-10-16',
-    '2015-10-17',
-    '2015-10-18',
-    '2015-10-19',
-    '2015-10-20',
-    '2015-10-21',
-    '2015-10-22',
-    '2015-10-23',
-    '2015-10-24',
-    '2015-10-25',
-    '2015-10-26',
-    '2015-10-27',
-    '2015-10-28',
-    '2015-10-29',
-    '2015-10-30',
-  ];
+function BrushYDemo() {
+  const displayName = 'BrushYDemo';
 
-  state = {
-    simple: {
-      startIndex: 0,
-      endIndex: this.data.length - 1,
-    },
-    gap: {
-      startIndex: 0,
-      endIndex: this.data.length - 1,
-    },
-  };
+  const bSteps = 100;
+  const range = { min: 0, max: 100 };
+  data.forEach(d => {
+    if (d.value < range.min) range.min = d.value;
+    if (d.value > range.max) range.max = d.value;
+  });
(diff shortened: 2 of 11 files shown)


.eslintrc
```diff
@@ -76,22 +76,16 @@
     "import/prefer-default-export": "off",
     "import/no-named-as-default": "error",
     "import/no-cycle": "error",
-    "import/no-default-export": "error",
+    "import/no-default-export": "off",
+    "import/no-unresolved": "off",
```
ckifer, line 80: Looks like these changes are unrelated to this PR, can you remove changes to the lint file?

demo/component/BrushYDemo.tsx
```diff
@@ -0,0 +1,51 @@
...
+  { day: '8', value: 20 },
+  { day: '9', value: 10 },
+  { day: '10', value: 70 },
+  { day: '11', value: 40 },
+  { day: '12', value: 20 },
+  { day: '13', value: 10 },
+  { day: '14', value: 10 },
+];
+
+function BrushYDemo() {
```
ckifer, line 21: We don't really use the demo project anymore, can you add this code to the storybook instead?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "recharts",
-  "version": "2.10.3",
+  "version": "2.10.4",
```
ckifer, line 3: Please don't bump the version unless an official recharts release, we'll handle this internally

ckifer (changes requested):
Really appreciate you taking the time to add this, I think we'll need to think on this a little more before full release though. We want to sure we test fully to catch any edge cases and aren't adding to our already large pile of tech debt. 

There are some changes in this PR that are outside of the context of the PR (version bump, eslint changes) that will need to be reverted if we are to merge this. 

Will also approvals on direction from some others as this doesn't look like the direction we want to continue to move in. 

Really appreciate the legwork!

package.json
kerguler: Of course, I'll revert this.

.eslintrc
kerguler: I included these as I was receiving errors while compiling locally. I will revert them.

demo/component/BrushYDemo.tsx
kerguler: Of course, as soon as I figure out how to do it :)

kerguler:
Thank you @ckifer for your lightning fast response. I need this component for the react app we are developing. If you decide not to include it in the master, could you please kindly advise the best approach to achieve this functionality? At the moment, I am keeping the changes in the fork and installing them using yalc.

.eslintrc
ckifer: Hmm odd, with your code or with other code? The current code in the codebase shouldn't have any errors with the lint settings besides maybe in the demo project (which we don't use 😅)

demo/component/BrushYDemo.tsx
ckifer: Should be similar code luckily! You can add this to the Brush examples: storybook - Examples - Cartesian - Brush

ckifer:
We definitely want the functionality eventually! We're just trying to take a different approach with some things (as you can see the generate function is very large and hard to follow). If we can get this to a good place we can maybe merge it, but we might also wait until after some more refactoring has been done to make it easier to understand/test and so we can have a more robust API. So it just might be a little bit before this gets merged/released with that said. 

I'll wait for your changes and some reviews from others and we can go from there. 

In terms of publishing yourself you can publish your own fork of recharts on npm if you'd like though it's always better to follow along with master since you'll miss out. You could publish your fork and use that until we merge this if you really need this functionality now. 

Thanks!

src/chart/generateCategoricalChart.tsx
```diff
@@ -1966,6 +1966,26 @@ export const generateCategoricalChart = ({
       });
     };
 
+    renderBrushY = (element: React.ReactElement) => {
+      const { margin, data } = this.props;
+      const { offset, updateId } = this.state;
+
+      // TODO: update brush when children update
+      return cloneElement(element, {
```
PavelVanecek, line 1974: Yes this is precisely the pattern we are trying to get rid of 😬 try to think of an API that does not require cloning, and replacing the props internally

PavelVanecek: Hm on the other hand it would make sense to have the same API as the existing Brush. After all it's the same thing, isn't it.

kerguler: @PavelVanecek I couldn't find an easy answer to the problem of rendering the same component twice. Now that I am a bit more familiar with the workings of the package, it might have been easier than I thought. However, I had to make some detailed conformational changes, and played with the endIndex range. So, a separate component made sense

.eslintrc
kerguler: Neither. The problem occurs when I try to commit with git. It won't let me unless the lint issues are resolved.

demo/component/BrushYDemo.tsx
kerguler: I'll add it as a new component as I couldn't figure out how to incorporate changes in the Brush component and have it rendered twice. I trust that the planned refactoring will enable this. Until then...

.eslintrc
ckifer: I think it's because of the changes to the demo components though, lint staged it linting when you commit and the lint errors are from the demo files

kerguler: [STARTED] Preparing lint-staged...
[SUCCESS] Preparing lint-staged...
[STARTED] Running tasks for staged files...
[STARTED] package.json — 7 files
[STARTED] *.{ts,tsx,js,jsx} — 3 files
[STARTED] eslint --fix
[FAILED] eslint --fix [FAILED]
[FAILED] eslint --fix [FAILED]
[FAILED] eslint --fix [FAILED]
[STARTED] Applying modifications from tasks...
[SKIPPED] Skipped because of errors from tasks.
[STARTED] Reverting to original state because of errors...
[SUCCESS] Reverting to original state because of errors...
[STARTED] Cleaning up temporary files...
[SUCCESS] Cleaning up temporary files...

✖ eslint --fix:

/Users/kamil/Projects/VEClim/recharts/src/cartesian/BrushY.tsx
  308:39  error  Unexpected empty arrow function                          @typescript-eslint/no-empty-function
  310:39  error  Unexpected empty arrow function                          @typescript-eslint/no-empty-function
  543:27  error  'isSlideMoving' is assigned a value but never used       @typescript-eslint/no-unused-vars
  543:42  error  'isTravellerMoving' is assigned a value but never used   @typescript-eslint/no-unused-vars
  543:61  error  'isTravellerFocused' is assigned a value but never used  @typescript-eslint/no-unused-vars

✖ 5 problems (5 errors, 0 warnings)

husky - pre-commit hook exited with code 1 (error)

kerguler: I'll fix these and see if it works

ckifer: Oh yeah those are actual lint errors that need fixed

kerguler: Okay, it committed without any errors and with BrushYDemo intact (might just be my luck). Just to note, the part I included in the storybook does nothing. It is the demo that connects BrushY with YAxis. I am not sure how to have the same functionality in the storybook, but it might not be necessary. Hance the reason to name it as a "dummy" component.

src/cartesian/BrushY.tsx
```diff
@@ -0,0 +1,570 @@
+/**
```
ckifer, line 1: I didn't even see this file... I wonder if we can do these changes without a new component and instead with an `orientation` prop or similar

kerguler: The problem is not being able to render the same component twice. If that's solved, than BrushY could be integrated into Brush

src/chart/generateCategoricalChart.tsx
nikolasrieble: I believe this does not add any extra effort for future migration, if it simply stays in line with the existing brush. We will have to migrate the brush either way.

src/cartesian/BrushY.tsx
nikolasrieble, line 1: This file is certainly the blocking thing here. 
We do not want to simply copy code with small variations. 

This will reduce our velocity going forward massively. 
Please do reuse the existing code for the Brush and extend it.

nikolasrieble (changes requested):
Great work with user value here, please do not copy paste the component but instead extend the existing one in alignment with the libraries concepts. 

i.e. we also have a CartesianAxis and then we have an XAxis component and a YAxis component. 

That is a pattern that seems well applicable here. What do you think?

src/cartesian/BrushY.tsx
nikolasrieble: The pattern to solve this is done for the Axis. A shared CartesianAxis exists, but then we also have XAxis and YAxis. Both of them internally use CartesianAxis.

src/cartesian/BrushY.tsx
kerguler: I agree that the existing pattern makes sense. Would it be acceptable to decommission Brush and create both an XBrush and a YBrush component?

ckifer: To be completely honest, in the next (or a future) major version I think we should remove Brush entirely and instead implement a recharts native pan and zoom feature. Brush is clunky, especially for Y values. 

(That's opinion only and probably won't happen)

What's the issue with removal of the once prop and just using Brush with an orientation prop?

nikolasrieble: Also possible, but I like the consistency of patterns between components that we would have if we had YBrush and XBrush

ckifer: Sure, lets get it to that state where components are shared and go from there

cesarzxk:
I believe that the brush could be accepted as a child of the x and y axes, behaving in the orientation of the axis.

ckifer:
Yeah I think that could make sense. Breaking API change of course. Once we get brush info into context all of this gets easier

crazy4chrissi:
I find this super useful for vertical barcharts (`layout="vertical"`). There, I would expect the BrushY to act similar as the Brush for horizontal barcharts, i.e. control the range of dimension values. However, the default `onChange` does not do this because BrushY does not call `this.handleBrushChange`. Would be cool if this would be called by default for vertical barcharts.

ckifer:
I'm going to close this as stale, but lets revisit in 3.1 or later

PR #4037, opened 2024-01-04
Sunburst chart

### LINKED ISSUE #3842: Sunburst Charts

If it's not a blocker, recharts team can choose to add it to charts list which seems to be only popular chart type missing from library.

Example provided in earlier issue: [Sunburst Chart](https://nivo.rocks/sunburst/)

```tsx
interface ISunburstData {
  name: string,
  value?: number,
  children?: ISunburstData[]
}

const child1: ISunburstData = {
  name: 'Child1',
  value: 60,
  children: [
    {
      name: 'Child1 > Child1',
      value: 30
    },
    {
      name: 'Child1 > Child2',
      value: 20
    },
    {
      name: 'Child1 > Child3',
      value: 20
    }
  ]
}
const child2: ISunburstData = {
  name: 'Child2',
  value: 40
}

<SunburstChart width={512} height={512}>
  <SunburstPie data={child1} />
  <SunburstPie data={child2} />
</SunburstChart>
```

csdiehl:
## Description

Adds an entirely new chart type, the Sunburst Chart - resolves #3842. 

Note: this is a WIP. I've added a Typescript component for a responsive sunburst chart, a story to demonstrate the functionality, and mock data. I still need to add event handlers, and a tooltip - I wanted to give this a chance to be reviewed before changes grow too large, and I'm looking for guidance on the best way to integrate these elements to match the style of existing components. 

## Related Issue

#3842 

## Motivation and Context

This adds a popular and useful new hierarchical chart type that is not addressed by any existing component in the library. 

## How Has This Been Tested?

I have not added tests yet. 

Current state of the chart - responsive dimensions and labels, still needs interactivity. 

Screenshot 2024-01-03 at 6 59 24 PM

## Types of changes

## Checklist:


package.json
```diff
@@ -123,6 +123,7 @@
     "chromatic": "^6.15.0",
     "concurrently": "^7.6.0",
     "cross-env": "^7.0.3",
+    "d3-scale": "^4.0.2",
```
ckifer, line 126: as it stands this upgrade will break our consumers because v4 moves to ESM only.

Can you use d3-scale from `victory-vendor/d3-scale` instead?

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,127 @@
...
+            stroke="#FFF"
+            strokeWidth={padding}
+            startAngle={start}
+            endAngle={start + arcLength}
+            innerRadius={innerR}
+            outerRadius={innerR + r}
+            cx={cx}
+            cy={cy}
+          />
+          <text
```
ckifer, line 97: there is a `Text` component, not sure if that can be re-used here

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,127 @@
...
+            fontSize=".875rem"
+            alignmentBaseline="middle"
+            textAnchor="middle"
+            stroke="#FFF"
+            strokeWidth={0.5}
+            paintOrder="stroke fill"
+            fill="#333"
+            fontFamily="sans-serif"
+            fontWeight="bold"
```
ckifer, line 106: some of these values are pretty arbitrary, probably we should allow the user to pass in their own Text props or a custom Text component

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,127 @@
...
+      const start = currentAngle;
+      // color priority - if there's a color on the individual point use that, otherwise use parent color or default
+      const fillColor = d?.fill ?? childColor ?? fill;
+      const [textX, textY] = polarToCartesian(innerR + (innerR + r - innerR) / 2, start + arcLength - arcLength / 2);
+      currentAngle += arcLength;
+      sectors.push(
+        <g>
+          <Sector
+            fill={fillColor}
+            stroke="#FFF"
```
ckifer, line 88: arbitrary stroke value should be configurable on high level props I think

ckifer (commented):
To be honest, this looks awesome. Super clean.

We could maybe factor some of the logic out into testable functions, but besides that and the ability to customize everything this looks great.

You will definitely face issues when integrating with the current Tooltip situation and ideally we don't fit the style of the existing components because it is mostly bad 😅. @PavelVanecek is working on moving to an approach using Context instead of the giant generator function which will make this a lot easier. 

Until then I think we can focus on getting this part to a good spot

csdiehl:
thanks for reviewing @ckifer! Makes sense - the tooltip integration was my biggest question. Looks like the API allows users to add a tooltip as a child component, but some charts, like treemap, also have their own tooltip functions. 

I'll focus on addressing your comments and adding unit tests for what is already here.

ckifer:
I would way we probably want closer to how Funnel works, if you want to create a similar function in the mean time I think that would be fine

src/chart/SunburstChart.tsx
nikolasrieble: 👍 We should evaluate all props to be maximum consistent with the existing API in other places.

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,136 @@
...
+  ringPadding?: number;
+  innerRadius?: number;
+  children?: React.ReactNode;
+  fill?: string;
+  stroke?: string;
+  textOptions?: TextOptions;
+}
+
+interface DrawArcOptions {
+  r: number;
```
nikolasrieble, line 37: I assume this is the radius? Lets please spell out everything to improve readability.

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,136 @@
...
+  innerR: number;
+  initialAngle: number;
```
nikolasrieble, line 39: Could you please add a comment, explaining in how far this angle is initial? Is it changing over time?

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const defaultTextProps = {
+  fontFamily: 'sans-serif',
+  fontWeight: 'bold',
+  paintOrder: 'stroke fill',
+  stroke: '#FFF',
+  fill: 'black',
+};
+
+function maxDepth(node: SunburstData): number {
```
nikolasrieble, line 51:
```suggestion
function getMaxDepthOf(node: SunburstData): number {
```

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,136 @@
...
+  if (!node.children || node.children.length === 0) return 1;
+
+  // Calculate depth for each child and find the maximum
+  const childDepths = node.children.map(d => maxDepth(d));
+  return 1 + Math.max(...childDepths);
+}
+
+function polarToCartesian(r: number, angleInDegrees: number): number[] {
```
nikolasrieble, line 59: This seems pretty generic. We likely already have this implemented elsewhere. Please check, or move into a generic file such as PolarUtils or CartesianUtils

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,136 @@
...
+  data?: SunburstData;
+  width?: number;
+  height?: number;
+  padding?: number;
+  ringPadding?: number;
+  innerRadius?: number;
+  children?: React.ReactNode;
+  fill?: string;
+  stroke?: string;
+  textOptions?: TextOptions;
```
nikolasrieble, line 33: Comments above each of these definitions will automatically land in storybook user facing documentation. We do not do this everywhere, but we totally should with all new Charts. 

This will show up in the ArgsTable in the description column.

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,136 @@
...
+  innerRadius = 50,
+  fill = '#333',
+  stroke = '#FFF',
+  textOptions = defaultTextProps,
+}: SunburstChartProps) => {
+  // get the max possible radius for a circle inscribed in the chart container
+  const outerRadius = Math.min(width, height) / 2;
+
+  const cx = width / 2,
+    cy = height / 2;
```
nikolasrieble, line 82: In other PolarCharts, cy and cx are also overridable. This is totally fine to iterate on, but lets make sure that the props are overall consistent with other PolarCharts.

src/chart/SunburstChart.tsx
nikolasrieble: I understand from reading the code - and wonder - should we also have an endAngle?

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,136 @@
...
+}
+
+interface DrawArcOptions {
+  r: number;
+  innerR: number;
+  initialAngle: number;
+  childColor?: string;
+}
+
+const defaultTextProps = {
```
nikolasrieble, line 43: Are these defaults also used in other places?

src/chart/SunburstChart.tsx
csdiehl: I added new props so now the user can configure startAngle, endAngle, cx, cy, innerRadius and OuterRadius. Added a new story to demonstrate a partial sunburst chart using a custom starting and ending angle. I think that's all the main polar props, but let me know if there are any I missed

src/chart/SunburstChart.tsx
csdiehl: seems like most of these already have default descriptions that fit pretty well, but I added comments for additional clarification. Let me know if anything is still unclear

src/chart/SunburstChart.tsx
csdiehl: I couldn't find an example of this anywhere, but seems nice to provide some default styling that stands out better on the chart background than the un-styled text.  I can remove if needed though

src/chart/SunburstChart.tsx
csdiehl: I think the remaining props needed to make it consistent are dataKey, event handlers and animation props (if those are needed for this chart type). What is the dataKey used for?

src/chart/SunburstChart.tsx
nikolasrieble: I agree with you and like this, but wonder whether we should tackle such an improvement separately, and instead focus on consistency for this new chart.

package.json
```diff
@@ -93,17 +93,16 @@
...
     "@storybook/react": "^7.6.3",
     "@storybook/react-webpack5": "^7.6.3",
+    "@storybook/source-loader": "^7.6.3",
     "@storybook/test-runner": "^0.16.0",
     "@storybook/testing-library": "^0.2.2",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^12.1.5",
     "@testing-library/user-event": "^14.3.0",
     "@types/d3-interpolate": "^3.0.1",
-    "@types/d3-scale": "^4.0.2",
```
nikolasrieble, line 106: Interesting, did we not need this? 🤔

storybook/stories/API/chart/SunburstChart.stories.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { ResponsiveContainer, SunburstChart, Tooltip } from '../../../../src';
+import { CategoricalChartProps } from '../props/ChartProps';
+import { hierarchy } from '../../data';
+
+export default {
+  argTypes: {
+    ...CategoricalChartProps,
```
nikolasrieble, line 8: This seems to lead to misleading descriptions. Instead of importing this group, we should only import the specific shared descriptions that are actually used. 

Otherwise, props such as barCategoryGap show up, which are not used.

nikolasrieble: Screenshot 2024-01-07 at 20 54 22

storybook/stories/data/Hierarchy.ts
```diff
@@ -0,0 +1,88 @@
+export interface SunburstData {
+  name: string;
+  value?: number;
+  children?: SunburstData[];
+  fill?: string;
+}
+
+const hierarchy: SunburstData = {
```
nikolasrieble, line 8: The file is called hierarchy, the type used here is SunburstData though. Should this exist as a separate file?

test/polar/Sunburst.spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+describe('<Sunburst />', () => {
+  const data = {
+    name: 'Root',
+    value: 100,
+    children: [
+      {
+        name: 'Child1',
+        fill: '#264653',
+        value: 30,
+        children: [
```
nikolasrieble, line 14: Could you please simplify the test as much as possible? The test data should be only as complex as need be.

storybook/stories/data/Hierarchy.ts
nikolasrieble: I think this file would make a lot of sense if we were to use a shared type between TreeMap and SunburstChart data? Is that possible / does it even make sense? 

If not, I would recommend to integrate this data into the SunburstChart story.

storybook/stories/API/chart/SunburstChart.mdx
```diff
@@ -0,0 +1,16 @@
+import { Canvas, Meta, ArgTypes } from '@storybook/blocks';
+import * as SunburstStories from './SunburstChart.stories';
+
+# Sunburst
+
+The sunburst is a hierarchical chart, similar to a treemap, plotted in polar coordinates.
```
nikolasrieble, line 6: A suggestion: It might be nice to write a bit more about the motivation, as to why you would want to use it. 
Given that you just spent some time implementing it, I am certain you have a good case ❤️

nikolasrieble (changes requested):
I tested it more thoroughly and reviewed the whole PR in detail. 
This is a great contribution, I do not see this interfering with any ongoing refactoring, and it adds direct user value. 

Thank you for putting in the effort and time ❤️ 💪 Great work. 

Please do finalise by fixing the ArgsTable as shown in the screenshot.

src/chart/SunburstChart.tsx
nikolasrieble: Given a dataset with multiple rows such as `[{a:1, b:2}. {a:3, b:4}]`
The dataKey for example for a Line component identifies which row to use for the line values. The dataKey on the XAxis identifies which row to use for the xAxis ticks.

test/polar/Sunburst.spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+        fill: '#e76f51',
+        value: 20,
+      },
+    ],
+  };
+
+  it('renders a sector for each data point', () => {
+    const { container } = render(<SunburstChart data={data} />);
+
+    expect(container.querySelectorAll('.recharts-sector')).toHaveLength(19);
```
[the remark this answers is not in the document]
ckifer: +1, wonder if there is something else we can do like assert that a specific sector is next to another?

src/chart/SunburstChart.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    childNodes.forEach(d => {
+      const arcLength = rScale(d.value);
+      const start = currentAngle;
+      // color priority - if there's a color on the individual point use that, otherwise use parent color or default
+      const fillColor = d?.fill ?? childColor ?? fill;
+      const { x: textX, y: textY } = polarToCartesian(0, 0, innerR + radius / 2, -(start + arcLength - arcLength / 2));
+      currentAngle += arcLength;
+      sectors.push(
+        <g>
```
ckifer, line 147: wonder if there are any accessibility concerns we can take care of up front? aria-labels, tabIndex, etc.?

ckifer (commented):
Looking good, added a few more comments

storybook/stories/data/Hierarchy.ts
csdiehl: I think it would make sense to use the same type for both. But I think the `size` property should be re-named `value` since it could represent any quantitative value, and it should support a custom fill color. 

For now, I'll just delete this file and integrate the data into the sunburst story to avoid changing treemap.

src/chart/SunburstChart.tsx
csdiehl: that's a good idea - I set these attributes on the g element that wraps each data point, and tested that the chart is now keyboard accessible

test/polar/Sunburst.spec.tsx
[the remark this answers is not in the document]
csdiehl: good point - I modified this test to check that the sectors are rendered in the correct order, with the correct category color

src/chart/SunburstChart.tsx
csdiehl: got it - I added this in, so instead of just `value`, the user can pass any data key

csdiehl:
Ok @ckifer and @nikolasrieble thanks for reviewing - I think I addressed all your comments, but let me know if I missed anything! I updated the props in the documentation to reflect only those that are currently available on this component

ckifer (commented):
I wonder if we could release this as experimental somehow.. We don't currently have a process for that but

ckifer (approved):
Looks awesome! Wondering if we would do well to add some sort of caveat to the chart that it is new or experimental, but I say lets merge it!

We can release it next minor and have the caveat exist in the changelog. 

OR we could not export it yet and test it internally a bit.

ckifer:
After looking at the deployed storybook you're right about the Text for sure, though not sure what we can do except leave it to the users to address themselves.

The Tooltip is definitely off a bit, we can keep iterating! 🚀

uson1x:
What a timing! We were just about to hunt for a graph library for sunburst charts and hear our beloved recharts adding it at the same moment.

I'm curious, maybe you have some ideas on how we could add animations to it for when you change the data?

In our case, we have a single layer (no children) of data to show and when user clicks on one of the segments, we want to async load new data for this segment from server and replace the existing single layer with a new single layer (but new numbers and colors). Would be nice to do it with some nice animations, so it feels more fluid.

Soo... I don't assume you had plans on working on sunburst animations next? 😄 
And I hoped maybe you could share some pointers on how we could attempt to add those ourselves?

Thank you so much!

csdiehl:
thanks @ckifer! I agree - we can fix anything that comes up in a separate PR. I'll leave it up to you about when to release. Looking forward to contributing more to the library! 

@uson1x there is an `onClick` prop - so you could pass a callback to that that updates your data. But as of right now, it wouldn't animate. I see that animations are supported in some other charts, so that's something we could likely add later.

pelletier197:
Hello! I am very happy this is part of the library, but I had a hard time finding it out, because there is no documentation available for this. Would be a great addition to help the feature get discovered.

ckifer:
There is an entry for the sunburst chart in the storybook https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-chart-sunburstchart--docs

We're kind of in transition with the recharts website and the storybook - feel free to make a PR to the website if you'd like to get this info there!

PR #3994, opened 2023-11-25
Separate cursor component

PavelVanecek:
## Description

Following up from @nikolasrieble's suggestion in https://github.com/recharts/recharts/pull/3852#discussion_r1390416484 I am moving towards merging Cursor and Tooltip component. Cursor depends on Tooltip props anyway so it's a natural refactor.

Baby steps. First I move the Cursor rendering as-is to its own component. In a followup PR, I will make it read props from Tooltip directly and remove `renderCursor` method from generateCategoricalChart.

## Related Issue

https://github.com/recharts/recharts/discussions/3717

## Motivation and Context

Small step towards breaking down generateCategoricalChart

## How Has This Been Tested?

`npm test`, storybooks

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] I have added a storybook story or extended an existing story to show my changes
- [X] All new and existing tests passed.

PR #4041, opened 2024-01-09
fix: cursor key throwing non-unique key error

ckifer:
## Description

See comment here https://github.com/recharts/recharts/pull/3994/files#r1445550533

Basically `renderByOrder` renders children in an array. React requires the elements in that array to have unique keys. previously the inline cloned element needed it, now `Cursor` needs it.

## Related Issue

N/A

## Motivation and Context

Prevent key error from making it into the next release

We need storybook to log these errors...

## How Has This Been Tested?

`npm test` - see errors
`npm run demo` - see errors

Errors are resolved after change

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1780,9 +1780,11 @@ export const generateCategoricalChart = ({
       // The cursor is a part of the Tooltip, and it should be shown (by default) when the Tooltip is active.
       const isActive: boolean = element.props.active ?? isTooltipActive;
       const { layout } = this.props;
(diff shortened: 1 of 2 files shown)


PavelVanecek (approved):
Oh no, sorry I missed this one. I wonder how can we test this so that a similar problem doesn't repeat? How difficult would it be to add tests for all components?

ckifer:
No worries, I'm not really sure but especially since storybook doesn't throw them they're so easy to miss... open to suggestions

ckifer:
It doesn't look like there's an easy way to test for this... snapshot tests maybe? But that still requires review

PR #4054, opened 2024-01-13
Chartlayoutcontext split

PavelVanecek:
## Description

OK so this is necessary if we want to render `XAxis` from context but not make it re-render everytime anything in the `CategoricalChartState` changes.

## Related Issue

https://recharts.slack.com/archives/C042Q5K5UDC/p1704889920155939

## Motivation and Context

Plan 5b! 

## How Has This Been Tested?

npm test

storybooks

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/XAxis.tsx
+++ b/src/cartesian/XAxis.tsx
@@ -2,7 +2,12 @@
  * @fileOverview X Axis
  */
 import type { FunctionComponent, SVGProps } from 'react';
+import React from 'react';
+import clsx from 'clsx';
+import { useViewBox, useXAxisOrThrow } from '../context/chartLayoutContext';
+import { CartesianAxis } from './CartesianAxis';
 import { BaseAxisProps, AxisInterval } from '../util/types';
+import { getTicksOfAxis } from '../util/ChartUtils';
 
 /** Define of XAxis props */
 interface XAxisProps extends BaseAxisProps {
@@ -31,7 +36,24 @@ interface XAxisProps extends BaseAxisProps {
 
 export type Props = Omit<SVGProps<SVGElement>, 'scale'> & XAxisProps;
 
-export const XAxis: FunctionComponent<Props> = () => null;
+export const XAxis: FunctionComponent<Props> = ({ xAxisId }: Props) => {
+  const viewBox = useViewBox();
+  const axisOptions = useXAxisOrThrow(xAxisId);
+
+  if (axisOptions == null) {
+    return null;
+  }
+
+  return (
+    // @ts-expect-error the axisOptions type is not exactly what CartesianAxis is expecting.
+    <CartesianAxis
+      {...axisOptions}
+      className={clsx(`recharts-${axisOptions.axisType} ${axisOptions.axisType}`, axisOptions.className)}
+      viewBox={viewBox}
+      ticksGenerator={(axis: any) => getTicksOfAxis(axis, true)}
+    />
+  );
+};
 
 XAxis.displayName = 'XAxis';
 XAxis.defaultProps = {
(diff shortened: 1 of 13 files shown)


src/context/chartLayoutContext.tsx
```diff
@@ -1,29 +1,98 @@
...
-export type ChartLayoutContextType = ChartLayoutContextInitializing | ChartLayoutContextTypeLoaded;
+function getKeysForDebug(object: Record<string, unknown>) {
+  const keys = Object.keys(object);
+  if (keys.length === 0) {
+    return 'There are no available ids.';
+  }
+  return `Available ids are: ${keys}.`;
+}
+
+export const useXAxisOrThrow = (xAxisId: string | number): XAxisProps => {
```
PavelVanecek, line 71: I already see a difference in behaviour between ReferenceLine and XAxis: while ReferenceLine throws an error and refuses to render, XAxis will return `null` and quietly not render anything.

I am tempted to unify this behaviour; i.e. make all the individual components more strict, and throw early with an explanation. What do you think?

src/cartesian/XAxis.tsx
```diff
@@ -31,7 +36,24 @@ interface XAxisProps extends BaseAxisProps {
...
+export const XAxis: FunctionComponent<Props> = ({ xAxisId }: Props) => {
+  const viewBox = useViewBox();
+  const axisOptions = useXAxisOrThrow(xAxisId);
+
+  if (axisOptions == null) {
+    return null;
+  }
+
+  return (
+    // @ts-expect-error the axisOptions type is not exactly what CartesianAxis is expecting.
```
ckifer, line 48: why/what are the consequences of this?

src/context/chartLayoutContext.tsx
ckifer: If there is a strict, breaking action a user can take (and it currently breaks them but they just don't know it) they should get notified as clearly as we can tell them. I vote for unification in the spirit of https://github.com/recharts/recharts/issues/3716

src/context/chartLayoutContext.tsx
```diff
@@ -1,29 +1,98 @@
...
+  const xAxisMap = useContext(XAxisContext);
+
+  invariant(xAxisMap != null, 'Could not find xAxisMap; are you sure this is rendered inside a Recharts context?');
```
ckifer, line 74: this is what we're telling the user. What does the user know about `xAxisMap` or Recharts context, our internal implementation? I think invariants should be directed as close to what a user can do to fix something as possible. Maybe the wording of this could be adjusted just a bit depending on when it actually happens? Same with Y below

ckifer (commented):
Few questions / one suggestion

src/cartesian/XAxis.tsx
PavelVanecek: Why? This is exactly the same type hole as in the original method here: https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.tsx#L1859 except in there it's hidden behind `as any`.

The consequence? Recharts axis types need some more love. When we're done with the context refactor, it will allow us to simplify and remove lot of extra code.

src/context/chartLayoutContext.tsx
PavelVanecek: Updated the error message

src/cartesian/XAxis.tsx
ckifer: Thanks!

ckifer (approved):
Very nice, LGTM!

ckifer:
Going to merge so we can keep moving

ckifer:
https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=510

@PavelVanecek we have some interesting diffs here
https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/story/examples-areachart--simple

The ticks are off by one

PavelVanecek:
Interesting! I will look into it.

ckifer:
Thanks!

Issue #4049, opened 2024-01-10
Radar: dot props event handler onMouseOver does not work.


nikolasrieble:
# Summary
The Radar component exposes `dot` props, which can take event handlers such as onMouseOver. 
In other components, such as the line, these do work. 

```
      <Radar
        dataKey="pv"
        fill="none"
        dot={{ onMouseEnter: (e) => console.log(e) }}
        strokeWidth={2}
        stroke="blue"
      />
```

Expected behaviour: Logs created when hovering over the dots. 

Actual behaviour: Nothing happens. 

# Reproduction

https://codesandbox.io/p/sandbox/recharts-issue-template-forked-hp9rs3

AnujSharma141:
hi @ckifer .. in `Radar.tsx` there is a `renderDots` function in which we have `const customDotProps = filterProps(dot);` ... filterProps has a `includeEvents` parameter which is type boolean... 
![image](https://github.com/recharts/recharts/assets/16597292/1db81863-40fa-46c3-8c4d-52a3b7f6accc)
.. now if we do `const customDotProps = filterProps(dot, true);` the events works the way it should

ckifer:
Sounds good!

PR #4056, opened 2024-01-16
fix: Radar dot events not working!

AnujSharma141:
## Description

`Radar.tsx` has `renderDots` function which has `const customDotProps = filterProps(dot);` ... filterProps has a includeEvents parameter which is type boolean and needs to be set true 

## Related Issue

Closes [#4049 ]

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Radar.tsx
+++ b/src/polar/Radar.tsx
@@ -227,7 +227,7 @@ export class Radar extends PureComponent<Props, State> {
   renderDots(points: RadarPoint[]) {
     const { dot, dataKey } = this.props;
     const baseProps = filterProps(this.props);
-    const customDotProps = filterProps(dot);
+    const customDotProps = filterProps(dot, true);
 
     const dots = points.map((entry, i) => {
       const dotProps = {


src/polar/Radar.tsx
```diff
@@ -227,7 +227,7 @@ export class Radar extends PureComponent<Props, State> {
   renderDots(points: RadarPoint[]) {
     const { dot, dataKey } = this.props;
     const baseProps = filterProps(this.props);
-    const customDotProps = filterProps(dot);
+    const customDotProps = filterProps(dot, true);
```
nikolasrieble, line 230: Should we make the second parameter non optional to require setting it everywhere? 
Not required for this fix, but wondering whether there are other cases where the events should exist but do not. 🤔

ckifer: I always prefer to pass options as objects so they are named. Ordered params can get confusing, but yeah making it required wouldn't be the worst thing ever I think. @AnujSharma141 would you like to do that as follow up?

ckifer (approved):
LGTM

src/polar/Radar.tsx
AnujSharma141: I also think it might be possible to get the same bug again if someone misses to add the parameter. We might need to see the occasions where the filterProps function is used but I think this can be done.

ckifer: Make it required, then you'll get errors anywhere it isn't set. Resolve the errors correctly and we should be good

PR #4059, opened 2024-01-17
Fix labels on XAxis

PavelVanecek:
## Description

Found the problem! So here's why my change broke it:

First I introduced `viewBox` computed as `x: offset.left, y: offset.top,` because that's what `renderReferenceElement` does. I did this when adding `ReferenceLine`.

Later when converting XAxis, I made a mistake to reuse the same viewBox. I should have noticed that the `renderAxis` method actually has `x: 0, y: 0` hardcoded, and uses width + height from props, not from offset.

## Related Issue

following up from https://github.com/recharts/recharts/pull/4054#issuecomment-1894774034

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/XAxis.tsx
+++ b/src/cartesian/XAxis.tsx
@@ -4,7 +4,7 @@
 import type { FunctionComponent, SVGProps } from 'react';
 import React from 'react';
 import clsx from 'clsx';
-import { useViewBox, useXAxisOrThrow } from '../context/chartLayoutContext';
+import { useChartHeight, useChartWidth, useXAxisOrThrow } from '../context/chartLayoutContext';
 import { CartesianAxis } from './CartesianAxis';
 import { BaseAxisProps, AxisInterval } from '../util/types';
 import { getTicksOfAxis } from '../util/ChartUtils';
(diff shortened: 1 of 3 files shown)


ckifer (approved):
Makes sense, thanks for the quick follow up

PavelVanecek:
I tried to write a unit test for this and here is how far I got:

```
  it('should render all ticks in an example if there is enough space for them all', () => {
    const mockDomRect: DOMRect = {
      width: 20,
      height: 20,
      x: 0,
      y: 0,
      bottom: 0,
      left: 0,
      right: 0,
      top: 0,
      toJSON() {
        throw new Error('Function not implemented.');
      },
    };
    // Recharts internally uses getBoundingClientRect to decide how many labels can fit on the axis
    vi.spyOn(Element.prototype, 'getBoundingClientRect').mockReturnValue(mockDomRect);
    const { container, debug } = render(
      <AreaChart width={500} height={400} data={pageData}>
        <XAxis dataKey="name" />
      </AreaChart>,
    );
    debug();
    const allLabels = container.querySelectorAll('.recharts-xAxis .recharts-text.recharts-cartesian-axis-tick-value');
    expect.soft(allLabels).toHaveLength(7);
    const allText = Array.from(allLabels).map(el => el.textContent);
    expect.soft(allText).toHaveLength(7);
    expect(allText).toEqual(['Page A', 'Page B', 'Page C', 'Page D', 'Page E', 'Page F', 'Page G']);
  });
```

BUT it still fails because `ticks[].coordinate` is computed from d3scales which somehow doesn't work in jsdom environment again.

I will leave this one to the storybooks and visual regression tests - that did end up working fine.

ckifer:
Ah, that stinks. But yeah I'll continue to check the snapshots them especially after larger changes like these. No worries

PR #4065, opened 2024-01-19
chore: set includeEvent property in filterProps as required!

AnujSharma141:
`includeEvent` parameter of `filterProps` function in `ReactUtils.ts` which is of type `boolean` was set as optional which caused breaking events in components when not set.

## Motivation and Context

based on discussions from [#4056 ]

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -315,7 +315,7 @@ export class Area extends PureComponent<Props, State> {
     }
 
     const { dot, points, dataKey } = this.props;
-    const areaProps = filterProps(this.props);
+    const areaProps = filterProps(this.props, false);
     const customDotProps = filterProps(dot, true);
 
     const dots = points.map((entry: AreaPointItem, i: number) => {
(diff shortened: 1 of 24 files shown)


PavelVanecek:
I'm a fan! Explicit parameters are better than implicit parameters.

However I wonder. What is the difference here? The only time the parameter is used is here https://github.com/recharts/recharts/blob/25528194d18be6b8907fc87750da395a547dc385/src/util/ReactUtils.ts#L297 as

```
(includeEvents && EventKeys.includes(key))
```

And when I try out this line of code, it turns out to work the same with `false` as it does with `undefined`:

image

What is the breaking change?

AnujSharma141:
hi @PavelVanecek .. this change comes from a follow-up from a bug fix [#4056 ] that caused mouse events to not work in a component as `includeEvents` was missed to set `true` while using `filterProps`, so to avoid the same in future @ckifer and @nikolasrieble suggested to make the parameter non-optional. We want to make sure that the cases that requires `true` should not be left out here, cases with `false` and `undefined` will work same.

ckifer:
@PavelVanecek there was no breaking change, this was just always broken. Making it required just requires us to set it so if it is intended to be there we don't miss it

ckifer (approved):
looks good!

Issue #3962, opened 2023-11-14
Consistent interface: ActiveIndex


nikolasrieble:
Currently a few components offer to programatically set the activeIndex, which enables our users to customise Tooltip behaviour and force showing the Tooltip without a mouse over. 

We should extend this to all components and unify the API. 

Goal: 
- [ ] Identify all components that should have activeIndex 
- [ ] Implement activeIndex for Line 
- [ ] ...

<img width="321" alt="Screenshot 2023-11-14 at 21 25 13" src="https://github.com/recharts/recharts/assets/27995606/e9926293-d427-495d-a88f-fa3c880bbadd">

nikolasrieble:
@andrewangelle You recently worked on this, would you consider picking this up as well?

andrewangelle:
@nikolasrieble Yea, sure. I'll comment back here with a proposed list of components before I start implementing.

andrewangelle:
@nikolasrieble I'm seeing Line, Area, Bar, Radar as needing activeIndex added to their api's. Agree?

Also, in regards to:
> which enables our users to customise Tooltip behaviour and force showing the Tooltip without a mouse over.

I wouldn't say it forces showing the tooltip without a mouse over. It forces showing the active shape/dot/bar without the Tooltip. Right? Or am I mis-understanding the ask?

nikolasrieble:
Indeed, it does not force showing the Tooltip. But if you force showing the Tooltip (on the Tooltip itself https://github.com/recharts/recharts/pull/3935) you need a way to also set the activeIndex - currently partially lacking. 
So, yes, it uses the activeShape, and passes the activeIndex on to the parent state. That would be the ask.

wmy94827:
> 我看到 Line、Area、Bar、Radar 需要将 activeIndex 添加到他们的 api 中。同意？
> 
> 此外，关于：
> 
> > 这使我们的用户能够自定义工具提示行为，并在不使用鼠标的情况下强制显示工具提示。
> 
> 我不会说它会在没有鼠标的情况下强制显示工具提示。它强制在没有工具提示的情况下显示活动形状/点/条。右？还是我误解了这个问题？

+1

MadsBuchmann:
Just to chime in with a use-case for this where it would be very useful. 

[...]

Issue #3797, opened 2023-09-28
Multiple data on same x-axis for line chart does not select the right data (dot) when  x-axis data type is number


Arti011:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/highlight-zomm-line-chart-forked-gws8xl?file=/src/App.tsx:4513-4682)

### Steps to reproduce
1.) Take the example of highlight and zoom line chart.
https://recharts.org/en-US/examples/HighlightAndZoomLineChart\

2.) Add data for same x-axis ( number) with different cost values
For Example :  { name: 15, cost: 1.1, impression: 300 },
  { name: 15, cost: 1.2, impression: 300 },
  { name: 15, cost: 1.3, impression: 300 },
  { name: 15, cost: -2, impression: 300 },
  { name: 15, cost: -3, impression: 300 },

3.) Try to select the lowest data (dot ) when name is 15. It can not be selected 

### What is expected?
It should be able to select the data less than cost 0 when number is 15.

### What is actually happening?
It does not select data(dot)  less than cost 0 when number is 15.

| Environment | Info |
|---|---|
| Recharts | v2.0.6 |
| React | 17.0.1 |
| System | MAC OS |
| Browser | Google Chrome  Version-109.0.5414.87  |

<img width="811" alt="Screenshot 2023-09-28 at 13 50 13" src="https://github.com/recharts/recharts/assets/146309122/d9c73716-e761-4c2b-b231-f63f12ccbaf3">

nikolasrieble:
The tooltip is computed based on the activeIndex, which itself does not handle duplicate values well. For every position on the xAxis, it will always show just a single value - specifically the first value that is found in the array which has the same xAxis value - in this case the original value. 

I have worked around this issue by implementing a custom tooltip that is activated on the mouse over. 

Another solution that I really like is https://stackoverflow.com/questions/64501313/recharts-how-is-it-possible-to-show-only-the-value-of-one-data-point-inside-a-t/67791178#67791178 by https://github.com/Muzietto

Arti011:
@nikolasrieble  : Thanks for the comment on this. I tried to implement this but its not working . Do u have a codesandbox (working example) link ?

nikolasrieble:
[...]

PR #4070, opened 2024-01-21
Enable developers to specify tooltip position before users interact

julianna-langston:
## Description

Added a `defaultIndex` prop to <strike>all chart types</strike> EDIT: the `<Tooltip />` element. If someone specifies `defaultIndex={2}`, the 2nd data item will show a tooltip. This includes, when relevant:
- The relevant data will show its hover state
- The vertical line / crosshair appears around the data point
- The floating tooltip label will appear in an appropriate location

Additionally, this works in tandem with the accessibilityManager. If a developer provides the prop `defaultIndex={2}`, and a keyboard-only user tabs to that chart, they will jump to the starting position. If the user presses the right arrow, they will move to index 3.

A story has been added under the storybook "Tooltip" section to demonstrate this functionality

Other behavioral notes:
- Supports vertical charts
- Setting `defaultIndex` doesn't tamper with the tooltip's usual interactions. Users can still move around the chart, and the tooltip will move with them.
- If the `defaultIndex` prop is updated, the chart handles the update as expected. (Hence, changing the `componentDidUpdate` lifecycle method in the code.
- If the chart is in a ResponsiveContainer, and the user hasn't interacted with the chart, and the chart gets resized... then the tooltip stays in the correct location.
- Extra handling for scatter plots, since they're weird. (They require a crosshair, which means that we need to know the y-value of the relevant point, but... we don't have that information in the props or state... so I dig it out of the graphicalItem associated with the scatter plot. Don't love this solution, but it's the only way I could find to support scatter plots for this.)

## Related Issue

https://github.com/recharts/recharts/issues/3962

## Motivation and Context

In my library, Chart2Music, we provide extra accessibility support to data visualizations by turning charts into music. As part of that process, we provide screen reader users the ability to access and interact with data in a chart, and visually synchronize the chart to the user's actions.

While we can provide visual syncing with other charting libraries, such as react-chartjs-2, it hasn't been available for recharts.

With this change, we'll be able to improve the UX for users with a Chart2Music+Recharts tech stack.

## How Has This Been Tested?

Added automated tests cover every line that was changed in this PR.

To identify edge cases, I made sure the property would work as expected on every example in storybook.

## Types of changes

## Checklist:


ckifer:
Why should this exist on each chart type and not on `Tooltip`? Also couldn't someone set a default index with `activeIndex`

whats the difference between active index and default index? I'd like to avoid bloat to the API if we can avoid it (the functionality isn't bloat of course, but we don't need two fields that accomplish the same thing)

julianna-langston:
My understanding was that `activeTooltipIndex` was internal. In this PR, `defaultIndex` is exposed to recharts consumers, so that they can set the index they want the tooltip to appear at.

I used the term `defaultIndex`, because I didn't want the tooltip to be locked to the provided index. I was concerned that providing `activeTooltipIndex` as a prop would communicate that there's where the tooltip would be, regardless of user interaction.

nikolasrieble:
I would like to also ask for review from @PavelVanecek because he recently worked on the Tooltip.

ckifer:
@julianna-langston gotcha makes more sense. I'll probably pull it down and check out more. 

Curious about how this and the active prop on Tooltip interact

src/chart/generateCategoricalChart.tsx
```diff
@@ -1103,6 +1104,69 @@ export const generateCategoricalChart = ({
         mouseHandlerCallback: this.triggeredAfterMouseMove,
         layout: this.props.layout,
       });
+      this.displayDefaultTooltip();
+    }
+
+    displayDefaultTooltip() {
+      const { defaultIndex } = this.props;
+      if (
+        typeof this.props.defaultIndex !== 'number' ||
```
ckifer, line 1113: any particular reason for the number check? TypeScript should take care of things

src/chart/generateCategoricalChart.tsx
```diff
@@ -1103,6 +1104,69 @@ export const generateCategoricalChart = ({
...
+        defaultIndex < 0 ||
+        defaultIndex > this.state.tooltipTicks.length
+      ) {
+        return;
+      }
+
+      const tooltipElem = (Array.isArray(this.props.children) ? this.props.children : [this.props.children]).find(
+        (child: any) => child.type.name === 'Tooltip',
+      );
```
ckifer, line 1122: There is a helper function `findChildByType` that could be used here, but also we can hopefully get rid of this soon and instead use context..

src/chart/generateCategoricalChart.tsx
```diff
@@ -1103,6 +1104,69 @@ export const generateCategoricalChart = ({
...
+        ? {
+            x: independentAxisCoord,
+            y: dependentAxisCoord,
+          }
+        : {
+            y: independentAxisCoord,
+            x: dependentAxisCoord,
+          };
+
+      // If there's a scatter plot, we'll want to grab the
```
ckifer, line 1145: nit: comment is cut off

src/chart/generateCategoricalChart.tsx
```diff
@@ -1471,8 +1543,11 @@ export const generateCategoricalChart = ({
       const mouse = this.getMouseInfo(e);
       const nextState: CategoricalChartState = mouse ? { ...mouse, isTooltipActive: true } : { isTooltipActive: false };
 
-      this.setState(nextState);
-      this.triggerSyncEvent(nextState);
+      // If the dev set a defaultIndex, don't set `isTooltipActive` to false
+      if (nextState.isTooltipActive || typeof this.props.defaultIndex !== 'number') {
```
ckifer, line 1547: don't think we need the number check here

src/chart/generateCategoricalChart.tsx
```diff
@@ -1499,9 +1574,11 @@ export const generateCategoricalChart = ({
      * @return {Object} no return
      */
     handleItemMouseLeave = () => {
-      this.setState(() => ({
-        isTooltipActive: false,
-      }));
+      if (typeof this.props.defaultIndex === 'number') {
```
ckifer, line 1577: why the number checks?

src/chart/generateCategoricalChart.tsx
```diff
@@ -1523,8 +1600,10 @@ export const generateCategoricalChart = ({
     handleMouseLeave = (e: any) => {
       const nextState: CategoricalChartState = { isTooltipActive: false };
 
-      this.setState(nextState);
-      this.triggerSyncEvent(nextState);
+      if (typeof this.props.defaultIndex !== 'number') {
```
ckifer, line 1603: same here

ckifer (changes requested):
Some nits

src/chart/generateCategoricalChart.tsx
```diff
@@ -1200,8 +1269,8 @@ export const generateCategoricalChart = ({
 
         const newState = {
           ...defaultState,
-          ...keepFromPrevState,
           ...updatesToState,
+          ...keepFromPrevState,
```
PavelVanecek, line 1273: Why this change?

src/chart/generateCategoricalChart.tsx
```diff
@@ -839,6 +839,7 @@ export interface CategoricalChartProps {
   accessibilityLayer?: boolean;
   role?: string;
   tabIndex?: number;
+  defaultIndex?: number;
```
PavelVanecek, line 842: Why is this a property of the chart? Could it be on the Tooltip instead?

PavelVanecek:
Okay code changes I commented in code. Now I am thinking library changes.

First, if the only difference between `defaultIndex` and `activeTooltipIndex` is that activeTooltipIndex is internal and undocumented, then let's fix it by making one external and adding documentation and tests. If we want a different behaviour then let's thinking about a name that better captures the difference.

Second, documentation - I think this can be easily confused and needs code comments, and examples, and tests.

Third, tests again! Let's aim for 100% code coverage. Can we get all the new changes tested? Perhaps I can help with some testing techniques to get everything under control. If it's impossible to have a unit test, then let's add more storybooks

src/chart/generateCategoricalChart.tsx
julianna-langston: So, I figured that if someone wanted the tooltip to be visible from the beginning, that they'd always want it to be visible. 
So, if `defaultIndex` was set, I don't want `isTooltipActive` to ever be false.
So, in the 2 `*MouseLeave` methods, I'd only allow `isTooltipActive=false` if `defaultIndex` _wasn't_ set. 

Using `typeof defaultIndex === "number"` was just how I was checking if it had been set.

At this point, I think I was just adding too much complexity. I think the decision to make the tooltips always visible should be separate from where the tooltip should appear by default. I've backed out these checks, to reduce the scope of this PR.

src/chart/generateCategoricalChart.tsx
julianna-langston: ditto

src/chart/generateCategoricalChart.tsx
julianna-langston: To protect against runtime errors from consumers who are using .jsx, not .tsx. I can imagine someone passing a `null` or something in the prop, and I don't want this method to choke and kill the whole chart.

src/chart/generateCategoricalChart.tsx
julianna-langston: It was to help manage if the tooltip should be updated when the image gets resize. I'll rewrite it for clarity.

src/chart/generateCategoricalChart.tsx
julianna-langston: I scoped the property to the chart, since it's the chart that knows where the tooltip is, where the tickmarks are, what index the tooltip should be on, etc. Do you feel strongly that the property should live on the Tooltip?

ckifer: I have the same question - as a user I would expect to set a default tooltip index on the tooltip rather than the chart rather than the high level chart. The technical places where things live we can figure out, but the API should align with where someone would expect to see it

src/chart/generateCategoricalChart.tsx
ckifer: JS users still see types and typehints, imo if they specify something that should be a number as a string that is on them for the most part. But all good to leave the checks

src/chart/generateCategoricalChart.tsx
PavelVanecek: It looks suspicious because it switched the order of overrides. "Keep" and "updates" are now in the opposite order of what the object destructuring does. Are we 100% certain there are no unintended side effects?

src/chart/generateCategoricalChart.tsx
PavelVanecek: Yes I'm thinking the same. From outside point of view, as a developer using Recharts, I have no knowledge of interest of the inner structure. If it's a property that controls where the Tooltip shows then that's a prop of Tooltip.

Or perhaps there are other use cases for this property that I am missing?

This decision is almost impossible to undo later so we need to get it right the first time

julianna-langston:
How do people feel about these 2 approaches to `defaultIndex` v `activeIndex`:

1. Have only `activeIndex`
Consumers use `activeIndex` to fully manage where the cursor appears. The cursor will _only_ appear at the `activeIndex` position, and if consumers want the tooltip to move with a user's mouse, they'll have to write something to manage updating the index, presumably with the `onMouseMove`. (Will need to figure out a way to tie that in with the accessibility manager.)

2. Separate external (`defaultIndex`) and internal (`activeIndex`)
When consumers set `defaultIndex`, the tooltip will appear at that position at first. Users can interact with the chart, and the tooltip will follow the user. Consumers will not be managing the tooltip position, so the tooltip may wind up moving to another index than the `defaultIndex`.

Note that 2 is what I implemented in this PR so far.

src/chart/generateCategoricalChart.tsx
ckifer: Agreed, especially on the hard-to-undo part. Lets think this through..

Say we look here https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-chart-barchart--docs

The chart itself already has:
* `defaultShowTooltip` - this should be on Tooltip too TBH. Also how does this new prop interact with it?

On [Tooltip](https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-component-tooltip--docs) we have:
* `active` - this sets the the tooltip to visible always, how does this interact with the new prop? How does it interact with defaultShowTooltip? This is already confusing I think...

ckifer: From what I can tell `defaultShowTooltip` does't work...

ckifer:
Lets do some pros and cons.

Con of both of these is that they rely on indices, which is something we'll need to move away from eventually. See https://github.com/recharts/recharts/issues/3797 and related issues

> Have only activeIndex

pros:

* gives the user a lot of customization power
* they can pair this with `active` on the tooltip to display the tooltip on first render / always if they want essentially giving the same functionality as 2

cons: 

* requires more management / customization to get it working (could also be a good thing)
* relies on events, our events aren't very optimized and cause a lot of renders


> Separate external (defaultIndex) and internal (activeIndex)


pros:

* gives really nice default functionality
* no need to manage index state
* less need for devs to give examples of how to accomplish this as its done for you
* we can control events internally

cons: 

* semi one-way door decision, if we add this we need to support it
* less customizable potentially
* potential for some undefined behavior with other prop conflicts


questions:

* what happens if you change the defaultIndex with state?

julianna-langston:
@PavelVanecek @ckifer 
This PR is ready for re-review.

**Current functionality**
Consumers can use `<Tooltip defaultIndex={8} />` to set the index they'd like the tooltip to start at.

**Notes**
I kept `defaultIndex` and `activeIndex` separate, because I have a slight preference for it. I think that combining the 2 causes a more onerous experience for developers (see previous comments). That said, if people feel strongly that they should be combined, I'm happy to do it. Just let me know.

I updated the API documentation, and I added test cases. I have full code coverage for every line I changed.

bcentdev:
I need exactly this! Good job! :)

src/chart/generateCategoricalChart.tsx
```diff
@@ -1103,6 +1103,70 @@ export const generateCategoricalChart = ({
...
+          }
+        : {
+            y: independentAxisCoord,
+            x: dependentAxisCoord,
+          };
+
+      // Unlike other chart types, scatter plot's tooltip positions rely on both X and Y coordinates. Only the scatter plot
+      // element knows its own Y coordinates.
+      // If there's a scatter plot, we'll want to grab that element for an interrogation.
+      const scatterPlotElement = this.state.formattedGraphicalItems.find(
```
ckifer, line 1145: I wish we didn't have to do special handling for this, but understand why. Just letting it be known

src/chart/generateCategoricalChart.tsx
```diff
@@ -1256,8 +1320,17 @@ export const generateCategoricalChart = ({
...
     };
 
-    // eslint-disable-next-line @typescript-eslint/no-empty-function
-    componentDidUpdate() {}
+    componentDidUpdate(prevProps: CategoricalChartProps) {
+      // Check to see if the Tooltip updated. If so, re-check default tooltip position
+      if (
+        !isChildrenEqual(
+          [findChildByType(prevProps.children, Tooltip)],
+          [findChildByType(this.props.children, Tooltip)],
```
ckifer, line 1328: yikes, I wish we didn't have to do this either 😅

ckifer (approved):
Looks good! Voiced some thoughts but totally understand why things need to be the way they are

ckifer:
🚢

PR #4074, opened 2024-01-22
CartesianGrid uses default props instead of generateCategoricalChart injection, also a functional component now

PavelVanecek:
## Description

I want the CartesianGrid to read data from context, and eventually completely remove `generateCategoricalChart.renderGrid` method. But before I do that, there's some opportunities for making things simpler that do not require the context even.

## Related Issue

## Motivation and Context

No more element cloning

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/CartesianGrid.tsx
+++ b/src/cartesian/CartesianGrid.tsx
@@ -1,7 +1,7 @@
 /**
  * @fileOverview Cartesian Grid
  */
-import React, { PureComponent, ReactElement, SVGProps } from 'react';
+import React, { ReactElement, SVGProps } from 'react';
 import isFunction from 'lodash/isFunction';
 
 import { warn } from '../util/LogUtils';
@@ -166,7 +166,7 @@ function renderLineItem(option: GridLineType, props: LineItemProps) {
 }
 
 function HorizontalGridLines(props: Props) {
-  const { x, width, horizontal, horizontalPoints } = props;
+  const { x, width, horizontal = true, horizontalPoints } = props;
 
   if (!horizontal || !horizontalPoints || !horizontalPoints.length) {
     return null;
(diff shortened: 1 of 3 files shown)


src/cartesian/CartesianGrid.tsx
```diff
@@ -285,129 +290,178 @@ function VerticalStripes(props: Props) {
...
+      <VerticalGridLines {...propsIncludingDefaults} verticalPoints={verticalPoints} />
+
+      <HorizontalStripes {...propsIncludingDefaults} horizontalPoints={horizontalPoints} />
+
+      <VerticalStripes {...propsIncludingDefaults} verticalPoints={verticalPoints} />
+    </g>
+  );
 }
+
+CartesianGrid.displayName = 'CartesianGrid';
```
ckifer, line 467: we can't do this because of https://github.com/recharts/recharts/issues/3615 :(

src/chart/generateCategoricalChart.tsx
```diff
@@ -1671,34 +1668,6 @@ export const generateCategoricalChart = ({
       }
     };
 
-    verticalCoordinatesGenerator = ({ xAxis, width, height, offset }: ChartCoordinate, syncWithTicks: Boolean) =>
```
ckifer, line 1674: niceeee

ckifer (changes requested):
Looks good, but we have one large problem with the displayName static field unfortunately...

src/cartesian/CartesianGrid.tsx
PavelVanecek: The linked issue says defaultProps but code says displayName, is that related?

I have removed defaultProps static property from this component completely - as far as I can tell it's never used. (Unlike Tooltip.defaultProps).

ckifer: lol sorry, I'm doing too much context switching today. Yeah keeping the display name should be fine.

Is CartesianGrid ever found by any of the `findByType` utils? That is a case where default params will not work

ckifer: Doesn't look like it 👍🏼 https://github.com/search?q=repo%3Arecharts%2Frecharts+ByType&type=code&p=1

ckifer (approved):

PR #4075, opened 2024-01-23
CartesianGrid reads chartWidth and chartHeight from context

PavelVanecek:
## Description

Another smallish change to the CartesianGrid

## Related Issue

## Motivation and Context

No more element cloning

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:


src/cartesian/CartesianGrid.tsx
```diff
@@ -44,20 +45,26 @@ export type VerticalCoordinatesGenerator = (
...
 interface CartesianGridProps extends InternalCartesianGridProps {
+  /**
+   * The x-coordinate of grid.
+   * If left undefined, it will be computed from the chart's offset and margins.
+   */
+  x?: number;
+  /**
+   * The y-coordinate of grid.
+   * If left undefined, it will be computed from the chart's offset and margins.
+   */
```
nikolasrieble, line 66: Nice, thank you for also adding the comments! ❤️

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -30,6 +30,179 @@ describe('<CartesianGrid />', () => {
...
+            width={100}
+            height={200}
+            fill="green"
+            verticalPoints={verticalPoints}
+            horizontalPoints={horizontalPoints}
+          />
+        </Surface>,
+      );
+      const horizontalLines = container.querySelectorAll('.recharts-cartesian-grid-horizontal line');
+      expect(horizontalLines).toHaveLength(5);
```
nikolasrieble, line 87:
```suggestion
      expect(horizontalLines).toHaveLength(horizontalPoints.length);
```

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -30,6 +30,179 @@ describe('<CartesianGrid />', () => {
...
+      horizontalLines.forEach(line => {
+        expect.soft(line).toHaveAttribute('x', '1');
+        expect.soft(line).toHaveAttribute('y', '2');
+        expect.soft(line).toHaveAttribute('x1', '1');
+        expect.soft(line).toHaveAttribute('x2', '101');
+      });
+      const verticalLines = container.querySelectorAll('.recharts-cartesian-grid-vertical line');
+      expect(verticalLines).toHaveLength(4);
```
nikolasrieble, line 95:
```suggestion
      expect(verticalLines).toHaveLength(verticalPoints.length);
```

test/cartesian/CartesianGrid.spec.tsx
nikolasrieble: Refering the test data directly would make the test more robust, since it would also succeed if the data was adapted, right?

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -30,6 +30,179 @@ describe('<CartesianGrid />', () => {
...
+  describe('layout and size when inherited from parent chart', () => {
+    const exampleMargin: Margin = {
+      bottom: 1,
+      left: 2,
+      right: 4,
+      top: 8,
+    };
+    it('should put x, y, width and height as coordinates to the background', () => {
+      const { container } = render(
+        <AreaChart margin={exampleMargin} width={500} height={400}>
```
nikolasrieble, line 144: I would suggest to always use the ComposedChart, since it is the most versatile. Or, we could make this a parametrised tests and simply test for all charts?

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -30,6 +30,179 @@ describe('<CartesianGrid />', () => {
...
+    top: 14,
+  };
+
+  const expectedOffset: ChartOffset = {
+    bottom: 11,
+    brushBottom: 11,
+    height: 175,
+    left: 12,
+    right: 13,
+    top: 14,
```
nikolasrieble, line 46: What would happen if these numbers were not to align with the chartMargin? If the alignment is required, why not just reuse it directly? 
```suggestion
   ...chartMargin
```

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -30,6 +30,179 @@ describe('<CartesianGrid />', () => {
...
+  describe('layout and size when set explicitly', () => {
+    it('should put x, y, width and height as coordinates to the background', () => {
+      const { container } = render(
+        <Surface width={500} height={500}>
+          <CartesianGrid
+            x={1}
+            y={2}
+            width={100}
+            height={200}
+            fill="green"
```
nikolasrieble, line 59:
```suggestion
```

nikolasrieble: This seems unrelated to the test. I believe it would help to keep tests as simple as possible to make them easy to understand.

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -30,6 +30,179 @@ describe('<CartesianGrid />', () => {
...
+    });
+
+    it('should put x, y, width and height as coordinates to all lines', () => {
+      const { container } = render(
+        <Surface width={500} height={500}>
+          <CartesianGrid
+            x={1}
+            y={2}
+            width={100}
+            height={200}
```
nikolasrieble, line 79: It seems these values are set in every test. Should we also extract them into a shared object?, i.e. 
```suggestion
            ...DEFAULT_TEST_PARAMETERS
```

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -30,6 +30,179 @@ describe('<CartesianGrid />', () => {
...
+            fill="green"
+            verticalPoints={verticalPoints}
+            horizontalPoints={horizontalPoints}
+            horizontalFill={['red', 'black']}
+            verticalFill={['green', 'blue']}
```
nikolasrieble, line 116: The fill seems unrelated to the test case.

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -30,6 +30,179 @@ describe('<CartesianGrid />', () => {
...
+    });
+  });
+
+  describe('layout and size when inherited from parent chart', () => {
+    const exampleMargin: Margin = {
+      bottom: 1,
+      left: 2,
+      right: 4,
+      top: 8,
+    };
```
nikolasrieble, line 141: Why do we use a different exampleMargin here? Why not just the top level chartMargin?

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -30,6 +30,179 @@ describe('<CartesianGrid />', () => {
...
+      const horizontalStripes = container.querySelectorAll('.recharts-cartesian-gridstripes-horizontal rect');
+      expect(horizontalStripes).toHaveLength(5);
+      horizontalStripes.forEach(stripe => {
+        expect.soft(stripe).toHaveAttribute('x', '2');
+        expect.soft(stripe).toHaveAttribute('width', '494');
+      });
+      const verticalStripes = container.querySelectorAll('.recharts-cartesian-gridstripes-vertical rect');
+      expect(verticalStripes).toHaveLength(5);
+      verticalStripes.forEach(stripe => {
+        expect.soft(stripe).toHaveAttribute('y', '8');
```
nikolasrieble, line 200: Interesting. Would you like to suggest us to use expect.soft always? 
Should we document this as a recommendation? 

I have seen this for the first time, but it does make sense. https://vitest.dev/api/expect.html#soft

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -310,28 +483,26 @@ describe('<CartesianGrid />', () => {
           ticks: ['x', 'y', 'x'],
         };
         render(
-          <Surface width={500} height={500}>
+          <AreaChart width={300} height={200} margin={chartMargin}>
```
nikolasrieble, line 486: So, is this a breaking change? This case no longer works with the Surface?

test/cartesian/CartesianGrid.spec.tsx
```diff
@@ -430,34 +587,27 @@ describe('<CartesianGrid />', () => {
...
-
           expect(horizontalCoordinatesGenerator).toHaveBeenCalledOnce();
           expect(horizontalCoordinatesGenerator).toHaveBeenCalledWith(
             {
-              yAxis: yAxisExpected,
+              yAxis: undefined,
               width: 300,
               height: 200,
-              offset,
+              offset: { bottom: 11, brushBottom: 11, height: 175, left: 12, right: 13, top: 14, width: 275 },
```
nikolasrieble, line 610: We have extract a constant in the top `expectedOffset`. Should we also use it here?

nikolasrieble (commented):
Lots of questions, only on testing.

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: This was easier than I expected

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: I want this to be a little bit explicit, and I want the tests to fail when someone updates the margin but does not update the offset, and vice versa. If we DRY then people might miss the relationship between margin and offset here.

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: This is required because without `fill` property, CartesianGrid does not render the background element

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: This is required because without `horizontalFill` and `verticalFill`, CategoricalChart will not render the stripes

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: But it's not required in the other two tests so I removed it from there

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: I find it useful because sometimes I want to see all failures in a test to have a full picture of what is going on - it helps me debug.

I think it's a personal preference so I wouldn't make any recommendations. The tool is there, use it if you like.

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: The CategoricalChart will no longer render on its own, it now requires a context coming from one of the generateCategoricalChart.

Do you think it's a breaking change? It should be fairly easy to fix (few null checks will do the trick), I just never expected that to be a use-case.

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: Done

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: Done! This was lot less easy :(

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: Well actually it still does render independently (or with Surface, to get the svg tag parent) in this pull request. But it will break in the next pull request.

nikolasrieble: I agree that this use case is not intended, but I want to raise awareness to breaking changes. If this tests fails with the surface, it strictly speaking is a breaking change - that being said: One that I am willing to accept for moving forward with improved state management. The benefits do outweigh the cost imho. 

But others might disagree. @ckifer

ckifer: Hmm, not sure how I feel. The components that don't need context should probably render without it if all the right data is supplied. I could see people rendering things separately or doing some sort of really custom thing with the recharts building block components. 

But I agree in principle, we should move forward else we'll just keep existing where we are

ckifer (approved):
PR looks good overall, one question.

re: breaking change. Lets ship it and keep moving, I don't want to stunt momentum by overthinking things. If we get tons of reports after next release we can address

test/cartesian/CartesianGrid.spec.tsx
nikolasrieble: Thank you for the context, I was not aware. This is helpful knowledge. ❤️

src/cartesian/CartesianGrid.tsx
```diff
@@ -44,20 +45,26 @@ export type VerticalCoordinatesGenerator = (
...
-  chartWidth?: number;
-  chartHeight?: number;
 }
 
 interface CartesianGridProps extends InternalCartesianGridProps {
+  /**
+   * The x-coordinate of grid.
+   * If left undefined, it will be computed from the chart's offset and margins.
+   */
+  x?: number;
```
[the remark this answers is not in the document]
PavelVanecek: They are documented on the Recharts.org website, that makes them external in my opinion.

test/cartesian/CartesianGrid.spec.tsx
PavelVanecek: Does our website or examples show or promote usage of individual components? Most of them won't work alone anyway because of their dependency on generateCategoricalChart state. So I assumed that's a design decision.

If we do want to support individual rendering: we will document and export the ChartLayoutContext. Then we get maximum flexibility and anyone can rendering everything without gCC.

Perhaps we should start another discussion and see if that's valuable for our users?

src/cartesian/CartesianGrid.tsx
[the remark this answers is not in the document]
ckifer: Got it. Some of them are denote "this is usually set internally", but agreed that external options are external options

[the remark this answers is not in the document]
PavelVanecek: Most of everything is set internally :( I think at some point we will want to revisit our external props and get rid of some that don't make sense. Or, at least mark them as deprecated, with a warning or something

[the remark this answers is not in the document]
ckifer: agreed

test/cartesian/CartesianGrid.spec.tsx
ckifer: So I know that people have used shapes on their own (Rectangle, Curve, etc.). `Text` probably too.

Others that I'm not sure about but may be useful to render without a greater context:

* Legend
* Label (?) 

Everything else is probably ok as long as they do render _individually_ within a chart component.

ex: I want this to render an XAxis without a chart

```
<ComposedChart height={200} width={200}>
  <XAxis {...props I need for axis to render} >
</ComposedChart
```

Issue #3555, opened 2023-04-27
Add screen reader support for chart values


julianna-ciq:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Screen reader users have no access to the data underlying a chart without the chart author making it available separately.

### What does the proposed API look like?
One possible solution to this problem would be to make the tooltip into a [live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions), so that as the tooltip content changed, the screen reader will read out the updates.

ckifer:
Released in 2.12!

PR #4077, opened 2024-01-24
Enable screen reader support with accessibilityLayer and default tooltip

julianna-langston:
## Description

- Made the default tooltip renderer aware of the `accessibilityLayer`
- If the `accessibilityLayer` is set for a chart, and the default tooltip is being used, that tooltip becomes a "live region". This means that a screen reader will automatically read out the content of the tooltip as it updates.
- If the `accessibilityLayer` is set for a chart, it will have `role="application"`, which will tell the screen reader (and user) that the area is interactive.

If `accessibilityLayer` has been set, the experience for a screen reader user will be:
1. They encounter the chart, which is announced as an interactive area.
2. If the user presses arrow keys, the tooltip will move, and the screen reader will read out the changing tooltip to the user.

## Related Issue

https://github.com/recharts/recharts/issues/3555

## Motivation and Context

This helps blind people have access to the same chart that sighted people can use.

## How Has This Been Tested?

I tested the interaction with Win11 + Chrome + NVDA. More combinations will be necessary before this PR goes in.
(NVDA is a windows-only open source screen reader).

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -2211,7 +2211,7 @@ export const generateCategoricalChart = ({
         // Set tabIndex to 0 by default (can be overwritten)
         attrs.tabIndex = this.props.tabIndex ?? 0;
         // Set role to img by default (can be overwritten)
-        attrs.role = this.props.role ?? 'img';
+        attrs.role = this.props.role ?? 'application';
         attrs.onKeyDown = (e: any) => {
           this.accessibilityManager.keyboardEvent(e);
           // 'onKeyDown' is not currently a supported prop that can be passed through
@@ -2239,7 +2239,7 @@ export const generateCategoricalChart = ({
             ref={(node: HTMLDivElement) => {
               this.container = node;
             }}
-            role={this.props.accessibilityLayer ? 'application' : 'region'}
+            role={attrs.role ?? 'region'}
           >
             <Surface {...attrs} width={width} height={height} title={title} desc={desc} style={FULL_WIDTH_AND_HEIGHT}>
               {this.renderClipPath()}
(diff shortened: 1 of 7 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -2210,7 +2211,7 @@ export const generateCategoricalChart = ({
         // Set tabIndex to 0 by default (can be overwritten)
         attrs.tabIndex = this.props.tabIndex ?? 0;
         // Set role to img by default (can be overwritten)
```
ckifer, line 2213: nit: need to update comment above

src/component/DefaultTooltipContent.tsx
```diff
@@ -140,8 +142,15 @@ export const DefaultTooltipContent = <TValue extends ValueType, TName extends Na
     finalLabel = labelFormatter(label, payload);
   }
 
+  const accessibilityAttributes = accessibilityLayer
+    ? ({
+        role: 'status',
+        'aria-live': 'assertive',
+      } as HTMLAttributes<HTMLDivElement>)
```
ckifer, line 149: can we do `Partial<HTMLAttributes<HTMLDivElement>>` instead of casting?

src/component/Tooltip.tsx
```diff
@@ -69,6 +70,7 @@ export class Tooltip<TValue extends ValueType, TName extends NameType> extends P
   static displayName = 'Tooltip';
 
   static defaultProps = {
+    accessibilityLayer: false,
```
ckifer, line 73: should this be an eternally facing prop if its already set on the chart?

ckifer (changes requested):

ckifer (approved):
Lets ship this. Nits can be addressed later

Issue #4052, opened 2024-01-12
Keyboard navigation with accessibilityLayer can break


dahlie:
Could not easily reproduce the error in Codesandbox, but we run into a bug where keyboard navigation with accessibility layer didn't work correctly if we used `setState()` in our `onMouseMove()` event handler.

After some investigation, I think the bug was caused because:

1. We had defined margins for our chart using inline object as a prop: `margin={{ top: ... }}`
2. We called `setState()` in our event handler, which caused the component to re-render itself
3. https://github.com/recharts/recharts/blob/70f730083bd8e06b44d5449e0aae68301336d885/src/chart/generateCategoricalChart.tsx#L1202 was trigged due to object references being different
4. Resulting `setDetails()` call only sets `offset` prop which means `coordinateList = []` default parameter in `AccessibilityManager` which effectively resets the previously set coordinates
5. This condition prevents keyboard navigation working https://github.com/recharts/recharts/blob/70f730083bd8e06b44d5449e0aae68301336d885/src/chart/AccessibilityManager.ts#L54 

We could fix the navigation by using stable object reference for `margin` props, but a proposed fix for this issue would be to either always provide `coordinateList`, `layout` and `offset` props when calling `setDetails()` or perhaps updating the `setDetails()` function to something like this:

```
public setDetails({
    coordinateList = null,
    container = null,
    layout = null,
    offset = null,
    mouseHandlerCallback = null,
  }: InitiableOptions) {
    this.coordinateList = coordinateList ?? this.coordinateList ?? []; // Use default array here instead
    this.container = container ?? this.container;
    this.layout = layout ?? this.layout;
    this.offset = offset ?? this.offset;
    this.mouseHandlerCallback = mouseHandlerCallback ?? this.mouseHandlerCallback;

    // Keep activeIndex in the bounds between 0 and the last coordinate index
    this.activeIndex = Math.min(Math.max(this.activeIndex, 0), this.coordinateList.length - 1);
  }
```

ckifer:
Thanks for the details and looking into it @dahlie! Do you want to open a PR for this? 

I probably prefer the second approach here but I'm confused on how it fixes the issue? Wouldn't `coordinateList` still have a length of 0 with using an empty array?

ckifer:
@julianna-langston fyi

dahlie:
Sure, I'll try to create a PR for the fix 👍🏻

julianna-langston:
Let me know if there's anything I can do to help

PR #4079, opened 2024-01-25
Fix default parameters for setDetails()

dahlie:
## Description

This PR is a small bug fix for how the default parameters are handled in `setDetails()` function. With the existing implementation, it is possible that `this.coordinateList` gets overridden with `[]` since it is not a nullish value.

## Related Issue

https://github.com/recharts/recharts/issues/4052

## Motivation and Context

This fix will make accessibility layer 

## How Has This Been Tested?

Unit tests are passing and Storybook story works as expected. Change is rather small so I'm not sure if a unit test for reproducing the issue is needed (it was a bit of a corner case anyways), but let me know if the tests should also be updated.

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/AccessibilityManager.ts
+++ b/src/chart/AccessibilityManager.ts
@@ -27,13 +27,13 @@ export class AccessibilityManager {
   private offset: InitiableOptions['offset'];
 
   public setDetails({
-    coordinateList = [],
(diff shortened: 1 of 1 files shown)


ckifer (approved):
Thanks for contributing!

PR #4082, opened 2024-01-25
type improvement

alexshenker:
When using custom labels, its become a sore to have to determine/remember what props the label function provides.
I'd love it if I can call the label function and know exactly what properties exist. But currently we just see it typed as 'any' which doesn't really make sense to me.

## Description

I've separated the properties defined in `PieSectorDataItem` into their own type called `PieSectorData` for more flexible usage. 
I've also defined a new type called `PieLabelProps` to be used as the extension for `P` in `type PieLabel<P>`.
The new type includes at the minimum, what we already know is returned in PieLabel props.
Since we utilize `GeometrySector` to help define `PieLabelProps` we may see some keys defined as optional that won't actually exist in PieLabelProps - this is because `GeometrySector` was not defined for this purpose. This is an imperfect solution.
Finally, we add `PieLabelRenderProps` which contains additional necessary items as well as `[key: string]: any` in its definition, ensuring there are no breaking changes. 

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -60,7 +60,7 @@ type PieLabelLine =
   | SVGProps<SVGPathElement>
   | boolean;
 
-type PieLabelProps = PieSectorData &
+export type PieLabelProps = PieSectorData &
   GeometrySector & {
     tooltipPayload?: any;
   } & PieLabelRenderProps;
@@ -82,6 +82,7 @@ export type PieSectorData = {
   dataKey?: string;
   payload?: any[];
 };
+
 export type PieSectorDataItem = SectorProps & PieSectorData;
 
 interface PieProps extends PieDef {
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -58,7 +58,7 @@ type PieLabelLine =
   | ((props: any) => ReactElement<SVGElement>)
   | SVGProps<SVGPathElement>
   | boolean;
-export type PieLabel<P = any> =
+export type PieLabel<P extends PieSectorDataItem = PieSectorDataItem> =
   | ReactElement<SVGElement>
   | ((props: P) => ReactNode | ReactElement<SVGElement>)
   | (SVGProps<SVGTextElement> & { offsetRadius?: number })
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -33,6 +33,7 @@ import {
   PresentationAttributesAdaptChildEvent,
   AnimationDuration,
   ActiveShape,
+  GeometrySector,
 } from '../util/types';
 import { Shape } from '../util/ActiveShapeUtils';
 
@@ -58,12 +59,19 @@ type PieLabelLine =
   | ((props: any) => ReactElement<SVGElement>)
   | SVGProps<SVGPathElement>
   | boolean;
-export type PieLabel<P extends PieSectorDataItem = PieSectorDataItem> =
+
+type PieLabelProps = PieSectorData &
+  GeometrySector & {
+    tooltipPayload?: any;
+  } & PieLabelRenderProps;
+
+export type PieLabel<P extends PieLabelProps = PieLabelProps> =
   | ReactElement<SVGElement>
   | ((props: P) => ReactNode | ReactElement<SVGElement>)
   | (SVGProps<SVGTextElement> & { offsetRadius?: number })
   | boolean;
-export type PieSectorDataItem = SectorProps & {
+
+export type PieSectorData = {
   percent?: number;
   name?: string | number;
   midAngle?: number;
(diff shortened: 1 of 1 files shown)


ckifer:
> which doesn't really make sense to me

Doesn't make sense to us either 😅. For context Recharts used to be a JS library with no types - types were hacked into the library when 2.0 came out with a TON of `any` usage. Current maintainers are still dealing with the fallout of that

src/polar/Pie.tsx
```diff
@@ -58,7 +58,7 @@ type PieLabelLine =
   | ((props: any) => ReactElement<SVGElement>)
   | SVGProps<SVGPathElement>
   | boolean;
-export type PieLabel<P = any> =
+export type PieLabel<P extends PieSectorDataItem = PieSectorDataItem> =
```
ckifer, line 61: does `PieSectorDataItem` contain the full set of properties that are possibly provided in `props`?

PavelVanecek:
Agree we need types here. I would love to also see a unit test that confirms which props are actually coming in. I know that many of these `any` are hiding surprises

src/polar/Pie.tsx
alexshenker: I've come up with a more accurate type for the PieLabel props - indeed I was missing a couple of props. 

I've attempted to reuse some already-defined types to build the `labelprops`, such as `GeometrySector`, but I couldn't find perfect matches - there are often some extra props defined, albeit optional, but defined nonetheless. So bare with me as this is an imperfect solution.

Below, I've separated the properties defined in `PieSectorDataItem` to their own type that I named `PieSectorData` in order to avoid looping in `SectorProps`, which contains some items Not available in labelprops.

I added the missing `tooltipPayload` prop. tooltipPayload is defined as `any` because I could not find a type definition that fits.

Finally, I've added the type `PieLabelRenderProps` which will help avoid a breaking change, since its type definition contains `[key: string]: any` by default. 

This is an imperfect solution, but would improve dev experience until a more perfect solution can be worked on. I understand that its a huge amount of work.

```ts
interface PieSectorData {
  percent?: number;
  name?: string | number;
  midAngle?: number;
  middleRadius?: number;
  tooltipPosition?: Coordinate;
  value?: number;
  paddingAngle?: number;
  dataKey?: string;
  payload?: any[];
}

type PieLabelProps = PieSectorData &
  GeometrySector & {
    tooltipPayload?: any;
  } & PieLabelRenderProps;
  
export type PieSectorDataItem = PieSectorData & SectorProps
...
...
export type PieLabel<P extends PieLabelProps = PieLabelProps>
```

alexshenker:
> > which doesn't really make sense to me
> 
> Doesn't make sense to us either 😅. For context Recharts used to be a JS library with no types - types were hacked into the library when 2.0 came out with a TON of `any` usage. Current maintainers are still dealing with the fallout of that

Thanks for the context, I definitely understand the pains of adding types after building a project in JS 😅

src/polar/Pie.tsx
```diff
@@ -58,12 +59,19 @@ type PieLabelLine =
   | ((props: any) => ReactElement<SVGElement>)
   | SVGProps<SVGPathElement>
   | boolean;
-export type PieLabel<P = any> =
+
+export type PieLabelProps = PieSectorData &
+  GeometrySector & {
```
ckifer, line 64: interesting... This may get passed but I wonder if we actually want it on the type here. Adding something to the type is saying "this belongs here" to consumers, the geometry sector feels out of place. As does tooltipPayload...

Hmm - thoughts @PavelVanecek ?

PavelVanecek: Agree that this looks out of place but if it's there then it's part of the external API. If that was intentional or accidental - we might never know.

If it's passed now, and it's passed as `any` then that's twice as bad. Adding types is an improvement - and it allows us to later mark it as obsolete/deprecated and start working towards removing it. If we decide that's the way to go of course.

src/polar/Pie.tsx
PavelVanecek: > I've attempted to reuse some already-defined types

I am not worried about reusing types and type definitions. All of those are compiled away anyway so it doesn't increase the size of javascript that browser has to download.

The types are for other developers mostly - so it's more important that it documents the intention, than reuse.

ckifer: +1 Also not worried about re-use. @alexshenker can you double check once more that there haven't been any extra types added here? Once that is confirmed this looks fine

src/polar/Pie.tsx
nikolasrieble: We would only make our removal of any more difficult, if we also at the same time would try to type things in meaningful ways 😓 
For now I think merely having types that do not include any would be a great step forward in reliability.

ckifer:
Changed the base for this PR to 3.x. Changes give good visibility to what props are available so going to merge! Apologies for this not being in 2.x

Issue #2819, opened 2022-04-06
`minPointSize` to accept a function instead of just a number?


markhicken:
- [X] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
When a `<Bar>` chart has very small and very large values, the small values don't get rendered. Using `minPointSize={1}` fixes this, but if the point value is `0`, the bar is still rendered at 1 pixel high. It would be ideal if `minPointSize` could ignore `0` values. 

In the following image, May has `0` values, but June and July have small values. In this case, May (the `0` values) should be rendered without the `minPointSize` prop or with it set to `0` while the others should use a value of `1`. 

<img width="392" alt="Screen Shot 2022-04-06 at 1 31 29 PM" src="https://user-images.githubusercontent.com/849930/162054414-689bf2f4-c636-4855-9792-e9001bf2e8c6.png">

One less-than-ideal workaround is to set `0` values to `null`, but this prevents the bar from rendering which means the tooltip also won't work. 

### What does the proposed API look like?
`minPointSize={(dataPoint) => dataPoint === 0 ? 0 : 1}`

lenzi-erickson:
I would also like to hide the bar when 0 or null, but otherwise have a minimum height. This would solve my problem as well. +1

sam-chatrlyst:
+ 1 for this feature, would be a great benefit in applying a fully circular radius

smaspe:
@fluffybone I see you're using a stackId, do you have `minPointSize` working with stacked bars?

sam-chatrlyst:
@smaspe - I managed to get something working with a custom `shape` prop where I built a rectangle. the way I did it was for each bar, work out if the previous bars that are being stacked onto are bigger or smaller than the minPointSize and then manually adjust the x coordinate to account for the minPointSize. I used the background. you can use `x - background.x` to workout the width.

![image](https://github.com/recharts/recharts/assets/130265631/78033037-a01b-4997-b3d6-d190f3a94345)

ckifer:
https://github.com/recharts/recharts/pull/4099 🚀 

Thanks for the comments, helps with visibility. Hopefully can get this in soon

rstreefland:
Hi @ckifer, when do you expect this feature to be released? I came across it in the [documentation](https://recharts.org/en-US/api/Bar#minPointSize) and it's exactly what I need to avoid the workaround posted above. Thanks in advance!

[...]

PR #4099, opened 2024-01-30
feat(Bar): determine minPointSize by value via function prop

ckifer:
## Description

Allow a function to be set for `minPointSize` which passes each _numerical_ datapoint and index as bars are rendered. Users can make decisions based off of their data values for what to use for `minPointValue`

Helps to be able to see small datapoints when other data is large.

Ex: `(value) => value === 0 ? 0 : 5` - this does not show 0 values, but has min point as 5 for 1 or 2 or 5. 

* _only_ allow numerical data to be passed as values to this function. If values are not numbers, throw. This keeps types tight. New feature so no breaking changes. If we do throw, use an invariant for good error messaging.
* Works with stacked and non-stacked data
* Storybook entry
* Unit tests

## Related Issue

https://github.com/recharts/recharts/issues/2819

## Motivation and Context

Linked issue, better bar chart quality

## How Has This Been Tested?

Storybook, unit tests to cover all changes

image

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation. - TODO

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -6,7 +6,6 @@ import clsx from 'clsx';
 import Animate from 'react-smooth';
 import isEqual from 'lodash/isEqual';
 import isNil from 'lodash/isNil';
-
 import { Props as RectangleProps } from '../shape/Rectangle';
 import { Layer } from '../container/Layer';
 import { ErrorBar, Props as ErrorBarProps, ErrorBarDataPointFormatter } from './ErrorBar';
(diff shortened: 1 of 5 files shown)


HHongSeungWoo (approved):
It looks good to me. I also think it would be good to update the docs.

PavelVanecek:
Isn't a log scale the solution for this? Feels like a lie almost. If the data is size X then it should stay size X so that I can compare it to the bar next to it. Well at least it's optional prop.

Would it do the same work if this prop was a number instead of a function?

PavelVanecek (approved):

ckifer:
To the log scale point... kind of? But log scale on bars isn't very pretty. The prop already exists so this is not adding something where a log scale should be used (at least that isn't already there)

Issue #4093, opened 2024-01-29
Tooltip sometimes re-appears after cursor leaves chart


mrcljx:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/recharts-issue-template-forked-84t5pf)

### Steps to reproduce
- Move your mouse quickly from left (outside of the charts) to the right (outside of the charts)
- Notice how the tooltip suddenly appears

### What is expected?
- Tooltip should stay hidden

### What is actually happening?
- Tooltip is visible

| Environment | Info |
|---|---|
| Recharts | v2.11 |
| React | React 18.2 |
| System | macOS 14.2 |
| Browser | Firefox 122.0 |

---
(I increased `throttleDelay` so that this is easier to reproduce.)

What seems to be happening is:

1. `handleMouseMove` is called -> will schedule `triggeredAfterMouseMove` in 16.6ms (60fps)
2. `handleMouseLeave` is called -> `isTooltipActive = false`
3. `triggeredAfterMouseMove` is called -> `isTooltipActive = true`

`triggeredAfterMouseMove` should probably not modify `isTooltipActive`

ckifer:
relevant line of code https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.tsx#L1534 if someone wants to pick this up.

This might be an easy change but there might also be consequences of removing that state change which will need to be tested.

Thanks for the report

ckifer:
Should be resolved in next minor or patch release

ckifer:
released in 2.12

PR #4100, opened 2024-01-31
fix: Tooltip sometimes re-appears after cursor leaves chart

HHongSeungWoo:
## Description

Cancel the throttled mousemove event when the mouseleave event is triggered

## Related Issue

#4093 

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1585,6 +1585,7 @@ export const generateCategoricalChart = ({
      * @return {Null} no return
      */
     handleMouseLeave = (e: any) => {
+      this.throttleTriggeredAfterMouseMove.cancel();
       const nextState: CategoricalChartState = { isTooltipActive: false };
 
       this.setState(nextState);


ckifer:
nice, thanks for the fix

ckifer (approved):

PavelVanecek:
Unit tests would be great 🙃

ckifer:
@HHongSeungWoo I think this may have broke https://github.com/recharts/recharts/issues/4424

HHongSeungWoo:
@ckifer It seems there is an issue with the timing of measuring the size of the tooltip element. It appears to have been resolved in the 3.x branch; should we modify the master branch?

ckifer:
Only if the fix is easy - otherwise we can wait until 3.0

Appreciate you looking, this was the only change I saw for tooltip between 2.11 and 2.12 but probably a different PR

HHongSeungWoo:
It seems that the issue was caused by https://github.com/recharts/recharts/commit/830bd0040cd0554f0b252c1e2e3bb8c457668ae8 :cry:

PR #4107, opened 2024-02-01
Draft: New: ignoreZeroSegments prop on pie

imagineLife:
## Description
**A Problem**:  When rendering segments of a pie that contain a "0" value, the lint error `react/no-array-index-key` appears.  
**This solution**: provide recharts consumers the ability to "tell" recharts to "skip" rendering segments of a pie where the value(s) are zero:  
- [x] new prop
- [x] check for prop presence and element value during segment mapping, and "skip" when appropriate 

Also, this address some unforeseen lint issues in the pie "demo" file.

## Related Issue

[4004](https://github.com/recharts/recharts/issues/4004)  

## Motivation and Context
One way forward out of the index-based keys currently in use in recharts.  
If this seems like a plausible addition, this same prop could be added to many (all?) other chart types where "0" values may cause index-based-key errors.  

## How Has This Been Tested?
_...tbd..._

- [ ] ~~Breaking change (fix or feature that would cause existing functionality to change)~~

## Checklist:
  - I added a comment above the prop to show a one-liner in the storybook prop definition

### DIFF
--- a/demo/component/PieChart.tsx
+++ b/demo/component/PieChart.tsx
@@ -1,8 +1,9 @@
-import React, { Component, ReactElement, ReactSVGElement } from 'react';
+import React, { Component } from 'react';
+// eslint-disable-next-line import/no-unresolved
 import { PieChart, Pie, Legend, Cell, Tooltip, ResponsiveContainer, Sector, Label, LabelList } from 'recharts';
 import { scaleOrdinal } from 'victory-vendor/d3-scale';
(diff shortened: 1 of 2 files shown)


ckifer:
One idea for solving the key issue would be to create a key map at render time. Both the key and value would be the react key we're creating. If it doesn't exist, add it to the map. If it does exist, increment a counter by 1 for that particular key and append the count. 

This is deterministic (will happen the same every render/every time unless data order changes). Who knows, maybe that's as bad as indices but just a thought

PR #4111, opened 2024-02-02
Bump Vitest and Node and React and Typescript

PavelVanecek:
## Description

Okay this started small and then snowballed.

## Motivation and Context

I wanted to try out vitest UI and then I bumped a little bit thing here and that required bumping another and here we are.

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:


package.json
```diff
@@ -147,17 +147,17 @@
     "storybook-addon-performance": "^0.17.1",
     "terser-webpack-plugin": "^5.1.1",
     "ts-loader": "^8.0.12",
-    "typescript": "4.4.4",
+    "typescript": "5.3.x",
```
ckifer, line 150: This is the only concerning change here to me. What are the consequences of upgrading TS this far? Does this mean consumers break on lesser versions? A lot of breaking changes between 4.4 and 5.3

ckifer (commented):
Not sure we can merge this until next major version. Which maybe we should just plan and do all the work for now...

nikolasrieble:
> Not sure we can merge this until next major version. Which maybe we should just plan and do all the work for now...

How about we start a 3.x branch?

PR #4127, opened 2024-02-03
[WIP] - Recharts issue 3113 filterProps

BuddhikaGeasman:
## Types of changes

## Checklist:


src/util/svgUtils/types.ts
```diff
@@ -0,0 +1,99 @@
+export type SVGElementAttributeTypes = {
+  '*': ReadonlyArray<string>;
```
ckifer, line 2: what is `*`

src/util/svgUtils/types.ts
```diff
@@ -0,0 +1,99 @@
+export type SVGElementAttributeTypes = {
```
ckifer, line 1: This list should directly reflect https://developer.mozilla.org/en-US/docs/Web/SVG/Element. If there are things that are no longer on that list we can probably clean the list up a bit

src/util/ReactUtils.ts
```diff
@@ -282,18 +207,22 @@ export const isValidSpreadableProp = (
...
-  svgElementType?: FilteredSvgElementType,
+  svgElementType?: SVGElementType,
 ) => {
   /**
    * If the svg element type is explicitly included, check against the filtered element key map
    * to determine if there are attributes that should only exist on that element type.
    * @todo Add an internal cjs version of https://github.com/wooorm/svg-element-attributes for full coverage.
    */
-  const matchingElementTypeKeys = FilteredElementKeyMap?.[svgElementType] ?? [];
+  const svgElementAttributesInCamelCase = getSVGElementAttributesInCamelCase(svgElementAttributes);
```
ckifer, line 217: the props we accept are also in camel case for the same reason (react), why do we need to do any transformation here?

src/util/ReactUtils.ts
```diff
@@ -282,18 +207,22 @@ export const isValidSpreadableProp = (
...
+  const matchingElementTypeKeys = svgElementAttributesInCamelCase?.[svgElementType] ?? [];
+
+  const { polygon, polyline, svg, ...rest } = svgElementAttributesInCamelCase;
```
ckifer, line 220: why do these three need to be extracted out?

ckifer (commented):
Some questions. I don't think we need to do any camel case transformation...

src/util/svgUtils/types.ts
BuddhikaGeasman: It is called the universal key. i kept it as it is, i wasn't use we use it. ps://github.com/wooorm/svg-element-attributes/blob/main/index.js#L7

src/util/svgUtils/types.ts
BuddhikaGeasman: Yes i would clean it, i was planning to do an update after i received some feedback!

src/util/svgUtils/types.ts
ckifer: we could probably omit this one

src/util/ReactUtils.ts
BuddhikaGeasman: this was me, messing with things!

src/util/ReactUtils.ts
BuddhikaGeasman: I created the `getSVGElementAttributesInCamelCase` cause the `svgElementAttributes` object structured as follows. I should have simply create the util to return the converted object without having to do the conversion multiple times. 
```js
circle: [
    'alignment-baseline',
    'baseline-shift',
    'clip',
    'clip-path',
    'clip-rule',
    'color',
    ...
    ]
    ```

ckifer: You can probably do the conversion and once and keep it stored as camel case, we don't need to do it at runtime

src/util/svgUtils/types.ts
BuddhikaGeasman: done

ckifer:
Going to close this for now since follow ups haven't been addressed. Can re-open later

Issue #4103, opened 2024-01-31
Bar Chart Shape Changes on Tooltip Hover


udia76:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/mixed-bar-chart-forked-2wdtzj?file=%2Fpackage.json%3A11%2C21)

### Steps to reproduce
Open the attached sandbox.
Hover over any bar in the chart to trigger the tooltip display.
Impact: This issue affects the visual consistency of the bar chart and may confuse users relying on the shape of the bars for data interpretation.

### What is expected?
When hovering over a bar in the chart, the tooltip should appear without altering the shape or appearance of the bar. (in v2.8.0 it worked, but break in v2.9.0 and above, no available fix in the latest version)

### What is actually happening?
The original square bar appears upon hovering instead of the desired shape.

| Environment | Info |
|---|---|
| Recharts | v2.9.0 |
| React | 18.2 |
| System | Mac Sonoma 14.1.1 (23B81) |
| Browser | Chrome 120 |

---
Worked until version 2.9.0. The issue is consistently reproducible across different datasets and charts.

ckifer:
Hi it looks like the sandbox link provided doesn't exist. 

In 2.9 an activeShape prop was added so that on hover the bar shape can be changed. activeShape=false should stop you from breaking but this change was supposed to be non breaking so we'll try to figure out why

ckifer:
Got it, labeled as a bug

ckifer:
I think sandboxes have been made private by default now...

But this is a reproduction [![Edit bar-chart-with-customized-shape (forked)](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/bar-chart-with-customized-shape-dusth)

ckifer:
@andrewangelle tagging as the contributor of `activeShape` for Bar

ckifer:
fixed in 2.12!

Issue #4101, opened 2024-01-31
BarChart with stacked bars and Tooltip with shared={false} break bars color


yonatandb:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[...]

PR #4139, opened 2024-02-06
fix: activeBar shouldn't be true by default

ckifer:
## Description

We, a few months ago, introduced changes for a consistent interface of `activeShape`

https://github.com/recharts/recharts/pull/3756

In doing so we created a regression when a bar `shape` is different than the default activeBar of `true` (and therefore a regular Rectangle component)

This causes components to change to Bar on hover - see [sandbox](https://codesandbox.io/p/sandbox/mixed-bar-chart-forked-2wdtzj?file=%2Fpackage.json%3A11%2C21)

Also noticed TS freaking out in some of our tests. This.. isn't good and means that the usability of some of these things is quite terrible ...

## Related Issue

#4103 and #4101

## Motivation and Context

- Fix regression

## How Has This Been Tested?

- Tests continue to pass besides the ones that should have broken
- Test by hand, everything else is the same

image
Still works

Does not break custom shape 
image

## Types of changes

## Checklist:

### DIFF
--- a/test/component/Tooltip.spec.tsx
+++ b/test/component/Tooltip.spec.tsx
@@ -71,7 +71,7 @@ describe('<Tooltip />', () => {
     );
 
     const chart = container.querySelector('.recharts-wrapper');
-    fireEvent.mouseOver(chart, { clientX: 200, clientY: 200 });
+    fireEvent.mouseOver(chart!, { clientX: 200, clientY: 200 });
 
     // After the mouse over event over the chart, the tooltip wrapper still is not set to visible,
     // but the content is already created based on the nearest data point.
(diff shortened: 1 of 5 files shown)


test/component/Tooltip.spec.tsx
```diff
@@ -71,7 +71,7 @@ describe('<Tooltip />', () => {
     );
 
     const chart = container.querySelector('.recharts-wrapper');
-    fireEvent.mouseOver(chart, { clientX: 200, clientY: 200 });
+    fireEvent.mouseOver(chart!, { clientX: 200, clientY: 200 });
```
ckifer, line 74: TS freaking out in this file. non-null assertions are better than errors...

ckifer:
#4101 activeBar also didn't take into account the `shared` tooltip property. Instead of rendering stacked Bars when using an activeShape it just renders one singular solid bar.

ckifer:
Going to ship it to resolve the bugs, feel free to review retroactively

PR #4152, opened 2024-02-07
test(types): add snapshot mechanism for our TS files, when TS declarations are updated our snapshots need to be too

ckifer:
## Description

We need a way to test our types.

We only really care about what ends up being vended to the consumer (d.ts files). One simple (albeit maybe annoying) way to do that is to run snapshots on their contents and double check the diffs when we make changes.

We still need to make sure that we don't use features from TS that are > 4.4 so our consumers don't break (if declaration files change) see https://github.com/microsoft/TypeScript/wiki/Breaking-Changes

## Related Issue

PR https://github.com/recharts/recharts/pull/4132 and others

## Motivation and Context

Make us more confident in TS changes

## How Has This Been Tested?

* npm test
* Compare diff of snaps from 4.4.4 to 4.9.5
  * the only difference is `declare` keyword removed which is fine, no functional changes

## Types of changes

- [X] Test

## Checklist:

- [ X] My code follows the code style of this project.


ckifer:
the CI workflow has --noEmit set, we'd need to change that

PR #4164, opened 2024-02-08
Test with different React versions in a matrix build

PavelVanecek:
## Description

Let's see what the builds do.

Following https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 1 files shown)


ckifer:
why is 16.8 failing 😅

PavelVanecek:
This is far too mysterious for us to merge. Let's support the latest version only.

PavelVanecek:
All the builds were failing randomly, it doesn't indicate anything. I
couldn't figure out why

PR #4205, opened 2024-02-18
Check for typescript errors in storybooks folder in CI

PavelVanecek:
## Description

Add npm script to run TS only in storybooks + fix all storybook TS errors + add the TS storybook script to CI checks to prevent regressions.

## Motivation and Context

So that we are alerted of anything that breaks TS types on our changes

## Types of changes

## Checklist:


src/polar/Pie.tsx
```diff
@@ -80,7 +80,7 @@ export type PieSectorData = {
   value?: number;
   paddingAngle?: number;
   dataKey?: string;
-  payload?: any[];
```
ckifer, line 83: ?

ckifer (approved):

ckifer:
LGTM but confused on the type change in Pie, was that intentional?

src/polar/Pie.tsx
PavelVanecek: I think this type was wrong from the start. In the storybook it receives an entry object, not array.

Issue #4214, opened 2024-02-21
Problem with accessibilityLayer in RTL


arturjackowiak:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://release--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-accessibility--docs](https://release--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-accessibility--docs)

### Steps to reproduce
Open chart in RTL

### What is expected?
On RTL the left arrow should focus the next point and the right arrow the previous one.

### What is actually happening?
On RTL the left arrow focus the previous point and the right arrow the next one.

| Environment | Info |
|---|---|
| Recharts | v2.12.1 |
| React | Any |
| System | Any |
| Browser | Any |

---
Function keyboardEvent is not adapted to RTL

public keyboardEvent(e: any) {
    // The AccessibilityManager relies on the Tooltip component. When tooltips suddenly stop existing,
    // it can cause errors. We use this function to check. We don't want arrow keys to be processed
    // if there are no tooltips, since that will cause unexpected behavior of users.
    if (this.coordinateList.length === 0) {
      return;
    }

    switch (e.key) {
      case 'ArrowRight': {
        if (this.layout !== 'horizontal') {
          return;
        }
        this.activeIndex = Math.min(this.activeIndex   1, this.coordinateList.length - 1);
        this.spoofMouse();
        break;
      }
      case 'ArrowLeft': {
        if (this.layout !== 'horizontal') {
          return;
        }
        this.activeIndex = Math.max(this.activeIndex - 1, 0);
        this.spoofMouse();
        break;
      }
      default: {
        break;
      }
    }
  }

ckifer:
OH, "RTL" stands for "right to left" not "react testing library" 🙃

julianna-langston:
How do you "open chart in rtl"? Is that a browser setting, or are we modifying the Rechart code?

arturjackowiak:
> How do you "open chart in rtl"? Is that a browser setting, or are we modifying the Rechart code?

In my app, we use 
```
const checkIsRtl = () => document.dir === 'rtl' || document.body.dir === 'rtl'
```

and then

```
<XAxis
                            reversed={isRtl}
```

```
<YAxis
                            orientation={isRtl ? 'right' : 'left'}
```

arturjackowiak:
Screenshot from my storybook 
[...]

PR #4225, opened 2024-02-22
accessibilityLayer should respect a reversed XAxis

julianna-langston:
## Description

The AccessibilityManager (which manages the behavior for the `accessibilityLayer` prop) will take into account if there is an `<XAxis reversed={true} />` child. If so, the left/right arrow keys will reverse, so that they align with the chart's visuals.

## Related Issue

https://github.com/recharts/recharts/issues/4214

## How Has This Been Tested?

Tested manually, and added test cases 

## Types of changes

## Checklist:

- [NA] My change requires a change to the documentation.
- [NA] I have updated the documentation accordingly.
- [NA] I have added a storybook story or extended an existing story to show my changes


julianna-langston:
@ckifer I can't remember, should this target `3.x` or `master`?

src/chart/generateCategoricalChart.tsx
```diff
@@ -1099,6 +1099,8 @@ export const generateCategoricalChart = ({
         coordinateList: this.state.tooltipTicks,
         mouseHandlerCallback: this.triggeredAfterMouseMove,
         layout: this.props.layout,
+        // Assuming there's only 1 <XAxis />, check to see if it has reversed={true}. If so, this is an RTL chart
+        ltr: this.state.xAxisMap?.['0']?.reversed === false,
```
ckifer, line 1103: this will break, the `xAxisId` can be changed and therefore so can `'0'` here. Its probably better to check all possible xAxes here

ckifer:
It could exist in both, if we want to backport it to 2.x as a bugfix I can release it in 2.12.2, I have something else to release as well

src/util/types.ts
```diff
@@ -1162,6 +1162,8 @@ export interface BaseAxisProps {
   label?: string | number | ReactElement | object;
   /** The HTML element's class name */
   className?: string;
+  /** If it's reversed */
+  reversed?: boolean;
```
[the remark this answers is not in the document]
julianna-langston: It is specifiable for the XAxis and YAxis, but not the ZAxis. Is this the wrong place to put it?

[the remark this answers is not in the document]
ckifer: Z Axis doesn't make sense to have this ever. It probably doesn't matter _too_ much though. Probably fine here

ckifer (commented):
Thanks for the changes, LGTM. 

@PavelVanecek from an API perspective do you think `reversed` should exist on the base axis? Imo Z axis is a totally different category and shouldn't really share any props, but since they already do... Better to be on the individual components maybe?

PavelVanecek:
I don't understand the purpose of reversed axis, any of them. If an axis is reversed then the whole chart should be reversed, no? So then I would do `data={data.reverse()}`, why separate prop?

julianna-langston:
It's for RTL languages (Arabic, hebrew)Hebrew. The charts origin would be in the bottom right, rather than the bottom left.

ckifer:
Yeah +1. Can't just change your data order if we always assume you read left to right. 

@julianna-langston I think we should move reversed to exist only on the XAxis. We don't need to support RTL vertically correct?

julianna-langston:
So, `reversed` already exists on the YAxis (as well as XAxis). I'm not trying to add it here... I was just getting typescript errors when I tried to rely on `xAxisMap` values having the `reversed` property, and I thought the type definitions were out of sync.

Since that change is going to leak into the ZAxus, I'll see if I can find another way to wrangle the types.

ckifer:
Lol oops my bad, I was thinking it was new. 

Maybe we could type narrow somehow to the specific axis types. Imo all good to put TS-expect-error there for now so we know to fix it later

PavelVanecek:
If it's for RTL reading should this be limited to xAxis only? Should we also revert Legend?

Also, reverse implies that there's one correct direction and one reversed direction. Which one is correct one depends on culture.

So should we then have a direction=ltr|rtl as a whole chart prop?

ckifer:
It was already a thing in YAxis before any of us, without thinking of RTL, etc. you were able to reverse both Axes. 

If we had a chart prop then yeah it should only affect XAxis. 

Chart prop could take precedence over reversed. 

As for Legend, we could ask the issue requester, he would know better than us. But it's easier to make a custom legend than a custom axis

src/chart/AccessibilityManager.ts
```diff
@@ -11,6 +11,8 @@ interface InitiableOptions {
   container?: HTMLElement;
   layout?: LayoutType;
   offset?: ContainerOffset;
+  /* Is the chart oriented left-to-right? */
```
PavelVanecek, line 14:
```suggestion
  /* Is the chart oriented left-to-right? true = left-to-right, false = right-to-left */
```

src/chart/generateCategoricalChart.tsx
```diff
@@ -1099,6 +1100,8 @@ export const generateCategoricalChart = ({
         coordinateList: this.state.tooltipTicks,
         mouseHandlerCallback: this.triggeredAfterMouseMove,
         layout: this.props.layout,
+        // Assuming there's only 1 <XAxis />, check to see if it has reversed={true}. If so, this is an RTL chart
```
PavelVanecek, line 1103: What happens if there are multiple axes?

src/util/ChartUtils.ts
```diff
@@ -1337,3 +1338,5 @@ export const getTooltipItem = (graphicalItem: ReactElement, payload: any) => {
     chartType,
   };
 };
+
+export const isAxisLTR = (axisMap: XAxisMap) => Object.values(axisMap ?? {}).some(({ reversed }) => !reversed);
```
PavelVanecek, line 1342: Can we have a unit test for this new function? Should be easy - true, false, undefined, expected results.

test/chart/AccessibilityLayer.spec.tsx
```diff
@@ -172,6 +195,83 @@ describe('AccessibilityLayer', () => {
     expect(mockMouseMovements.mock.instances).toHaveLength(0);
   });
 
+  test('Left/right arrow pays attention to if the XAxis is reversed', () => {
```
PavelVanecek, line 198: <3

test/chart/AccessibilityLayer.spec.tsx
```diff
@@ -172,6 +195,83 @@ describe('AccessibilityLayer', () => {
...
+        <YAxis />
+      </AreaChart>,
+    );
+
+    const svg = container.querySelector('svg');
+    assertNotNull(svg);
+    const tooltip = container.querySelector('.recharts-tooltip-wrapper');
+
+    expect(tooltip?.textContent).toBe('');
+    expect(mockMouseMovements.mock.instances).toHaveLength(0);
```
PavelVanecek, line 216:
```suggestion
    expect(mockMouseMovements).toHaveBeenCalledTimes(0);
```

This will show better errors if it fails

PavelVanecek (approved):

src/util/types.ts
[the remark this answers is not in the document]
julianna-langston: Removed it as unnecessary

src/chart/AccessibilityManager.ts
julianna-langston: Updated

src/util/ChartUtils.ts
julianna-langston: Added unit tests

test/chart/AccessibilityLayer.spec.tsx
julianna-langston: I updated this use of `.toHaveLength`, and all the other cases I had written like it, to use `.toHaveBeenCalledTimes`

ckifer (approved):
🚀🚢🚀🚢

ckifer:
@julianna-langston just merging 3.x into your branch so that coverage generates correctly

PR #4232, opened 2024-02-25
Add codecov config

PavelVanecek:
## Description

This will fail until we have the Secret installed as described here: https://app.codecov.io/gh/recharts/recharts/new.

I don't want to copy-paste the token here because it's not meant to be public.

## Motivation and Context

I hope that colorful dashboards will be a motivating force to add more tests. And Codecov is free for open-source projects.

## How Has This Been Tested?

I tried in my sandbox repository.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,12 @@ jobs:
       - name: Lint
         run: npm run lint
       - name: Test
-        run: npm test
+        run: npm run test-coverage
       - name: Serve Storybook and run tests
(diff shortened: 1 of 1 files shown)


PavelVanecek:
Hm I probably should only run it once, outside of the matrix

PavelVanecek:
> Hm I probably should only run it once, outside of the matrix

Ah nope it looks like it works fine. Well if it works it's not broken.

ckifer (approved):

PR #4234, opened 2024-02-25
Include only relevant folders in code coverage, also upload main branch

PavelVanecek:
## Description

From my other PR I can see that codecov does require uploads from main branch too so here it is.

Also I found out that vitest has nice and easy config for coverage folders so here it is.


PavelVanecek:
Much better this way.

PR #4240, opened 2024-02-28
Legend payload context

PavelVanecek:
## Description

Okay this is how all legend payload pushing can work. No more going through formattedGraphicalItems - this should be a step towards removing formattedGraphicalItems completely!

## Related Issue

https://github.com/recharts/recharts/pull/4229/files

## Motivation and Context

1. Once we get rid of formattedGraphicalItems we can also stop reading all graphical elements (Bar, Area, Pie, ...) from the DOM and stop reading their props and their defaultProps.
2. This is also a step towards removing cloning of Pie, Bar, and other graphical elements

## How Has This Been Tested?

tests + storybooks

## Types of changes

## Checklist:

### DIFF
--- a/src/context/legendPayloadContext.tsx
+++ b/src/context/legendPayloadContext.tsx
@@ -8,12 +8,17 @@ type DispatchPayload = {
   removeSupplier: (supplier: PayloadSupplier) => void;
 };
 
-export const LegendPayloadContext = createContext<Array<PayloadSupplier>>([]);
-export const LegendPayloadDispatchContext = createContext<DispatchPayload>({
+const LegendPayloadContext = createContext<Array<PayloadSupplier>>([]);
+const LegendPayloadDispatchContext = createContext<DispatchPayload>({
   addSupplier: () => {},
   removeSupplier: () => {},
 });
 
+/**
+ * Use this at the root of a chart where you want to have Legend
+ * @param children all of the chart goes here
+ * @returns ReactNode
+ */
 export const LegendPayloadProvider = ({ children }: { children: React.ReactNode }) => {
   const [payload, setPayload] = useState<Array<PayloadSupplier>>([]);
   const addSupplier = useCallback(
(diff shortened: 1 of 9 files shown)


src/context/legendPayloadContext.tsx
```diff
@@ -0,0 +1,82 @@
+import React, { createContext, useCallback, useContext, useEffect, useState } from 'react';
+import { Payload as LegendPayload } from '../component/DefaultLegendContent';
+
+type PayloadSupplier = () => Array<LegendPayload>;
```
ckifer, line 4: nit nit: why the longform `Array<>` over `[]`

src/context/legendPayloadContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+ * @returns ReactNode
+ */
+export const LegendPayloadProvider = ({ children }: { children: React.ReactNode }) => {
+  const [payload, setPayload] = useState<Array<PayloadSupplier>>([]);
+  const addSupplier = useCallback(
+    (payloadSupplier: PayloadSupplier) => setPayload(prev => [...prev, payloadSupplier]),
+    [],
+  );
+  const removeSupplier = useCallback(
+    (payloadSupplier: PayloadSupplier) => setPayload(prev => prev.filter(p => p !== payloadSupplier)),
```
ckifer, line 29: Don't we need a deep object equality check here or some sort of identifier?

src/context/legendPayloadContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+ * Use this inside every component that is adding items to the legend.
+ * This is a little bit convoluted because it needs to do caching and has to avoid re-renders so instead of just the array of items,
+ * it accepts a function to compute the items and the input in it.
+ *
+ * @param computeLegendPayload function that accepts input and returns Legend payload array
+ * @param input input to computeLegendPayload function
+ * @returns void - this does not return anything, only use it to write legend items
+ */
+export function useLegendPayloadDispatch<Input>(
+  computeLegendPayload: (input1: Input) => Array<LegendPayload>,
```
ckifer, line 58: input1? any more descriptive name we could go with?

src/context/legendPayloadContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+  input: Input,
+): void {
+  /*
+   * So this is a bit convoluted. If you are better at React than me (many are) then please refactor.
```
ckifer, line 62: lol, you are fantastic at react. This is a hard problem and imo this seems an elegant solution. We can always refactor later if there is a decent alternative, but not seeing it at this point

src/polar/Pie.tsx
```diff
@@ -155,6 +157,26 @@ type PieComposedData = PieCoordinate & {
...
+      type: legendType,
+      value: entry.name,
+      color: entry.fill,
+      payload: entry,
+    }),
+  );
+};
+
+function SetPiePayloadLegend(props: PiePayloadInputProps): null {
+  useLegendPayloadDispatch(computeLegendPayloadFromPieData, props);
```
ckifer, line 177: we can call this from the actual component later when we refactor Pie to a function yeah?

ckifer (changes requested):
few things, but looks good overall!

src/context/legendPayloadContext.tsx
PavelVanecek: Because I started with `ReadonlyArray` and then it turns out that Legend requires these to be mutable (because of the type definition, it doesn't actually mutate anything) so I deleted the `Readonly` part.

It behaves exactly the same anyway.

src/context/legendPayloadContext.tsx
PavelVanecek: `props`? 🤷 `input` is already used and typescript requires _some_ name.

This is a) never going to be a public API and b) the name is never returned, it is only in a type definition - so I didn't spend too much thinking about the name.

src/polar/Pie.tsx
PavelVanecek: Exactly. I am not yet ready to do that because I am afraid something somewhere reads Pie's `defaultProps`.

src/context/legendPayloadContext.tsx
ckifer: yeah I know, all good I think I prefer `[]` but doesn't matter at all :p

src/context/legendPayloadContext.tsx
ckifer: yeah makes sense, nbd

src/polar/Pie.tsx
ckifer: cool beans

src/context/legendPayloadContext.tsx
ckifer: sweet, thanks for the explanation!

ckifer (approved):
🚢

Issue #4239, opened 2024-02-27
ReferenceLine always positioned at 0 if value passed is not a string, console error thrown


justrealmilk:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
https://codesandbox.io/p/sandbox/wandering-cookies-5tkqwq

### Steps to reproduce
Create a chart (I use the ComposedChart) and draw some components. Add a reference line and specify `x="string"`. The shape of my data was as follows and I was trying to draw on the label value: 
```
{
    "id": 1918,
    "date": "2023-09-27T22:14:40.000Z",
    "crawled": 934330,
    "assessed": 934330,
    "frequency": 244882,
    "value": 244882,
    "label": "27/9/2023"
}
```

### What is expected?
Draw on the label.

### What is actually happening?
"NaN" is passed to the SVG component and several errors are thrown to the console

| Environment | Info |
|---|---|
| Recharts | v2.12.1 |
| React | 18.2.0 |
| System | Windows |
| Browser | Chrome |

---
I searched around and then thought I'd try a number instead but obviously an axis of timestamps is useless so I'll need to use the label formatter in place of converting my numbies to strings first

ckifer:
Can you give some code? What field are you using for your XAxis?

justrealmilk:
I've added an example out of frustration of a second issue https://codesandbox.io/p/sandbox/wandering-cookies-5tkqwq

ckifer:
Thank you 🙏🏼. Its hard to help without anything to go off of. 

Apologies for the frustration.

If an axis is type of category (default), then the item used in ReferenceLine must be one of those categories. If axis type is number, then the item used in ReferenceLine must be a number.

Based on the example, it looks like your axis is being treated as if it was categorical even though its a number (sometimes you want that, sometimes you don't. I'm not sure what you need here). Changing the axis type to number and then adjusting the domain allows me to see all of the reference lines.

```tsx
          <XAxis
            dataKey="label"
            strokeOpacity="0"
            tick={{ fill: 'var(--foreground)', fillOpacity: '0.6' }}
            tickLine={false}
            domain={([dataMin, dataMax]) => [dataMin, dataMax + 10000000]}
            type="number"
            // tickFormatter={(value) => new Date(value).toLocaleDateString(locale)}
          />
```


> if the particular value isn't one of the chosen ticks

This is a bug for sure, it shouldn't matter if it is a tick or not

[...]

PR #4244, opened 2024-02-29
fix: reference line should not render if its coord is NaN

ckifer:
## Description

Currently when a ReferenceLine is created and the scale generates `NaN` as a coordinate of the line (X or Y), the line still renders. It renders with a value of `NaN` which is invalid and should never be in the DOM. This gets treated as 0 and is rendered on the far edge of the chart.
image

If the main coord of the ReferenceLine is `NaN` we should prefer to discard it (at least for now) rather than rendering something that shouldn't be in the DOM.

As a follow up we could warn when this happens or for the case of duplicated categories, we could attempt to render lines at both categories somehow.

## Related Issue

https://github.com/recharts/recharts/issues/4239

## Motivation and Context

Attempts to mitigate some confusing behavior

## How Has This Been Tested?

- unit tests
- manual tests

see above

## Types of changes

  -  technically breaking in terms of current functionality

## Checklist:

- [ ] My change requires a change to the documentation. --> the documentation currently states that it is discarded but it is not

### DIFF
--- a/test/cartesian/ReferenceLine/ReferenceLine.spec.tsx
+++ b/test/cartesian/ReferenceLine/ReferenceLine.spec.tsx
@@ -246,9 +246,9 @@ describe('<ReferenceLine />', () => {
     const spy = vi.fn();
     const viewBox: CartesianViewBox = { x: 1, y: 2 };
     render(
-      <BarChart width={1100} height={250}>
-        <XAxis />
-        <YAxis />
+      <BarChart width={1100} height={250} data={data}>
+        <XAxis dataKey="name" />
+        <YAxis dataKey="uv" />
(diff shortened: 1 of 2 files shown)


PavelVanecek (approved):

test/cartesian/ReferenceLine/ReferenceLine.spec.tsx
```diff
@@ -352,4 +352,18 @@ describe('<ReferenceLine />', () => {
     );
     expect(container.querySelectorAll('.recharts-reference-line-line')).toHaveLength(0);
   });
+
+  test('does not return anything when there is a duplicated category', () => {
```
PavelVanecek, line 356: <3

PR #4251, opened 2024-03-01
Radar sets legend through context

PavelVanecek:
## Description

Also I found an edge case with how Pie was implemented. It all still works in 3.x branch anyway because the Legend component has a fallback to props.payload

## Related Issue

## Motivation and Context

No more cloning, no more reading DOM

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -20,7 +20,7 @@ import { Global } from '../util/Global';
 import { polarToCartesian, getMaxRadius } from '../util/PolarUtils';
 import { isNumber, getPercentValue, mathSign, interpolateNumber, uniqueId } from '../util/DataUtils';
 import { getValueByDataKey } from '../util/ChartUtils';
-import { Payload as LegendPayload } from '../component/DefaultLegendContent';
+import type { Payload as LegendPayload } from '../component/DefaultLegendContent';
 import {
   LegendType,
   TooltipType,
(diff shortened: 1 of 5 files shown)


src/polar/Pie.tsx
```diff
@@ -647,7 +647,14 @@ export class Pie extends PureComponent<Props, State> {
       !isNumber(innerRadius as number) ||
       !isNumber(outerRadius as number)
     ) {
-      return null;
+      /*
+       * This used to render `null`, but it should still set the legend because:
+       * 1. Hidden pie still renders a legend item (albeit with inactive color)
+       * 2. if a dataKey does not match anything from props.data, then props.sectors are not defined.
+       * Legend still renders though! Behaviour (2) is arguably a bug - and we should be fixing it perhaps?
```
ckifer, line 654: confused on the behavior. You say if it doesnt match anything from data then there are no sectors - that makes sense right? No data, no sectors. Where is the bug?

src/polar/Radar.tsx
```diff
@@ -84,6 +86,29 @@ type RadarComposedData = {
   isRange: boolean;
 };
 
+function getLegendItemColor(stroke: string | undefined, fill: string): string {
+  return stroke && stroke !== 'none' ? stroke : fill;
```
ckifer, line 90: where was this done previously? getMainColorOfGraphicItem?

src/polar/Pie.tsx
ckifer: Oh.. it still renders even though that is the case. Got it

ckifer (approved):

src/polar/Radar.tsx
PavelVanecek: Correct

src/polar/Pie.tsx
PavelVanecek: Legend shows but sectors don't.

PavelVanecek: I was also confused 😬

PR #4254, opened 2024-03-02
Update code coverage badge in README.md

PavelVanecek:
README link only

The previous one still works but has stopped receiving updates in 2016.


ckifer:
Yay!!

ckifer (approved):

PR #4259, opened 2024-03-04
3.x Bugfix: Legend should prefer its own iconType prop, rather than individual legendType, if both are provided

PavelVanecek:
## Description

I found a bug that my Legend context changes have caused, and fixed it.

The bug is only in 3.x branch so we don't need a release I think.

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -163,6 +163,9 @@ type PiePayloadInputProps = {
 };
 
 const computeLegendPayloadFromPieData = ({ sectors, legendType }: PiePayloadInputProps): Array<LegendPayload> => {
+  if (sectors == null) {
+    return [];
+  }
   return sectors.map(
     (entry: PieSectorDataItem): LegendPayload => ({
       type: legendType,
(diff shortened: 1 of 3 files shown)


test/component/Legend.spec.tsx
```diff
@@ -712,26 +712,29 @@ describe('<Legend />', () => {
...
-          });
+          assertExpectedAttributes(container, selector, expectedAttributes);
         },
       );
     });
+
+    it('should prefer Legend.iconType over Line.legendType', () => {
+      const { container } = render(
+        <LineChart width={500} height={500} data={numericalData}>
+          <Legend iconType="circle" />
```
PavelVanecek, line 729: The old code, inside `getLegendProps`, has `type: legendItem.props.iconType || item.props.legendType,`. I forgot to include that in my context :/ fixed now

src/polar/Pie.tsx
```diff
@@ -163,6 +163,9 @@ type PiePayloadInputProps = {
 };
 
 const computeLegendPayloadFromPieData = ({ sectors, legendType }: PiePayloadInputProps): Array<LegendPayload> => {
+  if (sectors == null) {
```
PavelVanecek, line 166: This is because Pie started throwing weird errors when called without its `data` prop. I imagine we might want it to fail with an invariant?

ckifer: Yeah I'd imagine so. Might as well! Loud and clear is better than a broken graph

ckifer (approved):

PR #4261, opened 2024-03-05
Area, Bar, and Line set their legend props through context

PavelVanecek:
## Description

These three have to be changed in one PR because they could appear in a ComposedChart together and the Legend would be confused.

There are changes this time - the legend renders in a different order, and if a legend content is passed as a custom function then that function is called 3 times instead of once. I imagine that will get better once we remove the element cloning.

## Related Issue

## Motivation and Context

No more element cloning!

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:


src/cartesian/Area.tsx
```diff
@@ -104,6 +106,29 @@ type AreaComposedData = ChartOffset & {
   isRange: boolean;
 };
 
+function getLegendItemColor(stroke: string | undefined, fill: string): string {
+  return stroke && stroke !== 'none' ? stroke : fill;
+}
```
ckifer, line 111: nit: could this be a common function?

test/component/Legend.spec.tsx
```diff
@@ -412,7 +412,7 @@ describe('<Legend />', () => {
     });
 
     it('should pass parameters to the function', () => {
-      expect.assertions(1);
+      // expect.assertions(1);
```
ckifer, line 415: nit: can remove this entirely?

ckifer (approved):
This is huge 🎉 🎉 🎉

test/component/Legend.spec.tsx
PavelVanecek: Ah yes I will remove

src/cartesian/Area.tsx
PavelVanecek: As I am untangling the generateCategoricalChart I am finding more and more unique behaviour of each graphical component so I am by default putting everything into separate function. Later once it's done there indeed might be an opportunity to extract some of these as shared code.

PR #4263, opened 2024-03-05
Scatter sets legend through Context

PavelVanecek:
This is the last graphical child! Next I can start thinning `getLegendProps`.


ckifer (approved):
🚀

PR #4273, opened 2024-03-09
getLegendProps does not compute payload

PavelVanecek:
## Description

Now that everything is in context we can stop passing payload via props.

## Related Issue

## Motivation and Context

One less use of `formattedGraphicalItems` and one step closer to cloning the Legend.

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:


ckifer (approved):
Red has never looked so good

ckifer:
@PavelVanecek starting here we had some visual diff in chromatic. My bad for not looking at it before

https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=693

Lots of chart margin shifts including an overlapping Legend

PavelVanecek:
> @PavelVanecek starting here we had some visual diff in chromatic. My bad for not looking at it before
> 
> https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=693
> 
> Lots of chart margin shifts including an overlapping Legend

🤦 this is because `Legend` no longer reads `payload` at all, and so on payload change it does not re-render and therefore it does not trigger `getBoundingClientRect` and `onBBoxUpdate`.

How far from 3.x release are we? I am thinking if I should revert this change, or try to rethink the bounding box updates instead.

ckifer:
I would say pretty far yet. I'd do what you need do!

PR #4277, opened 2024-03-10
Add stryker for mutation testing

PavelVanecek:
So in case anyone is wondering why do we need to write tests when we have 93% coverage already? Well Stryker will help explain. It is a tool that will try to mess around with the code a little bit, and run tests. If a test fails, all good! If no tests fail however, that's a live mutant and an evidence of insufficient testing. Usually it means that there is a test that executes the code path but has no `expect` on the result.

I added instructions how to run. Here is an example of what it produces:

Stryker replaced the condition `baseValue === 'dataMin'` with `false` and a unit test failed. This is good! Introducing what is effectively a bug should make a test fail:
image

Stryker replaced the condition `hasStack || isRange` with `hasStack && isRange`, and even though there are 221 tests executing this line of code, all of those tests passed! Even though this line will show with 100% perfect code coverage, the logic is in fact not covered at all:
image

I am not planning to introduce this tool to CI or PR checks because it takes waaaay too long. On my PC it takes 15 minutes to mutation-check a single file, and 9+ hours (estimated) to run the whole repository. I still find it useful to run on my local computer to discover missing coverage.


PavelVanecek:
We talked about mutation tests on Slack before: https://recharts.slack.com/archives/C042Q5K5UDC/p1707816250405689

ckifer (approved):
Super valuable add here. Will enable us to be more confident (or not 😅) that our line coverage actually means something

ckifer:
in case you dont see this @nikolasrieble fyi

PR #4287, opened 2024-03-13
Legend bounding box update fix

PavelVanecek:
## Description

Fixing the legend offset problem introduced in https://github.com/recharts/recharts/pull/4273

The problem is that after my changes, the component that reads and updates the `boundingClientRect` and the component that re-renders when payload changes, are no longer the same component.

In this PR I have moved both the bounding box, plus payload read, to the same component so that it can call `onBBoxUpdate` again.

In my opinion the `onBBoxUpdate` should not exist in the first place because all it does is it passes the width and height back to Legend as `width` and `height` props. But that's a refactor for another day.

## Related Issue

https://github.com/recharts/recharts/pull/4273

## Motivation and Context

Fixing 3.x regression

## How Has This Been Tested?

npm test
storybook

image
image

## Types of changes

## Checklist:

### DIFF
--- a/test/component/Legend.spec.tsx
+++ b/test/component/Legend.spec.tsx
@@ -1426,7 +1426,6 @@ describe('<Legend />', () => {
         align: LegendProps['align'];
         verticalAlign: LegendProps['verticalAlign'];
         layout: LegendProps['layout'];
-        expectedStyle: string;
         expectedStyleOnSecondRender: string;
       };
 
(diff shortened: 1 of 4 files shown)


test/component/Legend.spec.tsx
```diff
@@ -1339,7 +1426,6 @@ describe('<Legend />', () => {
         align: LegendProps['align'];
         verticalAlign: LegendProps['verticalAlign'];
         layout: LegendProps['layout'];
-        expectedStyle: string;
```
PavelVanecek, line 1342: As a bonus - the Legend now renders with the final styles, right away. While before it required render, then read from DOM, and adjustment on second render.

test/helper/mockGetBoundingClientRect.ts
```diff
@@ -17,10 +18,17 @@ export function mockGetBoundingClientRect(rect: Partial<DOMRect>) {
     right: 0,
     bottom: 0,
     left: 0,
-    toJSON: () => '{}',
     ...rect,
+  };
+  Element.prototype.getBoundingClientRect = () => ({
+    ...mockDomRect,
+    toJSON: () => JSON.stringify(mockDomRect),
```
PavelVanecek, line 25: Vitest will use this method to print difference if a test fails. So having the hardcoded `{}` was not helping.

ckifer (approved):
Looks great! Thanks for the fix

test/helper/mockGetBoundingClientRect.ts
ckifer: who knew, I'll put that one in my back pocket for later

Issue #4264, opened 2024-03-05
Radial bar components shifts the filled part relative to the unfilled background


AigiulGasymova:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[...]

PR #4295, opened 2024-03-15
fix: radial bar chart cx cy parsing

ckifer:
## Description

Fix issue where background and foreground of radial bar were not the same because parsing of strings to numbers were not the same. If we want to use `parseInt` we have to use it everywhere. Call to `Number` works the same as no cast as it retains the values without messing with a radix or actually "parsing"

## Related Issue

https://github.com/recharts/recharts/issues/4264

## Motivation and Context

fix bug https://github.com/recharts/recharts/issues/4264

## How Has This Been Tested?

- unit test red green

## Types of changes

## Checklist:


PavelVanecek (approved):
Nice!

PR #4298, opened 2024-03-15
Adding types instead of any

PavelVanecek:
I am trying to make sense on what's going on with events so I figured I could add some types first.

### DIFF
--- a/src/util/PolarUtils.ts
+++ b/src/util/PolarUtils.ts
@@ -1,7 +1,7 @@
 import isNil from 'lodash/isNil';
 import { getPercentValue } from './DataUtils';
 import { parseScale, checkDomainOfScale, getTicksOfScale } from './ChartUtils';
-import { Coordinate, ChartOffset, GeometrySector } from './types';
+import { Coordinate, ChartOffset, GeometrySector, RangeObj } from './types';
 
 export const RADIAN = Math.PI / 180;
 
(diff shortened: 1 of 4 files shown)


src/util/PolarUtils.ts
```diff
@@ -141,15 +141,17 @@ const reverseFormatAngleOfSetor = (angle: number, { startAngle, endAngle }: Geom
   return angle + min * 360;
 };
 
-export const inRangeOfSector = ({ x, y }: Coordinate, sector: GeometrySector) => {
+export const inRangeOfSector = ({ x, y }: Coordinate, sector: GeometrySector): RangeObj | null => {
   const { radius, angle } = getAngleOfPoint({ x, y }, sector);
   const { innerRadius, outerRadius } = sector;
 
   if (radius < innerRadius || radius > outerRadius) {
+    // @ts-expect-error usages of this method expect it to always return RangeObj, not boolean
```
ckifer, line 149: hmm, it looks like the code kind of handles the true/false cases no? All good

ckifer (approved):

src/util/PolarUtils.ts
PavelVanecek: Either it handles or we have not yet discovered the combination of props and data that breaks.

ckifer: got it. I tried to follow it a little bit and it _looks_ like it does? But allowing null to return should cover the falsy case and returning only true seems incorrect

PR #4272, opened 2024-03-09
Fix legend size

zhonglin94:
## Description
The chart shrinks when I apply a scale on the chart container

## Related Issue
[The issue](https://github.com/recharts/recharts/issues/4246)

## Motivation and Context
The [legend size(height and width)](https://github.com/okonet/react-container-dimensions/issues/5) provided by the `getBoundingClientRect` is changed if it's parent container is transformed using scale. If the container scale is 3x, then the legend height is significantly larger than usual, which makes the chart content visibly smaller than expected.

## How Has This Been Tested?
Firstly, I tested locally using the Recharts demo
Then I build a package and test in a third party application that facing the same problem

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] I have added a storybook story or extended an existing story to show my changes
- [x] All new and existing tests passed.

Issue #4246, opened 2024-02-29
The chart shrinks when I apply a larger scale on it's parent container


zhonglin94:
Hi Recharts experts!

The chart shrinks when I apply a larger scale on the parent container. The following is a video demonstrating the issue.

Thank you so much if you could take a look and give me some suggestions!

https://github.com/recharts/recharts/assets/32061513/e8619a28-5653-4621-b2fd-b11569e8f6d2

ckifer:
It looks like things are just adjusting from the margin on the chart. You could try reducing margin on the fly maybe or starting off with less?

zhonglin94:
> It looks like things are just adjusting from the margin on the chart. You could try reducing margin on the fly maybe or starting off with less?

Hi Ckifer, I did not see any margin that has been applied to the chart when I adjust the parent container's scale. I can reproduce this in the recharts official website.

https://github.com/recharts/recharts/assets/32061513/581fa470-fa65-4071-bafd-abe32ca5af08

ckifer:
These things do not look the same. Here with `scale` or `transform: scale` there is no movement in the chart size and the margin of the chart is unchanged. This is correct behavior. 

The first video is different than the second - what are you using for scale in the first video?

zhonglin94:
> These things do not look the same. Here with `scale` or `transform: scale` there is no movement in the chart size and the margin of the chart is unchanged. This is correct behavior.
> 
> The first video is different than the second - what are you using for scale in the first video?

The chart is placed inside the react flow's node and the transform of react flow view port is "**transform: translate(-3447.32px, -1646.51px) scale(2.85055)**" for example.

ckifer:
Unless you are using an old version of recharts, which may have this issue

RrYusheng:
I set the composedchart prop like this：
![code](https://github.com/recharts/recharts/assets/31916244/a21753f7-41f6-4601-a81e-7c6020ac7736)
but it seems still not working properly：
<img width="613" alt="image" src="https://github.com/recharts/recharts/assets/31916244/3d749a57-79b9-4398-9b4c-80c8ab47f72d">

RrYusheng:
looks like Yaxis is been compressed

ckifer:
You have static heights and widths on some of your components, those might be causing issues

zhonglin94:
> Unless you are using an old version of recharts, which may have this issue

[...]

PR #4307, opened 2024-03-18
fix: legend size, use offset height and width

ckifer:
## Description

Same PR as 
https://github.com/recharts/recharts/pull/4272
but for 3.x

- use offset height and width for legend bounding box to prevent scaling issues

## Related Issue

https://github.com/recharts/recharts/issues/4246

## Motivation and Context

https://github.com/recharts/recharts/issues/4246

## How Has This Been Tested?

- storybook visually
- uni test

## Types of changes

## Checklist:

### DIFF
--- a/test/component/Legend.spec.tsx
+++ b/test/component/Legend.spec.tsx
@@ -1,6 +1,7 @@
 import React, { CSSProperties } from 'react';
 import { render, screen } from '@testing-library/react';
 import { describe, expect, it, test, vi } from 'vitest';
+import { mockHTMLElementProperty } from '../helper/mockHTMLElementProperty';
 import {
   Area,
   AreaChart,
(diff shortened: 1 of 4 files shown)


test/component/Legend.spec.tsx
```diff
@@ -771,13 +790,13 @@ describe('<Legend />', () => {
                 animationDuration: 1500,
                 animationEasing: 'ease',
                 animationId: 0,
-                bottom: 5,
+                bottom: 65,
```
ckifer, line 793: unsure why all of these changed to be honest...

ckifer:
@zhonglin94 fyi

test/component/Legend.spec.tsx
PavelVanecek: Is it because the cleanup function from `mockHTMLElementProperty` is never called? Best to do that in `afterEach`. Check out the `restoreGetBoundingClientRect` and how it's used.

PavelVanecek (changes requested):

test/component/Legend.spec.tsx
ckifer: fixed, oops. 

So in an `afterEach` doesn't work well unless we want to make the mock specific for offsetHeight and offsetWidth. I tried restoring the entire `HTMLElement.prototype` but that didn't seem to work. 

I'm sure there's a better way but not enough time to get there at least atm :P

PavelVanecek (approved):

Issue #4055, opened 2024-01-14
ErrorBar does not animate


WillemSleegers:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-cartesian-errorbar--docs](https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-cartesian-errorbar--docs)

### Steps to reproduce
Simply check out the example in the Storybook.

### What is expected?
For the error bars to animate when they appear.

### What is actually happening?
The error bars simply pop in.

| Environment | Info |
|---|---|
| Recharts | v2.10.4 |
| React | I assume the latest |
| System | macOS 14.2.1 (23C71) |
| Browser | Version 17.2.1 (19617.1.17.11.12) |

ckifer:
I think this has always been the case. Labeled for now

PR #4311, opened 2024-03-18
Add animation for ErrorBar

ForestLinSen:
## Description

Added animation for ErrorBar component

## Related Issue

[ErrorBar does not animate](https://github.com/recharts/recharts/issues/4055)

## Motivation and Context

Previously the ErrorBar didn't have animation, now animation has been introduced to make it have a smoother visual effect

## How Has This Been Tested?

It was tested locally with Storybook and validated through unit tests

https://github.com/recharts/recharts/assets/41566276/32d7f85b-3e10-47de-87a8-ac5edd018072

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/ErrorBar.tsx
+++ b/src/cartesian/ErrorBar.tsx
@@ -134,6 +134,7 @@ export function ErrorBar(props: Props) {
         {...svgProps}
       >
         {lineCoordinates.map(coordinates => {
+          const lineStyle = isAnimationActive ? { transformOrigin: `${coordinates.x1 - 5}px` } : undefined;
           return (
             <Animate
               from="scale(0, 1)"
@@ -145,7 +146,7 @@ export function ErrorBar(props: Props) {
               duration={animationDuration}
               key={`line-${coordinates.x1}-${coordinates.x2}-${coordinates.y1}-${coordinates.y2}`}
             >
-              <line {...coordinates} style={{ transformOrigin: `${coordinates.x1 - 5}px` }} />
+              <line {...coordinates} style={lineStyle} />
             </Animate>
           );
         })}
@@ -164,7 +165,7 @@ ErrorBar.defaultProps = {
   layout: 'horizontal',
   isAnimationActive: true,
   animationBegin: 0,
-  animationDuration: 400,
+  animationDuration: 200,
   animationEasing: 'ease-in-out',
 };
 ErrorBar.displayName = 'ErrorBar';
--- a/src/cartesian/ErrorBar.tsx
+++ b/src/cartesian/ErrorBar.tsx
@@ -3,10 +3,11 @@
  */
 import React, { SVGProps } from 'react';
 import invariant from 'tiny-invariant';
+import Animate from 'react-smooth';
 import { Layer } from '../container/Layer';
 import { Props as XAxisProps } from './XAxis';
 import { Props as YAxisProps } from './YAxis';
-import { D3Scale, DataKey } from '../util/types';
+import { AnimationTiming, D3Scale, DataKey } from '../util/types';
 import { filterProps } from '../util/ReactUtils';
 import { BarRectangleItem } from './Bar';
 import { LinePointItem } from './Line';
@@ -43,12 +44,30 @@ interface ErrorBarProps extends InternalErrorBarProps {
    * Only accepts a value of "x" or "y" and makes the error bars lie in that direction.
    */
   direction?: 'x' | 'y';
+  isAnimationActive?: boolean;
+  animationBegin?: number;
+  animationDuration?: number;
+  animationEasing?: AnimationTiming;
 }
 
 export type Props = SVGProps<SVGLineElement> & ErrorBarProps;
 
 export function ErrorBar(props: Props) {
-  const { offset, layout, width, dataKey, data, dataPointFormatter, xAxis, yAxis, ...others } = props;
+  const {
+    offset,
+    layout,
+    width,
+    dataKey,
+    data,
+    dataPointFormatter,
+    xAxis,
+    yAxis,
+    isAnimationActive,
+    animationBegin,
+    animationDuration,
+    animationEasing,
+    ...others
+  } = props;
   const svgProps = filterProps(others, false);
 
   invariant(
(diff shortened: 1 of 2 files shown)


src/cartesian/ErrorBar.tsx
```diff
@@ -130,5 +162,9 @@ ErrorBar.defaultProps = {
   width: 5,
   offset: 0,
   layout: 'horizontal',
+  isAnimationActive: true,
```
ckifer, line 165: this is a breaking change (but this is the 3.x branch so that's ok)

src/cartesian/ErrorBar.tsx
```diff
@@ -43,12 +44,30 @@ interface ErrorBarProps extends InternalErrorBarProps {
    * Only accepts a value of "x" or "y" and makes the error bars lie in that direction.
    */
   direction?: 'x' | 'y';
+  isAnimationActive?: boolean;
+  animationBegin?: number;
+  animationDuration?: number;
+  animationEasing?: AnimationTiming;
```
ckifer, line 50: can we add tests for each of the added properties?

1. animation can be disabled
2. animation starts after delay
3. animationDuration can be modified
4. animationEasing can be modified

ckifer (changes requested):
Looking for a few more tests here if possible.

I'd like to play with this a little before we release it to get the best possible default. From the video the appearance seems maybe not fast enough or maybe that the animation should follow the direction of the error bar? I'm not sure

Either way thanks for doing this

ForestLinSen:
> Looking for a few more tests here if possible.
> 
> I'd like to play with this a little before we release it to get the best possible default. From the video the appearance seems maybe not fast enough or maybe that the animation should follow the direction of the error bar? I'm not sure
> 
> Either way thanks for doing this

Thanks! I've updated the PR with more tests and also updated the default animation duration

PavelVanecek:
I wish the animation expanded from the dot up and down <3 that would be amazing.

src/cartesian/ErrorBar.tsx
PavelVanecek: Why is it breaking? New feature, enabled by default, nothing breaks.

PavelVanecek: We will have it in 3.x release anyway so ... it doesn't really matter. But I don't see it as breaking.

test/cartesian/ErrorBar.spec.tsx
```diff
@@ -103,4 +116,90 @@ describe('<ErrorBar />', () => {
...
+      <LineChart data={barData} width={500} height={500}>
+        <Line isAnimationActive dataKey="uv">
+          <ErrorBar dataKey="uvError" />
+        </Line>
+      </LineChart>,
+    );
+
+    assertErrorBars(container, 0);
+
+    setTimeout(() => {
```
PavelVanecek, line 131: If we use `setTimeout` in a test then failure from here will be reported in some completely different test in completely different file and mightily confuse everyone. Please use fake timers instead: https://vitest.dev/guide/mocking.html#timers

test/cartesian/ErrorBar.spec.tsx
```diff
@@ -103,4 +116,90 @@ describe('<ErrorBar />', () => {
...
+      assertErrorBars(container, 4);
+      assertAnimationStyles(container, 'scale(1, 1)', 'transform 1000ms ease-in-out 0s');
+    }, 1000);
+  });
+
+  // animationEasing
```
PavelVanecek, line 188: nit: this comment is not required, the information is in test title already (is it AI prompt? looks like AI prompt 🙃 )

PavelVanecek (changes requested):
Looks good to me except the setTimeout in test - please switch it to fake timers

src/cartesian/ErrorBar.tsx
ckifer: By breaking I mean it changed default behavior. If someone expects error bars to animate but on a patch release now they do, then I would consider that "breaking" in the sense that it should be defaulted to false until the next major release. In this case it doesn't matter though

test/cartesian/ErrorBar.spec.tsx
ForestLinSen: sry added four test names that ckifer mentioned there as todo reminders, forgot to delete the last one

test/cartesian/ErrorBar.spec.tsx
ForestLinSen: thanks for the suggestion, tried to follow the document using `useFakeTimers` and `advanceTimersByTime` but failed to pass the test, could you give me any guidance on this?

```javascript
beforeAll(() => {
  vi.useFakeTimers();
});

afterAll(() => {
  vi.useRealTimers();
});

test('Renders Error Bars with animation', () => {
    const { container } = render(
      <LineChart data={barData} width={500} height={500}>
        <Line isAnimationActive dataKey="uv">
          <ErrorBar dataKey="uvError" />
        </Line>
      </LineChart>,
    );
      
    vi.advanceTimersByTime(200);
    assertErrorBars(container, 4);
    assertAnimationStyles(container, 'scale(1, 1)', 'transform 200ms ease-in-out 0s');
});
```

ForestLinSen:
> I wish the animation expanded from the dot up and down <3 that would be amazing.

Did try that initially but couldn't find a way to implement that without using css @keyframe (I guess we shouldn't add extra css files in this project?)

ckifer:
@ForestLinSen looks like tests are failing

ForestLinSen:
> @ForestLinSen looks like tests are failing

Hi @ckifer , I've updated the unit tests, could you approve the workflow again? Thx

ckifer (approved):
Change LGTM

test/cartesian/ErrorBar.spec.tsx
```diff
@@ -103,4 +126,94 @@ describe('<ErrorBar />', () => {
 
     expect(container.querySelectorAll('.recharts-errorBar')).toHaveLength(10);
   });
+
+  test('Renders Error Bars with animation', async () => {
+    mockAnimation();
```
ckifer, line 131: afterthought and nit: but do the mock and the cleanup in a before/afterEach block here since you do it the same for each test

ForestLinSen: good point! I've updated the code

ckifer (approved):
LGTM - going to merge it. If we want to adjust how the animation happens we probably can before 3.x is released

ckifer:
@ForestLinSen these tests failed after merge, could you take a look?

PR #4314, opened 2024-03-19
Fix axis padding calculation for 0 or 1 data point. Fixes #4313

### LINKED ISSUE #4313: XAxis is missing when padding is set and there is 0 or 1 data points
I stumbled upon this when investigating #3640

Repro: https://codesandbox.io/p/sandbox/tiny-bar-chart-forked-vk9fyv?file=%2Fsrc%2FApp.tsx%3A39%2C20

You can see how XAxis disappears when padding is set. It reappears when padding is removed.

graup:
The padding calculation did not take into account the case when there is only 0 or 1 data point, in which case the calculatedPadding is Infinite, resulting in `range()` to be set to Infinite, resulting in the ticks not being drawn properly.

I have fixed this by checking for Infinite before setting calculatedPadding.

This results in at least something being drawn – behaving as if padding was undefined. An even better fix would be to calculate padding without relying on the distance between values, but that would require a big refactor.

## Related Issue
#4313

## How Has This Been Tested?

I added a test case. It is failing without my fix, showing that the x-axis is not drawn at all.

N.B. the same test case can also show the bug in #3640 if we look at the rendered .recharts-rectangle (it is empty even though there should be one bar).

## Types of changes

## Checklist:

### DIFF
--- a/test/cartesian/XAxis.spec.tsx
+++ b/test/cartesian/XAxis.spec.tsx
@@ -1,4 +1,4 @@
-import { prettyDOM, render } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import React from 'react';
 import { ScatterChart, Scatter, LineChart, Line, XAxis, YAxis, BarChart, Bar } from '../../src';
 
@@ -165,8 +165,6 @@ describe('<XAxis />', () => {
         <YAxis dataKey="y" />
       </BarChart>,
     );
-
-    console.log(prettyDOM(container));
 
     const tick = container.querySelector('.xAxis .recharts-cartesian-axis-tick-value');
     expect(parseInt(tick?.getAttribute('x') as string, 10)).toEqual(180);
--- a/test/cartesian/XAxis.spec.tsx
+++ b/test/cartesian/XAxis.spec.tsx
@@ -167,7 +167,8 @@ describe('<XAxis />', () => {
     );
 
     const tick = container.querySelector('.xAxis .recharts-cartesian-axis-tick-value');
-    expect(parseInt(tick?.getAttribute('x') as string, 10)).toEqual(180);
+    expect(tick).toBeInTheDocument();
+    expect(tick?.getAttribute('x')).toEqual('180');
   });
 
   test('Render no ticks if type is category and data is empty', () => {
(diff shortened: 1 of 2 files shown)


test/cartesian/XAxis.spec.tsx
```diff
@@ -157,6 +157,19 @@ describe('<XAxis />', () => {
...
+    const { container } = render(
+      <BarChart width={300} height={300} data={data.slice(0, 1)}>
+        <Bar dataKey="y" isAnimationActive={false} />
+        <XAxis dataKey="x" type="number" domain={['dataMin', 'dataMax']} padding="gap" />
+        <YAxis dataKey="y" />
+      </BarChart>,
+    );
+
+    const tick = container.querySelector('.xAxis .recharts-cartesian-axis-tick-value');
+    expect(parseInt(tick?.getAttribute('x') as string, 10)).toEqual(180);
```
PavelVanecek, line 170: To make the test easier to read if it fails, can we do this instead?

```
const tick = container.querySelector('.xAxis .recharts-cartesian-axis-tick-value');
assertNonNull(tick);
expect(tick.getAttribute('x')).toEqual("180");
```

PavelVanecek (changes requested):
Thanks for the fix and for the test! I would just like to see a small change in the assertion area, all good elsewhere.

Once that's done would you please also consider adding the same fix to the `3.x` branch? @ckifer will prepare a bugfix release from master branch and we will want to have the fix also included in 3.x.

PavelVanecek:
I wonder if this problem is limited to Bar charts? Do you think it's valuable to have tests covering also other types of charts?

test/cartesian/XAxis.spec.tsx
graup: Good idea. I did it like this

```ts
expect(tick).toBeInTheDocument();
expect(tick?.getAttribute('x')).toEqual('180');
```

graup:
This problem will appear in all charts that use CartesianUtils I think. Only testing one is probably enough as the others will end up in the same code path. 

Happy to make a PR for the 3.0 branch

ckifer (approved):
LGTM. Like Pavel mentioned we'll need the same PR for 3.x so we don't forget about it. 

Thank you!

graup:
Done. Thanks for the quick response guys.

I've also been looking into #3853, will push an updated version of that PR soonish

PR #4323, opened 2024-03-21
Reset mocks automatically after each test

PavelVanecek:
## Motivation and Context

I don't like having to cleanup mocks manually after each call so I fixed it.


PavelVanecek:
Hm the mocks are still not completely independent. I suspect that vitest runs things in parallel, and mocking global prototype will never be compatible with that.

PavelVanecek:
I have an idea how to fix that. In the next PR. The changes from here are still useful so please consider merging it too! I will build on top.

test/component/__snapshots__/Legend.spec.tsx.snap
```diff
@@ -633,7 +633,7 @@ exports[`<Legend /> > content as a function > should pass parameters to the func
             0,
             -Infinity,
           ],
-          "height": 290,
+          "height": 270,
```
ckifer, line 636: Aren't these the wrong values?

ckifer (approved):
If you say this is good to go I trust you - a bit confused why the values changed

PavelVanecek:
I didn't change the implementation therefore it's not that. Some of the tests fail when run together and pass when run individually which means the mocks affect each other. I'll get a better solution.

PR #4329, opened 2024-03-22
Stop using offsetWidth and offsetHeight when calling getBoundingClientRect()

PavelVanecek:
## Description

I was wondering what this is doing and ... I can't find that.

## Related Issue

## Motivation and Context

This reassignment makes it so more difficult to mock and test different components. If we can get rid of it, testing gets easier.

## How Has This Been Tested?

There was one failing test - but that test wasn't actually testing anything, it was only mocking values and then expecting the mocked values back. jsdom doesn't do rendering so it won't do any scale either.

I looked at MDN and MDN says that it's getBoundingClientRect() that actually does include scale, and offsetWidth does not! So according to MDN this code was wrong from the beginning?

I also opened storybook and compared my branch to recharts.org storybook, and then added `scale: 2` to the main chart, and then added `scale: 2` to the Legend, and I can't find any difference.

## Types of changes

## Checklist:

### DIFF
--- a/src/util/useGetBoundingClientRect.ts
+++ b/src/util/useGetBoundingClientRect.ts
@@ -30,8 +30,6 @@ export function useGetBoundingClientRect(
     (node: HTMLDivElement | null) => {
       if (node != null && node.getBoundingClientRect) {
         const box = node.getBoundingClientRect();
-        box.width = node.offsetWidth;
-        box.height = node.offsetHeight;
         if (Math.abs(box.width - lastBoundingBox.width) > EPS || Math.abs(box.height - lastBoundingBox.height) > EPS) {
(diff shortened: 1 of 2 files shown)


ckifer:
lol this is interesting. 

Someone reported a bug with scaling, they fixed it themselves via this and were happy with it, but now since mocking is hard we're reverting it?
From what I saw there is a difference although small. It's easier to tell when zoomed in a lot

PavelVanecek:
Can you pls attach a screenshot of the difference? I couldn't find any.

ckifer:
image
definitely a diff in height and `cy` (red is with offset, green is without on same screen size same chart same zoom settings)

Visually with the same screen height 
image

PavelVanecek:
Thanks, I will look into this closer.

PR #4330, opened 2024-03-22
Tooltip props are set through context instead of element cloning

PavelVanecek:
## Description

All of these components still need a separate `renderTooltip()` function because the Tooltip itself doesn't react properly to change in payload and dimensions. Which is exactly the same problem as we found when refactoring Legend. I am working on that next.

## Related Issue

## Motivation and Context

No more element cloning

## How Has This Been Tested?

npm test
storybook

## Types of changes

## Checklist:


ckifer (approved):
Nice - I was going the right direction

PR #4332, opened 2024-03-22
Add script for comparing build difference

PavelVanecek:
## Description

Script that compares build output of all files

## Motivation and Context

I had the idea of using pre-processors to compile away code and I wanted to make sure I am not breaking anything - the goal was to not change the production output.

Turns out that the setup was so complicated that it's easier to just not do that and suffer with the mocks instead.

But this script might still be useful one day.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ npm-debug.log
 .justconf/
 types
 build
+build-baseline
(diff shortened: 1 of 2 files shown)


ckifer (approved):
Pretty nifty, nice!

PR #4340, opened 2024-03-24
Independent tooltip

PavelVanecek:
## Description

Tooltip no longer requires its own specialized method, instead can be rendered together with rest of children.

## Related Issue

## Motivation and Context

I don't want Tooltip to be special cloned element, I want Tooltip to be just another child element so that it can be deeply nested and wrapped in other custom elements.

## How Has This Been Tested?

npm test
storybook

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/SunburstChart.tsx
+++ b/src/chart/SunburstChart.tsx
@@ -1,13 +1,11 @@
 import React, { useState } from 'react';
 import { scaleLinear } from 'victory-vendor/d3-scale';
 import clsx from 'clsx';
-import { findChildByType } from '../util/ReactUtils';
 import { Surface } from '../container/Surface';
 import { Layer } from '../container/Layer';
 import { Sector } from '../shape/Sector';
 import { Text } from '../component/Text';
 import { polarToCartesian } from '../util/PolarUtils';
-import { Tooltip } from '../component/Tooltip';
 import { ViewBoxContext } from '../context/chartLayoutContext';
 import { doNotDisplayTooltip, TooltipContextProvider, TooltipContextValue } from '../context/tooltipContext';
 
@@ -190,11 +188,6 @@ export const SunburstChart = ({
 
   const layerClass = clsx('recharts-sunburst', className);
 
-  function renderTooltip() {
-    const tooltipComponent = findChildByType([children], Tooltip);
-
-    return tooltipComponent;
-  }
   function getTooltipContext(): TooltipContextValue {
     if (activeNode == null) {
       return doNotDisplayTooltip;
@@ -221,7 +214,6 @@ export const SunburstChart = ({
             {children}
             <Layer className={layerClass}>{sectors}</Layer>
           </Surface>
-          {renderTooltip()}
         </div>
       </TooltipContextProvider>
     </ViewBoxContext.Provider>
--- a/src/chart/SunburstChart.tsx
+++ b/src/chart/SunburstChart.tsx
@@ -211,8 +211,8 @@ export const SunburstChart = ({
           role="region"
         >
           <Surface width={width} height={height}>
-            {children}
             <Layer className={layerClass}>{sectors}</Layer>
+            {children}
           </Surface>
         </div>
       </TooltipContextProvider>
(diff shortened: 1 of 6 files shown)


src/chart/SunburstChart.tsx
```diff
@@ -218,10 +211,9 @@ export const SunburstChart = ({
           role="region"
         >
           <Surface width={width} height={height}>
-            {children}
             <Layer className={layerClass}>{sectors}</Layer>
+            {children}
           </Surface>
-          {renderTooltip()}
```
PavelVanecek, line 224: Sunburst now renders tooltip only once so I could enable tests for it!

I will fix the other charts too if we find this approach acceptable.

Tooltip still renders everywhere even on this branch.

src/component/TooltipBoundingBox.tsx
```diff
@@ -138,18 +112,21 @@ export class TooltipBoundingBox extends PureComponent<TooltipBoundingBoxProps, S
...
-        }}
-      >
-        {children}
-      </div>
+      <g>
+        <foreignObject x="0" y="0" width="100%" height="100%" style={{ pointerEvents: 'none' }}>
+          {/* This element allow listening to the `Escape` key. // See https://github.com/recharts/recharts/pull/2925 */}
+          <div
+            // @ts-expect-error typescript library does not recognize xmlns attribute
+            xmlns="http://www.w3.org/1999/xhtml"
```
ckifer, line 120: Why is this necessary?

src/component/TooltipBoundingBox.tsx
```diff
@@ -138,18 +112,21 @@ export class TooltipBoundingBox extends PureComponent<TooltipBoundingBoxProps, S
...
-        className={cssClasses}
-        style={outerStyle}
-        ref={node => {
-          this.wrapperNode = node;
-        }}
-      >
-        {children}
-      </div>
+      <g>
+        <foreignObject x="0" y="0" width="100%" height="100%" style={{ pointerEvents: 'none' }}>
```
ckifer, line 116: Any downsides of doing this? Doesn't the X and Y matter?

src/chart/SunburstChart.tsx
ckifer, line 224: Does being inside or outside the surface make differences? I guess this is why the foreign object is needed?

ckifer (commented):
Just some curiosities, looks great!

src/component/TooltipBoundingBox.tsx
PavelVanecek: No downside as far as I can tell.

src/component/TooltipBoundingBox.tsx
PavelVanecek: I added a comment in an attempt to explain

src/chart/SunburstChart.tsx
PavelVanecek: Yes it does - that's why the `foreignObject` and `xmlns` are necessary.

I suspect that this was the original motivation to have tooltip be a special element rendered outside of the regular children flow. For some reason the original author decided against using `foreignObject` even though it was available since version 1 of each browser? Anyway.

Also, in SunburstChart now (and soon in other charts) - it depends on the order. Tooltip no longer automagically renders above everything, but it now respects the order in its children array. This might be a breaking change for some, or fix for others - I think it's something we should mention in 3.0 changelog as breaking to be on the safe side.

ckifer: So you could potentially hide the tooltip behind other elements? That doesn't seem ideal but makes sense I guess

ckifer:
@PavelVanecek  Want to call out that storybook at least throws this

`Warning: <foreignObject /> is using incorrect casing. Use PascalCase for React components, or lowercase for HTML elements.`

when rendering. Not sure if that is always or only in storybook but worth checking on

PavelVanecek:
Yeah it's a warning when this svg element is rendered in HTML context. It's a warning only and it will go away once we finish the refactor.

ckifer:
Cool thanks

PR #4342, opened 2024-03-24
Types only: change DOMRect to BoundingBox

PavelVanecek:
## Description

This change is type only - it is my assessment that no code in recharts actually reads what the `getBoundingClientRect` returns, instead it always reads only `width` and `height` which are read from `offsetWidth` and `offsetHeight`, respectively.

Not sure why is it this way but my next step will be to remove the `getBoundingClientRect` call completely - it's useless.

## Related Issue

## Motivation and Context

The whole bounding rect reading code is very suspicious.

## How Has This Been Tested?

npm run check-types

## Types of changes

## Checklist:


ckifer (approved):
Sweet! Nice being that that call is expensive

PR #4356, opened 2024-03-27
refactor: PolarRadiusAxis to context

ckifer:
## Description

Move PolarRadiusAxis to context

That... was way faster than the first one 😅 

## Related Issue

## Motivation and Context

Less cloning,more context. Can't get rid of cloning in this PR but in the next

## How Has This Been Tested?

npm test, run storybook and check for diff

image

## Types of changes

- [x] Refactor

## Checklist:


src/context/chartLayoutContext.tsx
```diff
@@ -14,6 +14,7 @@ import { TooltipContextProvider, TooltipContextValue } from './tooltipContext';
 export const XAxisContext = createContext<XAxisMap | undefined>(undefined);
 export const YAxisContext = createContext<YAxisMap | undefined>(undefined);
 export const PolarAngleAxisContext = createContext<BaseAxisMap | undefined>(undefined);
+export const PolarRadiusAxisContext = createContext<BaseAxisMap | undefined>(undefined);
```
PavelVanecek, line 17: I wonder how much overlap is there between the PolarAngleAxis and PolarRadiusAxis?

PavelVanecek (approved):

src/chart/generateCategoricalChart.tsx
```diff
@@ -1802,19 +1802,6 @@ export const generateCategoricalChart = ({
       );
     };
 
-    renderPolarAxis = (element: any, displayName: string, index: number) => {
-      const axisType = get(element, 'type.axisType');
-      const axisMap = get(this.state, `${axisType}Map`);
```
PavelVanecek, line 1807: Nice and we now got rid of this string concatenation and reading stuff from state - that was super difficult to follow

src/context/chartLayoutContext.tsx
ckifer: they're basically the same in terms of the props they get back. What I'm not sure about is if the axis map comes back differently from the generator function or not. 

but I didn't want to mess with combining them just yet. Can do that later if we need

PR #4366, opened 2024-03-30
Add types to renderActivePoints

PavelVanecek:
No changes here, only types. I'm trying to make sense of how the active dots work.

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -1,7 +1,7 @@
 /**
  * @fileOverview Area
  */
-import React, { PureComponent, ReactElement, SVGProps } from 'react';
+import React, { PureComponent, SVGProps } from 'react';
 import clsx from 'clsx';
 import Animate from 'react-smooth';
 import isFunction from 'lodash/isFunction';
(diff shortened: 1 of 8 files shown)


src/cartesian/Area.tsx
```diff
@@ -66,8 +66,8 @@ interface AreaProps extends InternalAreaProps {
   tooltipType?: TooltipType;
   connectNulls?: boolean;
   // whether have dot in line
-  activeDot?: AreaDot;
-  dot?: AreaDot;
+  activeDot?: ActiveDotType;
```
ckifer, line 69: Should we not just have a dot type? dot isn't an "active" dot right?

ckifer (commented):
Confused how "dot" and "activeDot" are both active

src/cartesian/Area.tsx
PavelVanecek: I am also confused. The original were identical so I used the same.

ckifer: lol. Are there differences between DotProps and ActiveDotType?

PavelVanecek: `any` vs `any`

ckifer (approved):

PR #4369, opened 2024-03-31
Legend renders from Context

PavelVanecek:
## Description

No more Legend cloning

## Related Issue

## Motivation and Context

Let's get rid of element cloning

## How Has This Been Tested?

npm test
storybook

There are differences in every Legend - I can't see it but chromatic insists there is. I suspect it's because text antialiasing in SVG and HTML are different? Or something. See here: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=796

image

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -44,7 +44,6 @@ import {
   getDomainOfDataByKey,
   getDomainOfItemsWithSameAxis,
   getDomainOfStackGroups,
-  getLegendProps,
   getMainColorOfGraphicItem,
   getStackedDataOfItem,
   getStackGroupsByAxisId,
@@ -1808,22 +1807,13 @@ export const generateCategoricalChart = ({
      * @return {ReactElement}            The instance of Legend
      */
     renderLegend = (): React.ReactElement => {
-      const { children, width } = this.props;
+      const { children } = this.props;
       const legendItem = findChildByType(children, Legend);
       if (!legendItem) {
         return null;
       }
-      const margin = this.props.margin || {};
-      const legendWidth: number = width - (margin.left || 0) - (margin.right || 0);
-      const props = getLegendProps({
-        legendItem,
-        legendWidth,
-      });
 
-      return cloneElement(legendItem, {
-        ...props,
-        onBBoxUpdate: this.handleLegendBBoxUpdate,
-      });
+      return legendItem;
     };
 
     /**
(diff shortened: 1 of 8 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -1802,29 +1802,6 @@ export const generateCategoricalChart = ({
       );
     };
 
-    /**
-     * Draw legend
-     * @return {ReactElement}            The instance of Legend
-     */
-    renderLegend = (): React.ReactElement => {
```
ckifer, line 1809: 🚀

src/component/Legend.tsx
```diff
@@ -157,6 +164,12 @@ export class Legend extends PureComponent<Props, State> {
   }
 
   public render() {
-    return <LegendWrapper {...this.props} />;
+    return (
+      <g>
+        <foreignObject x="0" y="0" width="100%" height="100%" style={{ pointerEvents: 'none' }}>
```
ckifer, line 169: why does this need a foreignObject? Legend is always an SVG no?

Also why `pointerEvents: 'none'`?

ckifer (commented):
I guess I don't understand the addition of foreignObject. Otherwise looks fine

src/component/Legend.tsx
PavelVanecek: Yes so Legend is a `<div>`, same as Tooltip (both the wrapper and content are HTML elements, not SVG).

When I tried run tests without the `foreignObject` I found two problems, one our tooling fails because it only works with HTMLElement but not SVGElement. That's fixable, what's worse is that jsdom throws when recharts starts reading offsetWidth and offsetHeight:

```
TypeError: 'get offsetWidth' called on an object that is not a valid instance of HTMLElement.
    at SVGElement.get offsetWidth [as offsetWidth] (recharts/node_modules/jsdom/lib/jsdom/living/generated/HTMLElement.js:441:15)
    at recharts/src/util/useGetBoundingClientRect.ts:33:23
```

as jsdom is out of our reach, and I am not ready to refactor recharts to not use offsetWidth, I googled and added foreignObject and it all works.

Also without foreignObject the tooltip doesn't display at all in Sankey and other charts that do not have the `renderTooltip` method anymore. I suspect it's because of the offsetWidth problem but I didn't dig deeper into why it doesn't work.

Here's an example where embedded HTML breaks the SVG: https://codesandbox.io/p/sandbox/foreignobject-in-svg-ymvzlx?file=%2Findex.html%3A28%2C40

And this is what it looks like in my Firefox:
image

Now I suspect that this is the original reason of why both tooltip and legend are rendered outside of `<Surface>` (the `<svg>` root) in all charts. Since we don't want element cloning anymore we're going to have to do with foreignObject.

Why pointer-events: none? When I was fixing the Tooltip I noticed that mouse hover no longer works. I attributed that to the fact that there's now a `foreignObject` element stretched on top of the whole thing and I added pointer-events: none because the blank element is not meant to catch any mouse interaction in the first place and that fixed it. I can't get it to reproduce in codesandbox though.

ckifer: Thank you for the explanation!

We should make sure it's documented somewhere (comments I guess?) why we do this as I'm sure I'll ask the same question again in 2 years 😂

ckifer (approved):

Issue #4020, opened 2023-12-13
Weired behaviour when calculating yAxis `domain: ["auto", "auto"]`


liubko:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[...]

PR #4370, opened 2024-03-31
refactor: move recharts scale code into recharts repository

ForestLinSen:
## Description

Edit: This PR will now focus only on the code movement from `recharts-scale` as per the discussion

~~This PR is for an issue where minor changes in data could affect the Y-axis domain significantly(an example mentioned in the issue #4020 [here](https://codesandbox.io/p/sandbox/bar-chart-stacked-by-sign-forked-3hnrht?file=%2Fsrc%2FApp.tsx%3A22%2C17)). Various parameters impact this, but a relatively straightforward solution, after quite a few experiments, is to adjust the `stepRatioScale`.~~

~~This value was previously hardcoded to 0.05, it now supports adjustments among 0.01, 0.03, or 0.05. A value closer to 0 makes the Y-axis domain closely match the data's maximum value (it cannot be 0 to avoid division by zero errors). Higher values can cause the Y-axis to scale to excessively high values.~~

~~Since this change needs to modify the code in `recharts-scale`, after discussion with Coltin, I've moved the necessary code parts for this PR from `recharts-scale` directly into this project to make it easier to test and manage.~~

## Related Issue

[Weired behaviour when calculating yAxis domain: ["auto", "auto"]](https://github.com/recharts/recharts/issues/4020), 

## How Has This Been Tested?

Locally tested and unit test

https://github.com/recharts/recharts/assets/41566276/a1329272-e60b-4350-babf-308911c8bfc7

## Types of changes

## Checklist:


src/chart/generateCategoricalChart.tsx
```diff
@@ -1226,6 +1229,7 @@ export const generateCategoricalChart = ({
               props: nextProps,
               ...defaultState,
               updateId: 0,
+              // stepRatioControl
```
ckifer, line 1232: does this comment need removed?

src/chart/generateCategoricalChart.tsx
```diff
@@ -10,6 +10,7 @@ import clsx from 'clsx';
 // eslint-disable-next-line no-restricted-imports
 import type { DebouncedFunc } from 'lodash';
 import invariant from 'tiny-invariant';
+import { StepRatioControl } from '../util/scale/getNickTickValues';
```
ckifer, line 13: getNickTickValues? Do you mean getNiceTickValues?

storybook/stories/API/chart/BarChart.stories.tsx
```diff
@@ -47,6 +47,7 @@ export const BarInBar = {
 
 export const Stacked = {
   render: (args: Args) => {
+    console.log(args);
```
ckifer, line 50: nit: can remove the log here

storybook/stories/API/props/ChartProps.ts
```diff
@@ -181,6 +181,16 @@ toggling between multiple dataKey.`,
       category: 'General',
     },
   },
+  stepRatioControl: {
+    description: 'The value to control the range of y axis domain',
```
ckifer, line 185: I'm not sure if this is enough description for people to understand what this is and what it accomplishes

test/util/ChartUtils.spec.tsx
```diff
@@ -349,6 +349,22 @@ describe('getTicksOfScale', () => {
 
     expect(result?.niceTicks).toEqual([0, 0.25, 0.5, 0.75, 1]);
   });
+
+  it('of linear scale with specified domain with stepRatioControl of 0.03', () => {
```
ckifer, line 353: nit: I can't understand from the title what this test is doing

ckifer (changes requested):
Appreciate you digging into this. Looks like a few things need fixed/updated and the build is breaking. I'd really like to see if this affects the storybook diffs at all before merging, I can kick that off at some point.

Nice find, but does giving the user the ability to tune this parameter really solve the problem? By default the same thing will happen as before. Now the user has to understand this stepRatioControl prop and really know what it is doing... I barely know what it is doing 😅. There doesn't seem to be any better default we can change this to rather than making a new prop?

ForestLinSen:
> Appreciate you digging into this. Looks like a few things need fixed/updated and the build is breaking. I'd really like to see if this affects the storybook diffs at all before merging, I can kick that off at some point.
> 
> Nice find, but does giving the user the ability to tune this parameter really solve the problem? By default the same thing will happen as before. Now the user has to understand this stepRatioControl prop and really know what it is doing... I barely know what it is doing 😅. There doesn't seem to be any better default we can change this to rather than making a new prop?

thanks for the comments, still need to fix a few things later, but just want to discuss the default value first - I'm a little bit hesitant to change the default value, the number was set to 0.05 because we wanted to have nice numbers, like [100, 200, 300] instead of [80, 180, 280] (the values may be different but you get the point). It's only in cases where the data is large that minor changes cause significant shifts in the y domain, for smaller values it generally works well. So I'm inclined to set this as the default number, and give users the option to adjust it should they encounter any edge cases. What are your thoughts on this?

ckifer:
Having the option to change the behavior is better than not having the option to change it so to me, all good, it's a good increment. Ideally we would find a more intuitive way to adjust it - this number seems relatively arbitrary and isn't something easy to comprehend without knowing implementation details. 

No matter what we need the documentation to be crystal clear as to what this actually changes

ckifer (commented):
I think this is a great start (compared to JS code and no tests in the other repo). @PavelVanecek any thoughts or concerns here?

test/util/scale/utils.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, expect } from 'vitest';
+import { compose, curry, map, memoize, range, reverse } from '../../../src/util/scale/util/utils';
+
+describe('functional utilities', () => {
+  describe('curry', () => {
```
nikolasrieble, line 5: 💭 Love the tests, I am a bit sceptical as to whether we need these helper functions, and whether we could just get by with a simpler solution that just might be more verbose - but easier to read and maintain.

nikolasrieble (approved):
Pretty great work. It adds value for users, and is nicely documented.

I would like a separate example story just for this parameter. I.e. a chart with multiple yAxis, the same dataset, just showing how the values differ. Not a requirement. 

Code wise, I wonder whether some of the code could be simplified, but also not blocking.

src/util/scale/getNiceTickValues.ts
```diff
@@ -0,0 +1,216 @@
...
+export const calculateStep = (
+  min: number,
+  max: number,
+  tickCount: number,
+  allowDecimals: boolean,
+  correctionFactor = 0,
+  stepRatioControl: StepRatioControl = 0.05,
+): any => {
+  // dirty hack (for recharts' test)
+  if (!Number.isFinite((max - min) / (tickCount - 1))) {
```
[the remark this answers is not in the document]
ckifer: this isn't his code, this was copied from `recharts-scale` and the original author from that repo

test/util/scale/utils.spec.ts
ckifer: These were all brought over from recharts-scale

Your concern might be true (we don't need these), but we should get more test coverage first to ensure nothing breaks

src/util/scale/util/arithmetic.ts
```diff
@@ -0,0 +1,100 @@
...
+ * @param  {Decimal} step  步长
+ * @return {Array}         若干数值
+ */
+function rangeStep(start: Decimal, end: Decimal, step: Decimal) {
+  let num = new Decimal(start);
+  let i = 0;
+  const result = [];
+
+  // magic number to prevent infinite loop
+  while (num.lt(end) && i < 100000) {
```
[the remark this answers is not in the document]
ckifer: yeah this is weird, but also a recharts-scale artifact https://github.com/recharts/recharts-scale/blob/master/src/util/arithmetic.js#L46

src/util/scale/util/utils.ts
```diff
@@ -0,0 +1,99 @@
+const identity = (i: any) => i;
+
+export const PLACE_HOLDER = {
```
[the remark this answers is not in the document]
ckifer: recharts scale reference https://github.com/recharts/recharts-scale/blob/master/src/util/utils.js#L4

nikolasrieble:
Just sharing because I forgot to mention: The best way to do this change would have been a two step approach: 
(1) First PR only to move the recharts-scale into the repo. Add tests, etc.
(2) Second PR to build a new feature. 

With that approach, review would be quicker, reviewers would feel more comfortable.
Again, this is great, thank you. Just want to share feedback on how it could be even better ❤️

ForestLinSen:
> Just sharing because I forgot to mention: The best way to do this change would have been a two step approach: (1) First PR only to move the recharts-scale into the repo. Add tests, etc. (2) Second PR to build a new feature.
> 
> With that approach, review would be quicker, reviewers would feel more comfortable. Again, this is great, thank you. Just want to share feedback on how it could be even better ❤️

Thanks, that's a great suggestion. For now, I haven't moved all the code from rechart-scale, just the parts necessary for ticket computation. It might be worth creating another PR for all the remaining code and adding tests for it

ckifer:
Any concerns before merge @PavelVanecek?

package.json
```diff
@@ -79,7 +79,8 @@
     "react-smooth": "^4.0.0",
     "recharts-scale": "^0.4.4",
```
PavelVanecek, line 80: If we're merging recharts-scale inside the main package, can we remove the dependency now?

test/util/scale/getNiceTickValues.spec.ts
PavelVanecek, line 1: How does this compare to the previous tests? https://github.com/recharts/recharts-scale/blob/master/test/ticks/getNiceTickValues.test.js based on a quick scroll it looks like it used to have more? Did you decide to omit some of them?

storybook/stories/API/props/ChartProps.ts
```diff
@@ -181,6 +181,23 @@ toggling between multiple dataKey.`,
       category: 'General',
     },
   },
+  stepRatioControl: {
+    description: `This parameter controls the range of the y-axis domain. 
+    A lower value for this parameter results in the maximum y-axis value being closer to the highest data point in the dataset. 
+    For example, given the dataset [0, 400, 800, 1200, 1600], a stepRatioControl value of 0.05 would adjust the y-axis domain to [0, 2000]. 
+    Conversely, setting stepRatioControl to 0.1 brings the y-axis domain closer to [0, 1600]. 
+    Adjusting this parameter is recommended only when the default y-axis domain does not adequately represent the dataset.`,
```
PavelVanecek, line 189: What does this parameter do that `domain` doesn't?

ForestLinSen: Hi as the original [issue](https://github.com/recharts/recharts/issues/4020) mentioned it's for addressing the problem when the domain is set to ["auto", "auto"]

test/util/scale/getNiceTickValues.spec.ts
ForestLinSen: Good point, will add more tests based on this

storybook/stories/API/props/ChartProps.ts
PavelVanecek: Yeah, auto does what auto does, people who are not happy with the default auto behavior can go and provide their own behavior. Why add another prop that does the same thing?

package.json
ForestLinSen: didn't add all the code from `recharts-scale` just the parts necessary for the changes for now

storybook/stories/API/props/ChartProps.ts
ForestLinSen: I think one difference when setting to auto is for the "nice" value. Like, for a max value of 3374, 'dataMax + 10%' would be 3711 and auto with step ratio control can yield values like 3600 or 4000

PavelVanecek: The trouble is, what's nice or not is subjective.

In my opinion the real solution is to provide in the 'domain' prop the values that work for you in your chart and feel nice for you.

ckifer: This isn't about the domain, it's about what tick values the domain produce. Domain is still needed and these two properties do not do the same thing.

PavelVanecek: Aha! Can we find a different property that achieves the same and is more relatable? For example tick count, or ticks padding?

ckifer: Ok maybe I misspoke slightly, it controls what the values of the domain will be (the min, the max, and the in between round)m numbers) when the domain is set to auto. 

I agree that this property isn't relatable. I don't agree that the subjectivity of "nice" should prevent us from taking a subjective opinion on nice. We should have good defaults. We don't need perfect defaults but they should be solid - the behavior described in the linked issue is odd and worth changing. There might be a better way to change it than this, but I still think it's worth changing

PavelVanecek: I would prefer to just change the default then instead. We can release that as a breaking change in 3.x.

I'm expecting we won't be able to explain the new property and people will struggle to use it. And because it's a part of the public API then it will be impossible to change.

PavelVanecek: > controls what the values of the domain will be:
I agree we need some sort of control and customization for this

> the min, the max
This is the `domain` prop

> and the in between round)m numbers
Looking at recharts.org I can see that there already is `tickCount` prop on YAxis. There's also `ticks` and `minTickGap` which are easy to picture and understand the relationship between input and output. (`ticks={3}` means there will be 3 ticks.)

Then there's `scale` which is somewhat more abstract but on the other hand standardized in other charting libraries and experiences.

I don't think we need a new property and I am not at all convinced that values `0.01 | 0.03 | 0.05` are good values that communicate what they do, externally.

ckifer: I agree that this is hard to explain/understand. 

The properties you mention don't do the same thing as this though. 

Sure the min and max are set with domain, but when you set your own domain the generated ticks become non-round numbers like 554, 239, 38472847274, etc. 

When auto is used the ticks become 100, 250, 500, 750, 1000, etc. and they're still evenly spaced (if you specified your own domain, ticks, and tick count they would not be). 

The only way you could get the same result yourself is if you did some calculation to determine you domain using your data beforehand and make it always easily divisible by the number of ticks you have. Doing that is a lot tougher than just having recharts do it by default (and I'm not convinced that doing what I just said works 😅)

Just saying this to emphasize that it's something we should at least take some responsibility for, at some point. Let's wait until after 3.x and revisit unless we can come up with a less complicated way of improving this

ckifer: We can still certainly keep the recharts-scale movement

ckifer:
@ForestLinSen would you be able to remove the code that is specific to `stepRatioControl` and only treat this PR as recharts-scale movement? That was way we can unblock what most of this work was and merge it and keep the discussion going about stepRatioControl or other solutions to the tick problem. 

Feel free to do it here or close this and make a new PR

ForestLinSen:
@ckifer Sure, I have removed the code related to `stepRatioControl` and will make a new PR about the parameter after this one

Issue #4378, opened 2024-04-03
Apply `className` to the `label` and `labelLine` in `PieChart` component.


108yen:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
It would be possible to style each component with `className`.

### What does the proposed API look like?
I want to apply a `className` to a `PieChart` component, as in the following example
```tsx
        <PieChart {...args}>
          <Pie
            data={data}
            dataKey="uv"
            nameKey="name"
            innerRadius={50}
            outerRadius={80}
            label={{ className: 'label-custom-className' }}
            labelLine={{ className: 'label-line-custom-className' }}
          />
        </PieChart>
```

ckifer:
merged and will be in the next patch release! Thanks @108yen !

PR #4381, opened 2024-04-03
feat(pie-chart): apply `className` to the `label` and `labelLine`

108yen:
## Description

apply `className` to the `label` and `labelLine` in `PieChart` component.

## Related Issue

#4378

## Motivation and Context

Required if you want to assign `className` to the `PieChart` component.

## How Has This Been Tested?

Added unit test.

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -402,7 +402,8 @@ export class Pie extends PureComponent<Props, State> {
       return option(props);
     }
 
-    return <Curve {...props} type="linear" className="recharts-pie-label-line" />;
+    const className = clsx('recharts-pie-label-line', typeof option !== 'boolean' ? option.className : '');
+    return <Curve {...props} type="linear" className={className} />;
   }
 
   static renderLabelItem(option: PieLabel, props: any, value: any) {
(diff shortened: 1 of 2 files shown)


src/polar/Pie.tsx
```diff
@@ -402,7 +402,8 @@ export class Pie extends PureComponent<Props, State> {
       return option(props);
     }
 
-    return <Curve {...props} type="linear" className="recharts-pie-label-line" />;
+    const className = clsx('recharts-pie-label-line', typeof option !== 'boolean' ? option.className : '');
```
ckifer, line 405: nit: we kind of already have a typeguard for this right? 
https://github.com/recharts/recharts/commit/76bbdcb6d4c8e13f73f8ca7834bcf2a28143ac8d#diff-0030e00d80866b12434b3def516dcf03f12a37360fcaaa5fc6a7c32e0d04d17cR187

We can move the typeguard out of polar utils to somewhere that is more generic, and then make the function more generic so it works for every case.

And actually to make it a true "type guard" we could instead do

```tsx
export const hasCustomClassName = <T extends SVGElement>(
  option?: SVGProps<T> | ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>) | boolean,
): option is { className: string } =>
  !isValidElement(option) && !isFunction(option) && typeof option !== 'boolean' && 'className' in option;
```

Then `const className = clsx('recharts-pie-label-line', hasCustomClassName(option) ? option.className : '');

ckifer (approved):
Maybe we can do the suggestion later, this looks good how it is

PR #4390, opened 2024-04-04
feat(BarChart): support percentage for barSize. Fixes #3640

### LINKED ISSUE #3640: Automatic bar width is set to zero when there is only one data point

graup:
This adds support for percentage values for `barSize` as a workaround for #3640. See also my comment at https://github.com/recharts/recharts/issues/3640#issuecomment-2037082408

This is not really the perfect solution, but it allows users to at least have a good workaround.

If you know how many data points your axis domain expects, you can hard-code `barSize` to `100%/numberOfDataPoints` (e.g. `'10%'` for 10 data points).

You can also combine this with the automatic calculation, e.g. `barSize={data.length > 1 ? undefined : '10%'}`.

```
<BarChart
  width={500}
  height={300}
  data={[[4.5, 10]]}
  barSize="30%" /* When there's only one data point on a numerical domain, we cannot automatically calculate the bar size */
  margin={{
    top: 5,
    right: 30,
    left: 20,
    bottom: 5,
  }}
>
  <XAxis dataKey={v => v[0]} type="number" domain={[0, 10]} />
  <YAxis />
  <Tooltip />
  <Bar dataKey={v => v[1]} />
</BarChart>
```

Screenshot 2024-04-04 at 16 31 49


src/util/ChartUtils.ts
```diff
@@ -250,7 +252,10 @@ export const getBarSizeList = ({
         result[cateId].push({
           item: barItems[0],
           stackList: barItems.slice(1),
-          barSize: isNil(selfSize) ? globalSize : selfSize,
+          barSize:
+            isNil(selfSize) && isNil(globalSize)
```
ckifer, line 256: nit: can store `isNil(selfSize)` in a variable above this to prevent calling twice

ckifer (commented):
Cool thanks! - just will need to fix the tests and add tests

graup:
Thanks, I have addressed your comments and added new tests.

ckifer:
> Do you think we should backport this?

I guess it doesn't matter much to me 😅. If you or someone else wants this feature in 2.x then I don't think it violates SemVer enough to worry about it (before only numbers documented as an accepted value, if they provided a percentage it was outside of the API, now its documented and inside the API - if they break somehow its their fault)

src/util/ChartUtils.ts
```diff
@@ -247,10 +249,12 @@ export const getBarSizeList = ({
           result[cateId] = [];
         }
 
+        const barSize: string | number | undefined = isNil(selfSize) ? globalSize : selfSize;
```
ckifer, line 252: I wonder if we should use `invariant` and throw to the user an error message if their input is a string but not a percentage? 

None of the other recharts code does that I don't think for the items that accept percentages, but we probably should. All good for now

src/chart/generateCategoricalChart.tsx
```diff
@@ -842,22 +844,34 @@ export interface CategoricalChartProps {
...
+  angleAxis?: AngleAxisProps;
   angleAxisTicks?: Array<TickItem>;
 
-  radiusAxis?: BaseAxisProps;
+  radiusAxis?: RadiusAxisProps;
   radiusAxisTicks?: Array<TickItem>;
 };
 
+// Determine the size of the axis, used for calculation of relative bar sizes
+const getCartesianAxisSize = (axisObj: AxisObj, axisName: 'xAxis' | 'yAxis' | 'angleAxis' | 'radiusAxis') => {
```
ckifer, line 864: nit: `getCartesianAxisSize` seems generic enough that one could accidentally use this function in the future for something else without realizing how specific it is to calculating `barSize`. 

We should either name this slightly more specific (the comments are specific, so the function can be named specific too) or more generically and add JSDoc that helps future us understand what this could possibly be used for

ckifer (approved):
One bit of feedback but this looks good as is. 

I'll wait to merge until I hear back from you on if you'd like to backport or not as it'll be easier to include everything than to go through multiple PRs

ckifer:
Oh and I guess merge conflict needs fixed

src/chart/generateCategoricalChart.tsx
graup: Yeah I think this can be generic, I'll add JSDoc for it. Maybe there's a better word instead of `size`? Like if we think CSS flex-box, the size is kind of the length of the "main axis"?

graup:
> > Do you think we should backport this?
> 
>  If you or someone else wants this feature in 2.x then I don't think it violates SemVer enough to worry about it.

OK then yeah I'd be happy to have this in 2.x too. I'll make a backport PR after resolving the issue about `getCartesianAxisSize`.

ckifer (approved):
LGTM for now, going to merge - feel free to follow up with any small changes. Thank you!

PR #4399, opened 2024-04-06
Introducing Redux

PavelVanecek:
Okay this is ready for merging!

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1,4 +1,4 @@
-import React, { Component, cloneElement, ReactElement } from 'react';
+import React, { cloneElement, Component, ReactElement } from 'react';
 import isNil from 'lodash/isNil';
 import isFunction from 'lodash/isFunction';
 import range from 'lodash/range';
@@ -90,6 +90,8 @@ import { AngleAxisProps, RadiusAxisProps } from '../polar/types';
 import { ChartDataContextProvider } from '../context/chartDataContext';
 import { BrushStartEndIndex, BrushUpdateDispatchContext } from '../context/brushUpdateContext';
 import { ClipPath } from '../container/ClipPath';
+import { ChartOptions } from '../state/optionsSlice';
+import { RechartsStoreProvider } from '../state/RechartsStoreProvider';
 
 export interface MousePointer {
   pageX: number;
@@ -1066,7 +1068,7 @@ export const generateCategoricalChart = ({
     };
   };
 
-  return class CategoricalChartWrapper extends Component<CategoricalChartProps, CategoricalChartState> {
+  class CategoricalChartWrapper extends Component<CategoricalChartProps, CategoricalChartState> {
     static displayName = chartName;
 
     readonly eventEmitterSymbol: Symbol = Symbol('rechartsEventEmitter');
@@ -1796,7 +1798,7 @@ export const generateCategoricalChart = ({
 
       const key = element.key || '_recharts-cursor';
 
-      return <Cursor key={key} chartName={chartName} element={element} tooltipEventType={tooltipEventType} />;
+      return <Cursor key={key} element={element} tooltipEventType={tooltipEventType} />;
     };
 
     /**
@@ -2020,5 +2022,18 @@ export const generateCategoricalChart = ({
         </ChartDataContextProvider>
       );
     }
+  }
+
+  return function CategoricalChart(props: CategoricalChartProps) {
+    const options: ChartOptions = {
+      chartName,
+      defaultTooltipEventType,
+      validateTooltipEventTypes,
+    };
+    return (
+      <RechartsStoreProvider preloadedState={{ options }} reduxStoreName={props.id ?? chartName}>
+        <CategoricalChartWrapper {...props} />
+      </RechartsStoreProvider>
+    );
   };
 };
--- a/src/state/RechartsStoreProvider.tsx
+++ b/src/state/RechartsStoreProvider.tsx
@@ -1,0 +1,17 @@
+import React, { ReactNode, useRef } from 'react';
+import { Provider } from 'react-redux';
+import { createRechartsStore, RechartsRootState } from './store';
+
+type RechartsStoreProviderProps = {
+  children: ReactNode;
+  preloadedState?: Partial<RechartsRootState>;
+  reduxStoreName?: string;
+};
+
+export function RechartsStoreProvider({ preloadedState, children, reduxStoreName }: RechartsStoreProviderProps) {
+  const storeRef = useRef(null);
+  if (storeRef.current == null) {
+    storeRef.current = createRechartsStore(preloadedState, reduxStoreName);
+  }
+  return <Provider store={storeRef.current}>{children}</Provider>;
+}
--- a/src/state/RechartsStoreProvider.tsx
+++ b/src/state/RechartsStoreProvider.tsx
@@ -9,6 +9,17 @@ type RechartsStoreProviderProps = {
 };
 
 export function RechartsStoreProvider({ preloadedState, children, reduxStoreName }: RechartsStoreProviderProps) {
+  /*
+   * Why the ref? Redux official documentation recommends to use store as a singleton,
+   * and reuse that everywhere: https://redux-toolkit.js.org/api/configureStore#basic-example
+   *
+   * Which is correct! Except that is considering deploying Redux in an app.
+   * Recharts as a library supports multiple charts on the same page.
+   * And each of these charts needs its own store independent of others!
+   *
+   * The alternative is to have everything in the store keyed by the chart id.
+   * Which would make working with everything a little bit more painful because we need the chart id everywhere.
+   */
   const storeRef = useRef(null);
   if (storeRef.current == null) {
     storeRef.current = createRechartsStore(preloadedState, reduxStoreName);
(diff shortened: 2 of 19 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -2170,5 +2176,14 @@ export const generateCategoricalChart = ({
         </ChartDataContextProvider>
       );
     }
+  }
+
+  return function CategoricalChart(props: CategoricalChartProps) {
+    const store = useRef(createStore(props.id ?? chartName));
+    return (
+      <Provider store={store.current}>
+        <CategoricalChartWrapper {...props} dispatch={store.current.dispatch} />
```
PavelVanecek, line 2185: This way the CategoricalChartWrapper has access to the store too, and can dispatch actions. This will allow us to slowly migrate things from its state to redux state.

src/state/hooks.ts
```diff
@@ -0,0 +1,21 @@
+import { TypedUseSelectorHook, useSelector, ReactReduxContext } from 'react-redux';
+import type { RootState, AppDispatch } from './store';
+import { useContext } from 'react';
+
+const noopDispatch: AppDispatch = (): undefined => {};
+
+export const useAppDispatch = (): AppDispatch => {
```
PavelVanecek, line 7: The direct imports from `react-redux` throw when used outside of Redux context which is usually great but if we want to continue to have our individual components to work outside of Chart context (like the Area example we discussed earlier) then we need alternative hooks that return undefined instead of throwing.

src/state/store.ts
```diff
@@ -0,0 +1,28 @@
+import { Action, combineReducers, configureStore, Dispatch } from '@reduxjs/toolkit';
+import { legendReducer } from './legendSlice';
+
+const rootReducer = combineReducers({
+  // tooltip: tooltipReducer,
+  legend: legendReducer,
+});
+
+export const createStore = (chartName: string = 'Chart') => {
+  console.log('createStore', chartName);
```
PavelVanecek, line 10: TODO remove before merging

src/state/store.ts
```diff
@@ -0,0 +1,28 @@
+import { Action, combineReducers, configureStore, Dispatch } from '@reduxjs/toolkit';
+import { legendReducer } from './legendSlice';
+
+const rootReducer = combineReducers({
+  // tooltip: tooltipReducer,
```
PavelVanecek, line 5: TODO remove before merging

src/state/store.ts
```diff
@@ -0,0 +1,28 @@
...
+});
+
+export const createStore = (chartName: string = 'Chart') => {
+  console.log('createStore', chartName);
+  return configureStore({
+    reducer: rootReducer,
+    middleware: getDefaultMiddleware =>
+      getDefaultMiddleware({
+        serializableCheck: {
+          ignoredActions: ['legend/addSupplier', 'legend/removeSupplier'],
```
PavelVanecek, line 16: I'm considering turning this off completely - Recharts has a habit of passing everything everywhere so we might as well allow non-serializables in the state straight away.

src/state/store.ts
```diff
@@ -0,0 +1,28 @@
+import { Action, combineReducers, configureStore, Dispatch } from '@reduxjs/toolkit';
+import { legendReducer } from './legendSlice';
+
+const rootReducer = combineReducers({
+  // tooltip: tooltipReducer,
+  legend: legendReducer,
+});
+
+export const createStore = (chartName: string = 'Chart') => {
```
PavelVanecek, line 9: This is a function because we want to allow multiple stores on one page. Users can choose to have multiple charts and they should be independent too.

The tooltip synchronisation will continue working the same as it does now - via the emitter.

src/state/store.ts
```diff
@@ -0,0 +1,28 @@
...
+    middleware: getDefaultMiddleware =>
+      getDefaultMiddleware({
+        serializableCheck: {
+          ignoredActions: ['legend/addSupplier', 'legend/removeSupplier'],
+          ignoredPaths: ['legend.payloadSuppliers'],
+        },
+      }),
+    devTools: {
+      serialize: true,
+      name: 'recharts-' + chartName,
```
PavelVanecek, line 22: This is useful for debugging. It reuses already present `id` and `chartName` props.

nikolasrieble (commented):
I really like redux. That being said, I have also spent countless hours on debugging redux, and there are quite some things that do not fit well into redux in the recharts.

https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

Could you please write a short motivation as to what we would gain from using redux, and how we would handle non-serialisable state? 

I am worried that we add complexity to the codebase that in the worst case will slow us down, and add a larger bug surface. 

That being said, if done well, it can help us with state management, and nicely separate components from state processing (i.e. getting the tooltip separately from showing the tooltip, which would enable us to more easily iterate on the tooltip).

PavelVanecek:
Thanks @nikolasrieble.

We discussed this with @ckifer, I think the main benefits are:

1. One Provider instead of many
2. Built-in memoization
3. Tooling - redux browser extension for example

The main negative point is bundle size increase. Recharts bundle is 500kB, installing redux will add another 3kB on top of that so I think that's acceptable.

Redux indeed is not quite too happy putting non-serializables in the store. However, as the docs say:

> It's technically possible to insert non-serializable items into the store, but doing so can break the ability to persist and rehydrate the contents of a store, as well as interfere with time-travel debugging.

Recharts is not using persistence, and is not going to. Inability to time-travel is acceptable I think (we can't do that with hooks either).

I selected the legend payload as an example precisely because of the functions in state. I had to add the `serializable: true` and some exceptions to allow for it but in the end it works fine:

image

image

Yes I will try and avoid putting functions or Components or Elements into the state. I think internally we don't need to do that (definitely let's not put `formattedGraphicalItems` in there for example), but Recharts likes to store and forward _everything_, including event handlers and DOM references, in the external API. So we will need to save those somewhere. Perhaps we can have a separate vanilla Context just for functions and the like.

src/state/hooks.ts
ckifer: Later we'll continue to have a re-exported AppDispatch anyways for typing purposes so this is the right track

ckifer (commented):
This looks pretty good to me tbh. I just want to confirm that this works well when being published (via yalc or via alpha version, doesn't matter to me) and make sure we accept consequences and understand the direction we're going once we do this as Niko's comments call out

ckifer:
I want to make sure that we have a way of keeping the _power_ specifically, of recharts. I don't really care if we use redux, context, zustand, or anything else - I just want people to not be limited by what we change (even if these things can be done via vanilla SVG). We want to give them access to this data in a similar way to the Customized component (or expose the Provider and all the hooks to get the relevant values).

As long as we can do that and keep flexibility, all good to me. I'm super familiar with RTK and love working with it. If we establish a better interface (and documenting it) for this stuff than what currently exists, while still exposing the same information I think this will be great.

PavelVanecek:
Would you like to see a proof of concept for any of those? To demonstrate the path forward.

PavelVanecek:
Updated and ready for merging. It's a small start, only one slice in the store and only Cursor reading it. I will follow up in another PR.

ckifer (commented):
One question about the ref

src/state/RechartsStoreProvider.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { createRechartsStore, RechartsRootState } from './store';
+
+type RechartsStoreProviderProps = {
+  children: ReactNode;
+  preloadedState?: Partial<RechartsRootState>;
+  reduxStoreName?: string;
+};
+
+export function RechartsStoreProvider({ preloadedState, children, reduxStoreName }: RechartsStoreProviderProps) {
+  const storeRef = useRef(null);
```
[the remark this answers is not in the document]
PavelVanecek: That's right but that's considering a singleton app. In recharts, we want to be able to support multiple independent charts on one page.

The alternative is to have everything in the store keyed by the chart id. Which makes working with everything a little bit more painful because we need the chart id everywhere. And it makes synchronising charts easier.

storybook/stories/API/chart/Sankey.stories.tsx
```diff
@@ -26,7 +26,7 @@ export const Simple = {
   render: (args: Record<string, any>) => {
     return (
       <ResponsiveContainer width="100%" height={400}>
-        <Sankey data={nodeLinkData} {...args}>
+        <Sankey data={nodeLinkData} {...args} name="Sankey-Simple">
```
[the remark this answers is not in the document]
PavelVanecek: I use it to rename the store for easy debugging in the redux devtools extension.

No idea what's the purpose of this elsewhere.

src/state/RechartsStoreProvider.tsx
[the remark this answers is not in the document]
nikolasrieble: This is very nice reasoning. Could you please add that as a comment to the code? Otherwise I am afraid in the distant future someone will just be like: Oh, why is this a ref, lets fix it.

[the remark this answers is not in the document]
PavelVanecek: I added a comment. I tried to write a unit test to demonstrate that too.

storybook/stories/API/chart/Sankey.stories.tsx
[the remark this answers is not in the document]
PavelVanecek: Oh wait no! This is supposed to be id. I will remove it and re-add later when I add Redux provider to Sankey too.

ckifer (approved):
lets do it

PR #4400, opened 2024-04-06
Move Clip path to a component

PavelVanecek:
## Description

Use Component instead of an instance method

## Related Issue

## Motivation and Context

This wasn't blocking anything tbh. But it makes `generateCategoricalChart` a bit smaller.

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/scripts/snapshots/es6Files.txt
+++ b/scripts/snapshots/es6Files.txt
@@ -69,6 +69,7 @@
   "es6/context/accessibilityContext.js",
   "es6/container/Surface.js",
   "es6/container/Layer.js",
+  "es6/container/ClipPath.js",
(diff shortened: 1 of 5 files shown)


nikolasrieble (approved):
Nice, thank you. 

Just wondering - non blocking: ClipPath could be a non exported component inside of generateCategoricalChart.tsx. 
I do not think we will use it anywhere else, and we will not export it to users, right?

ckifer (approved):
LGTM,

@nikolasrieble i guess I'd rather make the generate file smaller then worry that it is exported, at least at the moment. We definitely don't want to export it from any index files

PR #4402, opened 2024-04-06
Add stories to show tooltip, and active dot

PavelVanecek:
## Description

Only storybooks.

## Related Issue

## Motivation and Context

I want to have some baseline screenshots before refactoring the active dots.

## How Has This Been Tested?

storybook

You will notice that the "Vertical" example renders cursor in a wrong position; this is only present when using `defaultIndex`. When mouse-hovering the cursor shows correct. I don't have ambitions of fixing that in this PR 😬 

image

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/Examples/RadarChart.stories.tsx
+++ b/storybook/stories/Examples/RadarChart.stories.tsx
@@ -91,7 +91,7 @@ export const RadarWithLegend: StoryObj = {
       <RadarChart data={data} width={360} height={360}>
         <PolarGrid gridType="circle" />
         <Legend />
-        <Tooltip />
+        <Tooltip defaultIndex={2} />
         <PolarRadiusAxis type="number" dataKey="r" />
 
         <PolarAngleAxis dataKey="angle" axisLineType="circle" type="number" domain={[0, 360]} />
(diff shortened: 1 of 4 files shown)


storybook/stories/Examples/RadarChart.stories.tsx
```diff
@@ -91,7 +91,7 @@ export const RadarWithLegend: StoryObj = {
       <RadarChart data={data} width={360} height={360}>
         <PolarGrid gridType="circle" />
         <Legend />
-        <Tooltip />
+        <Tooltip defaultIndex={2} />
```
nikolasrieble, line 94: The change seems unrelated to the Story `RadarWithLegend`. Should we possibly have a separate story for the Tooltip with the Radar Chart, which then has defaultIndex set? 

Or rather, should these examples with the Tooltip having non default values not be done on the Tooltip stories instead? 

I have no clear preference, but I think it would be great for documentation readers to always be only exposed to a single concept / component per story.

For the Tooltip defaultIndex, I suggest to change the Tooltip stories. Does that make sense?

ckifer: I guess we have to strike a balance here.

I think @PavelVanecek is trying to get stories in place that will break in snapshots if we do something that breaks the feature. That makes sense to me from a development perspective. It doesn't always make sense from a documentation perspective. 

I wish there was some way we could get the visual diffs without exposing the stories themselves.

PavelVanecek: I am after the visual diff.

ckifer (approved):

Issue #4382, opened 2024-04-03
Layer: TypeScript error with latest `@types/react` version


Lukas742:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://stackblitz.com/edit/vitejs-vite-raiemq?file=package.json,src%2FApp.tsx](https://stackblitz.com/edit/vitejs-vite-raiemq?file=package.json,src%2FApp.tsx)

### Steps to reproduce
When building `recharts` without `skipLibCheck: true` a TypeScript error is thrown with `v18.2.74` of `@types/react`. 
I believe the root cause is how TypeScript infers and represents types for `forwardRef`, but unfortunately I wasn't able to fix it myself.

1. Go to [StackBlitz](https://stackblitz.com/edit/vitejs-vite-raiemq?file=package.json,src%2FApp.tsx)
2. run `tsc` in terminal
3. See the following error:

```sh
node_modules/recharts/types/container/Layer.d.ts:7:108 - error TS2344: Type '"string" | "filter" | "values" | "fill" | "stroke" | "max" | "type" | "accumulate" | "offset" | "key" | "id" | "media" | "origin" | "height" | "width" | "end" | "name" | "min" | "alignmentBaseline" | ... 450 more ... | "onPointerLeaveCapture"' does not satisfy the constraint '"string" | "filter" | "values" | "fill" | "stroke" | "max" | "type" | "accumulate" | "offset" | "key" | "id" | "media" | "origin" | "height" | "width" | "end" | "name" | "min" | "alignmentBaseline" | ... 457 more ... | "onTransitionEndCapture"'.
  Type '"onPointerEnterCapture"' is not assignable to type '"string" | "filter" | "values" | "fill" | "stroke" | "max" | "type" | "accumulate" | "offset" | "key" | "id" | "media" | "origin" | "height" | "width" | "end" | "name" | "min" | "alignmentBaseline" | ... 457 more ... | "onTransitionEndCapture"'. Did you mean '"onPointerOverCapture"'?

7 export declare const Layer: React.ForwardRefExoticComponent<Pick<React.SVGProps<SVGGElement> & LayerProps, "string" | "ideographic" | "alphabetic" | "hanging" ...[shortened for readability]> & React.RefAttributes<unknown>>;
```

### What is expected?
Types for React v18 should be supported.

### What is actually happening?
Error is thrown.

| Environment | Info |
|---|---|
[...]

PR #4413, opened 2024-04-10
fix(layer-types): use SVGAttributes instead of SVGProps in forwardRef components, fix Pie refs

ckifer:
## Description

Compiling TS types for `SVGProps<T>` in a component that uses `forwardRef` enumerates the keys of T for some reason and can cause TS build issues when a user builds the project with @types/react version greater than the one we have installed.

This fixes the type error

## Related Issue

https://github.com/recharts/recharts/issues/4382

## How Has This Been Tested?

- Local `yalc publish`
- Install in a breaking project
- Run `tsc` and see successful build

## Types of changes

## Checklist:

### DIFF
--- a/src/container/Layer.tsx
+++ b/src/container/Layer.tsx
@@ -1,7 +1,4 @@
-/**
- * @fileOverview Layer
- */
-import React, { ReactNode, SVGProps } from 'react';
+import React, { ReactNode, SVGAttributes } from 'react';
 import clsx from 'clsx';
 import { filterProps } from '../util/ReactUtils';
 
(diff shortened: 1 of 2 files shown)


HHogg (approved):
Thanks a lot for pushing this through @ckifer. 

The HTMLElement -> SVGGElement changes might cause a type error for folks who have explicitly typed a ref like 

```ts
const ref = useRef<HTMLElement>(); 

<Layer ref={ref} />
```

...but it's unlikely and an easy and correct fix if so. 

Thanks again.

ckifer:
@HHogg thanks for the review. 

Yeah, the idea is that the change is, in reality, what the code should have reflected. HTMLElement is incorrect and therefore any user with the code mentioned above will be fixing a type issue in their codebase

PavelVanecek (approved):
Nice

Issue #4424, opened 2024-04-12
Tooltip first render


Kaiyrgali:
@ckifer Hi! Thank you very much for your work, it is very important for us

We noticed another problem in the new version (we don’t know in which version this happened, but version 2.0.6 does not have this problem)

When a Tooltip has isAnimationActive={false} and content arguments, the first render of the Tooltip is always in the corner of the chart.

Our tests make screenshot and detect the incorrect position of the tooltip
![image](https://github.com/recharts/recharts/assets/95644241/e1cc3dc6-48b3-4db5-b7bd-642fd0ce5204)

At the same time, all the correct arguments come to the tooltip props and it is not possible to cancel the incorrect position of the tooltip

https://codesandbox.io/p/sandbox/recharts-tooltip-problem-24ff35

![image](https://github.com/recharts/recharts/assets/95644241/62bf63a3-0599-41b7-bf38-1df7cdaeff54)

ckifer:
Hey @Kaiyrgali 

This issue is similar https://github.com/recharts/recharts/issues/4338#issue-2203919364

Also not sure when this became an issue, but I'll try to pin down when it did since this has been reported more than once.
Looks like in your case its pretty much tied to `isAnimationActive`

ckifer:
Looks like this occurred between 2.11 and 2.12 - can you try to install 2.11 and see if you experience the issue?

Kaiyrgali:
> Looks like this occurred between 2.11 and 2.12 - can you try to install 2.11 and see if you experience the issue?

@ckifer version "2.11.0" has the same problem (

ckifer:
Hmm weird why I couldn't reproduce. But try 2.9.3, I think we tracked it to 2.10 and above

PrabhaShanker:
this is an old issue, I have noticed the same in 2.4.5

ckifer:
Old as in pre-2.0.6? Or between the two somewhere?

I can't reproduce in 2.9.3 - if you can please feel free to make and share a codesandox, thanks!

ckifer:
fixed in 2.12.6

DylanTse520:
Hi @ckifer. Thanks for the great work! I noticed that if the animation is active, the tooltip still always comes in from top left corner before it gets to where the cursor is. This looks like an unexpected behavior and I'm wondering if this is a bug or if there is a way to keep the animation without it moving in from the top left corner.

[...]

PR #4425, opened 2024-04-13
fix: The box size of the Tooltip is 0 at the first rendering of TooltipBoundingBox

HHongSeungWoo:
## Description

When the size of the tooltip component changes, re-render the TooltipBoundingBox.

## Related Issue

#4424 

## Types of changes

## Checklist:


ckifer:
I thought the fix might be smaller than this 😅 

Appreciate the effort here, but if we change this here then we need to change it in 3.x as well - we don't want a regression later when people migrate major versions (even if this bug doesn't currently happen in 3.x). This seems snapper than what was currently being done..

HHongSeungWoo:
Can fix the bug just by moving lastBoundingBox to state. It has been fixed that way in version 3.x as well.

ckifer:
Yeah, could you do that here? We can use these other improvements on the 3.x branch and leave them out here just to prevent complicating things and diverging branches even further

HHongSeungWoo:
I cannot cherry-pick the implementation from version 3.x, but it seems to be the minimal modification

ckifer (approved):
Looks good, thanks!

PR #4428, opened 2024-04-15
Remove getItemByActiveIndex function and dependencies

PavelVanecek:
## Description

What's going on?

I discovered that the only reason that code ever goes to this branch is that it's missing the `activeTooltipIndex`. This is always set for `tooltipEventType: axis` but never for `item`. The reason is that `handleItemMouseEnter` function never set it.

Items that support `tooltipEventType: item` are:

- Bar
- Funnel
- Pie
- RadialBar
- Scatter

image

turns out, all of these already wrap their mouse events using a function call like this: `{...adaptEventsOfChild(this.props, entry, i)}`. And this will provide the index as a second argument! So I added the second argument into `handleItemMouseEnter` and now `activeTooltipIndex` is always set (and correctly) so the branch of code that reads DOM to figure out the index in case it's never set - is now a dead code so I could remove it.

This is also the first time (since I joined the project) that `generateCategoricalChart.tsx` is below 2000 lines of code 🥳 

## Related Issue

## Motivation and Context

I am on a quest of removing `renderGraphicChild` to allow removing all element cloning.

## How Has This Been Tested?

storybook diff - no diff
manual storybook testing
npm test

## Types of changes

## Checklist:

### DIFF
--- a/src/util/ActiveShapeUtils.tsx
+++ b/src/util/ActiveShapeUtils.tsx
@@ -1,14 +1,12 @@
-import React, { isValidElement, cloneElement, SVGProps } from 'react';
+import React, { cloneElement, isValidElement, SVGProps } from 'react';
 import isFunction from 'lodash/isFunction';
 import isPlainObject from 'lodash/isPlainObject';
 import isBoolean from 'lodash/isBoolean';
-import isEqual from 'lodash/isEqual';
 
 import { Rectangle } from '../shape/Rectangle';
 import { Trapezoid } from '../shape/Trapezoid';
 import { Sector } from '../shape/Sector';
 import { Layer } from '../container/Layer';
-import { GraphicalItem } from '../chart/generateCategoricalChart';
 import { Symbols, SymbolsProps } from '../shape/Symbols';
 
 /**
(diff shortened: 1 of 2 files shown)


src/util/ActiveShapeUtils.tsx
```diff
@@ -105,151 +103,3 @@ export function Shape<OptionType, ExtraProps, ShapePropsType>({
...
- *
- * @param {GetActiveShapeIndexForTooltip} arg an object of incoming attributes from Tooltip
- * @returns {number}
- *
- * To handle possible duplicates in the data set,
- * match both the data value of the active item to a data value on a graph item,
- * and match the mouse coordinates of the active item to the coordinates of in a particular components shape data.
- * This assumes equal lengths of shape objects to data items.
- */
-export function getActiveShapeIndexForTooltip({
```
ckifer, line 225: everything for active shape still works correctly even without all of this? I feel like we added this recently?

ckifer (approved):
Nice!! Thats a huge amount of stuff removed!

Curious if this ties us down to using indices forever?

I guess there are always better ways of doing that (showing tooltip by X,Y instead of by index) and we can re-implement when we get there

src/util/ActiveShapeUtils.tsx
PavelVanecek: All active shapes I'm aware of are working in storybook.

PavelVanecek:
> Curious if this ties us down to using indices forever?

The only thing the function was doing is it's computing an index. So we're in the same situation as before.

Given that the input is always a one dimensional array, I think index is okay.

src/util/ActiveShapeUtils.tsx
ckifer: ack, cool

Issue #4426, opened 2024-04-13
YAxis is missing if value = null || undefined


Kaiyrgali:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

@ckifer Hi! The problem starts in version v2.5.0, most likely from this pull request https://github.com/recharts/recharts/pull/3454

We use Yaxis even when there is no data = null || undefined. The user later selects a data source

Before version v2.5.0, the YAxis is rendered empty, without ticks
https://codesandbox.io/p/sandbox/yaxis-with-null-value-v2-4-3-pt4ws7

![image](https://github.com/recharts/recharts/assets/95644241/421d2196-c64f-46f6-82a7-51e53d8f8149)

Now the Yaxis is missing in our case
https://codesandbox.io/p/sandbox/yaxis-problem-with-null-value-v2-5-0-w9ffxd

![image](https://github.com/recharts/recharts/assets/95644241/14f55101-8240-4462-960f-198db1080cd8)

ckifer:
This has been in place for over a year without issue...

But it makes sense to render the line when there are no ticks so I'll check into that as free time allows

ckifer:
@Kaiyrgali the best I can do is address this behavior in the next major version https://github.com/recharts/recharts/pull/4433

See the PR for more explanation, but basically the behavior you relied on should never have worked reliably. It only did because of a bug that was fixed in the PR you linked

PavelVanecek:
Recharts 2.15:

<img width="491" height="331" alt="Image" src="https://github.com/user-attachments/assets/61d9e02f-493b-4705-8630-aafbd3796146" />

Recharts 3.1:

<img width="500" height="347" alt="Image" src="https://github.com/user-attachments/assets/a930cac0-deed-467e-9a3f-88dcd4fc8f08" />

I think this is fixed now!

PR #4433, opened 2024-04-16
fix(CartesianAxis): render the axis line when there are no ticks

ckifer:
## Description

Bug reported in linked issue where after recharts 2.5 the YAxis line would "disappear" with 0 data points or ticks, where it would previously show.

It would previously show because `NaN` was being calculated as a tick value and added to the dom as a tick item. This was fixed in https://github.com/recharts/recharts/pull/3454.

I would not call this a regression because as per the lines changed in this PR, the axis lines were never meant to show unless there were ticks. (See `!finalTicks || !finalTicks.length`). I believe however, that they should be able to.

Thus, this is a breaking change for both X and Y axes where even with 0 ticks or 0 valid data items, the axis lines will still render. 

For a user to return to the old behavior of no Axis, they only have to do something simple like `YAxis hide={data.length === 0} />`  or use React convention and conditionally render it

## Related Issue

https://github.com/recharts/recharts/issues/4426

## Motivation and Context

- It makes sense to show an axis line even when there are no ticks

## How Has This Been Tested?

- unit tests
- see screenshot

image

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/YAxis.tsx
+++ b/src/cartesian/YAxis.tsx
@@ -40,7 +40,6 @@ export const YAxis: FunctionComponent<Props> = ({ yAxisId }: Props) => {
   const width = useChartWidth();
   const height = useChartHeight();
   const axisOptions = useYAxisOrThrow(yAxisId);
-
   if (axisOptions == null) {
     return null;
   }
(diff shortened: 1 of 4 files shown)


src/cartesian/YAxis.tsx
```diff
@@ -40,7 +40,6 @@ export const YAxis: FunctionComponent<Props> = ({ yAxisId }: Props) => {
   const width = useChartWidth();
   const height = useChartHeight();
   const axisOptions = useYAxisOrThrow(yAxisId);
-
```
ckifer, line 43: oops

ckifer (commented):
I'm just going to ship this. Willing to defend it if anyone thinks its an issue

Issue #2458, opened 2021-02-22
Render Tooltip to React Portal to Prevent Clipping


Eric24:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Currently, if there are enough data series shown in the Tooltip to result in a Tooltip taller than the height of the chart and it's "chrome" (axes, legend, etc.), the tooltip is often clipped by the DOM element that contains the chart (note: it is not the chart itself that is clipping the Tooltip, but rarely if ever will a chart be the only element on a page by itself). By using a React Portal (https://reactjs.org/docs/portals.html), the Tooltip can be rendered independently of the chart and other elements in the DOM hierarchy, thus preventing the clipping.

### What does the proposed API look like?
No change to the API would be necessary. This should be a transparent, non-breaking change.

mrcljx:
For people who are looking for a baseline implementation (with `react-popper`):

```tsx
import { ReactNode, useEffect, useMemo, useRef, useState } from "react";
import { createPortal } from "react-dom";
import { usePopper } from "react-popper";
import { useEvent } from "react-use";
import type { TooltipProps } from "recharts/types/component/Tooltip";

function MyComponent() {
  return (
    <BarChart>
      <Tooltip content={<TooltipContent />} />
    </BarChart>
  }
}

function TooltipContent({ active }: TooltipProps) {
  return (
    <PopperPortal active={active}>
      CUSTOM TOOLTIP RENDERER
    </PopperPortal>
  );
}

function PopperPortal({
  active = true,
  children,
}: {
  active?: boolean;
  children: ReactNode;
}) {
  const [portalElement, setPortalElement] = useState<HTMLDivElement>();
  const [popperElement, setPopperElement] = useState<HTMLDivElement | null>();
  const virtualElementRef = useRef({
    getBoundingClientRect: INITIAL_GET_BOUNDING_CLIENT_RECT,
  });

  const { styles, attributes, update } = usePopper(
    virtualElementRef.current,
    popperElement,
    POPPER_OPTIONS
  );

  useEffect(() => {
    const el = document.createElement("div");
    document.body.appendChild(el);
    setPortalElement(el);
    return () => el.remove();
  }, []);

  useEvent("mousemove", ({ clientX: x, clientY: y }) => {
    if (!active) return;
    virtualElementRef.current.getBoundingClientRect =
      generateGetBoundingClientRect(x, y);
    update?.();
  });

[...]

PR #4435, opened 2024-04-16
Cursor now renders as a child of Tooltip

PavelVanecek:
## Description

Two fewer components being cloned.

This change follows Sankey and other non-generateCategoricalChart meaning, order of children now matters. So if people put Tooltip first, it renders behind other elements like Bars. SVG doesn't have z-index so that's not an option. I updated all storybooks to reflect that; we will need to put it as a breaking change.

This should also completely remove the `foreignObject` warning since every Tooltip is now rendered inside SVG.

## Related Issue

## How Has This Been Tested?

There are some minor visual differences, not sure why. Here's the diff results I ran from my computer: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=860

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -78,7 +78,6 @@ import {
 } from '../util/types';
 import { AccessibilityManager } from './AccessibilityManager';
 import { isDomainSpecifiedByUser } from '../util/isDomainSpecifiedByUser';
-import { Cursor } from '../component/Cursor';
 import { ChartLayoutContextProvider } from '../context/chartLayoutContext';
 import { AxisMap, CategoricalChartState } from './types';
 import { AccessibilityContextProvider } from '../context/accessibilityContext';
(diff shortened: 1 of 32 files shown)


ckifer (commented):
Code looks good 🚀

I'm concerned that people are going to hate this change, I think I would because from an external perspective it seems arbitrary. Why should I have to care that recharts uses SVGs under the hood and z index can't be set? 

I wonder if we could use react portals to always render where the user can see the tooltip no matter the position it's in?

PavelVanecek:
Yeah I was trying hard to think how to spin this as a feature but I don't have anything. I'll see if Portals will fit.

Alternatively we can have Tooltip be just a dummy component that pushes its props to the state and the actual rendering happens elsewhere. That will prevent the Tooltip from being used outside of the recharts context but perhaps that's less of a problem?

ckifer:
Going to ship this for now

ckifer (approved):

ckifer:
Related issue https://github.com/recharts/recharts/issues/2458

PR #4441, opened 2024-04-17
Tooltip portal

PavelVanecek:
## Description

This turned out to be easy now that tooltip is in context.

## Related Issue

https://github.com/recharts/recharts/pull/4435

https://github.com/recharts/recharts/issues/2458

## Motivation and Context

Make sure Tooltip always renders on top of other content, and Cursor always below

## How Has This Been Tested?

tests, storybooks

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Tooltip.tsx
+++ b/src/component/Tooltip.tsx
@@ -17,6 +17,8 @@ import { useAccessibilityLayer } from '../context/accessibilityContext';
 import { useGetBoundingClientRect } from '../util/useGetBoundingClientRect';
 import { Cursor, CursorDefinition } from './Cursor';
 import { useTooltipEventType } from '../state/selectors';
+import { useTooltipPortal } from '../context/tooltipPortalContext';
+import { createPortal } from 'react-dom';
 
 export type ContentType<TValue extends ValueType, TName extends NameType> =
   | ReactElement
@@ -105,6 +107,7 @@ function TooltipInternal<TValue extends ValueType, TName extends NameType>(props
   const accessibilityLayer = useAccessibilityLayer();
   const { active: activeFromContext, payload, coordinate, label } = useTooltipContext();
   const tooltipEventType = useTooltipEventType(shared);
+  const tooltipPortal = useTooltipPortal();
   /*
    * The user can set `active=true` on the Tooltip in which case the Tooltip will stay always active,
    * or `active=false` in which case the Tooltip never shows.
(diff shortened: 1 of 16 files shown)


src/component/TooltipBoundingBox.tsx
```diff
@@ -111,35 +111,18 @@ export class TooltipBoundingBox extends PureComponent<TooltipBoundingBoxProps, S
...
-     * The x and 0 and width and height are static and cover the whole SVG document, because
-     * the tooltip itself is positioned via context.
-     *
-     * pointer-events: none is necessary because this foreignObject covers the whole chart,
-     * and if it was catching mouse events it would steal them all before they reach the chart
-     * and tooltip would never open.
-     */
     return (
-      <g>
-        <foreignObject x="0" y="0" width="100%" height="100%" style={{ pointerEvents: 'none' }}>
```
PavelVanecek, line 129: Thanks to portal, Tooltip is rendered in HTML again and the foreignObject is no longer required.

ckifer:
image
Your description says the cursor is always below other content but it looks like it isn't from the storybook

src/component/Tooltip.tsx
```diff
@@ -113,6 +124,13 @@ function TooltipInternal<TValue extends ValueType, TName extends NameType>(props
    */
   const finalIsActive = activeFromProps ?? activeFromContext;
   const [lastBoundingBox, updateBoundingBox] = useGetBoundingClientRect(undefined, [payload, finalIsActive]);
+
+  const tooltipPortal = portalFromProps ?? tooltipPortalFromContext;
+  const cursorPortal = useCursorPortal();
+  if (tooltipPortal == null || cursorPortal == null) {
```
ckifer, line 130: what if someone doesn't want a cursor?

ckifer (changes requested):
Looks great! Just concerned about the ordering

src/component/Tooltip.tsx
PavelVanecek: The portal will be there always, the cursor inside will return null

PavelVanecek:
I think the screenshot is from build 863? I ran that one, then I realized I need to have a separate cursor portal, then I rerun. My bad, I should have generated a new diff, here it is: https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=66211568b767d61b79aa05d1

PavelVanecek:
For comparison here's a codesandbox I made on 2.x version - the tooltip wasn't clipped then. https://codesandbox.io/p/sandbox/tall-tooltip-739pqy?file=%2Fsrc%2FApp.tsx%3A62%2C16

ckifer:
Sweet sorry, I could've found out post-merge. Thanks!

ckifer (approved):

PR #4453, opened 2024-04-24
Automatically adjust stepRatioScale to fix weired behaviour when calculating yAxis-domain

ForestLinSen:
## Description

This PR is for an issue where minor changes in data could affect the Y-axis domain significantly(an example mentioned in the issue #4020 [here](https://codesandbox.io/p/sandbox/bar-chart-stacked-by-sign-forked-3hnrht?file=%2Fsrc%2FApp.tsx%3A22%2C17)). Various parameters impact this, but a relatively straightforward solution, after quite a few experiments, is to adjust the `stepRatioScale`.

This value was previously hardcoded to 0.05, it now can automatically adjust to make sure the calculated step is within the reasonable range

## Related Issue

[Weired behaviour when calculating yAxis domain: ["auto", "auto"]](https://github.com/recharts/recharts/issues/4020)

## How Has This Been Tested?

Locally tested and unit test

image
image

## Types of changes

## Checklist:


src/chart/generateCategoricalChart.tsx
```diff
@@ -802,6 +803,7 @@ export interface CategoricalChartProps {
   data?: any[];
   layout?: LayoutType;
   stackOffset?: StackOffsetType;
+  stepRatioControl?: StepRatioControl;
```
ckifer, line 806: one idea I had was, instead of making this a separate parameter that gets passed to every chart at the highest level why don't we associate it with the domain (since thats where it is used). For example, we can keep everything as-is but allow an object as part of domain specification.

`domain={'auto'}`

OR

`domain={[0, 10000]}`

OR

`domain={(min, max) => [min, max]}

OR

`domain={{ domain: 'auto' , settings: stepRatioControl: 0.1 }}` 
(or maybe somehow only allow it if domain is auto so there is no confusion)


Or if not the above then

`domainSettings={{ stepRatioControl: 0.1 }}`

on the X/YAxis?

This way there is less ambiguity against what the parameter affects/changes. Otherwise it seems arbitrary and hard to follow like the discussion in the other thread...

ForestLinSen: yeah I think combining this parameter with something is a good idea, but my concern is that it would require a lot of changes to the code that currently assumes the input value is an array containing two numeric values when the domain is set to `auto` - changing it to an object would add a lot messy. Another concern is that making it an object allows users to set any number, which might be problematic if the value is being set to too high/ low, without the enum type they would need an even better understanding of how to change the value. Anyway I like the idea, just wondering if there's a better way to combine this parameter

PavelVanecek:
I would like to do something similar to what the d3 scales are doing, or
what ggplot2 is doing. Instead of different parameters they accept whole
different function and then call it. That way we keep the API simple and
flexible.

Domain already has a support for it, so we can export various domain
functions and allow people swapping between them.

ckifer:
That would work for me, but we still have to expect a certain result to be returned and do something with it

ForestLinSen:
Got an idea: if we don't care too much about customization and just want to fix the bug, I could change the algorithm logic to check if the calculated step is within a reasonable range, if not the `stepRatioScale` should automatically minus 0.01 everytime until it satisfy the requirement. What do you guys think?

ckifer:
I'd be fine with that. What's the definition of a reasonable range?

ForestLinSen:
> I'd be fine with that. What's the definition of a reasonable range?

like if `calculatedStep * 0.6 > (max-min)/tickCount` then it's beyond the reasonable rage

ckifer:
can you give an example?

If had the domain from the original issue https://github.com/recharts/recharts/issues/4020 
what gets plugged into this formula and how does it help?

```
let data2 = [
  {
    date: "aaa",
    added: 3001,
    removed: -1000,
  },
  {
    date: "bbb",
    added: 2000,
    removed: -1000,
  },
  {
    date: "ccc",
    added: 1000,
    removed: -2001,
  },
];
```

min = -2001, max = 3001

ForestLinSen:
> can you give an example?
> 
> If had the domain from the original issue #4020 what gets plugged into this formula and how does it help?
> 
> ```
> let data2 = [
>   {
>     date: "aaa",
>     added: 3001,
>     removed: -1000,
>   },
>   {
>     date: "bbb",
>     added: 2000,
>     removed: -1000,
>   },
>   {
>     date: "ccc",
>     added: 1000,
>     removed: -2001,
>   },
> ];
> ```
> 
> min = -2001, max = 3001

so here max - min = 5001, giving the roughStep 5001/4 = 1250.25. The initially calculated step when a `stepRatioScale` of 0.05 is 2000, then we have 2000 * 0.7 > 1250.25, it should reduce the the`stepRatioScale` by 0.01 until it satisfies the requirement. The result after adding this check looks like this:

image

ckifer:
> 2000 * 0.7 > 1250.25 

where does 0.7 come from?

ForestLinSen:
> > 2000 * 0.7 > 1250.25 
> 
> 
> 
> where does 0.7 come from?

It's just the parameter we can choose to keep the calculated result within the reasonable range, ex. if it's 0.1 basically it has no effect at all because calculated result couldn't reach that high, if it's 0.9 the check mechanism could be too sensitive

ckifer:
Ok I finally took a second to step through the current code and understand it much better. This makes a lot of sense to me, I think for "auto" we should keep "auto" functionality subjective and within recharts. 

Lets try your idea out! And maybe we can add more comments as to what these functions are doing while we're at it. The Decimal library makes things a little hard to read

ckifer:
Alternatively, does directly using D3 help us at all? 
https://d3js.org/d3-array/ticks#tickStep

ForestLinSen:
I have updated the code - should be a smaller change now

ckifer:
Will take a look when I can, thanks

src/util/scale/getNiceTickValues.ts
```diff
@@ -45,12 +45,19 @@ export const getFormatStep = (roughStep: Decimal, allowDecimals: boolean, correc
...
-  const stepRatioScale = digitCount !== 1 ? 0.05 : 0.1;
-  const amendStepRatio = new Decimal(Math.ceil(stepRatio.div(stepRatioScale).toNumber()))
-    .add(correctionFactor)
-    .mul(stepRatioScale);
+  let stepRatioScale = digitCount !== 1 ? 0.05 : 0.1;
 
-  const formatStep = amendStepRatio.mul(digitCountValue);
+  let formatStep: Decimal;
+
+  do {
```
ckifer, line 52: this should be in a regular while loop I think? In a do-while 0.05 will always get decremented to 0.04 before the check

src/util/scale/getNiceTickValues.ts
```diff
@@ -45,12 +45,19 @@ export const getFormatStep = (roughStep: Decimal, allowDecimals: boolean, correc
...
+    const amendStepRatio = new Decimal(Math.ceil(stepRatio.div(stepRatioScale).toNumber()))
+      .add(correctionFactor)
+      .mul(stepRatioScale);
+
+    formatStep = amendStepRatio.mul(digitCountValue);
+
+    stepRatioScale -= 0.01;
+  } while (formatStep.mul(0.7).gt(roughStep) && stepRatioScale > 0.02);
```
ckifer, line 60: can we put 0.7 into a constant and name it so we know what it is later? Its a magic number at this point

ckifer (changes requested):
few things

ForestLinSen:
> few things

good points, have updated the code and also added another two constants

ckifer (approved):
this looks good to me!

@PavelVanecek tagging one more time for a look. This will cause defaults to change, it will be breaking, but it won't be as confusing as adding the ability to configure youself. It just tries to be a little more forgiving when calculating tick value steps

Scutua (approved):

ckifer:
I'm going to close this as stale for now

PR #4472, opened 2024-05-01
Add tests for Pie keyboard controls

PavelVanecek:
## Description

I wonder why is this feature only on Pie? Surely other charts should have keyboard controls too?

## Related Issue

## Motivation and Context

I want to refactor this next, but tests first.

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/test/helper/focus.ts
+++ b/test/helper/focus.ts
@@ -1,0 +1,18 @@
+import { assertNotNull } from './assertNotNull';
+
+export function focusTestHelper(
+  container: HTMLElement | SVGElement,
+  selector: string,
+  debug?: () => void,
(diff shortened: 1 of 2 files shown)


PavelVanecek:
And indeed, this code pre-dates the AccessibilityManager. Perhaps it's duplicated? I will try to remove it and see if AccessibilityManager alone can handle that. In a followup PR.

ckifer (approved):
We do get keyboard interaction on all of the categorical charts with accessibility layer, which we still need to default to true 😅. But this is a case of contributors adding the things they want and not worrying about the rest

PR #4474, opened 2024-05-02
chore: upgrade react, react types, test deps in devDependencies

ckifer:
## Description

upgrade dev deps. Specifically react types for longer term fix to linked issue

also upgrade:

- react (used in demo and what we're testing against technically)
- react-dom
- jsdom
- @testing-library/jest-dom
- @testing-library/react
- @testing-library/user-event

Some changes needed to be made to tests including `legacyRoot` in certain places to prevent failing tests in general. 
`ReactNode` now includes ReactPortal and `Iterator<ReactNode>`

## Related Issue

https://github.com/recharts/recharts/issues/4382

## Motivation and Context

- trying to get ahead of dependency upgrades for 3.x, uncovers some issues in our tests

## How Has This Been Tested?

npm test, chromatic build

https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=894

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] deps

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] I have added a storybook story or extended an existing story to show my changes

PR #4478, opened 2024-05-03
Use act to remove legacyRoot rendering from tests

PavelVanecek:
## Description

Followup from https://github.com/recharts/recharts/pull/4474


ckifer (approved):
👏🏼

PR #4482, opened 2024-05-04
Keyboard controls tests for Funnel Chart

PavelVanecek:
Tests only.

I will follow up with the other charts in next PR.

### DIFF
--- a/test/chart/AccessibilityLayer.spec.tsx
+++ b/test/chart/AccessibilityLayer.spec.tsx
@@ -7,713 +7,718 @@ import { getTooltip } from '../component/Tooltip/tooltipTestHelpers';
 import { PageData } from '../_data';
 
 describe.each([true, undefined])('AccessibilityLayer with accessibilityLayer=%s', accessibilityLayer => {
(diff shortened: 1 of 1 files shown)


test/chart/AccessibilityLayer.spec.tsx
```diff
@@ -1,546 +1,688 @@
...
+            <Legend />
+            <XAxis dataKey="name" />
+            <YAxis />
+          </AreaChart>
+        </div>
+      );
+    };
+
+    // Q: What is the point of this test?
+    // A: Well, I was thinking, what was a way this could break and would be hard to figure out?
```
ckifer, line 380: Cool! Future us thanks you

ckifer (approved):

test/chart/AccessibilityLayer.spec.tsx
PavelVanecek: (Wasn't my code, I only added indentation. I think Julianna added this one.)

ckifer: Oh hah, either way

PR #4485, opened 2024-05-05
PieChart keyboard tests, set accessibilityLayer to true in storybook args

PavelVanecek:
## Description

Tests and storybook only

## Related Issue

## Motivation and Context

I want to refactor Pie to use context next, but that happens to store all the DOM elements as refs for what looks like an a11y feature. I don't want to store refs but I don't want to break keyboard controls either.

Also turns out that the keyboard features on Pie do not show tooltip anyway.

Also https://github.com/recharts/recharts/discussions/4484

## How Has This Been Tested?

npm test, storybook

## Types of changes

## Checklist:


ckifer (approved):
Other charts should probably just implement a version of accessibility layer? If we have valid reasons to store refs we do and I think that's fine

PavelVanecek:
I was thinking if we have the defaultIndex prop then all charts could use that for accessibility purposes too?

PR #4496, opened 2024-05-06
Bar reads tooltip trigger handlers from context instead of cloning

PavelVanecek:
## Description

Tooltip event handlers are the last blocker before we can remove `renderGraphicChild`, and with it `renderMap` and `renderByOrder`.

This is how things could work; I used Bar as the first implementation because I think it's the most complex one. Others will follow in a separate PR if we agree this is the way to go.

## Related Issue

## Motivation and Context

Less element cloning

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -8,41 +8,46 @@ import isEqual from 'lodash/isEqual';
 import isNil from 'lodash/isNil';
 import { Props as RectangleProps } from '../shape/Rectangle';
 import { Layer } from '../container/Layer';
-import { ErrorBar, Props as ErrorBarProps, ErrorBarDataPointFormatter } from './ErrorBar';
+import { ErrorBar, ErrorBarDataPointFormatter, Props as ErrorBarProps } from './ErrorBar';
 import { Cell } from '../component/Cell';
 import { LabelList } from '../component/LabelList';
-import { uniqueId, mathSign, interpolateNumber } from '../util/DataUtils';
+import { interpolateNumber, mathSign, uniqueId } from '../util/DataUtils';
 import { filterProps, findAllByType } from '../util/ReactUtils';
 import { Global } from '../util/Global';
 import {
+  findPositionOfBar,
+  getBaseValueOfBar,
   getCateCoordinateOfBar,
+  getTooltipItem,
   getValueByDataKey,
   truncateByDomain,
-  getBaseValueOfBar,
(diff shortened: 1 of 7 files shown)


src/cartesian/Bar.tsx
```diff
@@ -194,6 +228,16 @@ function BarRectangles(props: BarRectanglesProps) {
 
   const { index: activeIndex } = useTooltipContext();
 
+  const onMouseEnterFromContext = useMouseEnterItemDispatch();
```
ckifer, line 231: why do we do all of this twice? Seems redundant. Is it so we're calling the events if we're in the background or the actual bar?

src/chart/generateCategoricalChart.tsx
```diff
@@ -1569,11 +1577,12 @@ export const generateCategoricalChart = ({
...
     /**
      * The handler of mouse entering a graphical item, such as bar, pie, scatter, funnel, ...
+     * This handler is used for `tooltipEventType: item` and both `Tooltip.trigger: hover` and `Tooltip.trigger: click`
      * @param el The active graphical element
      * @param index 0-based index of the active graphical element
-     * @returns undefined
+     * @return undefined
      */
-    handleItemMouseEnter = (el: any, index: number) => {
+    handleItemMouseEnter: ActivateTooltipAction = (el, index) => {
```
ckifer, line 1585: at this point what is the value of even having these here? Would all hell break loose if we just said lets define them in each graphical item (kind of like what we're doing above anyways) and be done with it? we could set tooltip state from there no?

ckifer (commented):
Just some questions - I think I agree with this approach, but it seems like we're specifying these events a lot of places

src/cartesian/Bar.tsx
PavelVanecek: Because both bars and backgrounds react to mouse events, and this is the only place that's functional component and allows hooks and so this is here twice.

There's some opportunity to DRY, I will iterate in next PR.

src/chart/generateCategoricalChart.tsx
PavelVanecek: generateCategoricalChart still needs the tooltip data in its internal state. It's using it for synchronising different charts and possibly other things too.

ckifer: ah right, forgot about synchronization

ckifer (approved):

PR #4497, opened 2024-05-07
test(reference-line): add test that will fail if we remove defaultProps

ckifer:
## Description

Add unit test that will fail when defaultProps are removed for referenceline. Realized that it seems everything is peachy still when I went to try to remove them again. Nope, still broken and this shows it 

## Related Issue

Don't really want to link the defaultProps issue

## Motivation and Context

testsss

## How Has This Been Tested?

tests pass

## Types of changes

## Checklist:


PavelVanecek (approved):

PavelVanecek:
I didn't even know it does that!

ckifer:
Yeah it matches reference items to axis ids and then for each reference item looks to see if it needs to extend the domain. 

I started writing the context but it's weird with everything being done inside the generator before any info is even passed to the components

PR #4505, opened 2024-05-08
Pie tooltip events in context

PavelVanecek:
## Description

Pie is another graphical component that uses context to handle tooltip mouse events.

Also includes fixes for the stories I broke in https://github.com/recharts/recharts/pull/4490

## Related Issue

## Motivation and Context

element cloning

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/Examples/Pie/PieWithNeedle.stories.tsx
+++ b/storybook/stories/Examples/Pie/PieWithNeedle.stories.tsx
@@ -1,12 +1,11 @@
 /* eslint-disable no-shadow */
 import React from 'react';
-import { Cell, Pie, PieChart, ResponsiveContainer } from '../../../../src';
+import { Cell, Pie, PieChart, ResponsiveContainer, Tooltip } from '../../../../src';
 
 export default {
   component: Pie,
 };
 
-const RADIAN = Math.PI / 180;
 const data = [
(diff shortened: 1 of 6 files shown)


PavelVanecek:
I am not mad that this change increased test coverage but I have no clue why in this function - I didn't aim for it

image

storybook/stories/Examples/Pie/PieWithNeedle.stories.tsx
```diff
@@ -71,6 +81,7 @@ export const PieWithNeedle = {
             fill="none"
             activeShape={Needle}
           />
+          <Tooltip defaultIndex={1} />
         </PieChart>
```
ckifer, line 85: I guess I do think it's kind of weird we're using tooltip to set the active shape state. Not a blocker but feel like it's not that intuitive

ckifer (approved):

storybook/stories/Examples/Pie/PieWithNeedle.stories.tsx
PavelVanecek: Yep I was wondering the same. I think the cause is the new methods for registering tooltipEventType - the new logic depends on tooltip being present in the DOM. Which will change later.

PR #4510, opened 2024-05-09
feat: add useDisplayedChartData to splice by start/end index

ckifer:
## Description

idk if I made up that we need this or if we actually do

Mostly just making this PR for review @PavelVanecek where should we store sliced data? Slice it only as we read it like this?

Move data to redux? idk 🙃 

## Types of changes

## Checklist:


PavelVanecek:
I think it's fine like this. Redux selectors will give us memoization out of the box.

PR #4513, opened 2024-05-09
All tooltip item events in context

PavelVanecek:
## Description

All tooltip item events are in context now.

I also fixed a problem that we have observed in https://github.com/recharts/recharts/pull/4505#discussion_r1594101983 - that is, the graphical components did not activate their `activeShape` when Tooltip was not present in the DOM.

With this PR, the `activeShape` does get activated on mouse hover, but we still need the Tooltip to set `defaultIndex` to make space for the now missing `activeIndex`.

## Related Issue

## Motivation and Context

No element cloning

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:


PavelVanecek:
```
renderGraphicChild = (element: React.ReactElement, displayName: string, index: number): ReactElement | null => {
  const item = this.filterFormatItem(element, displayName, index);
  if (!item) {
    return null;
  }

  return cloneElement(element, item.props);
};
```

Okay with this method now being super slim, we can start thinking on how to get rid of `static getComposedData`.

ckifer (changes requested):
Nicee

ckifer (approved):

PR #4517, opened 2024-05-10
Preparing for Tooltip payload refactor

PavelVanecek:
## Description

Okay this is a little bit all over the place. Basically I am trying to refactor the Tooltip payload functions into Redux store, and it's a fairly big change. Here I identified and committed few isolated changes that are not quite enough to move the tooltip yet but they pass tests so I think we can merge. See the commit messages for individual changes.

Let me know if you want me to split this into individual pull requests instead.

## Related Issue

## Motivation and Context

The tooltip refactor must happen first before we can remove formattedGraphicalItems which must happen before we can remove renderGraphicChild which we need to do before we can remove the element cloning.

## How Has This Been Tested?

npm test, storybook

## Types of changes

## Checklist:

### DIFF
--- a/src/context/chartDataContext.tsx
+++ b/src/context/chartDataContext.tsx
@@ -1,10 +1,22 @@
-import { createContext, useContext } from 'react';
+import { createContext, useContext, useEffect } from 'react';
+import { ChartData, setChartData } from '../state/chartDataSlice';
+import { useAppDispatch, useAppSelector } from '../state/hooks';
+import { RechartsRootState } from '../state/store';
 
-const ChartDataContext = createContext<any[] | undefined>(undefined);
 const DataStartIndexContext = createContext<number>(0);
 const DataEndIndexContext = createContext<number>(0);
 
-export const ChartDataContextProvider = ChartDataContext.Provider;
+export const ChartDataContextProvider = (props: { chartData: ChartData }): null => {
+  const { chartData } = props;
+  const dispatch = useAppDispatch();
+  useEffect(() => {
+    dispatch(setChartData(chartData));
+    return () => {
+      dispatch(setChartData([]));
+    };
+  }, [chartData, dispatch]);
+  return null;
+};
 export const DataStartIndexContextProvider = DataStartIndexContext.Provider;
 export const DataEndIndexContextProvider = DataEndIndexContext.Provider;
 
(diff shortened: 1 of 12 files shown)


src/state/hooks.ts
```diff
@@ -30,8 +38,12 @@ export const useAppDispatch = (): AppDispatch => {
  */
 export function useAppSelector<T>(selector: (state: RechartsRootState) => T): T | undefined {
   const context = useContext(ReactReduxContext);
-  if (context) {
```
PavelVanecek, line 33: This is an interesting one.

The old code was not enough to re-run the selector on state change; the Redux context itself never actually changes, so React was never calling this hook for the second time.

The right way is to use store.subscribe and store.getState() which conveniently is precisely what `useSyncExternalStoreWithSelector` does. I used the shim to allow us to continue supporting React 16. After we decide to support only React 18+ we can import that from `react` directly.

src/context/chartDataContext.tsx
```diff
@@ -24,7 +36,8 @@ export const DataEndIndexContextProvider = DataEndIndexContext.Provider;
  *
  * @return data array for some charts and undefined for other
  */
-export const useChartData = () => useContext(ChartDataContext);
+export const useChartData = (): ChartData | undefined =>
```
PavelVanecek, line 39: Brush is the only one that uses this hook and it is tested well so I am fairly confident this works just as before.

src/context/chartDataContext.tsx
```diff
@@ -1,10 +1,22 @@
...
+import { ChartData, setChartData } from '../state/chartDataSlice';
+import { useAppDispatch, useAppSelector } from '../state/hooks';
+import { RechartsRootState } from '../state/store';
 
-const ChartDataContext = createContext<any[] | undefined>(undefined);
 const DataStartIndexContext = createContext<number>(0);
 const DataEndIndexContext = createContext<number>(0);
 
-export const ChartDataContextProvider = ChartDataContext.Provider;
+export const ChartDataContextProvider = (props: { chartData: ChartData }): null => {
```
ckifer, line 9: this isn't context anymore though?

src/state/hooks.ts
ckifer: very interesting.. 
I don't have great knowledge of this hook and how it works

ckifer (approved):

src/context/chartDataContext.tsx
PavelVanecek: Correct, I will migrate the other properties too and update names. Later.

PR #4518, opened 2024-05-11
Tooltip data in redux, Line first

PavelVanecek:
## Description

Okay this is how I think Tooltip can work. The tooltip state is not used anywhere yet, that will come in next PR. If we can merge as-is I will follow up.

Good news is - Tooltip works, it reads data from graphical children without the need to lookup all the children in generateCategoricalChart.

## Related Issue

## Motivation and Context

Element cloning

## How Has This Been Tested?

npm test - passes
storybooks - working

if I uncomment the Tooltip code and try - LineChart works, tests fail because of two unfinished tasks:
- Tooltip uses either context payload, or props payload, not both - so the test that has ComposedChart with Area + Bar + Line fails. This will be fixed when all three Area + Bar + Line use redux, can't have that done with just one.
- It's missing the part where it should read props from Tooltip and use those
- ~Brush slicing is not yet implemented~ okay I fixed and pushed this one

I will fix both.

## Types of changes

## Checklist:

### DIFF
--- a/src/state/tooltipSlice.ts
+++ b/src/state/tooltipSlice.ts
@@ -1,0 +1,75 @@
+import { createSlice, current, PayloadAction } from '@reduxjs/toolkit';
+import { NameType, Payload, ValueType } from '../component/DefaultTooltipContent';
+
+/**
+ * One Tooltip can display multiple TooltipPayloadEntries at a time.
+ */
+export type TooltipPayloadEntry = Payload<ValueType, NameType>;
+
+/**
+ * So what happens is that the tooltip payload is decided based on the available data, and the dataKey.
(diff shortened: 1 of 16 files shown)


src/state/tooltipSlice.ts
PavelVanecek, line 1: I tried to explain which is which, let me know if I should add more details.

src/util/ChartUtils.ts
```diff
@@ -1330,6 +1338,35 @@ export const getTooltipItem = (
   };
 };
 
+export function getTooltipEntry({
```
PavelVanecek, line 1341: This no longer depends on `displayName` which will enable us to switch to functional components.

ckifer (approved):

PR #4520, opened 2024-05-12
Brush storybook diff

PavelVanecek:
## Description

Fixing the bug I introduced in https://github.com/recharts/recharts/pull/4518

### DIFF
--- a/storybook/stories/Examples/cartesian/Brush/Brush.stories.tsx
+++ b/storybook/stories/Examples/cartesian/Brush/Brush.stories.tsx
@@ -1,4 +1,3 @@
-import { expect, within, userEvent } from '@storybook/test';
 import React, { useState } from 'react';
 import {
   ComposedChart,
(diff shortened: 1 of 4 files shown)


storybook/stories/Examples/cartesian/Brush/Brush.stories.tsx
```diff
@@ -61,24 +60,6 @@ export const ControlledBrush = {
       </>
     );
   },
-  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
-    const user = userEvent.setup();
-    const { getByLabelText } = within(canvasElement);
```
ckifer, line 66: Why remove this test? Pretty sure it's still a valid use case no?

ckifer (approved):

storybook/stories/Examples/cartesian/Brush/Brush.stories.tsx
ckifer: Oh I see the message now. I don't necessarily agree? Asserting that certain things with in a real browser environment is still important. Maybe not this, but things that measure in the dom for example

PavelVanecek: The test runs on page load, makes the storybook blink twice. The failure is not recorded or displayed. As such it doesn't serve the purpose.

I'm supportive of real browser tests (we will need them to test touch I think) but our storybook setup as is today is not it.

PR #4521, opened 2024-05-12
Area + Bar send their tooltip data to redux store

PavelVanecek:
## Description

Following up on the idea from https://github.com/recharts/recharts/pull/4518

This is still failing tests - to start actually using the Redux data in Tooltip, I need to migrate the tooltipAxis next.

## Related Issue

## Motivation and Context

No element cloning.

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Tooltip.tsx
+++ b/src/component/Tooltip.tsx
@@ -115,7 +115,7 @@ function TooltipInternal<TValue extends ValueType, TName extends NameType>(props
   const viewBox = useViewBox();
   const accessibilityLayer = useAccessibilityLayer();
   const { active: activeFromContext, payload: payloadFromProps, coordinate, label } = useTooltipContext();
-  // TODO this will fail tests until Area, Bar, and Line all push their own payloads
+  // TODO this will fail tests until Tooltip props and tooltipAxis are in Redux
   // const payloadFromContext = useAppSelector(selectTooltipPayload);
   // const payload = payloadFromContext?.length > 0 ? payloadFromContext : payloadFromProps;
   const payload = payloadFromProps;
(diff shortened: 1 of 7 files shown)


src/component/Tooltip.tsx
```diff
@@ -115,7 +115,7 @@ function TooltipInternal<TValue extends ValueType, TName extends NameType>(props
   const viewBox = useViewBox();
   const accessibilityLayer = useAccessibilityLayer();
   const { active: activeFromContext, payload: payloadFromProps, coordinate, label } = useTooltipContext();
-  // TODO this will fail tests until Area, Bar, and Line all push their own payloads
+  // TODO this will fail tests until Tooltip props and tooltipAxis are in Redux
```
ckifer, line 118: What even is tooltipAxis? That seems like something we should rename? Is it an axis or is it tooltip information?

ckifer (approved):

src/component/Tooltip.tsx
PavelVanecek: One of the axes is selected based on layout and dice roll and that's treated as more important than others. I'm following what was already there.
Seems to me that most people only use one axis anyway so this doesn't cause too many problems? We can look into it... Later.

ckifer: Ah right. We can use axisIds to match axes the same as before though yeah?

PavelVanecek: I am planning to use the exact same logic as was in generateCategoricalChart, only move it to Redux so that it no longer requires direct DOM access.

PR #4524, opened 2024-05-13
Moving Axes to Redux

PavelVanecek:
## Description

One step closer in my quest of rendering Tooltip from Redux store alone.

Tooltip is not yet functional - for that it needs Tooltip.shared and Scatter - but it's almost there. Builds are green so I figured we might as well merge it.

## Related Issue

## Motivation and Context

No direct DOM access

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/src/state/selectors.ts
+++ b/src/state/selectors.ts
@@ -1,7 +1,9 @@
+import { createSelector } from '@reduxjs/toolkit';
 import { useAppSelector } from './hooks';
 import { RechartsRootState } from './store';
-import { TooltipPayload } from './tooltipSlice';
+import { TooltipPayload, TooltipState } from './tooltipSlice';
 import { getTooltipEntry, getValueByDataKey } from '../util/ChartUtils';
+import { ChartDataState } from './chartDataSlice';
 
 export const useChartName = (): string => {
   return useAppSelector((state: RechartsRootState) => state.options.chartName);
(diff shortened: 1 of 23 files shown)


ckifer (approved):

ckifer:
Nice

src/state/selectors.ts
```diff
@@ -26,29 +31,60 @@ function getSliced<T>(arr: ReadonlyArray<T>, startIndex: number, endIndex: numbe
...
+      return undefined;
+    }
+    const { chartData, dataStartIndex, dataEndIndex } = chartDataState;
+
+    /*
+     * If a payload has data specified directly from the graphical item, prefer that.
+     * Otherwise, fill in data from the chart level, using the same index.
+     */
+    return tooltipItemPayloads.map(({ dataDefinedOnItem, settings }) => {
+      const finalData = dataDefinedOnItem ?? chartData;
```
ckifer, line 72: this will need changed to a check for length or `dataDefinedOnItem` will need to be null/undefined

PavelVanecek: Possible. Do we have all the cases covered with tests?

ckifer: no https://github.com/recharts/recharts/blob/3.x/src/cartesian/Scatter.tsx#L205
😅

ckifer: https://github.com/search?q=repo%3Arecharts%2Frecharts%20data%3A%20%5B%5D&type=code

Those first three there will all return `dataDefinedOnItem` every time

PavelVanecek: That might actually explain some of the failures I saw.

src/state/selectors.ts
```diff
@@ -26,29 +31,60 @@ function getSliced<T>(arr: ReadonlyArray<T>, startIndex: number, endIndex: numbe
...
+      if (tooltipAxis?.dataKey && !tooltipAxis?.allowDuplicatedCategory) {
+        tooltipPayload = findEntryInArray(sliced, tooltipAxis.dataKey, activeLabel);
+      } else {
+        tooltipPayload = sliced?.[activeIndex];
+      }
+      return getTooltipEntry({
+        tooltipEntrySettings: settings,
+        dataKey: settings?.dataKey,
+        payload: tooltipPayload,
+        value: getValueByDataKey(tooltipPayload, settings?.dataKey),
```
ckifer, line 86: I think this also needs to be `settings?.dataKey ?? tooltipAxis?.dataKey`

Else no value will be rendered in tooltip when the axis has the dataKey (at least for Scatter)

ckifer: Also really wonder how Cell is going to work :/

src/state/selectors.ts
```diff
@@ -26,29 +31,60 @@ function getSliced<T>(arr: ReadonlyArray<T>, startIndex: number, endIndex: numbe
...
+  selectTooltipAxis,
+  selectActiveLabel,
+  (
+    tooltipState: TooltipState,
+    chartDataState: ChartDataState,
+    tooltipAxis: BaseAxisProps | undefined,
+    activeLabel: string | undefined,
+  ) => {
+    const { activeIndex, tooltipItemPayloads } = tooltipState;
+    if (activeIndex === -1) {
```
ckifer, line 62: activeShape isn't getting reset anymore
https://63da8268a0da9970db6992aa-pkiagidnpj.chromatic.com/?path=/story/examples-scatterchart--simple

https://63da8268a0da9970db6992aa-pkiagidnpj.chromatic.com/?path=/docs/api-cartesian-bar--docs

Hover over those

PavelVanecek: It gets reset when I hover on another element, and it does not reset (but it should) on mouse leave, is that correct?

src/state/selectors.ts
ckifer: `Scatter` was returning nothing for Tooltip data. Now it is (once the dataKey issue is good to go too)

We also fix a bug where Scatter tooltips never get color styles when we do this

src/state/selectors.ts
ckifer: correct yep

PavelVanecek: I will have a look

src/state/selectors.ts
PavelVanecek: And what's worse, the `getComposedData` will re-inject the `data` prop and overwrite whatever it was that was there before so we lose access to it.

Soooo a new blocker to this blocker, we need to remove `getComposedData` first.

src/state/selectors.ts
PavelVanecek: https://github.com/recharts/recharts/pull/4527

src/state/selectors.ts
PavelVanecek: Also this is, in the tooltipAxis case, applying the dataKey twice? This looks like it shouldn't work at all

Issue #4216, opened 2024-02-21
Enable CSS transition for sectors on hover in PieChart with `isAnimationActive` as `true`


beltranrengifo:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/pie-chart-with-hover-on-sector-gd9nkm)

### Steps to reproduce
- if `isAnimationActive` is true, hover any sector, check the fill transition is not applied
- if `isAnimationActive` is false, hover any sector, check the fill transition is applied

### What is expected?
the fill transition works in both cases

### What is actually happening?
the fill transition only works if `isAnimationActive` is false

| Environment | Info |
|---|---|
| Recharts | v2.12.1 |
| React | 16.14.0 |
| System | 14.3 (23D56) |
| Browser | Chrome Version 121.0.6167.184 (Official Build) (arm64) |

---
I posted this in SO as I thought it was not an issue but I got no response so I'm opening this here. Thank you!

ckifer:
Looks like this is a bug as far as I can tell. When animation is not active, the sectors do not re-render on hover. When animation is active, they do. Probably a rendering optimization to be done here.

beltranrengifo:
> Looks like this is a bug as far as I can tell. When animation is not active, the sectors do not re-render on hover. When animation is active, they do. Probably a rendering optimization to be done here.

Thanks for your inputs @ckifer 
Let me know how I can help, in case this is a bug for new people in this repo 🙌

ckifer:
https://github.com/recharts/recharts/blob/3.x/src/polar/Pie.tsx#L536

This is _probably_ where the issue is, but the fix isn't clear. I personally wouldn't recommend this as a first issue to take up, but feel free to look into it if you'd like

ckifer:
Seems to work fine for me? What about it isn't working

ckifer:
Some improvements were made to this in 3.x alpha, it seems to still cancel rendering but its smoother than it was

PavelVanecek:
See https://recharts.github.io/en-US/examples/PieChartWithCustomizedLabel/

PR #4528, opened 2024-05-14
Fix custom Pie animation bug

ForestLinSen:
## Description

This PR is for the issue with `PieChart` that custom animations do not work when `isAnimationActive` is set to true

## Related Issue

[Enable CSS transition for sectors on hover in PieChart with isAnimationActive as true](https://github.com/recharts/recharts/issues/4216)

## Motivation and Context

After digging into the code, I found that the `key` in the `Animate` component causes the problem. The dynamic key triggers the process of remounting the component, which overrides the custom animation. It also adds rendering burden: if a user sets the mouse hover animation, every time the user hovers over one piece of the pie chart, the entire pie chart will re-render. 

## How Has This Been Tested?

Tested locally

https://github.com/recharts/recharts/assets/41566276/f9662c73-8985-4c85-9b78-af8d91dc7b2a

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -578,9 +578,9 @@ export class Pie extends PureComponent<Props, State> {
   }
 
   renderSectorsWithAnimation() {
-    const { sectors, isAnimationActive, animationBegin, animationDuration, animationEasing, animationId } = this.props;
+    const { sectors, isAnimationActive, animationBegin, animationDuration, animationEasing } = this.props;
 
-    const { prevSectors, prevIsAnimationActive } = this.state;
+    const { prevSectors } = this.state;
 
     return (
       <Animate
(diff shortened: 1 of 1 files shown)


ckifer:
does this not cause duplicate key errors? https://github.com/recharts/react-smooth/blob/master/src/Animate.js#L285C50-L285C64

ForestLinSen:
> does this not cause duplicate key errors? https://github.com/recharts/react-smooth/blob/master/src/Animate.js#L285C50-L285C64

it didn't cause key errors - from my understanding unique key is the requirement for the child in a list; key for Animate here is more like a way to manage the component lifecycle, which causes some troubles in this case

ckifer:
It's still being used to render a list in the code I linked there. Those don't need keys?

ForestLinSen:
have a quick look at the code, correct me if I'm wrong - seems that it just wraps all the children in a <div> and there's no dynamic list being generated there?

ckifer:
From the react docs

>If you return an element or an array of elements with keys from fn, the returned elements’ keys will be automatically combined with the key of the corresponding original item from children. When you return multiple elements from fn in an array, their keys only need to be unique locally amongst each other.

So I think this uses the keys from the original children and we shouldn't need to worry about it, cool!

PR #4530, opened 2024-05-15
Move two functions to separate file, add types

PavelVanecek:
I started with huge ambitions here (as evidences by the branch name) but quickly ran out of steam. Everything depends on everything, I don't see a path forward here.

What I did here is a small improvement and ready to merge but it's not a breathtaking change.


ckifer (approved):
Yeah it's way overwhelming... :/

PR #4531, opened 2024-05-15
Scatter sends item data to Redux store

PavelVanecek:
## Description

Scatter is a little bit different from others - it always pushes an array of tooltip payloads because it always shows two values (one for X, second for Y coordinate).

## Related Issue

## Motivation and Context

Render tooltip from redux only, delete generateCategoricalChart code

## How Has This Been Tested?

npm test, storybook

https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=951

## Types of changes

## Checklist:

### DIFF
--- a/src/state/selectors.ts
+++ b/src/state/selectors.ts
@@ -1,11 +1,11 @@
 import { createSelector } from '@reduxjs/toolkit';
 import { useAppSelector } from './hooks';
 import { RechartsRootState } from './store';
-import { TooltipPayload, TooltipState } from './tooltipSlice';
+import { TooltipPayload, TooltipPayloadEntry, TooltipState } from './tooltipSlice';
 import { getTicksOfAxis, getTooltipEntry, getValueByDataKey } from '../util/ChartUtils';
 import { ChartDataState } from './chartDataSlice';
 import { selectTooltipAxis } from '../context/useTooltipAxis';
-import { BaseAxisProps, TickItem } from '../util/types';
+import { BaseAxisProps, DataKey, TickItem } from '../util/types';
 import { findEntryInArray } from '../util/DataUtils';
 
 export const useChartName = (): string => {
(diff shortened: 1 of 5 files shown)


src/state/selectors.ts
```diff
@@ -47,44 +47,94 @@ const selectActiveLabel = createSelector(
...
-    /*
-     * If a payload has data specified directly from the graphical item, prefer that.
-     * Otherwise, fill in data from the chart level, using the same index.
-     */
-    return tooltipItemPayloads.map(({ dataDefinedOnItem, settings }) => {
-      const finalData = dataDefinedOnItem ?? chartData;
+  const init: Array<TooltipPayloadEntry> = [];
 
-      const sliced = getSliced(finalData, dataStartIndex, dataEndIndex);
+  return tooltipItemPayloads.reduce((agg, { dataDefinedOnItem, settings }): Array<TooltipPayloadEntry> => {
```
ckifer, line 83: hmm this combination does not play nice with `tooltipEventType === 'item'`. We need to know which graphical child (or payload) we are targeting rather than always returning all of them

ckifer: Need something saying "I'm explicitly showing the data for this graphical item". Does each graphical item also dispatch that info?

PavelVanecek: About `tooltipEventType`: the Tooltip sends `shared` property and the selector then selects the appropriate data array based on that.

Each graphical item dispatches all of its data and tooltip settings, and Tooltip then decides which one does it display. This will make it trivial to support `item` and `axis` tooltip types everywhere, and also allow us to support multiple tooltips with different types.

PavelVanecek: Hm but yes I think I see the next problem with `tooltipEventType` 😬 fortunately it's covered by tests!

PavelVanecek: Okay I think I got it. Here's what I would prefer to do:

1. Extend the tooltip state so that it stores more information:
  - Hovered item index
  - Hovered item dataKey
  - Clicked item index
  - Clicked item dataKey
  - same 4 variants for the axis interactions
2. set the tooltip index + dataKey from inside graphical items, instead of the generateCategoricalChart central
3. Update the selector so that it chooses the correct payload, based on the state + tooltip props (shared, trigger)

I have a proof of concept ready and it works with Bar shared=true,false, but it will need some polishing before it's ready to merge. Maybe over the weekend.

What do you think?

ckifer: Yeah that's the direction I started to go too. Need more information about the active item.

Concept sounds good, we'll probably need that info other places anyways

PR #4541, opened 2024-05-17
BREAKING CHANGE: Make `react-is` a peer dependency

eps1lon:
## Description

Make `react-is` a peer dependency instead of a direct dependency

## Motivation and Context

`react-is` needs to be the same version as the used version of React. So far these dependencies could be out of sync since the related internals didn't changed. However, in React 19 related internals did change. Using `react-is<19` would not recognize elements from `react>=19` and vice versa. So generally, `react-is` should be installed by library users just like React. That way, users can make sure the versions match without having to resort to `resolutions` or `overrides` which wouldn't even work for libraries. 

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,10 @@ NPM is the easiest and fastest way to get started using Recharts. It is also the
 
 ```sh
 # latest stable
-$ npm install recharts
+$ npm install recharts react-is
 ```
+
+`react-is` needs to match the version of your installed `react` package.
 
 ### umd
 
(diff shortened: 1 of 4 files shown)


ckifer:
uhh I think this PR was made from the wrong branch. Can you try switching to 3.x and then make the change?

README.md
```diff
@@ -45,16 +45,19 @@ NPM is the easiest and fastest way to get started using Recharts. It is also the
 
 ```sh
 # latest stable
-$ npm install recharts
+$ npm install recharts react-is
 ```
 
+`react-is` needs to match the version of your installed `react` package.
```
ckifer, line 51: peer dependencies are automatically installed, this doesn't need to be in the readme

ckifer (commented):
just one thing, but otherwise LGTM. Thanks for seeing this

README.md
eps1lon: Not by Yarn. This behavior can also be disabled in NPM and PNPM.

ckifer: Hmm...

Hopefully we can just remove the usage of react-is before we release react 19 support. The way its currently used is not ideal/wanted anyways since it supports some ani patterns.

Anyways, appreciate the PR!

ckifer (approved):

PR #4545, opened 2024-05-19
Extract RechartsWrapper to its own component

PavelVanecek:
## Description

I want to use hooks and dispatch here later so it has to be a functional component.

There are charts that are not using generateCategoricalChart, and they have almost the same component and could reuse this one too. With one difference, aria role. This uses `region` the other charts use `application`, which one is correct? @julianna-ciq what would be the correct labelling? I was studying the spec but as far as I can tell, neither is a good fit for interactive charts.

## Related Issue

## Motivation and Context

Use redux instead of element cloning.

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/scripts/snapshots/es6Files.txt
+++ b/scripts/snapshots/es6Files.txt
@@ -108,6 +108,7 @@
   "es6/chart/SunburstChart.js",
   "es6/chart/ScatterChart.js",
   "es6/chart/Sankey.js",
+  "es6/chart/RechartsWrapper.js",
   "es6/chart/RadialBarChart.js",
   "es6/chart/RadarChart.js",
   "es6/chart/PieChart.js",
--- a/scripts/snapshots/libFiles.txt
+++ b/scripts/snapshots/libFiles.txt
@@ -108,6 +108,7 @@
   "lib/chart/SunburstChart.js",
   "lib/chart/ScatterChart.js",
   "lib/chart/Sankey.js",
+  "lib/chart/RechartsWrapper.js",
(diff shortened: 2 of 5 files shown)


ckifer (approved):

ckifer:
Tagging @julianna-langston  as well, I think she has two accounts

julianna-langston:
> With one difference, aria role. This uses region the other charts use application, which one is correct?

The chart should have 1 role, `role="application"`. It should be at the top, in whatever is the container of the chart.

When it comes to `role`s, the only thing we care about is what a screen reader thinks about it. When a screen reader sees `role="application"`, it knows that thing, and all of its descendants, are interactive. In other words, we use `role="application"` because it *makes* the element interactive.

Now, our charts are only interactive *sometimes*. If `accessibilityLayer={false}` (current default, won't be default in 3.x) and there's a `<Tooltip />`, then it's interactive. Otherwise, it's not interactive. In that case, philosophically it'd be better to have `role="img"` at the container level, but I think that'll be difficult to implement, and the impact to a user is minimal.

Let me know if you'd like me to elaborate on how exactly `role="application"` works for screen readers.

PavelVanecek:
Thank you @julianna-langston , I made the change in new PR.

I think it would be valuable to have chart behaviour from a11y point of view written down in a wiki in the repository, and we would use it twice:
1. When making changes or adding tests, we would refer to the document to see what is the correct behaviour
2. When choosing libraries, devs could check out the document and see what exactly is supported or not

Issue #4548, opened 2024-05-19
Move chart synchronisation to Redux


PavelVanecek:
Chart synchronisation is now handled in `CategoricalChartWrapper.applySyncEvent` and `CategoricalChartWrapper.handleReceiveSyncEvent` and `CategoricalChartWrapper.triggerSyncEvent` and relies on direct DOM access. That will block #2788.

Instead of DOM access, write something different. Perhaps Redux middleware will work?

ckifer:
I started writing these tests. Some interesting findings 🙃

ckifer:
So I don't think Middleware works for this right? You can't share state across unrelated stores - you can share a store instance or even wrap providers in a parent provider but I don't think middleware helps?

Unless you mean still using event emission and using middleware as the "when" to dispatch those events?

PavelVanecek:
Indeed middleware alone is not good enough. Event emissions + middleware works.

I am not attached to any special mechanism as long as it doesn't use `updateStateOfAxisMapsOffsetAndStackGroups` and other direct DOM access functions.

ckifer:
Tried to just move this out to its own function component rendered in the generator that copies the current implementation. Issue with that is that tooltip still calculates its own activeIndex in getTooltipData and overrides any of the current redux state.

Something like this... meh

```tsx
import React, { useEffect } from 'react';
import { setActiveMouseOverItemIndex } from '../state/tooltipSlice';
import { setDataStartEndIndexes } from '../state/chartDataSlice';
import { useAppDispatch, useAppSelector } from '../state/hooks';
import { newEventCenter, SYNC_EVENT, SyncEventData } from '../util/Events';
import { CartesianViewBox, Coordinate, LayoutType } from '../util/types';
import { selectTooltipPayload, selectTooltipTicks } from '../state/selectors';

type SyncProviderProps = {
  syncId: string | number;
  syncMethod: 'index' | 'value' | Function;
  eventEmitterSymbol: Symbol;
  layout: LayoutType;
};

const originCoordinate: Coordinate = { x: 0, y: 0 };

export const SyncProvider = ({
  syncId,
  syncMethod,
  eventEmitterSymbol,
  layout,
}: SyncProviderProps): React.ReactNode => {
  const dispatch = useAppDispatch();
  const { offset } = useAppSelector(state => state.layout);
  const tooltipTicks = useAppSelector(selectTooltipTicks);
  // const tooltipPayload = useAppSelector(state => selectTooltipPayload(state));

[...]

PR #4554, opened 2024-05-19
test(synchronized): add base tests for two chart tooltip/cursor sync

ckifer:
## Description

Write some base tests for chart synchronization. Very very similar to other existing tooltip tests

## Related Issue

## Motivation and Context

write tests so we don't break things

## How Has This Been Tested?

 npm test

## Types of changes

- [X] tests

## Checklist:

### DIFF
--- a/test/component/Tooltip/Tooltip.sync.spec.tsx
+++ b/test/component/Tooltip/Tooltip.sync.spec.tsx
@@ -1,0 +1,382 @@
+import React, { ComponentType, ReactNode } from 'react';
+import { describe, expect, it, test } from 'vitest';
+import { fireEvent, queryByText, render } from '@testing-library/react';
+
+import {
+  Area,
+  AreaChart,
+  Bar,
+  BarChart,
+  Brush,
+  CartesianGrid,
+  ComposedChart,
+  Legend,
+  Line,
+  LineChart,
+  PolarAngleAxis,
+  PolarGrid,
+  PolarRadiusAxis,
+  Radar,
+  RadarChart,
+  RadialBar,
+  RadialBarChart,
+  Tooltip,
+  XAxis,
+  YAxis,
+} from '../../../src';
+import { PageData } from '../../_data';
+import { getTooltip, showTooltip } from './tooltipTestHelpers';
+import {
+  areaChartMouseHoverTooltipSelector,
+  barChartMouseHoverTooltipSelector,
+  composedChartMouseHoverTooltipSelector,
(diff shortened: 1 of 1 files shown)


test/component/Tooltip/Tooltip.sync.spec.tsx
```diff
@@ -0,0 +1,382 @@
...
+          <Wrapper syncId="tooltipSync">
+            <Tooltip />
+          </Wrapper>
+        </>,
+      );
+      const wrappers = container.querySelectorAll('.recharts-wrapper');
+      const wrapperTwo = wrappers[1];
+
+      // target the first chart to show the tooltip
+      showTooltip(container, mouseHoverSelector, debug);
```
PavelVanecek, line 291: How do we know this targets the first chart? Should it instead do `showTooltip[wrappers[0], ...`?

test/component/Tooltip/Tooltip.sync.spec.tsx
```diff
@@ -0,0 +1,382 @@
...
+
+      // target the second chart to see if it has the synchonized tooltip showing
+      const tooltip = getTooltip(wrapperTwo);
+      expect(tooltip).toBeVisible();
+
+      const tooltipContentName = wrapperTwo.querySelector('.recharts-tooltip-item-name');
+      const tooltipContentValue = wrapperTwo.querySelector('.recharts-tooltip-item-value');
+      expect(tooltipContentName).not.toBeNull();
+      expect(tooltipContentValue).not.toBeNull();
+      expect(tooltipContentName).toBeInTheDocument();
```
PavelVanecek, line 301: Can we also `expect`:

- the tooltip content (name + value)
- same for the other tooltip? both of them should display, right? and each with different content depending on their respective `dataKey`s

PavelVanecek (approved):
Nice, a great start. The tooltip sync support is worse than I expected. Something with potential to improve in 3.2 or somewhere.

I have some comments but let's follow up on those in the next PR, and merge this!

test/component/Tooltip/Tooltip.sync.spec.tsx
ckifer: Yeah probably

test/component/Tooltip/Tooltip.sync.spec.tsx
ckifer: Sure. And probably different content given different data, but the same indices

Issue #4549, opened 2024-05-19
Move Tooltip state to Redux


PavelVanecek:
Tooltip payloads and active and everything relies on direct DOM access and blocks #2788. Refactor to use Redux instead.

PR #4560, opened 2024-05-21
Add couple more selectors

PavelVanecek:
## Description

Moving things to redux store

## Related Issue

## Motivation and Context

I want to move the `getMouseInfo` function to Redux and there's some more information I need there, so this move.

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/RechartsWrapper.tsx
+++ b/src/chart/RechartsWrapper.tsx
@@ -1,5 +1,7 @@
 import React, { CSSProperties, DOMAttributes, forwardRef, ReactNode, Ref } from 'react';
 import clsx from 'clsx';
+import { setContainer } from '../state/layoutSlice';
+import { useAppDispatch } from '../state/hooks';
 
 export type RechartsWrapperProps = {
   children: ReactNode;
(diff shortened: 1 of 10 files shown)


test/context/chartLayoutContext.spec.tsx
```diff
@@ -1325,3 +1327,87 @@ describe('ChartLayoutContextProvider', () => {
...
+    };
+    render(
+      <ComposedChart width={100} height={200} margin={{ top: 10, right: 20, bottom: 30, left: 40 }}>
+        <Customized component={Comp} />
+        <Brush height={13} />
+      </ComposedChart>,
+    );
+  });
+
+  it.todo('should include width of YAxis');
```
PavelVanecek, line 1400: We will need these if we want to move offset calculation to redux completely

PavelVanecek:
Oh no. I will fix the type error tomorrow.

src/context/chartLayoutContext.tsx
```diff
@@ -310,8 +308,10 @@ export const useViewBox = (): CartesianViewBox => {
   return useContext(ViewBoxContext);
 };
 
+const manyComponentsThrowErrorsIfOffsetIsUndefined: ChartOffset = {};
```
ckifer, line 311: lol

src/state/layoutSlice.ts
```diff
@@ -16,9 +28,15 @@ const chartLayoutSlice = createSlice({
     setLayout(state, action: PayloadAction<LayoutType>) {
       state.layoutType = action.payload;
     },
+    setContainer(state, action: PayloadAction<HasDomRect | null>) {
+      state.container = castDraft(action.payload);
```
ckifer, line 32: You dispatch an HTML element but only type one function?

src/chart/RechartsWrapper.tsx
```diff
@@ -12,18 +14,24 @@ export type RechartsWrapperProps = {
...
-      {...wrapperEvents}
-      ref={ref}
-    >
-      {children}
-    </div>
-  ),
+  ({ children, width, height, className, style, wrapperEvents }: RechartsWrapperProps, ref: Ref<HTMLDivElement>) => {
+    const dispatch = useAppDispatch();
+    const innerRef = (node: HTMLDivElement | null) => {
+      dispatch(setContainer(node));
```
ckifer, line 20: Hmm, we probably shouldn't be dispatching an element into the store? But all good

ckifer (approved):

src/state/layoutSlice.ts
ckifer: I guess that's to prevent us from doing stupid stuff. But the other properties will still be there

PavelVanecek: The interface is there to make mocks and unit testing easier, and to show which properties of the element is recharts using.

src/chart/RechartsWrapper.tsx
PavelVanecek: Correct, Redux will consider this an antipattern. I tried to only dispatch the domRect, and sizes (which is the correct Redux way) but then the tooltip will not react to resize events. We can solve that with Resize observer... Later.

ckifer: Ack, makes sense

PR #4542, opened 2024-05-17
2.x: Support react 19

eps1lon:
Best reviewed without whitespace changes

## Description

`defaultProps` are no longer supported on function components. To simplify migration, the components that do rely on the original defaultProps behavior were migrated to class components.

Libraries also need to manually apply `defaultProps` if they expect that these are resolved by the JSX runtime. In React 19, defaultProps are only resolved by the time the component instances may access props not when a parent renders or reads from the JSX e.g. via `children.props`.

## Motivation and Context

Adds support for React 19 to `recharts@2.x` without having to patch published files. 

Complete React 19 support would require https://github.com/recharts/recharts/pull/4541 but that's a breaking change. Consumers of 2.x will have to use `resolutions` or `overrides` with their package managers to ensure the correct `react-is` version is used

## How Has This Been Tested?

We've been using `recharts` in various (internal) surfaces at Vercel. We've applied this patch to our codebase and charts have been working when they were previously broken.

The following components we use explicitly:
* `ResponsiveContainer`
* `LineChart`
* `CartesianGrid`
* `XAxis`
* `Tooltip`
* `YAxis`
* `Legend`
* `Line`
* `PieChart`
* `Cell`
* `Pie`
* `Sector`

We also use [`SparkAreaChart` from `@tremor/react`](https://github.com/tremorlabs/tremor/blob/0a6c9a08381feae148286ab2075497dc347ce587/src/components/spark-elements/SparkAreaChart/SparkAreaChart.tsx) which uses `recharts` to implement that chart.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] I have added a storybook story or extended an existing story to show my changes

PR #4561, opened 2024-05-21
React 19 compatible usage of defaultProps

eps1lon:
Forward port of https://github.com/recharts/recharts/pull/4542 for 3.x


src/cartesian/ErrorBar.tsx
```diff
@@ -52,23 +52,27 @@ interface ErrorBarProps extends InternalErrorBarProps {
...
 
+ErrorBar.defaultProps = {
+  stroke: 'black',
+};
 export function ErrorBar(props: Props) {
   const {
-    offset,
-    layout,
-    width,
+    offset = 0,
```
ckifer, line 60: this causes visual regression https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=664cfc9eadd585053f5b885c

Since props are read from the React component before the component is actually called, these won't be set correctly

ckifer (changes requested):
This will have to be reverted to a class as well I'm afraid

src/cartesian/ReferenceDot.tsx
```diff
@@ -88,27 +88,36 @@ export function ReferenceDot(props: Props) {
...
-
-  if (React.isValidElement(option)) {
-    dot = React.cloneElement(option, props);
-  } else if (isFunction(option)) {
-    dot = option(props);
-  } else {
-    dot = <Dot {...props} cx={props.cx} cy={props.cy} className="recharts-reference-dot-dot" />;
-  }
+// eslint-disable-next-line react/prefer-stateless-function -- requires static defaultProps
+export class ReferenceDot extends React.Component<Props> {
```
ckifer, line 92: looks like this needs moved up

ckifer (changes requested):
There were some type errors as well before the current lint error, try running https://github.com/recharts/recharts/blob/3.x/package.json#L30 and see if type checking passes

ckifer:
Going to close this since we are going try to get rid of this in 3.x before releasing

PR #4566, opened 2024-05-22
Redux middleware that takes mouse events and turns them into data coordinates

PavelVanecek:
## Description

This is basically exactly the same thing that `generateCategoricalChart` is doing except it sources its data from Redux store instead of scouting DOM elements.

## Related Issue

## Motivation and Context

Element cloning

## How Has This Been Tested?

npm test, storybooks

## Types of changes

## Checklist:


src/state/mouseEventsMiddleware.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const mouseMoveAction = createAction<MousePointer>('mouseMove');
+
+export const mouseMoveMiddleware = createListenerMiddleware();
+
+mouseMoveMiddleware.startListening({
+  actionCreator: mouseMoveAction,
+  effect: (action: PayloadAction<MousePointer>, listenerApi: ListenerEffectAPI<RechartsRootState, AppDispatch>) => {
+    const mousePointer = action.payload;
+    const activeIndex: number | undefined = selectActiveIndexFromMousePointer(listenerApi.getState(), mousePointer);
```
PavelVanecek, line 30: I was wondering how expensive it is to run this chonk of a selector on every mouse move, and turns out that ... it's less than 1 millisecond on my 6 years old laptop.

There is definitely an opportunity to optimize here but this naive version works just fine.

Screenshot 2024-05-22 at 9 07 13 PM

Screenshot 2024-05-22 at 9 08 14 PM

PR #4573, opened 2024-05-24
2.x: Fix issues with React 19

eps1lon:
https://github.com/recharts/recharts/pull/4542 was incomplete with regards to `defaultProps`. This PR ensures the test suite is green in React 19 as well as well as fixing all new warnings introduced in React 19.


storybook/preview.ts
```diff
@@ -2,11 +2,10 @@ import { Preview } from '@storybook/react';
 import './global.css';
 import { DocsPage, DocsContainer } from '@storybook/addon-docs';
 // eslint-disable-next-line import/no-extraneous-dependencies
-import { withPerformance } from 'storybook-addon-performance';
+// import { withPerformance } from 'storybook-addon-performance';
```
PavelVanecek, line 5: Doesn't work with React19?

test/cartesian/Bar.spec.tsx
```diff
@@ -278,7 +278,9 @@ describe.each(chartsThatSupportBar)('<Bar /> as a child of $testName', ({ ChartE
             x: expect.any(Number),
             y: 50,
           },
-          {}, // this object arrives as a second argument, I am not sure where that comes from
+          // React always calls function components with undefined as the second argument
```
PavelVanecek, line 281: 💡 thanks for clarifying this! was a mystery to me

test/cartesian/ReferenceDot.spec.tsx
```diff
@@ -17,7 +17,7 @@ describe('<ReferenceDot />', () => {
     { name: '201112', uv: 4.3, pv: 0 },
   ];
 
-  test('Render 1 dot and 1 label in ReferenceDot', () => {
+  test.only('Render 1 dot and 1 label in ReferenceDot', () => {
```
PavelVanecek, line 20: Same here, remove .only pls

src/cartesian/Area.tsx
```diff
@@ -256,7 +256,8 @@ export class Area extends PureComponent<Props, State> {
       dotItem = option(props);
     } else {
       const className = clsx('recharts-area-dot', typeof option !== 'boolean' ? option.className : '');
-      dotItem = <Dot {...props} className={className} />;
+      const { key, ...rest } = props;
+      dotItem = <Dot key={key} {...rest} className={className} />;
```
eps1lon, line 260: Spreading `key` is deprecated. `<div {...props} key={key} />` would be an alternative to avoid the warning but does deopt to `createElement` instead of the new, fast JSX runtime.

PavelVanecek:
Not sure why but I had to approve the CI run like four times for this PR. Perhaps if you do a regular push instead of force pushes then it will allow you to run without me pushing buttons.

And thanks for looking into this!

eps1lon:
> Not sure why but I had to approve the CI run like four times for this PR. Perhaps if you do a regular push instead of force pushes then it will allow you to run without me pushing buttons.

GitHub wants approval for every single push. You could change in the repo settings that only new users on GitHub need approval not new contributors.

test/cartesian/ReferenceDot.spec.tsx
eps1lon: Wasn't ready for review.

storybook/preview.ts
eps1lon: Based off of @ckifer branch. I don't know why he removed that. Though the commit was marked as "revert later" so I don't think that's relevant. I think the idea was just to force React 19, check that unit tests pass, revert React 19 and merge.

ckifer: Exactly. I had to upgrade master to storybook 8 to get things to behave. Idk why this works in our 3.x branch but not here, maybe I just didn't try hard enough. But either way it doesn't matter it was just to get R19 force installed so I could run the tests

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   build-and-test:
     name: Build and Test on ${{ matrix.node-version }}
     runs-on: ubuntu-latest
-    
+
     strategy:
       matrix:
         node-version: [16.x, 18.x, 20.x]
```
ckifer, line 15: CI failing because of node16, but I can fix that real quick

ckifer (approved):
really appreciate all of the work here @eps1lon 
I wouldn't have been able to get to it in a reasonable amount of time at all.

The branches are so diverged now that I'll just merge all of this including the storybook upgrade, not too worried about it as long as it deploys.

I will merge this, release another alpha version, wait for a little more testing, and then add R19 to peerDeps when the stable version is released.

ckifer:
https://www.npmjs.com/package/recharts/v/2.13.0-alpha.1

Issue #4583, opened 2024-05-26
Move Axis map to Redux


PavelVanecek:
`generateCategoricalChart` -> `getAxisMap` depends on a direct DOM access to read Axis configuration. Instead make each axis element to push whatever config is needed to Redux store and read it from there.

PR #4606, opened 2024-05-30
Axis domain by data key selector

PavelVanecek:
## Description

There is a ton to unwrap here in the axis area but it's looking possible.

## Related Issue

## Motivation and Context

Remove direct DOM access

## How Has This Been Tested?

npm test

the new hooks and selectors are not yet used anywhere outside of unit tests so I am not running storybook diff

## Types of changes

## Checklist:

### DIFF
--- a/src/state/axisMapSlice.ts
+++ b/src/state/axisMapSlice.ts
@@ -1,5 +1,5 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
-import { AxisDomain, AxisType, ScaleType } from '../util/types';
+import { AxisDomain, AxisDomainType, AxisType, DataKey, ScaleType } from '../util/types';
 import { RechartsScale } from '../util/ChartUtils';
 
 export type AxisId = string | number;
@@ -13,10 +13,18 @@ export type XAxisPadding = { left?: number; right?: number } | 'gap' | 'no-gap';
 export type AxisSettings = {
   id: AxisId;
   scale: ScaleType | RechartsScale | undefined;
-  type: 'number' | 'category';
+  type: AxisDomainType;
   padding: XAxisPadding;
   domain: AxisDomain | undefined;
   allowDataOverflow: boolean;
+  /**
+   * The axis functionality is severely restricted without a dataKey
+   * - but there is still something left, and the prop is optional
+   * so this can also be undefined even in real charts.
+   * There are no defaults.
+   */
+  dataKey: DataKey<any> | undefined;
+  allowDuplicatedCategory: boolean;
 };
 
 // TODO add the remaining properties here
--- a/src/state/axisMapSlice.ts
+++ b/src/state/axisMapSlice.ts
@@ -1,8 +1,9 @@
-import { PayloadAction, createSlice } from '@reduxjs/toolkit';
-import { AxisType, ScaleType } from '../util/types';
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { AxisDomain, AxisType, ScaleType } from '../util/types';
 import { RechartsScale } from '../util/ChartUtils';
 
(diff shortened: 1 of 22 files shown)


src/state/axisMapSlice.ts
```diff
@@ -12,7 +13,18 @@ export type AxisId = string | number;
...
+  padding: XAxisPadding;
+  domain: AxisDomain | undefined;
+  allowDataOverflow: boolean;
+  /**
+   * The axis functionality is severely restricted without a dataKey
+   * - but there is still something left, and the prop is optional
+   * so this can also be undefined even in real charts.
+   * There are no defaults.
+   */
+  dataKey: DataKey<any> | undefined;
```
ckifer, line 26: Don't axes grab onto specific dataKeys under certain conditions? Like if you specific on a graphical item then Y axis is defaulted. Vice versa for X axis if layout is swapped

src/state/axisSelectors.ts
```diff
@@ -16,6 +17,62 @@ const unknownScale: ParsedScaleReturn = {
...
+      return undefined;
+    }
+    // Array.prototype.flat has good support everywhere except IE11: https://caniuse.com/?search=flat
+    const allDataSquished = graphicalItemsData
+      .flat(1)
+      .concat(chartData.chartData)
+      .map(entry => getValueByDataKey(entry, axisSettings.dataKey));
+
+    if (axisSettings.type === 'number') {
+      const onlyNumbers = allDataSquished.map(Number).filter(n => Number.isNaN(n) === false);
```
ckifer, line 66: Doesn't isNaN only check for the actual value NaN? I think this filter would include strings, etc?

src/state/graphicalItemsSlice.ts
```diff
@@ -24,9 +34,15 @@ const graphicalItemsSlice = createSlice({
     removeBar(state) {
       state.countOfBars -= 1;
     },
+    addGraphicalItemData(state, action: PayloadAction<ChartData>) {
+      state.graphicalItemData.push(action.payload);
+    },
+    removeGraphicalItemData(state, action: PayloadAction<ChartData>) {
+      state.graphicalItemData = state.graphicalItemData.filter(item => item !== action.payload);
```
ckifer, line 41: Is this comparing an entire array?

test/state/axisSelectors.spec.tsx
```diff
@@ -64,3 +64,249 @@ describe('selectAxisScale', () => {
...
+      const spy = vi.fn();
+      const Comp = (): null => {
+        const result = useAppSelector(state => selectDomainOfDataByKey(state, 'xAxis', '0'));
+        spy(result);
+        return null;
+      };
+      render(
+        <BarChart data={PageData} width={100} height={100}>
+          <XAxis dataKey="uv" type="number" />
+          <Customized component={Comp} />
```
ckifer, line 113: Won't these all fail later? Why can't we use renderHook?

ckifer (approved):
Nice!! Some questions

src/state/axisMapSlice.ts
PavelVanecek: Indeed, graphical items have their own axis maps. That comes next

src/state/graphicalItemsSlice.ts
PavelVanecek: It is, on purpose.

src/state/axisSelectors.ts
PavelVanecek: I will add a test

test/state/axisSelectors.spec.tsx
PavelVanecek: Yes all of these will fail. They will be trivial to replace however.

ckifer (approved):

PR #4616, opened 2024-06-02
Functions for parsing domain from user input and data

PavelVanecek:
## Description

none of this is new - it was just spread over few places. I merged three functions into one and put tests and types on it.

## Related Issue

## Motivation and Context

I will use this function in Redux scale selector.

## How Has This Been Tested?

npm test

image

## Types of changes

## Checklist:

### DIFF
--- a/src/util/isDomainSpecifiedByUser.ts
+++ b/src/util/isDomainSpecifiedByUser.ts
@@ -1,5 +1,7 @@
+import { MAX_VALUE_REG, MIN_VALUE_REG } from './ChartUtils';
 import { isNumber } from './DataUtils';
-import { AxisDomain, AxisDomainType } from './types';
+import { getNiceTickValues } from './scale';
+import { AxisDomain, AxisDomainType, NumberDomain } from './types';
 
 /**
  * Takes a domain and user props to determine whether he provided the domain via props or if we need to calculate it.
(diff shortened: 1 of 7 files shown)


src/util/isDomainSpecifiedByUser.ts
```diff
@@ -28,3 +30,198 @@ export function isDomainSpecifiedByUser(
...
+ *
+ * @return [min, max] domain if it's well-formed; undefined if the domain is invalid
+ */
+export function parseNumericalUserDomain(
+  userDomain: AxisDomain | undefined,
+  dataDomain: NumberDomain | undefined,
+  allowDataOverflow: boolean,
+  allowDecimals: boolean,
+): NumberDomain | undefined {
+  if (!allowDataOverflow && dataDomain == null) {
```
PavelVanecek, line 161: It's not the prettiest function ever but I want to focus on the selectors rather than refactoring this 🙃 it's well covered so it should be easy to rewrite later if we ever want to.

test/util/isDomainSpecifiedByUser.spec.ts
```diff
@@ -85,3 +89,184 @@ describe('isDomainSpecifiedByUser', () => {
     expect(isDomainSpecifiedByUser(undefined as never, true, 'number')).toBe(false);
   });
 });
+
+describe('parsing axis domain provided by user', () => {
+  // These are valid domains without any further questions
+  const validCases: MyTestCases = [
```
PavelVanecek, line 95: Please focus on the test cases when reviewing. Please let me know if I missed some

src/util/isDomainSpecifiedByUser.ts
```diff
@@ -28,3 +30,198 @@ export function isDomainSpecifiedByUser(
...
+      finalMin = providedMin;
+    } else if (typeof providedMin === 'function') {
+      try {
+        finalMin = providedMin(dataDomain?.[0]);
+      } catch (ex) {
+        /* ignore the exception and compute domain from data later */
+      }
+    } else if (typeof providedMin === 'string' && MIN_VALUE_REG.test(providedMin)) {
+      const value = +MIN_VALUE_REG.exec(providedMin)[1];
+      finalMin = dataDomain?.[0] - value;
```
ckifer, line 194: Doesn't the strings that are allowed allow for math? Like more than subtraction you could do "dataMin + 400" and get that as your data min. Or "dataMax - 400"

Your test cases seem to indicate that doesn't work. Maybe we should invariant somewhere in here if people do something wrong like that? I guess we're not trying to change current behavior here..

PavelVanecek: If you look closely on the regular expressions you will find that they allow expanding the domain but not contracting: https://github.com/recharts/recharts/blob/6e9f9fe5b06ab939135a48f4c68e124d89e3f337/src/util/ChartUtils.ts#L1295

I was a bit surprised by that but I decided against changing the regex, and I copy pasted this chunk of code from its original location: https://github.com/recharts/recharts/blob/6e9f9fe5b06ab939135a48f4c68e124d89e3f337/src/util/ChartUtils.ts#L1312

PavelVanecek: I have some test cases for math that expands the domain: https://github.com/PavelVanecek/recharts/blob/9c7a275b59887ac2c649228ef526169d22bade16/test/util/isDomainSpecifiedByUser.spec.ts#L140

And also for math that contracts the domain: https://github.com/PavelVanecek/recharts/blob/9c7a275b59887ac2c649228ef526169d22bade16/test/util/isDomainSpecifiedByUser.spec.ts#L132

ckifer: Cool cool, I figured you didn't change it I just always tried to use it this way myself and totally thought it worked 😂. All good as long as the function versions allow you to do whatever you want

PavelVanecek: With the test suite it should now be much easier to add new behaviour. I'm expecting that one of the 500 open issues is asking for that anyway.

But. Let's release 3.0 first!

test/util/isDomainSpecifiedByUser.spec.ts
```diff
@@ -85,3 +89,186 @@ describe('isDomainSpecifiedByUser', () => {
...
+    { domain: [-100, 100], expected: [-100, 100] },
+    { domain: [() => 1, () => 2], expected: [1, 2] },
```
ckifer, line 97: I think we're missing a case for 

([dataMin, dataMax]) => [dataMin, dataMax]

I see individual functions for each but there's also a version where the whole thing is a function and both get passed to it

PavelVanecek: Here is a test case for dataMin, dataMax: https://github.com/PavelVanecek/recharts/blob/9c7a275b59887ac2c649228ef526169d22bade16/test/util/isDomainSpecifiedByUser.spec.ts#L144

Here is a test case for one function that accepts an array: https://github.com/PavelVanecek/recharts/blob/9c7a275b59887ac2c649228ef526169d22bade16/test/util/isDomainSpecifiedByUser.spec.ts#L147

The tests are split because by using `dataMin` and `dataMax` it forces the selector to iterate all data. This is mentioned in code as a performance optimization: https://github.com/recharts/recharts/blob/6e9f9fe5b06ab939135a48f4c68e124d89e3f337/src/chart/generateCategoricalChart.tsx#L360

Now, I have my doubts on the effectiveness of such optimization (since it has to iterate the data _anyway_, the chart cannot render without iterating it all) but again ... let's keep it as-is until we find a good reason to change it.

ckifer: When domain is a function it gives the array of dataMin dataMax as a parameter to that function. Will send an example in a sec

PavelVanecek: Does this case cover it?

ckifer: Kind of but not really since `d` isn't data but rather an array of `[dataMin, dataMax]` so you could just make that case `([dMin]) => [dMin, 999]`

and then would want another case for 
`([dMin, dMax]) => [dMin, dMax + 100]` or something like that

PavelVanecek: I can add that

ckifer: Sweet looks good thanks

ckifer (approved):
LGTM! Crazy we couldn't really test this super important stuff in isolation before 😅

PR #4624, opened 2024-06-04
Add data selectors, use them in the domain calculator

PavelVanecek:
## Description

Getting closer to replacing XAxis data with redux store

## Related Issue

## Motivation and Context

No DOM access

## How Has This Been Tested?

npm test

## Types of changes

## Checklist:

### DIFF
--- a/src/state/dataSelectors.ts
+++ b/src/state/dataSelectors.ts
@@ -1,0 +1,44 @@
+import { createSelector } from '@reduxjs/toolkit';
+import { RechartsRootState } from './store';
+import { ChartData, ChartDataState } from './chartDataSlice';
+import { DataKey } from '../util/types';
+import { getValueByDataKey } from '../util/ChartUtils';
+
+/**
+ * This is a "cheap" selector - it returns the data but doesn't iterate them, so it is not sensitive on the array length.
+ * @param state RechartsRootState
+ * @returns data defined on the chart root element, such as BarChart or ScatterChart
+ */
+export const selectChartDataWithIndexes = (state: RechartsRootState): ChartDataState => state.chartData;
+
+/**
+ * This is a "cheap" selector - it returns the data but doesn't iterate them, so it is not sensitive on the array length.
+ * @param state RechartsRootState
+ * @returns data defined on the chart graphical items, such as Line or Scatter or Pie
+ */
+export const selectAllGraphicalItemsData: (state: RechartsRootState) => ReadonlyArray<ChartData> = (
+  state: RechartsRootState,
+) => state.graphicalItems.graphicalItemData;
+
+/**
+ * This selector will return all data there is in the chart: graphical items, chart root, all together
+ * and all processed by a single DataKey.
+ * Useful for figuring out an axis domain (because that needs to know of everything),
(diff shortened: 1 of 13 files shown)


src/state/dataSelectors.ts
```diff
@@ -0,0 +1,44 @@
+import { createSelector } from '@reduxjs/toolkit';
+import { RechartsRootState } from './store';
+import { ChartData, ChartDataState } from './chartDataSlice';
+import { DataKey } from '../util/types';
+import { getValueByDataKey } from '../util/ChartUtils';
+
+/**
+ * This is a "cheap" selector - it returns the data but doesn't iterate them, so it is not sensitive on the array length.
```
ckifer, line 8: what do you mean by 

> it is not sensitive on the array length

This is just a simple selector, it is not ever going to be sensitive to anything?

src/state/dataSelectors.ts
```diff
@@ -0,0 +1,44 @@
...
+  state: RechartsRootState,
+) => state.graphicalItems.graphicalItemData;
+
+/**
+ * This selector will return all data there is in the chart: graphical items, chart root, all together
+ * and all processed by a single DataKey.
+ * Useful for figuring out an axis domain (because that needs to know of everything),
+ * not useful for rendering individual graphical elements (because they need to know which data is theirs and which is not).
+ *
+ * This function will discard the original indexes, so it is also not useful for anything that depends on ordering.
```
ckifer, line 29: ah nvm this helps with the above

PavelVanecek: Some of the selectors are iterating user data and some are not. I'm not yet decided if that's the performance optimisation we need.

src/util/ChartUtils.ts
```diff
@@ -674,6 +675,9 @@ export type AxisPropsNeededForTicksGenerator = {
   isCategorical?: boolean;
   categoricalDomain?: ReadonlyArray<AxisTick>;
   type?: 'number' | 'category';
+  /**
+   * The range appears to be only used in Angle Axis - needs further investigation
```
ckifer, line 679: I would expect all charts to have both a domain and a range, but maybe thats just my secondary school math classes speaking 😅. I know everything is a "domain" in recharts

test/state/axisSelectors.spec.tsx
```diff
@@ -294,43 +302,100 @@ describe('selectDomainOfDataByKey', () => {
...
+        'May',
+        '',
+        'Jun',
+        '',
+        'Jul',
+        '',
+        'Aug',
+        '',
+        // no idea where the string 'undefined' comes from - the new implementation doesn't have it, breaking or not.
+        'undefined',
```
ckifer, line 387: 😅

ckifer (approved):
🚀

src/util/ChartUtils.ts
PavelVanecek: In all categorical charts, this variable is undefined. Perhaps it's computed somewhere else, or it's included in the scale already? Investigation continues

ckifer: Nah I'm just saying from an external perspective, like d3 lets you define a domain and a range. Idt it does anything in recharts besides what you say

PR #4625, opened 2024-06-04
Add more XAxis domain tests

PavelVanecek:
spoiler alert: Recharts ignores this combination of props

### DIFF
--- a/test/cartesian/XAxis.spec.tsx
+++ b/test/cartesian/XAxis.spec.tsx
@@ -69,7 +69,7 @@ describe('<XAxis />', () => {
   it('Render ticks with tickFormatter', () => {
     const { container } = render(
       <LineChart width={400} height={400} data={lineData} margin={{ top: 20, right: 20, bottom: 20, left: 20 }}>
-        <XAxis dataKey="name" tickFormatter={(value, i) => `${i}`} />
+        <XAxis dataKey="name" tickFormatter={(_, i) => `${i}`} />
         <Line type="monotone" dataKey="uv" stroke="#ff7300" />
       </LineChart>,
     );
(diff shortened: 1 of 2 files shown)


test/cartesian/XAxis.spec.tsx
```diff
@@ -373,190 +373,243 @@ describe('<XAxis />', () => {
...
-          <XAxis dataKey="x" type="number" domain={spy} allowDataOverflow />
-        </BarChart>,
-      );
-      expectXAxisTicks(container, ['-500', '-250', '0', '250', '500']);
-      expect(spy).toHaveBeenCalledTimes(1);
-      expect(spy).toHaveBeenCalledWith([100, 170], true);
+      it('should default to dataMin, dataMax ANYWAY when domain is provided as strings? I do not understand this behaviour.', () => {
+        const { container } = render(
+          <Component>
+            {/* omg recharts just completely ignores this when it's defined as strings */}
```
ckifer, line 529: I mean, the domain is supposed to be numbers and it isn't. Would you expect it to parse the strings as numbers or maybe throw an error?

test/cartesian/XAxis.spec.tsx
```diff
@@ -590,20 +643,100 @@ describe('<XAxis />', () => {
...
+      it('should display every third tick with interval = 2', () => {
+        const { container } = render(
+          <BarChart width={300} height={300} data={data}>
+            <XAxis dataKey="x" type="number" interval={2} />
+          </BarChart>,
+        );
+        expectXAxisTicks(container, ['0', '135']);
+      });
+
+      it('should add more ticks with tickCount and then reduce them again with interval', () => {
```
ckifer, line 675: Interval can also be certain string values if we want tests for those I.e "preserveStart", etc.

ckifer (approved):
Such awesome tests 🚀

test/cartesian/XAxis.spec.tsx
PavelVanecek: It parses strings elsewhere

test/cartesian/XAxis.spec.tsx
PavelVanecek: I couldn't find what difference it makes. There's an example in docs where three out of four axes are exactly the same. Can you think of good values that will make a nice test?

ckifer: `preserveEnd` always preserves the last tick and then tries to compute ticks that fit from there
`preserveStart` always preserves the first tick and then tries to compute ticks that fit from there
`preserveStartEnd` tries to do both and create ticks that fit in the middle
`equidistantPreserveStart` tries to do the same as `preserveStart` but it doesn't insert random ticks on one side but not the other

https://codesandbox.io/p/sandbox/line-chart-of-different-axis-intervals-d2kjmp?file=%2Fsrc%2FApp.tsx%3A96%2C27

Play with the width of that and you'll see the differences

image

test/cartesian/XAxis.spec.tsx
ckifer: true that, I think D3 is what parses strings though? At least string data..

ckifer: Yep. Although I would always always steer users away from strings if I could either way

PR #4627, opened 2024-06-05
Add tests for interval string constants

PavelVanecek:


ckifer (approved):
Equidistant isn't the same thing but similar. All good though

PR #4631, opened 2024-06-06
Move Tooltip state redux

ckifer:
## Description

In order to move synchronized to redux I need to move tooltip state to redux as opposed to context.

Everything renders as normal here... I will move Sankey as well before marking this ready. Sunburst is way behind so I'm not going to touch it.

## Related Issue

## Motivation and Context

Be able to access tt state via selectors

## How Has This Been Tested?

npm test

## Types of changes

- [X] Refactor

## Checklist:

### DIFF
--- a/src/chart/Treemap.tsx
+++ b/src/chart/Treemap.tsx
@@ -18,7 +18,7 @@ import { Global } from '../util/Global';
 import { findChildByType, validateWidthHeight, filterProps } from '../util/ReactUtils';
 import { AnimationDuration, AnimationTiming, DataKey } from '../util/types';
 import { ViewBoxContext } from '../context/chartLayoutContext';
-import { TooltipContextProvider, TooltipContextValue } from '../context/tooltipContext';
+import { SetTooltipInfo, TooltipContextValue } from '../context/tooltipContext';
 import { CursorPortalContext, TooltipPortalContext } from '../context/tooltipPortalContext';
 import { RechartsWrapper } from './RechartsWrapper';
 import {
@@ -868,6 +868,7 @@ export class Treemap extends PureComponent<Props, State> {
     const { width, height, className, style, children, type, ...others } = this.props;
     const attrs = filterProps(others, false);
     const viewBox = { x: 0, y: 0, width, height };
+    const ttContext = this.getTooltipContext();
 
     return (
       <RechartsStoreProvider preloadedState={{ options }} reduxStoreName={this.props.className ?? 'Treemap'}>
@@ -878,33 +879,36 @@ export class Treemap extends PureComponent<Props, State> {
               args={{ props: this.props, currentRoot: this.state.currentRoot }}
             />
             <ViewBoxContext.Provider value={viewBox}>
-              <TooltipContextProvider value={this.getTooltipContext()}>
-                <RechartsWrapper
-                  className={className}
-                  style={style}
-                  width={width}
-                  height={height}
-                  ref={(node: HTMLDivElement) => {
-                    if (this.state.tooltipPortal == null) {
-                      this.setState({ tooltipPortal: node });
-                    }
-                  }}
-                >
-                  <Surface {...attrs} width={width} height={type === 'nest' ? height - 30 : height}>
-                    <g
-                      className="recharts-cursor-portal"
-                      ref={(node: SVGElement) => {
-                        if (this.state.cursorPortal == null) {
-                          this.setState({ cursorPortal: node });
-                        }
-                      }}
-                    />
-                    {this.renderAllNodes()}
-                    {children}
-                  </Surface>
-                  {type === 'nest' && this.renderNestIndex()}
-                </RechartsWrapper>
-              </TooltipContextProvider>
+              <SetTooltipInfo
+                active={ttContext.active}
+                activeCoordinate={ttContext.coordinate}
+                activeLabel={ttContext.label}
+              />
+              <RechartsWrapper
+                className={className}
+                style={style}
+                width={width}
+                height={height}
+                ref={(node: HTMLDivElement) => {
+                  if (this.state.tooltipPortal == null) {
+                    this.setState({ tooltipPortal: node });
+                  }
+                }}
+              >
+                <Surface {...attrs} width={width} height={type === 'nest' ? height - 30 : height}>
+                  <g
+                    className="recharts-cursor-portal"
+                    ref={(node: SVGElement) => {
+                      if (this.state.cursorPortal == null) {
+                        this.setState({ cursorPortal: node });
+                      }
+                    }}
+                  />
+                  {this.renderAllNodes()}
+                  {children}
+                </Surface>
+                {type === 'nest' && this.renderNestIndex()}
+              </RechartsWrapper>
             </ViewBoxContext.Provider>
           </TooltipPortalContext.Provider>
         </CursorPortalContext.Provider>
--- a/src/component/Cursor.tsx
+++ b/src/component/Cursor.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement, cloneElement, createElement, isValidElement, SVGProps } from 'react';
 import clsx from 'clsx';
+import { useAppSelector } from '../state/hooks';
 import { ChartCoordinate, ChartOffset, LayoutType, TooltipEventType } from '../util/types';
 import { Curve } from '../shape/Curve';
 import { Cross } from '../shape/Cross';
@@ -98,14 +99,17 @@ export function CursorInternal(props: CursorConnectedProps) {
  */
 export function Cursor(props: CursorProps) {
   const tooltipAxisBandSize = useTooltipAxisBandSize();
-  const { coordinate, payload, index } = useTooltipContext();
+  // TODO: move index to redux. Not sure about payload...?
+  const { payload, index, coordinate: coordFromContext } = useTooltipContext();
+  const tooltipPropsFromRedux = useAppSelector(state => state.tooltip);
+  console.log(tooltipPropsFromRedux);
   const offset = useOffset();
   const layout = useChartLayout();
   const chartName = useChartName();
   return (
     <CursorInternal
       {...props}
-      coordinate={coordinate}
+      coordinate={tooltipPropsFromRedux?.activeCoordinate ?? coordFromContext}
       index={index}
       payload={payload}
       offset={offset}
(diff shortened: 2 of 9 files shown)


src/context/chartLayoutContext.tsx
```diff
@@ -88,6 +89,13 @@ export const ChartLayoutContextProvider = (props: ChartLayoutContextProviderProp
   };
 
   const dispatch = useAppDispatch();
+  /**
+   * This throws an error no matter where I put it (here, useEffect here, separate component)
+   * How is this fixed? Eventually we'll get rud of the chart context provider altogether but...
+   *
+   * Warning: Cannot update a component (`TooltipInternal`) while rendering a different component (`ChartLayoutContextProvider`).
```
ckifer, line 96: @PavelVanecek any thoughts about this? I need to set this state, but can't without conflicting with other things being set..

src/state/tooltipSlice.ts
```diff
@@ -131,9 +131,16 @@ export type TooltipState = {
    * may render multiple tooltip payloads.
    */
   tooltipItemPayloads: ReadonlyArray<TooltipPayloadConfiguration>;
+  // whether or not the Tooltip is active (from Redux' perspective)
+  active: boolean;
```
PavelVanecek, line 135: I am afraid this information (and all the other activeX) needs to be more granular. Tooltip can be activated based on clicks or hovers, and the trigger is defined on the Tooltip element. Axis and item "tooltipEventType" also have different activation rules.

And if you look closely, there is already activeIndex and `activeClick` + `activeHover`, four of them, in the objects above: 

```
  axisInteraction: {
    activeClick: boolean;
    activeHover: boolean;
```

and there's a Redux selector selecting that: `selectIsTooltipActive`. Can we use that instead of adding a new property?

PavelVanecek: If we have this information in state split down for interactions, we unlock two things:

1) Graphical elements can send events without knowing what is the Tooltip trigger, and
2) ability to have multiple tooltips with different behaviour.

src/chart/Treemap.tsx
```diff
@@ -868,6 +868,7 @@ export class Treemap extends PureComponent<Props, State> {
     const { width, height, className, style, children, type, ...others } = this.props;
     const attrs = filterProps(others, false);
     const viewBox = { x: 0, y: 0, width, height };
+    const ttContext = this.getTooltipContext();
```
PavelVanecek, line 871: Treemap is one of the easier ones (only allows interaction with item, has no axes), I encourage you to pick up Line or RadialBar first. Once we have those covered we will know that solution will also work with Treemap. But not the other way around.

src/context/chartLayoutContext.tsx
PavelVanecek: We already have `selectIsTooltipActive`, once we have corresponding `selectTooltipCoordinate` then we can delete this dispatch completely.

Until then this can stay as a warning - it's notifying us of discarded re-renders. Not perfect but acceptable for what is effectively a feature branch.

I am not yet sold on if we need `activeLabel` at all.

src/state/tooltipSlice.ts
ckifer: lol my eyes literally glazed over all of the existing state in this slice 😅

src/chart/Treemap.tsx
ckifer: This PR swaps to redux for all categorical charts and treemap.

Or at least it swaps what is currently being set from the generator in context set in redux. Not necessarily end state

src/state/tooltipSlice.ts
ckifer: > ability to have multiple tooltips with different behaviour

At the moment though we could only have two though yeah? One with click behavior one with hover behavior? Unless we have a tooltipId or something

PavelVanecek: In master branch, the renderMap has a "once" property which makes it lookup and render the first Tooltip and ignore others. So, one max.

ckifer: I mean in the current state of 3.x. With `activeClick...` and `activeMouseOver...` we could only have one of each of those two types of Tooltips. Else we would need to store more info

PavelVanecek: Ah right. There can be as many as we want. Maybe one for each coordinate. Or one tooltip for each tick on a Line idk.

src/component/Cursor.tsx
```diff
@@ -98,16 +101,17 @@ export function CursorInternal(props: CursorConnectedProps) {
...
   const layout = useChartLayout();
   const chartName = useChartName();
   return (
     <CursorInternal
       {...props}
-      coordinate={coordinate}
+      coordinate={props.coordinate}
       index={index}
-      payload={payload}
+      payload={props.payload}
```
ckifer, line 114: @PavelVanecek was there a reason these came from from context when they can just be passed in?

PavelVanecek: This was a refactor in progress. Cursor was first and Tooltip second. This change should be safe.

src/state/mouseEventsMiddleware.ts
```diff
@@ -27,9 +28,11 @@ mouseMoveMiddleware.startListening({
...
   effect: (action: PayloadAction<MousePointer>, listenerApi: ListenerEffectAPI<RechartsRootState, AppDispatch>) => {
     const mousePointer = action.payload;
-    const activeIndex: TooltipIndex = selectActiveIndexFromMousePointer(listenerApi.getState(), mousePointer);
-    if (activeIndex != null) {
-      listenerApi.dispatch(setMouseOverAxisIndex({ activeIndex, activeDataKey: undefined }));
+    const activeProps = selectActivePropsFromMousePointer(listenerApi.getState(), mousePointer);
+    if (activeProps?.activeIndex != null) {
+      listenerApi.dispatch(setMouseOverAxisIndex({ activeIndex: activeProps.activeIndex, activeDataKey: undefined }));
+    } else {
+      listenerApi.dispatch(mouseLeaveChart());
```
ckifer, line 35: this is needed to clear tooltip state when mouse moves out of the `inRange` function, but not out of the svg

src/state/selectors.ts
```diff
@@ -105,10 +115,32 @@ export function selectActiveIndex(
...
+    activeCoordinate = tooltipState.axisInteraction.activeClickCoordinate;
+  }
+  return activeCoordinate;
+}
+
+export const selectActiveLabel = createSelector(
   selectTooltipTicks,
   selectActiveIndex,
   (tooltipTicks: ReadonlyArray<TickItem>, activeIndex: TooltipIndex): string | undefined => {
+    console.log(activeIndex);
```
ckifer, line 143: oops

src/state/tooltipSlice.ts
```diff
@@ -168,38 +187,64 @@ const tooltipSlice = createSlice({
...
-      action: PayloadAction<{ activeIndex: TooltipIndex; activeDataKey: DataKey<any> | undefined }>,
+      action: PayloadAction<{
+        activeIndex: TooltipIndex;
+        activeDataKey: DataKey<any> | undefined;
+        activeMouseOverCoordinate?: ChartCoordinate;
+      }>,
     ) {
       state.axisInteraction.activeHover = true;
       state.axisInteraction.activeMouseOverAxisIndex = action.payload.activeIndex;
-      state.axisInteraction.activeMouseOverAxisDataKey = action.payload.activeDataKey;
```
ckifer, line 194: also oops

ckifer:
@PavelVanecek alright this is more the direction I think. Will fix/add tests tomorrow, but feel free to review in the meantime. Everything working as it should in the storybook

https://github.com/recharts/recharts/assets/25180830/51c28dff-d589-404c-803c-411df875d9bb

src/state/mouseEventsMiddleware.ts
PavelVanecek: This should be a comment in code

src/state/selectors.ts
```diff
@@ -241,18 +272,28 @@ export const selectIsTooltipActive = (
   state: RechartsRootState,
   tooltipEventType: TooltipEventType,
   trigger: TooltipTrigger,
+  defaultIndex: number | undefined,
 ): boolean => {
-  if (tooltipEventType === 'axis') {
+  const isTooltipActive = createSelector(selectActiveCoordinate, (coordinate: ChartCoordinate) => {
```
PavelVanecek, line 277: I believe the createSelector call should be at the root level so that it is evaluated once, instead of on every run.

Now how to validate that belief. I suppose we should be able to write a unit test to demonstrate that? There is already a test that a selector is stable, that might show the difference?

PavelVanecek (commented):
Yep I like the approach. I was imagining that the activeCoordinate will be more difficult to port, not sure why. I see some failing tests, I wonder how difficult is it going to be to fix it.

src/state/selectors.ts
ckifer: I need the values passed in initially but I'm not sure how to do that when calling createSelector at the root level since it only returns the results of the previous selectors and not any input you pass. I could return the input in the previous selector but that ruins the selectors purpose

ckifer:
Welp lets try this again. Pesky `shared` prop...

src/state/selectors.ts
PavelVanecek: You can have what redux docs call "input selector", that is a function that ignores state and only returns the user input. That way you get the initial arguments in the combiner too. Check out how we're using it here: https://github.com/recharts/recharts/blob/3.x/src/state/dataSelectors.ts#L27

Alternatively, if you prefer, you can skip the `createSelector` part, and call selectors directly. `const isTooltipActive = selectActiveCoordinate(state, ...`. But beware that this way we lose the memoization that `createSelector` comes with.

ckifer: Updated to do the first way, thanks for the explanation

ckifer:
@PavelVanecek thoughts on this? Need to merge it to continue without making this huge. But can wait if you're about to make a ton of changes with the axis stuff

test/component/Tooltip/Tooltip.visibility.spec.tsx
```diff
@@ -723,12 +723,12 @@ describe('Tooltip visibility', () => {
     expect(container.querySelector('.recharts-tooltip-cursor')).toBeVisible();
 
     // Data should be displayed in the Tooltip payload
-    expect(tooltip.textContent).toBe('100stature : 100cmweight : 200kg');
+    expect(tooltip.textContent).toBe('stature : 100cmweight : 200kg');
```
PavelVanecek, line 726: The title disappeared? Why?

PavelVanecek (approved):
lgtm except that one sus missing Tooltip title

PavelVanecek:
@ckifer merge this when you're ready, it will take me couple more days to get my branch ready, don't wait for it.

test/component/Tooltip/Tooltip.visibility.spec.tsx
ckifer: The title is not there in 2.x. Same with the storybook diff removing what is essentially the index from the tooltip title on the shared bar and radial charts. Scatter never had a title

PR #4646, opened 2024-06-11
selectAxisRange selector plus dependencies

PavelVanecek:
## Description

Okay this took me a few days to figure out but I'm getting somewhere. When I attempt to switch XAxis to use Redux scale, only the vertical chart fails to render - everything else is exactly the same as in master branch. I will do the vertical chart next.

Some of the code here is XAxis only. YAxis will come after.

## Related Issue


PavelVanecek:
Oh no, toSorted is only supported in node20+. I'll revert that part.

src/state/axisSelectors.ts
```diff
@@ -94,27 +176,192 @@ export const selectAxisDomain = (
     return selectNumericalDomain(state, axisSettings, axisType, axisId);
   }
 
-  const allDataSquished = selectAllDataSquished(state, axisType, axisId, axisSettings.dataKey);
+  const allDataSquished = selectAllDataSquished(state, axisType, axisId);
   return computeCategoricalDomain(allDataSquished, axisSettings);
 };
 
+/**
```
ckifer, line 183: man where did all this stuff exist before 😅

src/util/ChartUtils.ts
```diff
@@ -667,7 +667,7 @@ export interface RechartsScale {
 
 export type AxisPropsNeededForTicksGenerator = {
   duplicateDomain?: ReadonlyArray<AxisTick>;
-  realScaleType?: 'scaleBand' | 'band' | 'point' | 'linear';
+  realScaleType?: 'scaleBand' | string;
```
ckifer, line 670: ?

src/state/axisSelectors.ts
```diff
@@ -1,55 +1,136 @@
 import { createSelector } from '@reduxjs/toolkit';
-import { selectChartLayout } from '../context/chartLayoutContext';
-import { ParsedScaleReturn, parseScale } from '../util/ChartUtils';
-import { AxisDomain, AxisType, CategoricalDomain, LayoutType, NumberDomain } from '../util/types';
-import { AxisId, AxisSettings } from './axisMapSlice';
-import { selectChartName } from './selectors';
+import range from 'lodash/range';
```
ckifer, line 2: people don't like that we use lodash, but I guess idc since maintaining our own implementation would suck

ckifer (approved):

src/util/ChartUtils.ts
PavelVanecek: This function accepts all scales, not sure why was it limited to only these four before. All usages were hidden behind `any`.

src/state/axisSelectors.ts
PavelVanecek: I copied from the previous place, the lodash question is not really the topic of the day now. We can later replace it with d3 or our own impl.

src/state/axisSelectors.ts
PavelVanecek: Yeah it's wild. What I discovered so far is:

- scale gets created in `generateCategoricalChart` first .. somewhere
- then `formatAxisMap` takes it and changes a little
- then the ticks generator also takes the scale and changes its domain based on the ticks it generated

I think it's a prime example of why we should wrap the d3 scales in something immutable 😬

ckifer: yeah...

PR #4656, opened 2024-06-13
Consider null as 0 in stacked area chart with connectNulls

seiyab:
## Description
Resolves https://github.com/recharts/recharts/issues/2176
Note that this PR wants to be merged into master branch.

## Related Issue
https://github.com/recharts/recharts/issues/2176

## Motivation and Context

https://github.com/recharts/recharts/issues/2176

## How Has This Been Tested?

Screenshot pasted below.

Screenshot for new story:
Screenshot 2024-06-13 at 16 41 24

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -186,6 +186,7 @@ export class Area extends PureComponent<Props, State> {
     dataKey: Props['dataKey'];
   }) => {
     const { layout } = props;
+    const { connectNulls } = item.props;
     const hasStack = stackedData && stackedData.length;
     const baseValue = Area.getBaseValue(props, item, xAxis, yAxis);
     const isHorizontalLayout = layout === 'horizontal';
@@ -206,7 +207,7 @@ export class Area extends PureComponent<Props, State> {
         }
       }
 
-      const isBreakPoint = value[1] == null || (hasStack && getValueByDataKey(entry, dataKey) == null);
+      const isBreakPoint = value[1] == null || (hasStack && !connectNulls && getValueByDataKey(entry, dataKey) == null);
 
       if (isHorizontalLayout) {
         return {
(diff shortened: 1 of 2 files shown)


ckifer:
This is (even if unintentionally or "fixing" something) is a breaking change. We'll need to do it in the 3.x branch

src/cartesian/Area.tsx
```diff
@@ -206,7 +207,7 @@ export class Area extends PureComponent<Props, State> {
         }
       }
 
-      const isBreakPoint = value[1] == null || (hasStack && getValueByDataKey(entry, dataKey) == null);
+      const isBreakPoint = value[1] == null || (hasStack && !connectNulls && getValueByDataKey(entry, dataKey) == null);
```
ckifer, line 210: can you write a unit test for this too?

ckifer (changes requested):
Change is good on 3.x. Would like a unit test if you could though

seiyab:
Hmm... I'm wondering whether we can merge it into v2. Because I'm still stuck in v1 because of https://github.com/recharts/recharts/issues/2176 😭 . Or is migration from v1 to v3 easy?
I understand it's a breaking change in a sense so I'll change to v3 if you deny it...

seiyab:
Anyway, I'll open a PR for v3, too. And I'll try to add an unit test.

ckifer:
For now, let's try to get it merged in v3 since the change is small. Let me think about it for v2. It helps that it is behind "connectNulls" but still changes behavior if someone uses the prop. 

The upgrade from 2.x to 3.x should be easy. 1.x to 3.x I'm not sure.. it'll probably be harder. I wasn't around during 1.x so harder to estimate

seiyab:
@ckifer 
Could you consider merging it into v2, or releasing v3-alpha?
We are still sticking in v1, which doesn't explicitly support React v17 or later.

ckifer:
Hi, this is merged into 3.x and will be released in 3.0. 3.0 is not ready for an alpha version unfortunately, but it might be so soon. 

It is a breaking change in 2.0 so we can't merge it. I will probably close this PR.

PR #4658, opened 2024-06-13
Add new tests for XAxis domain

PavelVanecek:
## Description

I found why the vertical XAxis story was misbehaving, I added a test and will add fix later.

## Related Issue


ckifer (approved):
Wild

PR #4669, opened 2024-06-17
Domain selector for XAxis in vertical chart

PavelVanecek:
## Description

This was as much fun as getting my teeth pulled but I got the behaviour to be the same as generateCategoricalCharts, but in Redux.

Next, ErrorBars, then stacking, and then we can replace `scale` in XAxis.

## Related Issue

### DIFF
--- a/src/state/axisSelectors.ts
+++ b/src/state/axisSelectors.ts
@@ -11,7 +11,7 @@ import {
   numericalDomainSpecifiedWithoutRequiringData,
   parseNumericalUserDomain,
 } from '../util/isDomainSpecifiedByUser';
-import { ChartData } from './chartDataSlice';
+import { AppliedChartData, ChartData } from './chartDataSlice';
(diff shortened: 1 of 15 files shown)


src/state/axisSelectors.ts
```diff
@@ -105,39 +118,64 @@ export const selectAllDataSquished: (
...
-    } else {
-      finalData = chartData.slice(dataStartIndex, dataEndIndex + 1);
+      return graphicalItemsData;
     }
-    return finalData.map(entry => getValueByDataKey(entry, axisSettings.dataKey));
+    return chartData.slice(dataStartIndex, dataEndIndex + 1);
+  },
+);
+
+export const selectAllAppliedValues = createSelector(
```
ckifer, line 132: nit: maybe want a comment on this one?

ckifer (approved):

Issue #4671, opened 2024-06-18
ClipDot doesn't work in the latest version


anasmi:
## Do you want to request a feature or report a bug?
Bug report

## What is the current behavior?
`clipDot:false` doesn't prevent a dot cut off when allowDataOverflow is set for Y axis (dots on xAxis are cut off)
(updating property value doesn't have any effect). Also, example ins storybook for the feature doesn't seem to work either (there dots are not shown in full on yAxis)
Original fix was introduced here https://github.com/recharts/recharts/commit/91801b1386698be4522ca5077455f3b812c99f95#diff-1dee178d6dc13ab8722aca748856acfb8062f660e9c1fe17edda38f8e5953168
## Code to reproduce : 

```
import React from "react";
import {
  AreaChart,
  Area,
  XAxis,
  YAxis,
  CartesianGrid,
  Tooltip,
  Legend,
  ResponsiveContainer,
  Line,
  LineChart
} from "recharts";

const data = [
  {
    name: "Page A",
    uv: 4000,
    pv: 2400,
    amt: 2400
  },
  {
    name: "Page B",
    uv: 3000,
    pv: 1398,
    amt: 2210
  },
  {
    name: "Page C",
    uv: 2000,
    pv: 9800,
    amt: 2290
  },
  {
    name: "Page D",
    uv: 0,
    pv: 3908,
    amt: 2000
  },
  {
    name: "Page E",
    uv: 1890,
    pv: 0,
    amt: 2181
  },
  {
    name: "Page F",
    uv: 0,
    pv: 3800,
    amt: 2500
  },
  {
    name: "Page G",
    uv: 3490,
    pv: 4300,
    amt: 2100
  }
];

const SimpleLineChart = () => (
  <LineChart
    width={600}
    height={300}
    data={data}
    margin={{
      top: 5,
      right: 30,
      left: 20,
      bottom: 5
    }}
  >
    <CartesianGrid strokeDasharray="3 3" />
    <XAxis dataKey="name" />
    <YAxis domain={[0, 7500]} allowDataOverflow />
    <Tooltip />
    <Legend />
    <Line
      type="monotone"
      dataKey="pv"
      stroke="#8884d8"
      dot={{clipDot:false}}
    />
    <Line type="monotone" dataKey="uv" stroke="#82ca9d" 
      dot={{clipDot:false}}/>
  </LineChart>
);

export default function App() {
  return <SimpleLineChart />;
}

```
[...]

PR #4674, opened 2024-06-18
fix: clipDot setting fails to take effect if set as the only DotProp

ckifer:
## Description

Fix clipDot not working when props are of the form `dot={{ clipDot: false }}` or similar

## Related Issue

https://github.com/recharts/recharts/issues/4671

## Motivation and Context

Fix a bug

## How Has This Been Tested?

See tests, see storybook

image

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -29,7 +29,7 @@ import {
   TickItem,
   AnimationDuration,
 } from '../util/types';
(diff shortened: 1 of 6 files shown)


PavelVanecek (approved):
Can we also have a test that it does clip when clipDot is true?

ckifer:
These tests made some bad assumptions, revision incoming

ckifer:
aight better now. Will port this to 3.x too

PR #4678, opened 2024-06-20
fix: clipDot setting fails to take effect if set as the only DotProp (3.x)

ckifer:
## Description

Fix clipDot not working when props are of the form `dot={{ clipDot: false }}` or similar

This time for 3.x

**ALSO**: The tests in `master` are wayyyy wrong and don't actually check for anything at all :(
These ones are legit

## Related Issue

https://github.com/recharts/recharts/issues/4671

## Motivation and Context

Fix a bug

## How Has This Been Tested?

See tests, see storybook

image

## Types of changes

## Checklist:

### DIFF
--- a/test/cartesian/Line.spec.tsx
+++ b/test/cartesian/Line.spec.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { render } from '@testing-library/react';
+import { scaleLinear } from 'victory-vendor/d3-scale';
 import { Surface, Line } from '../../src';
 
 describe('<Line />', () => {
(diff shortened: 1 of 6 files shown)


test/cartesian/Line.spec.tsx
```diff
@@ -45,6 +46,50 @@ describe('<Line />', () => {
     );
 
     expect(container.querySelectorAll('.recharts-line-curve')).toHaveLength(1);
+    expect(container.querySelectorAll('.recharts-line-dot')).toHaveLength(0);
+  });
+
+  it('Does not render clip dot when clipDot is false', () => {
```
ckifer, line 52: name needs slightly updated but I'm lazy :(

PavelVanecek (approved):
Yeah I was wondering what the master tests are doing - it checks if an element is present in the DOM? That doesn't really tell us much.

Better now.

I would still go even further - and have `expect` for attributes on the clip path. There's some math going on that could have assertions: https://app.codecov.io/gh/recharts/recharts/pull/4678/indirect-changes

PR #4690, opened 2024-06-21
Changes to get clean tests

julianna-langston:
## Description

Once the eslint-plugin-jsx-a11y dependency was updated to v6.9.0, we started failing tests. For my own edification, I wanted to see what was necessary to get those tests back to a clean state.

## Related Issue

Dependency bump PR:
https://github.com/recharts/recharts/pull/4681

## Motivation and Context

To start a discussion about the changes

## How Has This Been Tested?

Tests ran clean locally. We'll see if they also run clean in GHA.

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/SunburstChart.tsx
+++ b/src/chart/SunburstChart.tsx
@@ -182,7 +182,7 @@ export const SunburstChart = ({
       const { x: textX, y: textY } = polarToCartesian(0, 0, innerR + radius / 2, -(start + arcLength - arcLength / 2));
       currentAngle += arcLength;
       sectors.push(
-        <g aria-label={d.name} tabIndex={0}>
+        <g>
           <Sector
             onClick={() => handleClick(d)}
             onMouseEnter={e => handleMouseEnter(d, e)}
--- a/src/shape/Sector.tsx
+++ b/src/shape/Sector.tsx
@@ -247,5 +247,5 @@ export const Sector: React.FC<Props> = sectorProps => {
     path = getSectorPath({ cx, cy, innerRadius, outerRadius, startAngle, endAngle });
   }
 
-  return <path {...filterProps(props, true)} className={layerClass} d={path} role="img" />;
+  return <path {...filterProps(props, true)} className={layerClass} d={path} />;
 };
--- a/test/chart/AccessibilityScans.spec.tsx
+++ b/test/chart/AccessibilityScans.spec.tsx
@@ -33,7 +33,7 @@ import { PageData as data, exampleSankeyData, exampleTreemapData, exampleSunburs
 const svgTagMustHaveLabelViolation = expect.objectContaining({
   description: 'Ensures <svg> elements with an img, graphics-document or graphics-symbol role have an accessible text',
   help: '<svg> elements with an img role must have an alternative text',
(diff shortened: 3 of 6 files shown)


src/chart/SunburstChart.tsx
```diff
@@ -182,7 +182,7 @@ export const SunburstChart = ({
       const { x: textX, y: textY } = polarToCartesian(0, 0, innerR + radius / 2, -(start + arcLength - arcLength / 2));
       currentAngle += arcLength;
       sectors.push(
-        <g aria-label={d.name} tabIndex={0}>
+        <g>
```
ckifer, line 185: this doesn't concern me too much since it is a new-er chart

src/shape/Sector.tsx
```diff
@@ -247,5 +247,5 @@ export const Sector: React.FC<Props> = sectorProps => {
     path = getSectorPath({ cx, cy, innerRadius, outerRadius, startAngle, endAngle });
   }
 
-  return <path {...filterProps(props, true)} className={layerClass} d={path} role="img" />;
+  return <path {...filterProps(props, true)} className={layerClass} d={path} />;
```
ckifer, line 250: as long as this doesn't have any negative impact, all good to me. I think we've removed these roles other places?

julianna-langston:
The jsx-a11y tests threw errors for a couple of things.

1. The Sunburst chart had a `<g aria-label="something" tabIndex={0}>`. Either there should be a `role` to go with that aria-label, or there shouldn't be an aria-label. I elected to remove the aria-label, and I think we should have a subsequent PR to make sure that sunburst charts can work with the `accessibilityLayer` prop, but that's just because it was the easiest thing to do. @csdiehl , what do you think?
2. There were a bunch of cases of `role="img"` for subsections of the chart. Either there should be an `aria-label` to go with those roles, or the roles should be removed. I elected to remove the `role`, since I don't believe it adds anything, and could potentially conflict with the `accessibilityLayer` prop.

ckifer:
I think your points are solid - we shouldn't have this half-supported stuff that may conflict with how we systematically do things.

Just for clarity though, `Sector` is used mostly in Polar charts and Polar charts do not currently work with accessibilityLayer, correct?

julianna-langston:
> Polar charts do not currently work with accessibilityLayer, correct?

I went through the examples in storybook. It looks like `accessibilityLayer` is supported for the following types:

- [x] AreaChart
- [x] BarChart
- [x] ComposedChart
- [ ] FunnelChart
- [x] LineChart
- [ ] PieChart
- [ ] RadarChart
- [ ] RadialBarChart
- [ ] Sankey
- [ ] ScatterChart
- [ ] SunburstChart
- [ ] Treemap

I actually couldn't get tooltips to work with RadarChart or ScatterChart, so I might be missing something there. I think both of those charts should be able to work with the accessibilityLayer, once I figure out how to get tooltips up and running.

Once I have the bandwidth, FunnelChart, PieChart, RadialBarChart, and SunburstChart should be straightforward to work on. 

Treemap and Sankey would require more... experimentation.

ckifer:
got it, thank you! I'm going to merge this then, lgtm

ckifer (approved):

PR #4693, opened 2024-06-22
Error bars in state

PavelVanecek:
## Description

ErrorBars set their settings to Redux. This is preparing for future PR where axisDomain selector will read these and extend the domain based on errobar domain.

## Related Issue


src/cartesian/Bar.tsx
```diff
@@ -311,6 +312,8 @@ function BarRectangles(props: BarRectanglesProps) {
 
 const defaultMinPointSize: number = 0;
 
+const emptyArray: Array<never> = [];
```
ckifer, line 315: What's the reason for the empty arrays?

src/context/CartesianGraphicalItemContext.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+type GraphicalItemContextProps = {
+  data: ChartData;
+  xAxisId: AxisId;
+  yAxisId: AxisId;
+  dataKey: DataKey<any>;
+  children: React.ReactNode;
+};
+
+export const CartesianGraphicalItemContext = ({
```
ckifer, line 28: Why not redux for all of this?

storybook/stories/API/props/utils.ts
```diff
@@ -5,6 +5,7 @@ export const getStoryArgsFromArgsTypesObject = (argsTypes: StorybookArgs): Recor
   const args: Record<string, unknown> = {};
   Object.keys(argsTypes).forEach((key: string) => {
     const defaultValue = argsTypes[key]?.defaultValue ?? argsTypes[key]?.table?.defaultValue;
+    // TODO this erases zeroes and `False` values, perhaps it should check for undefined instead?
     if (defaultValue) {
```
ckifer, line 9: Probably

ckifer (approved):
Nice thanks!

src/cartesian/Bar.tsx
PavelVanecek: So that the reference doesn't change between renders.

src/context/CartesianGraphicalItemContext.tsx
PavelVanecek: We need to keep track which error bar belongs to which graphical item because that's where the axis id is defined. It does end up in the redux store in the end, the context here is necessary to avoid yet another layer of direct Dom access when rendering the error bars.

storybook/stories/API/props/utils.ts
PavelVanecek: Couple of stories break when I do that. I'll fix it in separate PR

PavelVanecek:
I need to have a look at that

PR #4699, opened 2024-06-25
fix: allow to opt out of area dot fallback

### LINKED ISSUE #4697: Area printing a dot when having only one point
We have charts with an Area showing min/max values that looks like this:
Screenshot from 2024-06-25 11-24-45

But due to the following condition:
https://github.com/recharts/recharts/blob/bfd179194d77d6a46ba265681e9cad705ee56921/src/cartesian/Area.tsx#L669

Instead of just not print anything when there is only one point, it prints a dot:
Screenshot from 2024-06-25 11-24-24

It feels strange for our users since it looks like there is a max value but no min one which is wrong.

The condition seems introduced in this commit but it is very old and I'm not confident in saying what it does and if it is safe to juste remove the condition to avoid this problem:
https://github.com/recharts/recharts/commit/ab452bcbc77b75ea76632a2a293abd98dcfba98c

I can create a PR for its removal if you think it is safe to or maybe add a prop named `noSingleDotFallback` to avoid this behavior in cases like mine.

nfroidure:
fix #4697

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -666,7 +666,7 @@ export class Area extends PureComponent<Props, State> {
             </defs>
           ) : null}
           {!hasSinglePoint ? this.renderArea(needClip, clipPathId) : null}
-          {(dot || hasSinglePoint) && this.renderDots(needClip, clipDot, clipPathId)}
+          {(dot || (hasSinglePoint && dot !== false)) && this.renderDots(needClip, clipDot, clipPathId)}
           {(!isAnimationActive || isAnimationFinished) && LabelList.renderCallByParent(this.props, points)}
         </Layer>
         <ActivePoints


src/cartesian/Area.tsx
```diff
@@ -666,7 +666,7 @@ export class Area extends PureComponent<Props, State> {
             </defs>
           ) : null}
           {!hasSinglePoint ? this.renderArea(needClip, clipPathId) : null}
-          {(dot || hasSinglePoint) && this.renderDots(needClip, clipDot, clipPathId)}
+          {(dot || (hasSinglePoint && dot !== false)) && this.renderDots(needClip, clipDot, clipPathId)}
```
ckifer, line 669: so I think if we were going to do this we would just remove the `hasSinglePoint` check altogether. Then when dot is falsy the dots wouldn't render. Adding an additional condition only checking for false is confusing.

ckifer (changes requested):
This is a breaking change so we'd have to change it in the next major rather than a patch/minor (which we're working on and is the branch you PR'ed to so all good). 

But still not sure if we should change it or not. Ideally it is configurable, which it currently is you just have to provide a function to the `dot` prop.

Otherwise, we need _something_ to be rendered when there is one datapoint I think. Thats why this was here initially.

@PavelVanecek @nikolasrieble any thoughts?

src/cartesian/Area.tsx
nfroidure: Since the  `dot` prop is optional, it leaves some field for people using default to have the default behavior.

My thoughts are that the dots should not be there at all when not set to `true` but trying to fit with the will to not change the default behavior.

PavelVanecek:
We're going to need tests for this one - all the combinations of single dot multiple dot, dot true false.

The code might be confusing (and it wouldn't be the first here) but let's look at the tests once we have them and see if the external behaviour is intuitive and then go from there.

src/cartesian/Area.tsx
ckifer: its defaulted to false here https://github.com/recharts/recharts/blob/3.x/src/cartesian/Area.tsx#L171

so this would still be the case as the new default behavior unless the default itself changed.

nfroidure: Thanks for the feedback. So the only way to do this is to just remove `hasSinglePoint` condition.

nfroidure:
Since I could workaround the problem, I let maintainers see if it is still relevant if so, I can update the PR with the breaking change or just close it otherwise. I think it makes sense to though.

ckifer:
So just considering what other charting libraries do for a moment.

Highcharts as an example by default will show dots when there is one datapoint (and just in general)
https://jsfiddle.net/9to71hf4/8/

This can be disabled and the one datapoint is only shown on hover
https://jsfiddle.net/h8e75ysg/8/

I think this is a valid approach, with the `dot={false}` breaking change. Our issue is that `false` is the default value of dot. As a breaking change in 3.x we could change the default value to `undefined`, make explicit false never render the dot, and keep the current default behavior (not value) the same.

`undefined` = default don't show dots, but show one when there is one datapoint
`true` = always show
`false` = never show (except on hover, but that is controlled with `activeDot`

ckifer:
Considering your specific issue though 
> It feels strange for our users since it looks like there is a max value but no min one which is wrong.

Wouldn't it be better if a min value was rendered too when there is a range? Even though it wouldn't shade between them it'll at least show the user the data that it is supposed to

ckifer:
I think I'm going to close this as stale since there is a relatively easy workaround for this issue. If we get more asks for it we can revisit

PR #4712, opened 2024-06-30
Stacked data domains in Redux

PavelVanecek:
## Description

Couple of changes here:
- selectAxisDomain no longer considers niceTicks
- new separate selector for niceTicks
- selectAxisDomain now uses stack groups to get the right domain
- selectCartesianItemsSettings now uses hide+includeHidden props
- hidden Bar now reports its settings too
- and a bunch of tests

## Related Issue

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -666,6 +666,7 @@ export class Bar extends PureComponent<Props, State> {
         xAxisId={this.props.xAxisId}
         yAxisId={this.props.yAxisId}
         dataKey={this.props.dataKey}
+        stackId={this.props.stackId}
       >
         <Layer className={layerClass}>
           <ReportBar />
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -55,6 +55,7 @@ import { TooltipPayloadConfiguration } from '../state/tooltipSlice';
 import { SetTooltipEntrySettings } from '../state/SetTooltipEntrySettings';
 import { ReportBar } from '../state/ReportBar';
 import { CartesianGraphicalItemContext } from '../context/CartesianGraphicalItemContext';
+import { SetCartesianGraphicalItem } from '../state/SetCartesianGraphicalItem';
 
 export interface BarRectangleItem extends RectangleProps {
   value?: number | [number, number];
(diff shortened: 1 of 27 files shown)


src/cartesian/Bar.tsx
```diff
@@ -646,6 +647,16 @@ export class Bar extends PureComponent<Props, State> {
     if (hide || !data || !data.length) {
       return (
         <>
+          <SetCartesianGraphicalItem
+            // Bar does not allow setting data directly on the graphical item (why?)
+            data={null}
```
ckifer, line 652: hmm, maybe because of the iterations with stacked items? the cases might be too complicated if the data is different?

src/state/graphicalItemsSlice.ts
```diff
@@ -42,6 +43,14 @@ export type CartesianGraphicalItemSettings = {
    * One graphical item can have multiple error bars. This probably only makes sense in Scatter.
    */
   errorBars: ReadonlyArray<ErrorBarsSettings> | undefined;
+  stackId: StackId | undefined;
+  /**
+   * Why not just stop pushing the graphical items to state when they are hidden?
+   * Well some components decide to continue showing them anyway.
```
ckifer, line 49: right - this is useful for having the `hide` prop control visibility on legend click. `includeHidden` as well - if the graphical item wasn't there it couldn't use that dataset to keep the domain the same

src/util/ChartUtils.ts
```diff
@@ -1272,23 +1280,23 @@ export const getStackedDataOfItem = <StackedData>(
...
 
-const getDomainOfSingle = (data: Array<Array<any>>): number[] =>
-  data.reduce(
-    (result: number[], entry: Array<any>): number[] => [
-      min(entry.concat([result[0]]).filter(isNumber)),
-      max(entry.concat([result[1]]).filter(isNumber)),
-    ],
-    [Infinity, -Infinity],
-  );
+const getDomainOfSingle = (data: Array<Array<any>>): number[] => {
```
ckifer, line 1283: `getDomainOfSingle` what? Ik this is preexisting but maybe we can add a comment here

ckifer (approved):
that is a lot of tests :o noice

src/util/ChartUtils.ts
PavelVanecek: Of single stack. I will delete this soon.

ckifer: Cool cool

Issue #4717, opened 2024-07-02
v2.9.3+ tooltip does not render on right half of Line graph


Forfold:
## Do you want to request a feature or report a bug?
Bug, [reference PR to dep update](https://github.com/target/goalert/pull/3861)

## What is the current behavior?
After upgrading from `v2.9.2`, we noticed that our tooltip does not render on the right half of our Line graph. This happens on an upgrade to `v2.9.3` or any future release. [This PR seems to be the only change?](https://github.com/recharts/recharts/pull/3932) When I hover over the line that does not render a tooltip, my console log in the tooltip's `content` prop _does_ still show.

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar.

**[Codesandbox Link](https://codesandbox.io/p/sandbox/simple-line-chart-forked-ply72f?file=%2Fsrc%2FApp.tsx)**

Left side of graph, mouseover a line:
![Screenshot 2024-07-02 at 8 58 50 AM](https://github.com/recharts/recharts/assets/11381794/129a3ae9-7b97-478c-95cb-f502f9bf49b6)

Right side of graph, mouseover a line:
![Screenshot 2024-07-02 at 8 58 56 AM](https://github.com/recharts/recharts/assets/11381794/b3a22af1-df22-4e02-9241-d993aab5966e)

## What is the expected behavior?
Tooltip renders as expected

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Chrome, Yes, `v2.9.2`

ckifer:
Remove the `data` prop from the high level chart. Then your Tooltip will render normally.

This shouldn't have been a breaking change though.. still looking into it

Forfold:
Oooh that was simple. Thank you @ckifer! Hoping you can find the breaking change soon.

This was the only change I noticed from an upgrade to 2.9.2 to 2.9.3:

```
const { data = chartData } = child.props;
```

->

```
const data = (child.props.data ?? chartData).slice(state.dataStartIndex, state.dataEndIndex + 1);
```

ckifer:
No problem!

Your `data` array that you pass into the high level chart is not in the form that recharts knows how to react to since it consists of nested series. None of the datakeys that you reference in the chart itself are within that structure. If all of your data a flat array of objects and your series were instead different keys per line
i.e.
```
data = [
  {
    name: 'xAxis1'
    series1: 100
    series2: 300
  }
[...]

PR #4718, opened 2024-07-03
Fix: Modify to ensure data is not sliced if the active index is outside the dataIndex range.

HHongSeungWoo:
## Related Issue

#4717 

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -240,7 +240,13 @@ const getTooltipContent = (
      */
     let data = child.props.data ?? chartData;
 
-    if (data && state.dataStartIndex + state.dataEndIndex !== 0) {
+    if (
+      data &&
+      state.dataStartIndex + state.dataEndIndex !== 0 &&
+      // https://github.com/recharts/recharts/issues/4717
+      // The data is sliced only when the active index is within the start/end index range.
+      state.dataEndIndex - state.dataStartIndex >= activeIndex
+    ) {
       data = data.slice(state.dataStartIndex, state.dataEndIndex + 1);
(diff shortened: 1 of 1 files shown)


src/chart/generateCategoricalChart.tsx
```diff
@@ -681,6 +687,7 @@ export const createDefaultState = (props: CategoricalChartProps): CategoricalCha
   const brushItem = findChildByType(children, Brush);
   let startIndex = 0;
   let endIndex = 0;
+
   if (props.data && props.data.length !== 0) {
     endIndex = props.data.length - 1;
```
ckifer, line 692: This probably needs to reference the combined graphical items data arrays if they are set. 

We can probably do that in 3.x so this works no matter where data is set. 

@PavelVanecek fyi

src/chart/generateCategoricalChart.tsx
```diff
@@ -240,7 +240,13 @@ const getTooltipContent = (
      */
     let data = child.props.data ?? chartData;
 
-    if (data && state.dataStartIndex + state.dataEndIndex !== 0) {
+    if (
+      data &&
+      state.dataStartIndex + state.dataEndIndex !== 0 &&
+      // https://github.com/recharts/recharts/issues/4717
+      // The data is sliced only when the active index is within the start/end index range.
+      state.dataEndIndex - state.dataStartIndex >= activeIndex
```
ckifer, line 248: This gets us back to the previous behavior as far as I know. 

I think it still won't play nice with brush but that's a different issue

ckifer (approved):

ckifer:
Thank you

PR #4719, opened 2024-07-03
refactor: start moving defaultIndex functionality to selectors

ckifer:
## Description

I want to move this defaultIndex init to redux to remove DOM access of `Tooltip` and this function from the generator. However, how do we support this for Scatter and other charts?

Hunting for some ideas before continuing

## Motivation and Context

remove dom access and move to redux

## How Has This Been Tested?

it hasn't yet!

## Types of changes

- [X] Refactor

## Checklist:

### DIFF
--- a/src/state/selectors.ts
+++ b/src/state/selectors.ts
@@ -124,9 +124,11 @@ export function selectActiveCoordinate(
   state: RechartsRootState,
   tooltipEventType: TooltipEventType,
   trigger: TooltipTrigger,
+  defaultIndex: number | undefined,
 ): ChartCoordinate | undefined {
   const tooltipState: TooltipState = selectTooltipState(state);
   let activeCoordinate: ChartCoordinate;
+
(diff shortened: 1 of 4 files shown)


src/state/selectors.ts
```diff
@@ -138,6 +140,35 @@ export function selectActiveCoordinate(
   } else {
     activeCoordinate = tooltipState.axisInteraction.activeClickCoordinate;
   }
+
+  // the following works for bar, area, line, but does not work for polar or scatter charts :(
```
PavelVanecek, line 144: Is this inherent to Scatter and polar or is it because of the axis/item tooltip type?

ckifer: Everything else for axisType is being selected correctly I think?

I can use axis type to check things, but that y coordinate still doesn't exist in the data. 

I might need to reset some state 💡

PR #4723, opened 2024-07-04
Simplify append offset of legend

ckifer:
## Description

Remove DOM access one thing at a time.

_technically_ this doesn't remove DOM access, but it does remove an unnecessary call to getChildByType and a lot of other stuff that isn't really being used for anything. Yay!

Also we don't need to worry about compat anymore in 3.x so remove export and remove `getLegendProps`

## Motivation and Context

Above. Remove some code.

## How Has This Been Tested?

npm test

Wait for chromatic diff

## Types of changes

- [X] Refactor

## Checklist:


src/chart/generateCategoricalChart.tsx
```diff
@@ -736,8 +736,7 @@ const calculateOffset = (
   }
 
   if (legendItem && prevLegendBBox) {
-    // @ts-expect-error margin is optional in props but required in appendOffsetOfLegend
-    offset = appendOffsetOfLegend(offset, props, prevLegendBBox);
+    offset = appendOffsetOfLegend(offset, legendItem, prevLegendBBox);
```
ckifer, line 739: we only passed in props to get `legendItem` again using `findChildByType`, but we quite literally already have it above. Why find it twice? Don't!

src/util/getLegendProps.ts
ckifer, line 1: none of the props that are created from merging props and the result of `getWidthOrHeight` are actually used. So we really don't need this

ckifer:
No diff https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1112

ckifer:
Codecov failing because lines were removed

PavelVanecek (approved):
Nice!

PR #4725, opened 2024-07-04
fix(polar): set paddingAngle 0 when pie data length <= 1

syi0808:
# Set paddingAngle 0 when pie data length <= 1

## Description
If datas length is less or equal then 1 in props, set force paddingAngle to 0.

## Motivation and Context
In my project using recharts's pie chart. I expected that setting paddingAngle would not create a gap when there was 1 data, but it did not.
According to https://recharts.org/en-US/api/Pie#paddingAngle. paddingAngle is "The angle between two sectors.", So when there was just 1 data, it should not create gap.

## How Has This Been Tested?
recharts has no e2e test. So i tested on storybook.

### before
Screenshot 2024-07-04 at 9 39 34 PM
### after
Screenshot 2024-07-04 at 9 53 51 PM

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -352,11 +352,12 @@ export class Pie extends PureComponent<Props, State> {
       return null;
     }
 
-    const { cornerRadius, startAngle, endAngle, paddingAngle, dataKey, nameKey, tooltipType } = item.props;
+    const { cornerRadius, startAngle, endAngle, dataKey, nameKey, tooltipType } = item.props;
     const minAngle = Math.abs(item.props.minAngle);
     const coordinate = Pie.parseCoordinateOfPie(item, offset);
     const deltaAngle = Pie.parseDeltaAngle(startAngle, endAngle);
     const absDeltaAngle = Math.abs(deltaAngle);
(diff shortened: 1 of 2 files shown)


ckifer:
I think the title is wrong? <= 1 right? Looks like the code is correct

ckifer (commented):
Makes sense to me. Could we add a unit test?

syi0808:
> Makes sense to me. Could we add a unit test?

sorry to make confuse you. I have trouble to writing test.
We can't access rendered component's state, so I have one idea. test to snapshot from rendered path element in pie component. Would that be a good way?

ckifer:
Snapshot tests aren't generally a good idea, nor are testing internal state. 

What you can do is render the pie test that the it does not look like it did previously. Find the element, find the path attribute, asset on that

syi0808:
In rechart latest version, when set paddingAngle to 360, path element disappeared.
So i tested if path is rendered when setting data.length to 1 and paddingAngle to 360.

ckifer:
That works. As long as it was failing before the change and is passing after. 

Will potentially help prevent future regressions

syi0808:
force push just remove test.only syntax.

ckifer (approved):
LGTM as long as it passes CI

ckifer:
Thank you

PR #4729, opened 2024-07-06
Reference elements extend domain

PavelVanecek:
## Description

I think this should be the last one required for axis scales in redux.

## Related Issue


ckifer (approved):
Nice 🚀

PR #4730, opened 2024-07-07
XAxis uses scale (and domain and range) from redux instead of generator

PavelVanecek:
## Description

This breaks tooltip synchronisation and it will remain broken until that's moved to Redux store too. There is no visual diff though! https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1129

## Related Issue

### DIFF
--- a/src/cartesian/XAxis.tsx
+++ b/src/cartesian/XAxis.tsx
@@ -109,6 +109,7 @@ const XAxisSettingsDispatcher = (props: Props) => {
         allowDecimals={props.allowDecimals}
         tickCount={props.tickCount}
         includeHidden={props.includeHidden ?? false}
+        reversed={props.reversed}
       />
       <XAxisImpl {...props} />
     </>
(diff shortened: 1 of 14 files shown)


PavelVanecek:
I may have missed the `reversed` prop. I will add more tests to the main branch before merging.

ckifer:
Well I can fix the synchronization part! It's just not perfectly perfect

PavelVanecek:
Yeah let's keep the sync for later PR.

PavelVanecek:
Weird, this test works on my local but fails in CI. What's going on?

ckifer (approved):

ckifer:
Looks like we have a few issues 
https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1131

* composed bars are rendering on the Y Axis
* difference in ticks
* extendDomain/tick calculation isn't working for error bars (looks like this maybe wasn't working before either, not totally sure though)

PavelVanecek:
ErrorBars domain never extended the X axis, only Y axis. I added that in the selector but apparently that alone was not good enough.

I'll look at the ticks.

ckifer:
Oh huh, guess I'm thinking of reference elements

PavelVanecek:
Okay interesting. The scatter is now actually _more correct_ because the axis domain takes into account the error bars. But unfortunately the ticks don't, and the graphical elements also don't. This might take a while to resolve.

Do you prefer to revert this change and get the storybook back to baseline, or keep it as "rejected" and roll forward?

PR #4732, opened 2024-07-07
XAxis uses niceTicks from redux instead of generator

PavelVanecek:
## Description

Following up from problems in https://github.com/recharts/recharts/pull/4730 here I am fixing two of them:

1. nice ticks in generator were excluded for scale=time, but included in redux
2. axis domain (and therefore ticks) were extended for XAxis in redux, but not extended in generator

I found that while trying to read niceTicks from redux and then I figured I might just as well commit that part.

## Related Issue

https://github.com/recharts/recharts/pull/4730

### DIFF
--- a/src/state/axisSelectors.ts
+++ b/src/state/axisSelectors.ts
@@ -607,11 +607,59 @@ export const selectAxisDomain = (
   return selectNumericalDomain(state, axisSettings, axisType, axisId);
 };
 
+const unknownScale: ParsedScaleReturn = {
+  scale: undefined,
+  realScaleType: undefined,
+};
+
+/**
+ * This returns an axis scale and type (linear, log, etc.) for the given axis.
+ * This is before applying the correct domain and range!
+ *
+ * That's on one hand silly, yes, but on the other hand also required
+ * because the final domain depends on nice ticks,
+ * and nice ticks depend on the scale type.
+ * With this shared dependency we cannot create the scale and set its domain in one step.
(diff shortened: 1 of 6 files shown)


src/state/axisSelectors.ts
```diff
@@ -607,11 +607,59 @@ export const selectAxisDomain = (
...
+/**
+ * This returns an axis scale and type (linear, log, etc.) for the given axis.
+ * This is before applying the correct domain and range!
+ *
+ * That's on one hand silly, yes, but on the other hand also required
+ * because the final domain depends on nice ticks,
+ * and nice ticks depend on the scale type.
+ * With this shared dependency we cannot create the scale and set its domain in one step.
+ *
+ * Usually you would never want to use this selector, ever, outside of this circular dependency problem.
```
ckifer, line 624: Hmm, I wonder what our path away from this is then

ckifer (approved):
Nice nice. Prefer to roll forwards it is

src/state/axisSelectors.ts
PavelVanecek: For now, no path away, it's fine and it can stay like this. An implementation detail, never exported.

If we choose to separate `parseScale` into two methods, one that decides on the `realScaleType` and second which actually creates the scale function/object then we might work around it. But then we get a new problem of transforming from `realScaleType` to scale function so I was in no hurry in doing that.

Issue #4266, opened 2024-03-05
RadarChart PolarAngleAxis getting cut off.


kenperregaux:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/simple-radar-chart-forked-qfjnvf?file=%2Fsrc%2FApp.tsx&from-embed=)

### Steps to reproduce
Wrapping only happens if there is a space in the label. I would expect it to wrap regardless. Using scaleToFit removes the label entirely with or without a width.

### What is expected?
Wrapping regardless of label containing spaces or scaleToFit to work.

### What is actually happening?
See example above...text wraps only if label contains a space OR scaleToFit to work.

| Environment | Info |
|---|---|
| Recharts | v2.7.3 |
| React | 18.2 |
| System | Windows 11 Pro |
| Browser | Chrome Version  122.0.6261.71 |

ckifer:
I would generally say this is an issue for `margin` on `RadarChart` to solve

[![Edit simple-radar-chart (forked)](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/simple-radar-chart-forked-f786ck?file=%2Fsrc%2FApp.tsx%3A66%2C6)

`margin` however, does not take effect when using static `cy`, `cx`, and `outerRadius`. If you'd like to change those values while still having them be relative, you can change them as percentage values rather than whole numbers.

kenperregaux:
That is fine until you make the chart smaller (say width 350) then you have to play with the margins until it fits? This makes no sense. If I use scaleToFIt or set a width on the Text it should wrap automatically (not just when the label has spaces).

ckifer:
is using percentages an issue?

Recharts doesn't have any magic/dynamic `scaleToFit` solutions for any charts at the moment. `scaleToFit` is supposed to be based off of the `Text` component's `width` property. If you have a chart that changes size margin won't adjust automatically unless you make it dynamic yourself (which works well).

That being said, `Text` itself needs some work it seems (apologies, legacy inheritor of recharts. this functionality has been the same a long time before me). 

For now I would just use something custom like the following or use a `foreignObject` to insert HTML and use CSS for line breaking:
```jsx
        tick={(props) => {
          const { payload, ...rest } = props;
[...]

PR #4739, opened 2024-07-08
Add surface to text story - that way we can test angle and scaling

razzeee:
## Description

Setting `scaleToFit` on `text` to `true` causes the text to become invisible (or so small that you can't see it)

See https://github.com/recharts/recharts/issues/4266

IMO - even after this, `scaleToFit` doesn't do what the name suggests, but allowing the text to show is a good start.

## Related Issue

Partially related https://github.com/recharts/recharts/issues/4266

## Motivation and Context

I'm hoping to solve overflowing labels in a radar chart with this, we'll see if it helps - but it's still a bug anyway.

## How Has This Been Tested?

Ran the storybook and played around with it.

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Text.tsx
+++ b/src/component/Text.tsx
@@ -240,7 +240,7 @@ export const Text = ({
   if (scaleToFit) {
     const lineWidth = wordsByLines[0].width;
     const { width } = props;
-    transforms.push(`scale(${(isNumber(width as number) ? (width as number) / lineWidth : 1) / lineWidth})`);
+    transforms.push(`scale(${isNumber(width as number) ? (width as number) / lineWidth : 1})`);
   }
   if (angle) {
     transforms.push(`rotate(${angle}, ${x}, ${y})`);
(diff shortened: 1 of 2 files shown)


ckifer (approved):
Good start, thank you! 

Even if this doesn't do exactly what we want, getting the storybook correct is one thing. Not scaling down to microscopic text is another so good stuff 🚀

src/component/Text.tsx
```diff
@@ -240,7 +240,7 @@ export const Text = ({
   if (scaleToFit) {
     const lineWidth = wordsByLines[0].width;
     const { width } = props;
-    transforms.push(`scale(${(isNumber(width as number) ? (width as number) / lineWidth : 1) / lineWidth})`);
+    transforms.push(`scale(${isNumber(width as number) ? (width as number) / lineWidth : 1})`);
```
ckifer, line 243: We should probably have a unit test in place for this in general even if it can't be perfect. Not a blocker for this for now, but if we can add it we should to prevent regressions

razzeee: I think it would be easier to add this, if it's actually fixed? I guess we could check that it never gets too small now, but if the actual behavior is in place, I imagine, you could test for that?

ckifer: this is good for now though

PR #4740, opened 2024-07-08
If we `scaleToFit` we don't want to truncate and ellipse the string, …

razzeee:
…instead return all words

## Description

Basically makes `scaleToFit` do what it's supposed to do. Will make the text both smaller or bigger, depending on the width.

## Related Issue

None

## Motivation and Context

Follow up to https://github.com/recharts/recharts/pull/4739

## How Has This Been Tested?

Only via storybook

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Text.tsx
+++ b/src/component/Text.tsx
@@ -99,7 +99,7 @@ const calculateWordsByLines = (
   const findLongestLine = (words: Array<Words>): Words =>
     words.reduce((a: Words, b: Words) => (a.width > b.width ? a : b));
 
-  if (!shouldLimitLines) {
+  if (!shouldLimitLines || scaleToFit) {
     return originalResult;
   }
 


src/component/Text.tsx
```diff
@@ -99,7 +99,7 @@ const calculateWordsByLines = (
   const findLongestLine = (words: Array<Words>): Words =>
     words.reduce((a: Words, b: Words) => (a.width > b.width ? a : b));
 
-  if (!shouldLimitLines) {
+  if (!shouldLimitLines || scaleToFit) {
```
ckifer, line 102: again, we should add a unit test for these cases. Where before it did add ellipses, now it doesn't, etc.

ckifer (approved):
Checked the functionality by adding the check locally, definitely better than now I think. LGTM

Issue #4735, opened 2024-07-08
Y Axis Tick overflow calculation is wrong


bernardobelchior:
## Do you want to request a feature or report a bug?

Report a bug

## What is the current behavior?

When using a custom component for the `YAxis` `tick` prop that sets the `width`, `maxLines` and `fontSize` props of the `Text` component, the label is still cut off in some cases. If we pass the `style` prop, the label is still cut, but it shows an ellipsis (`...`). 

Without `style={{ fontSize: "1.2rem" }}` prop:

![image](https://github.com/recharts/recharts/assets/12778398/1371ef20-3bb0-4922-8a7e-470d2dffb52f)

With `style={{ fontSize: "1.2rem" }}` prop:

![image](https://github.com/recharts/recharts/assets/12778398/60a2643f-7407-4929-b375-85f980084182)

[...]

PR #4743, opened 2024-07-09
Avoid adding ellipsis if text does not overflow

bernardobelchior:
## Description

Avoid adding an ellipsis (`...`) if the text does not overflow. This should not only fix an issue, but also improve performance slightly as we don't need to check for overflow for tick labels that we know will fit. 

## Related Issue

https://github.com/recharts/recharts/issues/4735

## Motivation and Context

There seems to be a small bug in the ellipsis logic for the `Text` component, more specifically in `calculateWordsByLines`. 

The bug happens because sometimes adding a new character creates a narrower string than adding the suffix `…`. In that case, Recharts is adding a suffix, but adding the last character would not overflow. I tested this suspicion with the code below:

```ts
const suffix = '…';
console.log(
  'March',
  getStringSize('March', style).width,
  '|',
  'Marc' + suffix,
  getStringSize('Marc' + suffix, style).width,
);
// March 32.328125 | Marc… 38.328125
```
This can be replicated by running the `Text` Storybook story and setting the arguments below: 

<details>
<summary>Storybook Story Arguments</summary>

```js
{
  maxLines: 1,
  textAnchor: 'start',
  verticalAnchor: 'start',
  angle: 0,
  width: 33,
  y: 50,
  x: 50,
  fontSize: '12px',
  letterSpacing: '0.2px',
  content: 'March',
  style: {
    fontSize: '12px',
    letterSpacing: '0.2px',
  },
}
```

</details>

The algorithm should checking if the initial content doesn't overflow, and if it doesn't then return that value immediately. 
Instead, it's always running the `while` loop to check where to trim, and it is actually running code that the following comment says it shouldn't: 

https://github.com/recharts/recharts/blob/0179beddc1070d0723cedc290c6dcb09122bc0e6/src/component/Text.tsx#L153-L155

This happens because if text doesn't overflow, `trimmedResult` will be `undefined`, so `originalResult` will be returned. 
If we check for overflow early, we can avoid running the `while` loop altogether. 

## How Has This Been Tested?

Added unit tests and tested manually in Storybook `Text` story, using the following arguments:

<details>
<summary>Storybook Story Arguments</summary>

```js
{
  maxLines: 1,
  textAnchor: 'start',
  verticalAnchor: 'start',
  angle: 0,
  width: 33,
  y: 50,
  x: 50,
  fontSize: '12px',
  letterSpacing: '0.2px',
  content: 'March',
  style: {
    fontSize: '12px',
    letterSpacing: '0.2px',
  },
}
```

</details>

Before:

image

After:

image

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Text.tsx
+++ b/src/component/Text.tsx
@@ -167,7 +167,7 @@ const getWordsWithoutCalculate = (children: React.ReactNode): Array<Words> => {
 
 type GetWordsByLinesProps = Pick<Props, 'width' | 'scaleToFit' | 'children' | 'style' | 'breakAll' | 'maxLines'>;
 
-const getWordsByLines = ({ width, scaleToFit, children, style, breakAll, maxLines }: GetWordsByLinesProps) => {
+export const getWordsByLines = ({ width, scaleToFit, children, style, breakAll, maxLines }: GetWordsByLinesProps) => {
   // Only perform calculations if using features that require them (multiline, scaleToFit)
   if ((width || scaleToFit) && !Global.isSsr) {
     let wordsWithComputedWidth: Array<WordWithComputedWidth>, spaceWidth: number;
--- a/src/component/Text.tsx
+++ b/src/component/Text.tsx
@@ -103,6 +103,11 @@ const calculateWordsByLines = (
     return originalResult;
   }
 
+  const overflows = originalResult.length > maxLines || findLongestLine(originalResult).width > Number(lineWidth);
+  if (!overflows) {
+    return originalResult;
+  }
+
   const suffix = '…';
 
   const checkOverflow = (index: number): [boolean, Words[]] => {
--- a/test/component/Text.spec.tsx
+++ b/test/component/Text.spec.tsx
@@ -1,7 +1,10 @@
 import { render, screen } from '@testing-library/react';
 import React from 'react';
+import { vi } from 'vitest';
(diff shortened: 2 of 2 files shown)


ckifer:
If we can let's try for a unit test. Might have to mock some things but at least getting the code path to execute and asserting we did have ellipses and now we don't

bernardobelchior:
> If we can let's try for a unit test. Might have to mock some things but at least getting the code path to execute and asserting we did have ellipses and now we don't

Sure 👍 as soon as I have some time, I'll add some tests

bernardobelchior:
@ckifer added some tests. The `getWordsByLines` is very complex with many different permutations, so I added some. Let me know if there's something you think is important that we should add. I'd like to add more tests, but have limited time, so left some (most?) permutations untested, but I guess this is already an improvement when it comes to increasing code coverage. Let me know what you think!

ckifer:
Focusing on the ones you've added should be good enough. I'll review when I get a chance

ckifer (approved):
Thank you for the tests! 

LGTM although I haven't spent very much time in this component. Not a risky change or anything though so 🚢

bernardobelchior:
> Thank you for the tests!
> 
> LGTM although I haven't spent very much time in this component. Not a risky change or anything though so 🚢

Thank you 🙏

Issue #4757, opened 2024-07-11
Provide all payloads to Tooltip when the chart contains multiple X Axis


alexbalonperin:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/line-chart-width-multi-series-forked-zd2hk5)

### Steps to reproduce
Hover on the chart to trigger the tooltip and notice that only the payload related to the first X-Axis are visible/available.

### What is expected?
When multiple X-Axis are used, all data series should be available in the payload of the tooltip.

### What is actually happening?
Only the data series for the first X-Axis are available

| Environment | Info |
|---|---|
| Recharts | v2.12.7 |
| React | 16.14.0 |
| System | Linux Pop!_OS 22.04 LTS  |
| Browser | Chrome Version 126.0.6478.126 (Official Build) (64-bit) |

---
I looked at existing issues but I don't think there was anything that was exactly like this.
However, the following issues seem related:
- https://github.com/recharts/recharts/issues/2348
- https://github.com/recharts/recharts/issues/4108

sroy8091:
@ckifer @PavelVanecek I think this is also fixed in this my PR https://github.com/recharts/recharts/pull/6116

ckifer:
We only changed Area chart - things still might need adjusted for the others

Anatole-Godard:
Hey,
Same issue around here. Has anyone know any workaround? Thanks

PR #4758, opened 2024-07-11
Fix: show all series data on tooltip when multiple X-Axis

alexbalonperin:
## Description
Fixes https://github.com/recharts/recharts/issues/4757

I am using the state xAxisMap to access the domain of each x axis and create the missing payloads.
Not sure that's the best solution but it works and I couldn't really find a more elegant solution without major refactoring.

## How Has This Been Tested?
I have added a story for the line chart. Also, all automated tests are passing.

Before the fix:
Screenshot from 2024-07-12 12-58-46

After the fix:
Screenshot from 2024-07-12 13-00-02

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -285,8 +285,19 @@ const getTooltipData = (state: CategoricalChartState, chartData: any[], layout: 
   const activeIndex = calculateActiveTickIndex(pos, ticks, tooltipTicks, axis);
 
   if (activeIndex >= 0 && tooltipTicks) {
-    const activeLabel = tooltipTicks[activeIndex] && tooltipTicks[activeIndex].value;
-    const activePayload = getTooltipContent(state, chartData, activeIndex, activeLabel);
+    const activeLabel = tooltipTicks[activeIndex] && tooltipTicks[activeIndex].value; // for backward compatibility
+    let activePayload: any[];
+    if (state.xAxisMap) {
+      activePayload = Object.values(state.xAxisMap).reduce<any[]>((acc, xAxis) => {
+        const label: TickItem['value'] | undefined = Array.isArray(xAxis.domain)
+          ? xAxis.domain[activeIndex]
+          : tooltipTicks[activeIndex]?.value;
+        const tooltipContent: any[] | null = getTooltipContent(state, chartData, activeIndex, label);
+        return [...acc, ...(tooltipContent ?? [])];
+      }, []);
+    } else {
+      activePayload = getTooltipContent(state, chartData, activeIndex, activeLabel);
+    }
     const activeCoordinate = getActiveCoordinate(layout, ticks, activeIndex, rangeData);
 
     return {
--- a/test/chart/AreaChart.spec.tsx
+++ b/test/chart/AreaChart.spec.tsx
@@ -1,3 +1,4 @@
+import { screen, waitFor } from '@testing-library/dom';
 import { fireEvent, render } from '@testing-library/react';
 import React, { ComponentProps, FC } from 'react';
 import { vi, SpyInstance } from 'vitest';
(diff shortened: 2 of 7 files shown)


ckifer:
Will need to make sure this doesn't break anything. Seems like it doesn't at first glance, but won't be good if it does - will also need to consider this in 3.x

ckifer:
Also could you screenshot what this makes your TT look like and paste it in the PR? ty!

src/chart/generateCategoricalChart.tsx
```diff
@@ -285,8 +285,19 @@ const getTooltipData = (state: CategoricalChartState, chartData: any[], layout:
   const activeIndex = calculateActiveTickIndex(pos, ticks, tooltipTicks, axis);
 
   if (activeIndex >= 0 && tooltipTicks) {
-    const activeLabel = tooltipTicks[activeIndex] && tooltipTicks[activeIndex].value;
-    const activePayload = getTooltipContent(state, chartData, activeIndex, activeLabel);
+    const activeLabel = tooltipTicks[activeIndex] && tooltipTicks[activeIndex].value; // for backward compatibility
+    let activePayload: any[];
+    if (state.xAxisMap) {
```
ckifer, line 290: thinking about this again, this does not work if you have a chart in the opposite layout.

Trying it, this causes a pretty big regression where the tooltip does not show at all

ckifer (changes requested):
image
we can't do this unless we can satisfy all cases - or at least not break things

unit tests for both cases could help cover us

ckifer:
code for the above breaking behavior
```jsx
        <LineChart width={500} height={300} layout="vertical">

          <Tooltip />
          <Legend />
          {finDataOverTime.map((s, sIdx) => {
            return (
              <>
                <YAxis
                  key={`yaxis-${s.name}-`}
                  // domain={['auto', 'auto']}
                  orientation={yAxisOrientations[sIdx]}
                  dataKey={dataKey}
                  allowDuplicatedCategory={false}
                  yAxisId={sIdx}
                  type="category"
                />
                {categories.map((category, cIdx) => {
                  const dataRange = [].concat(
                    ...s.data.map(d =>
                      d[category]
                        ? [
                            {
                              [dataKey]: d[dataKey],
                              [category]: d[category],
                            },
                          ]
                        : [],
                    ),
                  );
                  return (
                    <>
                      <XAxis
                        key={s.name}
                        // dataKey={dataKey}
                        orientation={xAxisOrientations[cIdx]}
                        xAxisId={cIdx}
                        type="number"
                        domain={['auto', 'auto']}
                      />
                      <Line
                        key={`line-${s.name}-${category}`}
                        dataKey={category}
                        data={dataRange}
                        name={`${s.name}: ${category}`}
                        xAxisId={cIdx}
                        yAxisId={sIdx}
                        stroke={colors[cIdx]}
                        fill={colors[cIdx]}
                        strokeDasharray={`5 ${sIdx > 0 ? 5 : ''}`}
                      />
                    </>
                  );
                })}
              </>
            );
          })}
        </LineChart>
      </ResponsiveContainer>
```

src/chart/generateCategoricalChart.tsx
alexbalonperin: Nice catch. I didn't think of the vertical layout. Will add unit tests for both vertical and horizontal layouts.

alexbalonperin: What's interesting is that the regular "Vertical" storybook seems to work fine and show the tooltip as expected. So it's only when there are multiple axis that things go wrong apparently.

I have been trying to write tests for the usecases that you mentioned but I am struggling (first time using react testing library). 

Here is what I was hoping to do:
- Copy the following test: "Renders tooltip when Tooltip item is added"
- Add extra x-axis, y-axis and area to mimic the storybook that you posted
- Fire a mouseover event on one of the area elements: `fireEvent.mouseOver(area)`
- Check for visibility of the tooltip wrapper: `expect(tooltipWrapper).toBeVisible()`

Unfortunately, this doesn't seem to work. I used the testing library screen.debug() to check the state of the dom but there is no change. Also, I added the storybook that you posted and I don't see any mouse event listener on any element so I am a bit confused. :thinking: 

I can't push broken tests so unless I copy the whole thing here you won't see my attempt.

ckifer: You can push broken tests with --no-verify

test/chart/AreaChart.spec.tsx
```diff
@@ -149,6 +150,182 @@ describe('AreaChart', () => {
...
+      const area = container.querySelectorAll('.recharts-surface')[0];
+
+      expect(tooltip).toHaveLength(1);
+      expect(tooltipWrapper).toHaveLength(1);
+
+      await waitFor(() => {
+        fireEvent.mouseOver(area);
+        expect(tooltipWrapper).toBeVisible();
+        screen.debug();
+      });
```
ckifer, line 255: For these you can consider re-using some utilities we have in place 

https://github.com/recharts/recharts/blob/979d921631364d33696d9a5b6f5fba561af5bf5e/test/component/Tooltip/tooltipTestHelpers.tsx#L17

https://github.com/recharts/recharts/blob/979d921631364d33696d9a5b6f5fba561af5bf5e/test/component/Tooltip/tooltipTestHelpers.tsx#L29

ckifer: and assert the payload with the second one

ckifer (commented):
Just a few things, thanks for the patience

ckifer:
Going to close this since this PR is to master, lets get this in 3.x instead

PR #4767, opened 2024-07-14
Refresh axis stories

PavelVanecek:
## Description

- Remove props that do nothing
- Update positions so that label no longer covers ticks
- Fix default values and storybook controls

## Related Issue

## Motivation and Context

There will be some visual diff, I want it to be in its own PR so we can tell it wasn't caused by the other changes.

image

image


ckifer (approved):
lgtm

PR #4768, opened 2024-07-15
refactor: move funnel tooltip coordinate and index to redux

ckifer:
## Description

Small one moving the rest of Funnel state to redux (coordinate)

Wheras previously the tooltip did not "fly in" every time we go out of the chart, now it does. Might want to address that but I think its just because we clear state after mouse out. Can address later if needed.

## Related Issue

## Motivation and Context

## How Has This Been Tested?

- tests pass
- no storybook diff
- if I remove context fallback I can both click and hover to make tooltip work as normal

image

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Tooltip.tsx
+++ b/src/component/Tooltip.tsx
@@ -127,8 +127,8 @@ function TooltipInternal<TValue extends ValueType, TName extends NameType>(props
   const dispatch = useAppDispatch();
 
   useEffect(() => {
-    dispatch(setTooltipSettingsState({ shared }));
-  }, [dispatch, shared]);
+    dispatch(setTooltipSettingsState({ shared, trigger }));
+  }, [dispatch, shared, trigger]);
 
   const viewBox = useViewBox();
   const accessibilityLayer = useAccessibilityLayer();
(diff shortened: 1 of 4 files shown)


src/component/Tooltip.tsx
```diff
@@ -127,8 +127,8 @@ function TooltipInternal<TValue extends ValueType, TName extends NameType>(props
   const dispatch = useAppDispatch();
 
   useEffect(() => {
-    dispatch(setTooltipSettingsState({ shared }));
-  }, [dispatch, shared]);
+    dispatch(setTooltipSettingsState({ shared, trigger }));
+  }, [dispatch, shared, trigger]);
```
ckifer, line 131: need this to get the active index

src/numberAxis/Funnel.tsx
```diff
@@ -115,7 +116,9 @@ function getTooltipEntrySettings(props: FunnelProps): TooltipPayloadConfiguratio
 
 function FunnelTrapezoids(props: FunnelTrapezoidsProps) {
   const { trapezoids, shape, activeShape, allOtherFunnelProps } = props;
-  const { index: activeIndex } = useTooltipContext();
+  const activeItemIndex = useAppSelector(state =>
+    selectActiveIndex(state, 'item', state.tooltip.settings.trigger, undefined),
```
ckifer, line 120: hardcoded item cause I'm lazy and its the only thing funnel supports, can use the actual type later

PavelVanecek (approved):

PR #4782, opened 2024-07-18
fix: do not clear coordinate on item or chart leave

ckifer:
## Description

small fix for Redux tooltip animations

If we reset coordinate on every item and mouse out dispatch then we jarr the user when they hover again as the Tooltip always animates from the corner 0,0. This change keeps Tooltip animation the same as 2.x

## Related Issue

## Motivation and Context

Animations be the same that they are in 2.x. Also less stuttering/flying in from 0,0

## How Has This Been Tested?

- storybook

image

## Types of changes

## Checklist:

### DIFF
--- a/src/state/tooltipSlice.ts
+++ b/src/state/tooltipSlice.ts
@@ -210,17 +210,16 @@ const tooltipSlice = createSlice({
       state.itemInteraction.activeMouseOverCoordinate = action.payload.activeMouseOverCoordinate;
     },
     mouseLeaveChart(state) {
+      // clear everything except for coordinate - we don't always want to animate from 0,0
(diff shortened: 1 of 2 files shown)


ckifer:
Ideally we probably just don't animate if we don't have a coordinate but

PavelVanecek (approved):
We should be able to have a test for this. Even if it's only on the selector level.

1. Start with empty coordinate
2. Mouse over - coordinate selector now returns
3. Mouse out - coordinate selector returns the last known coordinate

ckifer:
Yeah agree - will add

ckifer:
ok better!

PavelVanecek (approved):

PR #4784, opened 2024-07-18
Reference elements use redux

PavelVanecek:
## Description

Continuing on the quest of removing all references to the generator axisMap. It's getting easier because all the selectors are already in place.

I also had to add the concept of "implicit axis". In the generator this is modelled as "getAxisMapByItems" in case there are no axis DOM elements. In that case we still want to render _something_ in the chart, with sensible defaults, so that's what the implicit axis is for.

There is one breaking change here, Reference elements no longer render inside the Brush panorama: https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=6699050c997f5df5a8727981 this will remain broken until we move the Brush cloning to Redux. I think.

## Related Issue


test/cartesian/ReferenceLine/ReferenceLine.spec.tsx
```diff
@@ -268,80 +268,71 @@ describe('<ReferenceLine />', () => {
...
-        </BarChart>,
-      );
-    }).toThrowError(
-      'Invariant failed: Could not find xAxis by id "this ID definitely does not exist anywhere" [string]. Available ids are: 0',
+  it('should not render anything when passed in invalid xAxisId', () => {
+    const { container } = render(
+      <BarChart width={1100} height={250}>
+        <XAxis />
+        <YAxis />
+        <ReferenceLine xAxisId="this ID definitely does not exist anywhere" />
```
ckifer, line 305: Shouldn't we still invariant on these? Now we're back to someone not knowing why their stuff doesn't render

test/cartesian/ReferenceLine/ReferenceLine.spec.tsx
```diff
@@ -354,18 +345,21 @@ describe('<ReferenceLine />', () => {
...
         <YAxis />
-        {/* reference line is currently unable to differentiate between duplicates on categorical axes. Which one do I render on? The first? Both? */}
         <ReferenceLine y={firstDataItem.name} ifOverflow="extendDomain" />
       </BarChart>,
     );
-    expect(container.querySelectorAll('.recharts-reference-line-line')).toHaveLength(0);
+    const allLines = container.querySelectorAll('.recharts-reference-line-line');
+    expect(allLines).toHaveLength(1);
+    // Is this correct? That looks like a category, not pixel coordinate.
+    expect(allLines[0]).toHaveAttribute('y', '201102');
```
ckifer, line 361: That's weird. Pixel values should be in the 100s. 

Maybe something to look into

ckifer (approved):
Nice! It's still possible to mess up axis ids and will be even more so if we add them to grid and tooltip. Why no invariants?

test/cartesian/ReferenceLine/ReferenceLine.spec.tsx
PavelVanecek: I was considering a more hollistic approach. Something like:
- all components should be permissive by default - if something is wrong, return undefined, never throw
- there should be a "strict mode" prop where each of these invariants will throw an error. Like, when chart width is zero.

What do you think?

ckifer: Personally if it's not working I think I'd like to know about it no matter what, but I do like the idea of modes. 

Sounds good then! We can make an issue to track that

PR #4785, opened 2024-07-18
New selectors for various dimensions and layouts

PavelVanecek:
## Description

New selectors, not used yet. I need this in place because next I want to refactor `calculateOffset` into a selector and that needs all of this information in the state first.

## Related Issue

## Motivation

`calculateOffset` is another user of `xAxisMap` and `yAxisMap` and I want to get rid of those.


src/state/brushSlice.ts
```diff
@@ -0,0 +1,22 @@
...
+const initialState: BrushSettings = {
+  height: 0,
+};
+export const brushSlice = createSlice({
+  name: 'brush',
+  initialState,
+  reducers: {
+    setBrushHeight(state, action: PayloadAction<number>) {
+      state.height = action.payload;
+    },
```
ckifer, line 16: Computing total offset with these causes some fun re-render issues since brush and legend are lower in the tree. Axes currently do the same thing

test/cartesian/XAxis.spec.tsx
```diff
@@ -3978,7 +3978,7 @@ describe('<XAxis />', () => {
       expect(itemDataSpy).toHaveBeenCalledTimes(3);
       expect(displayedDataSpy).toHaveBeenLastCalledWith(pageData);
       // oof
-      expect(axisDomainSpy).toHaveBeenCalledTimes(21);
+      expect(axisDomainSpy).toHaveBeenCalledTimes(24);
```
ckifer, line 3981: lol

ckifer (approved):

src/state/brushSlice.ts
PavelVanecek: I'm not sure how to deal with that yet. Ideally we would have some sort of notification when all children in the React tree are rendered. Does that API exist? Not sure.

ckifer: useLayoutEffect? Idk if that solves this case though

We'll be able to see better how to address it once the generator is a function component I think

PR #4795, opened 2024-07-21
CartesianGrid uses redux axes instead of generator axes

PavelVanecek:
## Description

Removing another dependency on generator axisMap.

This adds new props, `xAxisId` and `yAxisId` to CartesianGrid. Before, it used to select "first" available axis. Trouble is, since the axes are stored in an object, there is no definition of what "first" means, and different browsers and environments are allowed to iterate object values in different ways. I very much don't like that so let's add an explicit axis IDs, and default it to 0, same as all other components do.

## Related Issue


PavelVanecek:
Actually hold on. I think there might be some visual diff. I need to look into it before merging.

PavelVanecek:
Okay fixed: https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=669e5efbb2044d25526c2477

there is only one diff, an empty chart now renders gridlines. I think that's acceptable.

Also this is a breaking change - CartesianGrid, with multiple axes, now requires an axis ID, same as other elements. Before, it just picked one at random.

PR #4798, opened 2024-07-22
Upgrade dependencies to the latest version

abnud11:
This PR upgrades all dependencies in `package.json` to the latest version except ESLint which is upgraded to v8 since v9 is not compatible yet with the plugins used in this project.
## Description

The updates other than `package.json` are:
1- husky no longer needs to source from it inside the hooks, also `husky install` changed to just `husky`.
2- with the update from prettier v2 to v3, formatting with the new version of prettier changed a couple of files, you'll notice that the change in those files is only formatting-related.
3- the update to `eslint-config-airbnb` and other eslint plugins gave new errors since new rules were introduced, these rules were either set to `off` or `warn` or fixed. The only file I didn't know how to fix is `src/util/isDomainSpecifiedByUser.ts` where I had to put disable comments on eslint rule until I figure out how to actually fix these errors.
4- the latest version for vitest deprecated CJS build so renaming `vitest.config.ts` into `vitest.config.mts` was required
5- updating webpack-cli needed an update to scripts using it, especially removing `--entry` command line option since it resulted in an error about duplicate entry point.
6- chromatic was updated to v11, it was confirmed that it works with my chromatic project but you'll probably need to update the chromatic project you use on your end to make it work.
7- The update to vitest also resulted in Typescript errors since some types changed so they were fixed.

## Motivation and Context

Keeping dependencies up-to-date removes Tech dept, ensures security vulnerabilities are fixed, and ensures that current docs applies instead of trying to navigate old docs for old versions.
## How Has This Been Tested?
I ran `npm run lint`, `npm run lint-test` , `npm run lint-storybook`, `npm test` , `npm run test-storybook`, `npm run check-types`, `npm run storybook`, `npm run build-storybook`, `npm run test-storybook:ci`, `npm run chromatic`, `npm run build`, `npm run test-build-output`, `npm run demo` to make sure all have succeeded. I ran `npm run demo` and fixed errors resulting from the upgrade from react-router-dom v5 to v6.

I also ran `npm run analyse` to see that the size of the package didn't change because of upgrade.

## Types of changes

## Checklist:


PavelVanecek:
Thank you @abnud11 ! Who are you, out of curiosity? GitHub account created in May 2024 and jumped straight into chores and upgrades in couple opensource repositories.

PavelVanecek:
The -9k lines in coverage diff is supicious. What happened there?

abnud11:
@PavelVanecek I looked at an example of one of the not covered lines, it's line 103 of `src/cartesian/ReferenceArea.tsx`

That line is not a false negative, it's truely uncovered by tests since no tests put a scale of null on `XAxis` or `YAxis`.

In my opinion, the previous version of `vitest` and its companion `@vitest/coverage-v8` reported 100% coverage which was a mistake, upgrading from v1 to v2 fixed a lot of issues in coverage report which now reports uncovered lines correctly.

This is both good and bad, it means that we thought our tests covered our code when it didn't. also it means that we need to add multiple tests to cover the new uncovered lines.

abnud11:
@PavelVanecek I'm not sure what happened, vitest did mention that on upgrade from v1 to v2, coverage report will be different since they now Exclude empty lines, Typescript Types and other things from coverage report

Maybe it's related to that, I need you to approve the workflow again to further investigate that

PavelVanecek:
Ah right that would explain. Would you mind if I ask you to create a separate PR with only that change so that we can confirm?

abnud11:
@PavelVanecek I understand you, I rollbacked vitest to v1, I will upgrade it to v2 in a later PR but then we will probably need to configure coverage threshold

abnud11:
@PavelVanecek rolled vitest back to v1, can you approve the workflows again so we can see the difference?

PavelVanecek:
https://app.codecov.io/github/recharts/recharts/commit/8a5d57e0a7b90dd9c9671057c424d0de173138ad/indirect-changes this still shows massive improvement. Yes it looks like it's skipping empty lines and TS types. Looks like the rollback didn't work.

Anyway. I think that explains the mystery of missing lines well enough.

PR #4825, opened 2024-07-29
Allow selecting ZAxis scale from redux

PavelVanecek:
## Description

This was almost there, only had to wire through the zAxisId and fix tests and types. I need to select the scale so that I can use it in the scatter points selector next.

## Related Issue


src/cartesian/Area.tsx
```diff
@@ -463,6 +463,7 @@ class AreaWithState extends PureComponent<Props, State> {
             data={this.props.data}
             xAxisId={xAxisId}
             yAxisId={yAxisId}
+            zAxisId={0}
```
ckifer, line 466: does is matter that these don't care about z-axis? Why can't we just leave it undefined?

ckifer (approved):

src/cartesian/Area.tsx
PavelVanecek: Could be undefined too. I used zero so that I don't have to change the type to `string | number | undefined`.

ckifer: sounds good, as long as it doesn't cause issues when someone adds a ZAxis to a component that doesn't accept a ZAxis

PavelVanecek: I can imagine we would want to add it later. It could be represented with a color or perhaps a thickness of a line. Until then it will be ignored.

PR #4827, opened 2024-07-29
Error bars use direction instead of layout

PavelVanecek:
## Description

Small refactor to make it easier to break the Error bar - axis domain - scatter points - error bar cyclic dependency.

## Related Issue

### DIFF
--- a/src/cartesian/ErrorBar.tsx
+++ b/src/cartesian/ErrorBar.tsx
@@ -4,9 +4,7 @@
 import React, { Component, SVGProps } from 'react';
 import Animate from 'react-smooth';
 import { Layer } from '../container/Layer';
-import { Props as XAxisProps } from './XAxis';
-import { Props as YAxisProps } from './YAxis';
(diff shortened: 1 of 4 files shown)


src/cartesian/ErrorBar.tsx
```diff
@@ -102,12 +97,12 @@ function ErrorBarImpl(props: Props) {
   }
 
   // ErrorBar requires type number XAxis, why?
-  if (props.direction === 'x' && xAxis.type !== 'number') {
+  if (direction === 'x' && xAxis.type !== 'number') {
```
ckifer, line 100: So this is a breaking change where the user now specifies X or Y instead of layout?

ckifer (approved):

src/cartesian/ErrorBar.tsx
PavelVanecek: Layout is injected through element cloning, it's not an external prop. Also it entirely depends on direction anyway.

PR #4829, opened 2024-07-30
chore: upgrade vitest to v2

abnud11:
This PR upgrades vitest to v2
## Description

Only one change currently(didn't require much), `package.json` and `package-lock.json`, two packages got updated to v2, `vitest` and `@vitest/coverage-v8`
## How Has This Been Tested?

but running `npm test` and `npm run check-types`

## Types of changes

## Checklist:

The only problem is the change in coverage report, vitest v2 enables `coverage.ignoreEmptyLines` by default, see https://vitest.dev/guide/migration#v8-coverage-s-coverage-ignoreemptylines-is-enabled-by-default

Because of that coverage is less in this PR than previous code, this is expected since we have fewer total lines now but it fails the CI, the solutions to this:
1- disable `coverage.ignoreEmptyLines`, I don't agree with this since ignoring empty lines, Typescript types, etc is very reasonable to me.
2- accept the PR even when codecov fails, not sure if that's feasible
3- do you know an option to pass to codecov to make it succeed? I searched for that, we can tell codecov a threshold but that's not the problem here, the problem is coverage is less in this PR than branch 3.x, I didn't find an option for codecov regarding that and vitest v2 change.

Let me know if you have a solution to this problem, if we go with option 2, the coverage report will change in branch `3.x` but will codecov compare future PRs to the new coverage report? I saw from Codecov docs that it updates coverage report only when CI succeeds not when CI fails.


ckifer (approved):
I can just merge the PR and then the new number will be the baseline going forwards. 

This LGTM I think, no issues ignoring empty lines. 

Thanks!

PR #4834, opened 2024-07-30
Area, Bar, Line and Scatter report their settings before rendering

PavelVanecek:
## Description

This has two benefits, one: fewer renders, two: helps to avoid the errorbar - scatter circular dependency.

## Related Issue


ckifer (approved):
Looks like some conflicts

PR #4840, opened 2024-07-31
Make scale selector stable

PavelVanecek:
## Description

This is necessary to prevent infinite re-rendering loop in Scatter.

## Related Issue


ckifer (approved):
Nice

PR #4842, opened 2024-07-31
Timescale fix

PavelVanecek:
## Description

Fixes regression https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=66aa465be316cf9b344d1a0b introduced in https://github.com/recharts/recharts/pull/4840

## Related Issue

### DIFF
--- a/src/state/selectors/axisSelectors.ts
+++ b/src/state/selectors/axisSelectors.ts
@@ -870,12 +870,20 @@ function getD3ScaleFromType(realScaleType: string | undefined) {
 }
 
 function combineScaleFunction(
+  axis: BaseAxis,
   realScaleType: string | undefined,
   axisDomain: NumberDomain | CategoricalDomain,
   axisRange: [number, number],
 ): RechartsScale | undefined {
+  if (axisDomain == null) {
+    return undefined;
+  }
+  if (typeof axis.scale === 'function') {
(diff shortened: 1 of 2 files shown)


src/state/selectors/axisSelectors.ts
```diff
@@ -870,12 +870,20 @@ function getD3ScaleFromType(realScaleType: string | undefined) {
...
   realScaleType: string | undefined,
   axisDomain: NumberDomain | CategoricalDomain,
   axisRange: [number, number],
 ): RechartsScale | undefined {
+  if (axisDomain == null) {
+    return undefined;
+  }
+  if (typeof axis.scale === 'function') {
+    // @ts-expect-error we're going to assume here that if axis.scale is a function then it is a d3Scale function
+    return axis.scale.copy().domain(axisDomain).range(axisRange);
```
ckifer, line 883: I guess we'll find out when we release if this is an issue 😅

ckifer (approved):

src/state/selectors/axisSelectors.ts
PavelVanecek: This has always been the case so it shouldn't be an issue.

PavelVanecek: Except before my changes this assumption was hidden behind an ocean of `any`s 😬

ckifer: Of course it was haha

PR #4863, opened 2024-08-07
selectScatterPoints

PavelVanecek:
## Description

Adds new selector, and uses the selector in Scatter. The old `getComposedData` function is still around and used in Tooltip, but Scatter itself does not use it anymore.

There are some breaking changes:

- Scatter doesn't work in Brush panorama anymore - I will need to fix that before we can release 3.0
- ZAxis range had this behaviour where it used to get the first digit of the range and used that as a default size in an absence of a dataKey. That is gone, if you want to change the size then give it a dataKey. In my opinion this can stay.
- Scatter used to display `null` values _somehow, somewhere_ and then Tooltip had a `filterNull` prop that further defined what happens with the nulls. Now it doesn't display nulls at all and the `filterNull` accordingly does not do anything. In my opinion, this too can stay and we should remove the `filterNull` completely. If anyone wants to display empty strings in a chart then they can pass empty strings instead of relying on Recharts to convert nulls to empty strings.

## Related Issue

This is the latest visual diff report: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1265

### DIFF
--- a/src/cartesian/Scatter.tsx
+++ b/src/cartesian/Scatter.tsx
@@ -594,7 +594,6 @@ export class Scatter extends Component<InternalProps> {
     displayedData: any[];
   }): ScatterComposedData => {
     const cells = findAllByType(item.props.children, Cell);
-    console.log('getComposedData', { xAxisTicks });
     const points: ReadonlyArray<ScatterPointItem> = computeScatterPoints({
       displayedData,
       xAxis,
--- a/src/cartesian/Scatter.tsx
+++ b/src/cartesian/Scatter.tsx
@@ -2,7 +2,7 @@
  * @fileOverview Render a group of scatters
  */
 // eslint-disable-next-line max-classes-per-file
-import React, { Component, PureComponent, ReactElement } from 'react';
+import React, { Component, PureComponent, ReactElement, useMemo } from 'react';
 import Animate from 'react-smooth';
 
 import isNil from 'lodash/isNil';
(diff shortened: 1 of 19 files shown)


ckifer:
Where would null values even render? I thought scatter only supported numbers?

But I'd say I'm in agreement with these three. 

People always surprise me with use cases for null values, but as long as we document the breaking change it shouldn't be an issue. we may get bug reports but we'll just give them a way to get back to what they had

PavelVanecek:
In categorical axis in 2.x, null becomes <empty string> category.

In Scatter it also causes all sorts of error messages with duplicate key (as the key becomes `-NaN-NaN-`, with passing invalid attributes to an SVG element ('y1 should be a number, instead received NaN'). Overall it didn't feel like something intentional so I did not aim to reproduce that behaviour.

src/cartesian/Scatter.tsx
```diff
@@ -573,16 +584,17 @@ export class Scatter extends Component<InternalProps> {
...
+    yAxis?: BaseAxisWithScale;
+    zAxis?: ZAxisWithScale;
     xAxisTicks: TickItem[];
     yAxisTicks: TickItem[];
     item: ScatterWithState;
     bandSize: number;
     displayedData: any[];
   }): ScatterComposedData => {
     const cells = findAllByType(item.props.children, Cell);
+    console.log('getComposedData', { xAxisTicks });
```
PavelVanecek, line 597:
```suggestion
```

src/state/selectors/scatterSelectors.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (
+      displayedData == null ||
+      xAxis == null ||
+      yAxis == null ||
+      xAxisTicks?.length === 0 ||
+      yAxisTicks?.length === 0
+    ) {
+      return undefined;
+    }
+    return computeScatterPoints({
```
ckifer, line 86: Will we follow the same pattern for the rest of the items or is this unique to scatter? Selector per chart type with the function that does the work exported from the component?

ckifer (approved):
LGTM, just a pattern question

src/state/selectors/scatterSelectors.ts
PavelVanecek: Yes I think that's the path forward. There might be some opportunities for code deduplication.

PR #4866, opened 2024-08-07
Update chromatic.yml to use latest action

PavelVanecek:
## Description

Chromatic now recommends to use `@latest` so let's use that: https://www.chromatic.com/docs/github-actions/#automate-chromatic-with-github-actions


PavelVanecek:
I tried searching for the changelog but couldn't find one. Here's another recommendation to use the `latest` tag: https://github.com/chromaui/action/releases

ckifer:
Sounds good

PR #4871, opened 2024-08-08
Scatter storybook docs update: remove internal props, add comment

PavelVanecek:
These props no longer do anything in 3.x branch so let's remove them from storybook too.

### DIFF
--- a/storybook/stories/API/cartesian/Scatter.stories.tsx
+++ b/storybook/stories/API/cartesian/Scatter.stories.tsx
@@ -465,49 +465,6 @@ const ReactiveProps = {
   },
 };
 
-const OtherProps = {
-  xAxis: {
-    table: {
-      category: 'Other',
-    },
-  },
-  yAxis: {
-    table: {
-      category: 'Other',
-    },
-  },
-  zAxis: {
-    table: {
-      category: 'Other',
-    },
-  },
-  top: {
-    table: {
-      category: 'Other',
-    },
-  },
-  left: {
-    table: {
-      category: 'Other',
-    },
-  },
-};
-
-const InternalProps = {
-  points: {
-    description: 'The coordinates of all the scatters.',
-    table: {
-      category: 'Internal',
-    },
-  },
-  activeIndex: {
-    description: "The active scatter's index in all the scatters. Active here refers to activeShape and the Tooltip.",
-    table: {
-      category: 'Internal',
-    },
-  },
-};
-
 const [surfaceWidth, surfaceHeight] = [600, 300];
 
 const GeneralPropsForScatter = {
@@ -526,10 +483,6 @@ export default {
     ...EventHandlers,
     ...AnimationProps,
     ...ReactiveProps,
-    ...OtherProps,
-    ...InternalProps,
-    // Deprecated
-    dangerouslySetInnerHTML: { table: { category: 'Deprecated' }, hide: true, disable: true },
   },
 };
 
--- a/storybook/stories/API/props/CartesianComponentShared.ts
+++ b/storybook/stories/API/props/CartesianComponentShared.ts
@@ -76,7 +76,7 @@ export const General: StorybookArgs = {
   yAxisId,
 };
 
-export const points: StorybookArg = {
+const points: StorybookArg = {
   description:
     'The coordinates of points in the line, usually calculated internally. In most cases this should not be used.',
   table: {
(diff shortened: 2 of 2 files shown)


storybook/stories/API/cartesian/Scatter.stories.tsx
```diff
@@ -526,10 +483,6 @@ export default {
     ...EventHandlers,
     ...AnimationProps,
     ...ReactiveProps,
-    ...OtherProps,
-    ...InternalProps,
-    // Deprecated
-    dangerouslySetInnerHTML: { table: { category: 'Deprecated' }, hide: true, disable: true },
```
ckifer, line 532: these no longer get spread on the element?

storybook/stories/API/cartesian/Scatter.stories.tsx
```diff
@@ -465,49 +465,6 @@ const ReactiveProps = {
...
-  },
-  left: {
-    table: {
-      category: 'Other',
-    },
-  },
-};
-
-const InternalProps = {
-  points: {
```
ckifer, line 497: people might mention that they use points to do some manual work but if they do that they can just use `Dot`

storybook/stories/API/cartesian/Scatter.stories.tsx
```diff
@@ -465,49 +465,6 @@ const ReactiveProps = {
...
-    description: 'The coordinates of all the scatters.',
-    table: {
-      category: 'Internal',
-    },
-  },
-  activeIndex: {
```
ckifer, line 503: will also need a way to perform the same controls as with active index (defaultIndex?)

ckifer (approved):

storybook/stories/API/cartesian/Scatter.stories.tsx
PavelVanecek: I expect that whoever needs that amount of control might also render SVG directly. I would consider adding it back as an optional prop if there is strong demand.

storybook/stories/API/cartesian/Scatter.stories.tsx
PavelVanecek: These were never external props, whatever came from the outside was overwritten by the generator. It happened to be documented as "Internal, do not use" because Recharts typescript types did not handle the difference between real props, and props coming from cloning.

Since we no longer have cloning we also don't have the type problem so we can remove this stuff from the docs.

storybook/stories/API/cartesian/Scatter.stories.tsx
PavelVanecek: The activeIndex is coming through tooltipContext, I suspect I should have removed it from the docs long ago. `defaultIndex` should propagate to the same property.

storybook/stories/API/cartesian/Scatter.stories.tsx
ckifer: yep

storybook/stories/API/cartesian/Scatter.stories.tsx
ckifer: I thought they ended up on the high level SVG. Anyways, as long as they don't exist in the types then all good

storybook/stories/API/cartesian/Scatter.stories.tsx
ckifer: defaultIndex as it exists is not the same as activeIndex. So it will either need to become like it, or we'll need another property on tooltip?

PavelVanecek: What I'm trying to say is that this prop has not been used on Scatter for a while, and is following whatever index is active in Tooltip.

storybook/stories/API/cartesian/Scatter.stories.tsx
PavelVanecek: Speaking of dangerouslySetInnerHTML in particular, let's remove it in 3.x too? It's marked as deprecated.

storybook/stories/API/cartesian/Scatter.stories.tsx
ckifer: sounds good

storybook/stories/API/cartesian/Scatter.stories.tsx
ckifer: yeah absolutely. Ideally, I'd only want to include implicit SVG props that are actually useful. Thats easier said than done though since there are so many

PR #4873, opened 2024-08-08
Axes, Reference elements, and Scatter compute their position using panorama dimensions

PavelVanecek:
## Description

This adds Brush Panorama support to Redux and fixes two visual diffs.

Latest VR: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1277

## Related Issue


ckifer (approved):
awesome way to take care of this

PR #4886, opened 2024-08-12
selectArea selector

PavelVanecek:
## Description

Area now reads its config from Redux, just like Line and Scatter.

Visual diff: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1293

## Related Issue

### DIFF
--- a/src/state/selectors/areaSelectors.ts
+++ b/src/state/selectors/areaSelectors.ts
@@ -1,0 +1,153 @@
+import { createSelector } from '@reduxjs/toolkit';
+import { Series } from 'victory-vendor/d3-shape';
+import { Coordinate, DataKey } from '../../util/types';
+import { BaseValue, computeArea } from '../../cartesian/Area';
+import { selectAxisWithScale, selectStackGroups, selectTicksOfGraphicalItem, StackGroup } from './axisSelectors';
+import { RechartsRootState } from '../store';
+import { AxisId } from '../axisMapSlice';
+import { selectChartLayout } from '../../context/chartLayoutContext';
+import { selectChartDataWithIndexes } from './dataSelectors';
+import { getBandSizeOfAxis, isCategoricalAxis, StackId } from '../../util/ChartUtils';
+import { ChartData } from '../chartDataSlice';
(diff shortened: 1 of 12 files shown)


test/chart/AreaChart.spec.tsx
```diff
@@ -128,6 +153,225 @@ describe('AreaChart', () => {
...
+      {
+        coordinate: 470,
+        index: 6,
+        offset: 0,
+        value: 'Page G',
+      },
+    ]);
+    expect(xAxisTicksSpy).toHaveBeenCalledTimes(3);
+
+    // For some reason this assertion always fails but never shows what's the difference.
```
ckifer, line 250: hm, maybe it is in a different order?

ckifer (approved):

src/state/selectors/areaSelectors.ts
```diff
@@ -0,0 +1,155 @@
...
+const selectXAxisTicks = (state: RechartsRootState, xAxisId: AxisId, _yAxisId: AxisId, isPanorama: boolean) =>
+  selectTicksOfGraphicalItem(state, 'xAxis', xAxisId, isPanorama);
+
+const selectYAxisWithScale = (state: RechartsRootState, _xAxisId: AxisId, yAxisId: AxisId, isPanorama: boolean) =>
+  selectAxisWithScale(state, 'yAxis', yAxisId, isPanorama);
+
+const selectYAxisTicks = (state: RechartsRootState, _xAxisId: AxisId, yAxisId: AxisId, isPanorama: boolean) =>
+  selectTicksOfGraphicalItem(state, 'yAxis', yAxisId, isPanorama);
+
+const selectBandSize = createSelector(
```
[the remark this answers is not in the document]
PavelVanecek: I am still understanding how all of those work right now. Once all graphical items are running on Redux I will try to DRY and reuse those selectors.

test/chart/AreaChart.spec.tsx
PavelVanecek: Different property order will show in the diff. It might be some object property that doesn't get serialised.

PR #4900, opened 2024-08-16
refactor(tooltip): return event callback from context

decovicdev:
## Description

`useMouseEnterItemDispatch`,  `useMouseLeaveItemDispatch`  and `useMouseClickItemDispatch`
return function that gets `entry` and `index` and returns a dom event callback function to make the API
simpler and fewer function definitions. 

## Motivation and Context

working on a boxplot and noted that tooltip event dispatchers could be made simpler

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -201,18 +201,13 @@ function BarBackground(props: BarBackgroundProps) {
           return null;
         }
 
-        const onMouseEnter = (e: React.MouseEvent<SVGPathElement, MouseEvent>) => {
-          // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
-          onMouseEnterFromContext(entry, i, e);
-        };
-        const onMouseLeave = (e: React.MouseEvent<SVGPathElement, MouseEvent>) => {
-          // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
-          onMouseLeaveFromContext(entry, i, e);
-        };
-        const onClick = (e: React.MouseEvent<SVGPathElement, MouseEvent>) => {
-          // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
-          onClickFromContext(entry, i, e);
-        };
+        // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
+        const onMouseEnter = onMouseEnterFromContext(entry, i);
+        // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
+        const onMouseLeave = onMouseLeaveFromContext(entry, i);
+        // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
+        const onClick = onClickFromContext(entry, i);
+
         const barRectangleProps: BarRectangleProps = {
           option: backgroundFromProps,
           isActive: i === activeIndex,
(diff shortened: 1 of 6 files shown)


src/cartesian/Bar.tsx
```diff
@@ -201,18 +201,13 @@ function BarBackground(props: BarBackgroundProps) {
           return null;
         }
 
-        const onMouseEnter = (e: React.MouseEvent<SVGPathElement, MouseEvent>) => {
```
ckifer, line 204: These were left this way on purpose so refactoring doesn't need done when we need to add to them

It's easier to have them explicitly defined than to implicitly inherit parameters

ckifer (commented):
Hi, thanks for taking initiative and making a PR. 

I personally don't think we need this change, especially while we're in progress working on 3.0 and these functions could get added to in a heartbeat. 

@PavelVanecek  thoughts?

PavelVanecek (approved):

PavelVanecek:
We're going to run into problems once we remove the two tsexpecterror because each graphical item sends different SVGElement event and the inheritance in this part is not great.

On the other hand the existing types are already not great so let's merge this and remove a few characters.

PavelVanecek:
Agreed we don't _need_ it but it's a good first contribution and we don't exactly have too many contributors so let's support that and merge <3

ckifer:
Sounds good 

Thanks @decovicdev !

ckifer (approved):

decovicdev:
thanks @PavelVanecek and @ckifer. 



- is the issue missing tooltipPayload and tooltipPosition from each graphical item type or the different SVG elements?
because events from those SVG elements are the same and SVGElement extends other svg elements

``` ts
type TooltipTriggerInfo<T extends TooltipPayloadType> = {
  tooltipPayload: T;
  tooltipPosition: Coordinate;
  cx: number;
  cy: number;
};

```

- all elements I checked implement `tooltipPayload` & `tooltipPosition` is `cx`,`cy` needed? and should `tooltipPayload` & `tooltipPosition` be included in type for each graphical item

src/cartesian/Bar.tsx
decovicdev: rather than this
```ts
(event: React.MouseEvent<SVGElement>) => {
    onMouseClickFromProps?.(data, index, event);
    onMouseClickFromContext?.(data, index, event);
    dispatch(
      setActiveClickItemIndex({
        activeIndex: String(index),
        activeDataKey: dataKey,
        activeClickCoordinate: data.tooltipPosition,
      }),
    );
  }
```

```ts
<T extends React.MouseEvent<SVGElement>> (event:T) => {
    onMouseClickFromProps?.(data, index, event);
    onMouseClickFromContext?.(data, index, event);
    dispatch(
      setActiveClickItemIndex({
        activeIndex: String(index),
        activeDataKey: dataKey,
        activeClickCoordinate: data.tooltipPosition,
      }),
    );
  }
```
so now the event can be SVGPathElement MouseEvent

ckifer: I think that might work? We definitely want to be more specific rather than less (prefer the actual element type)

decovicdev:
```ts 
<T extends SVGElement> (event:React.MouseEvent<T>) => {
    onMouseClickFromProps?.(data, index, event);
    onMouseClickFromContext?.(data, index, event);
    dispatch(
      setActiveClickItemIndex({
        activeIndex: String(index),
        activeDataKey: dataKey,
        activeClickCoordinate: data.tooltipPosition,
      }),
    );
  }
  ```
  
  `onMouseEnterFromContext<SVGPathElement>(entry,i)`  you can use inference or be very explicit

PR #4902, opened 2024-08-18
selectPieSectors, shared computePieSectors

ckifer:
## Description

selectPieSectors

For now this still uses `getComposedData` through the generator.

Write the selector and the shared `computePieSectors`, but leave using the selector/fixing tests/setting legend items without sectors for later

## Motivation and Context

- All graphical items in redux

## How Has This Been Tested?

no diff https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1326

## Types of changes

## Checklist:


src/polar/Pie.tsx
```diff
@@ -202,35 +205,35 @@ type PieComposedData = PieCoordinate & {
...
+// };
+
+// const computeLegendPayloadFromPieData = ({ sectors, legendType }: PiePayloadInputProps): Array<LegendPayload> => {
+//   if (sectors == null) {
+//     return [];
+//   }
+//   return sectors.map(
+//     (entry: PieSectorDataItem): LegendPayload => ({
+//       type: legendType,
+//       value: entry.name,
```
PavelVanecek, line 220: How hard would it be to recreate this using the raw `displayedData` instead of sectors? If I read it correctly then it might be as simple as

1.
```
value: getValueByDataKey(entry, dataKey, 0),
name: getValueByDataKey(entry, nameKey, i),
color: entry.fill,
payload: entry // this part will be breaking but w/e it was never documented anyway
```

src/polar/Pie.tsx
```diff
@@ -779,10 +753,149 @@ export class Pie extends PureComponent<InternalProps, State> {
...
+  const presentationProps = filterProps(props, false);
+
+  const sectorsAndCoord = useAppSelector(state => {
+    // @ts-expect-error result of filterProps is not PieProps, but is treated as such
+    return selectPieSectors(state, pieSettings, presentationProps, cells) ?? {};
+  });
+  const sectors = sectorsAndCoord?.sectors;
+
+  // this causes way too many re-renders
+  // useLegendPayloadDispatch(computeLegendPayloadFromPieData, { ...props, sectors });
```
PavelVanecek, line 831: Any clue what causes the re-renders? Do the sectors depend on offset?

PavelVanecek (approved):

src/polar/Pie.tsx
PavelVanecek: Ah right yes it does. Well we _could_ compute the sectors in two steps, first just the data, then feed the data (plus name plus fill color) into the legend and sector position selectors, then feed the output of sector positions into Pie.

That, or break the API that was never documented anyway and compute Legend from raw data instead of sectors. I am fan of this option because that's what I wanted to do either way. Why should we expose internal representation of the payload and pretend it's given by the user?

test/chart/PieChart.spec.tsx
```diff
@@ -201,7 +201,7 @@ describe('<PieChart />', () => {
     expect(container.querySelectorAll('.recharts-pie-sector')).toHaveLength(6);
   });
 
-  test('Renders legend when all the values are 0', () => {
+  test.skip('Renders legend when all the values are 0', () => {
```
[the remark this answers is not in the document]
ckifer: Ah, didn't know that. Was just skipping to get some eyes on it so yep I can change that if I still can't get them to pass

src/polar/Pie.tsx
ckifer: Yeah I was trying to get away without the breaking change for whatever reason. 

There's more info in entry in Pie than some other cases - for example "percentage" is a computed property that exists on sector that exists only to give more data to custom tooltip and legend

src/polar/Pie.tsx
ckifer: Yes, but I haven't quite figured out what's up there. I commented out where I pass offset in because it was causing tests to fail - I need to go back to current 3.x and see if offset is actually getting passed or not from the generator

ckifer: Agreed on the second option for the most part, was just trying to get away with the least work :p. 

The thing is, we should give the user the data (computed or otherwise) they need when they go make a custom legend. But it should be explicitly documented rather than just stuffing whatever is there into the "payload"

src/polar/Pie.tsx
ckifer: If I did this though I wouldn't be able to get the `Cell` `fill` for use in `Legend`, another breaking change which is a pretty big use-case of cell

ckifer: I think scatter is broken in this way too

src/polar/Pie.tsx
```diff
@@ -305,17 +307,15 @@ function PieSectors(props: PieSectorsProps) {
   });
 }
 
-const getRealPieData = (item: Pie): RealPieData[] => {
-  const { data, children } = item.props;
+const getRealPieData = (item: Pie, cells: ReactElement[]): RealPieData[] => {
```
PavelVanecek, line 310: I see that `cells` are being injected later, do we need to pass them as an argument here?

src/polar/Pie.tsx
```diff
@@ -391,117 +400,107 @@ const renderLabelItem = (option: PieLabel, props: any, value: any) => {
...
-    nameKey: 'name',
-    blendStroke: false,
-    rootTabIndex: 0,
+export function computePieSectors({
+  pieSettings,
+  displayedData,
+  cells,
+  offset,
+}: {
+  displayedData: any[];
```
PavelVanecek, line 409: Should this be `RealPieData[]`? Since it's expecting the result of `getRealPieData`

src/polar/Pie.tsx
PavelVanecek, line 310: I was in horror when I discovered that Recharts likes to randomly mutate arrays here and there so I am trying to switch everything to `ReadonlyArray` since to help me find those places. Is it a big change to do that here too?

PavelVanecek (approved):
Approach LGTM, I have some minor comments on types/unused params.

I wonder what happens when you do try to use that selector?

src/polar/Pie.tsx
```diff
@@ -112,7 +113,7 @@ interface InternalPieProps extends PieDef {
   hide?: boolean;
   /** the input data */
   data?: any[];
-  sectors?: PieSectorDataItem[];
+  sectors?: Readonly<PieSectorDataItem[]>;
```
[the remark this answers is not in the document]
ckifer: ah, I thought they were the same. it should be `ReadonlyArray<T>`. My version cannot be assigned to but it can be mutated with methods, the latter cant do either

src/polar/Pie.tsx
ckifer: Yes so that the cell information can be added to each sector right below this (L322 is hidden, need to expand it)

`cells` are injected later in the selector, but still need to be added to the data for `getComposedData` to continue to work

src/polar/Pie.tsx
ckifer: `RealPieData` is `any[]` so technically yes 😅

src/polar/Pie.tsx
ckifer: nope - done

ckifer:
Fixed the readonlyarray instances and added more, I'll merge this after CI and follow up after

Issue #4591, opened 2024-05-27
Add tests for ZAxis


PavelVanecek:
The only test is that it "don't render anything" which is good but we need more. The storybook has something so let's also add tests for the dot size, and tooltip payloads.

qlthd:
Interested to contribute on this issue, you can assign it to me @PavelVanecek @ckifer

ckifer:
@qlthd thank you! Assigned

qlthd:
@ckifer i think i need permission to repo to push my code

ckifer:
@qlthd we generally do fork development. Just push to your fork and open a cross-fork PR

PR #4915, opened 2024-08-22
Add tests for ZAxis #4591

dillonreedy:
Added more ZAxis tests

## Description

Tests on the following
1. Domain and range
1. Scaling
1. Children
1. Handling undefined data
1. rerender

## Related Issue

https://github.com/recharts/recharts/issues/4591

## Motivation and Context

V3.0 changes

## How Has This Been Tested?

Running our testing suite

## Types of changes

## Checklist:

### DIFF
--- a/test/cartesian/ZAxis.spec.tsx
+++ b/test/cartesian/ZAxis.spec.tsx
@@ -28,6 +28,113 @@ describe('<ZAxis />', () => {
       yAxis: 1400,
     },
   ];
+
+  it('renders with the correct number of children', () => {
+    render(
+      <Surface width={500} height={500}>
+        <ZAxis dataKey="x" name="stature" unit="cm" />
+      </Surface>,
+    );
+    const svg = document.querySelector('svg');
+    assertNotNull(svg);
+
+    expect(svg).toBeInTheDocument();
+    expect(svg.children).toHaveLength(2);
+  });
+
+  it('displays the correct tooltip content when hovering over scatter symbols', async () => {
+    const { container } = render(
+      <ScatterChart height={400} width={400}>
+        <XAxis dataKey="xAxis" type="number" />
+        <YAxis dataKey="yAxis" />
+        <ZAxis dataKey="zAxis" name="test name" range={[0, 2000]} unit="km" />
+        <CartesianGrid />
+        <Scatter data={data} name="pageData" />
+        <Tooltip />
+      </ScatterChart>,
+    );
+
+    const svg = document.querySelector('svg');
+    const gContainers = svg.querySelectorAll('.recharts-layer.recharts-scatter-symbol');
+
+    const firstShape = gContainers[0].querySelector('path');
+    await userEvent.hover(firstShape);
+    const tooltip = container.querySelector('.recharts-tooltip-wrapper');
+    expect(tooltip).toBeVisible();
+    expect(tooltip).toHaveTextContent(`xAxis : ${data[0].xAxis}yAxis : ${data[0].yAxis}test name : ${data[0].zAxis}km`);
+  });
+
+  it('should handle missing or undefined data gracefully', () => {
+    const dataWithMissingValues = [
+      {
+        name: 'Page A',
+        xAxis: 590,
+        yAxis: 1000,
+        zAxis: 400,
+      },
+      {
+        name: 'Page B',
+        xAxis: 490,
+        zAxis: undefined,
+        yAxis: 1400,
+      },
+    ];
+
+    const { container } = render(
+      <ScatterChart height={400} width={400}>
+        <XAxis dataKey="xAxis" type="number" />
+        <YAxis dataKey="yAxis" />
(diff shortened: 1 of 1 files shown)


test/cartesian/ZAxis.spec.tsx
```diff
@@ -28,6 +28,146 @@ describe('<ZAxis />', () => {
...
+      <ScatterChart width={400} height={400}>
+        <XAxis dataKey="xAxis" />
+        <YAxis dataKey="yAxis" />
+        <ZAxis dataKey="zAxis" range={[100, 1000]} />
+        <Scatter data={data} />
+      </ScatterChart>,
+    );
+
+    // Instead of accessing __data__, directly check SVG elements or use testing-library utilities
+    const zAxisTicks = container.querySelectorAll('.recharts-cartesian-axis-tick');
```
ckifer, line 43: I don't think these are necessarily z axis ticks - they're probably not since z axis doesn't have any ticks. Since z axis changes scatter size we'll have to assert that somehow

test/cartesian/ZAxis.spec.tsx
```diff
@@ -28,6 +28,146 @@ describe('<ZAxis />', () => {
...
+        <XAxis dataKey="xAxis" />
+        <YAxis dataKey="yAxis" />
+        <ZAxis dataKey="zAxis" scale="sqrt" />
+        <Scatter data={data} />
+      </ScatterChart>,
+    );
+
+    // Since we cannot directly access __data__, focus on the observable behavior
+    const symbols = container.querySelectorAll('.recharts-layer.recharts-scatter-symbol');
+    // Perform assertions based on the rendered symbols' attributes
```
ckifer, line 61: We're saying we're doing assertions based attributes but it doesn't look like we are

test/cartesian/ZAxis.spec.tsx
```diff
@@ -28,6 +28,146 @@ describe('<ZAxis />', () => {
...
+    expect(symbols.length).toBe(2);
+  });
+
+  it('renders with the correct number of children', () => {
```
ckifer, line 65: This most likely won't be doable in 3.x - surprised that it adds any children here

test/cartesian/ZAxis.spec.tsx
```diff
@@ -28,6 +28,146 @@ describe('<ZAxis />', () => {
...
+
+  it('should update the chart when data changes', () => {
+    const { rerender, container } = render(
+      <ScatterChart height={400} width={400}>
+        <XAxis dataKey="xAxis" type="number" />
+        <YAxis dataKey="yAxis" />
+        <ZAxis dataKey="zAxis" name="test name" range={[0, 2000]} unit="km" />
+        <Scatter data={data} name="pageData" />
+      </ScatterChart>,
+    );
```
ckifer, line 139: Should we assert the initial as well?

ckifer (commented):
Some comments - mostly just I think we need a good way of testing if z axis is doing what it should or not. Maybe by checking the value of the radius of each dot element or something. 


Thank you!!

test/cartesian/ZAxis.spec.tsx
```diff
@@ -28,6 +28,146 @@ describe('<ZAxis />', () => {
...
+        <CartesianGrid />
+        <Scatter data={data} name="pageData" />
+        <Tooltip />
+      </ScatterChart>,
+    );
+
+    const svg = document.querySelector('svg');
+    const gContainers = svg.querySelectorAll('.recharts-layer.recharts-scatter-symbol');
+
+    const firstShape = gContainers[0].querySelector('path');
```
PavelVanecek, line 93: Check out the test helper methods `showTooltip` and `expectTooltipPayload`, might be easier to use.

test/cartesian/ZAxis.spec.tsx
```diff
@@ -28,6 +28,146 @@ describe('<ZAxis />', () => {
...
+        <YAxis dataKey="yAxis" />
+        <ZAxis dataKey="zAxis" name="test name" range={[0, 2000]} unit="km" />
+        <Scatter data={dataWithMissingValues} name="pageData" />
+      </ScatterChart>,
+    );
+
+    const svg = container.querySelector('svg');
+    const gContainers = svg.querySelectorAll('.recharts-layer.recharts-scatter-symbol');
+
+    expect(gContainers).toHaveLength(2); // Only one valid point should be rendered
```
PavelVanecek, line 128: Assertion says length 2, comment says "only one", which is it?

PavelVanecek (changes requested):
Thanks for the PR! Coltin said it before so I'll just +1 to the sentiment: we need more expectations on what the SVG looks like, what are attribute values, x and y positions, width height everything. So on top of checking children length, please also check their attributes. That alone is not a blocker though, I would be happy to merge this because some tests are better than no tests.

However, and Coltin said that too, ZAxis does not render any ticks. So these tests are testing X and Y axis ticks. Let's fix that before merging.

ckifer:
I'm going to close this PR as stale. @dillonreedy feel free to re-open when you're able to

PR #4927, opened 2024-08-26
selectBarRectangles

PavelVanecek:
## Description

Bar reads rectangles from Redux.

This one was a bit more complicated. The bar stack size list is indexed by the DOM elements. Since we don't have DOM elements order and will not have it, we need a new key. I selected dataKey which works fine but it means that we drop support for duplicated dataKeys in chart. This decision broke some stories so I updated them but it means another breaking change for 3.0 release.

Visual diff: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1343 I added some comments in chromatic UI.

Also the panorama rendering is different now, and a fix will require large-ish refactor too. I have decided to add that in followup PR if that's okay.

## Related Issue


ckifer:
> dataKey which works fine but it means that we drop support for duplicated dataKeys in chart

Does this only mean duplicated dataKeys for Bar components or all components?

Something people will do with a composed chart is have a bar connected with a line that has the same dataKey, which I think is important to continue support for.

PavelVanecek:
Only bars. The combination you describe still works fine.

ckifer:
Ok I think that is acceptable. If someone for whatever reason wants the same Bar to exist twice they can go into custom Rectangle-land

PavelVanecek:
That, or add a duplicate property in the data array.

PavelVanecek:
Sorry for the 4k diff. I can break it down to smaller PRs if you prefer.

ckifer:
All good, going to work through it during a break in my work day

src/state/selectors/barSelectors.ts
```diff
@@ -0,0 +1,495 @@
...
+
+type BarCategory = {
+  stackId: StackId | undefined;
+  /**
+   * List of dataKeys of items stacked at this position.
+   * All of these Bars are either sharing the same stackId,
+   * or this is an array with one Bar because it has no stackId defined.
+   *
+   * This structure limits us to having one dataKey only once per stack which I think is reasonable.
+   * People who want to have the same data twice can duplicate their data to have two distinct dataKeys.
```
ckifer, line 94: What happens if someone does try to do this? Silent failure? explosion?

src/state/selectors/barSelectors.ts
```diff
@@ -0,0 +1,495 @@
...
+    axis = selectAxisWithScale(state, 'xAxis', xAxisId, isPanorama);
+    ticks = selectTicksOfGraphicalItem(state, 'xAxis', xAxisId, isPanorama);
+  } else {
+    axis = selectAxisWithScale(state, 'yAxis', yAxisId, isPanorama);
+    ticks = selectTicksOfGraphicalItem(state, 'yAxis', yAxisId, isPanorama);
+  }
+  return getBandSizeOfAxis(axis, ticks, true) ?? maxBarSize ?? 0;
+};
+
+const selectAxisBandSize = (
```
ckifer, line 191: whats the difference between bar band size and axis band size?

src/state/selectors/barSelectors.ts
```diff
@@ -0,0 +1,495 @@
...
+  const len = sizeList.length;
+  if (len < 1) return null;
+
+  let realBarGap = getPercentValue(barGap, bandSize, 0, true);
+
+  let result: ReadonlyArray<BarWithPosition>;
+  const initialValue: ReadonlyArray<BarWithPosition> = [];
+
+  // whether is barSize set by user
+  // Okay but why does it check only for the first element? What if the first element is set but others are not?
```
ckifer, line 225: Hmm, maybe this only accepts props from the higher `BarChart` so all items after should be the same as the first? idk

src/state/selectors/barSelectors.ts
```diff
@@ -0,0 +1,495 @@
...
+      sum -= (len - 1) * realBarGap;
+      realBarGap = 0;
+    }
+    if (sum >= bandSize && fullBarSize > 0) {
+      useFull = true;
+      fullBarSize *= 0.9;
+      sum = len * fullBarSize;
+    }
+
+    const offset = ((bandSize - sum) / 2) >> 0;
```
ckifer, line 242: `>>`? Is a right shift intentional here? If so, why?

ckifer (approved):
LGTM! Just some questions

src/state/selectors/barSelectors.ts
ckifer: > Right shifting any number x by 0 returns x converted to a 32-bit integer. Do not use >> 0 to truncate numbers to integers; use [Math.trunc()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#using_bitwise_no-ops_to_truncate_numbers) instead.

src/cartesian/Bar.tsx
```diff
@@ -524,24 +539,26 @@ function computeBarRectangles({
...
-  const stackedDomain = stackedData ? numericAxis.scale.domain() : null;
+  // @ts-expect-error this assumes that the domain is always numeric, but doesn't check for it
+  const stackedDomain: ReadonlyArray<number> = stackedData ? numericAxis.scale.domain() : null;
   const baseValue = getBaseValueOfBar({ numericAxis });
 
   return displayedData.map((entry, index) => {
     let value, x, y, width, height, background;
 
     if (stackedData) {
+      // @ts-expect-error d3 stack series type says it's an array, but actually it's a tuple
```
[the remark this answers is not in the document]
PavelVanecek: d3 returns an array, recharts code assumes a tuple, typescript says there's a mismatch in the `.length` property (`2` vs `number`). We can fix, later.

src/state/selectors/barSelectors.ts
PavelVanecek: I don't understand this well enough to be able to explain. I reproduced what was in the generator.

src/state/selectors/barSelectors.ts
PavelVanecek: If it's only from the root BarChart then why is this a property on every item in the sizeList? I left this comment as an opportunity for improvement.

src/state/selectors/barSelectors.ts
PavelVanecek: Copy paste from the generator.

PavelVanecek:
> Something people will do with a composed chart is have a bar connected with a line that has the same dataKey, which I think is important to continue support for.

I looked at the storybooks and I believe what you describe here is in the "Line Bar Horizontal" example and it works the same as before: 

image

src/state/selectors/barSelectors.ts
ckifer: Does it accept the same property from both BarChart and Bar and then add the former to every item if none specified on the latter?

Just making things up but that's possible

src/state/selectors/barSelectors.ts
ckifer: Oof, we should get rid of now imo if we can since it's pretty confusing. But all good, I've never seen it there before

ckifer:
Sounds good

ckifer (approved):

src/state/selectors/barSelectors.ts
PavelVanecek: Agreed but honestly it's quite low on my list of priorities. We have other things to do first imho.

ckifer: Yep no worries

PR #4932, opened 2024-08-27
Graphical items report if they are in panorama our outside

PavelVanecek:
## Description

Usually graphical elements render the same regardless of panorama or not, but Bar and barSizeList means that each new bar will collide with all other bars. So we need to know if it's in panorama or not so that main area does not squish panorama bars and vice versa.

## Related Issue

Followup from https://github.com/recharts/recharts/pull/4927 where I broke this thing and now I am fixing it.

Also https://github.com/recharts/recharts/issues/4583


ckifer (approved):
ahh I was wondering why that panorama story changed

PR #4942, opened 2024-08-29
Split Radar into three components so that we can use hooks.

PavelVanecek:
## Description

No logic changes here, only broke down the one big class.

## Related Issue

### DIFF
--- a/test/component/Legend.spec.tsx
+++ b/test/component/Legend.spec.tsx
@@ -2480,7 +2480,6 @@ describe('<Legend />', () => {
       const { container } = render(
         <RadarChart width={500} height={500} data={numericalData}>
           <Legend />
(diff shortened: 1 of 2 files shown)


test/component/Legend.spec.tsx
```diff
@@ -2480,7 +2480,6 @@ describe('<Legend />', () => {
       const { container } = render(
         <RadarChart width={500} height={500} data={numericalData}>
           <Legend />
-          {/* @ts-expect-error TypeScript correctly points out that dataKey is required but I want a test for this anyway */}
```
ckifer, line 2483: no longer required?

ckifer (approved):

test/component/Legend.spec.tsx
PavelVanecek: Not sure why. I will change Radar props later so I will take a look.

PR #4946, opened 2024-08-30
refactor(Pie): removed unused functions from Pie

ckifer:
## Description

Removes unused functions in Pie

(add Legend to Scatter story to see Tooltip issue)

## Motivation and Context

- remove dead code

## How Has This Been Tested?

- everything else still works

## Types of changes

## Checklist:


PavelVanecek (approved):
How come that getComposedData is not used? I thought the generator still calls it and caches results.

ckifer:
It only calls it if it exists. Once it no longer exists the generator returns no formattedItems. 

Everything else is still happening up until that point though. 

I don't think the results need cached unless they're being used for something else? The only thing I can think of is customized and we're going to break that anyways

Issue #4858, opened 2024-08-06
Recharts 2.13.0-alpha.4 - Encountered two children with the same key


jzombie:
## What is the current behavior?

Experiencing a "Warning: Encountered two children with the same key" error when using the `PieChart` / `Pie` combination, that I wasn't previously experiencing on the non-alpha version.

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar (template: https://codesandbox.io/p/sandbox/simple-line-chart-v25dl4).

![Screenshot 2024-08-06 at 3 53 28 PM](https://github.com/user-attachments/assets/6de8d9cd-680f-4dc5-8766-eab67a4cccb9)

```ts
import React from "react";

import {
  Cell,
  LabelList,
  Pie,
  PieChart,
  PieLabelRenderProps,
  ResponsiveContainer,
  Tooltip,
} from "recharts";

import useStoreStateReader from "@hooks/useStoreStateReader";

import formatByteSize from "@utils/formatByteSize";

// TODO: Centralize somewhere else
const COLORS = ["#8884d8", "#82ca9d", "#ffc658", "#ff8042", "#00C49F"];

// Define a type for the data entries
interface CacheDetail {
  key: string;
  size: number;
}

// Define a type for the props used in the customized label
interface CustomizedLabelProps extends PieLabelRenderProps {
  cx: number;
  cy: number;
  midAngle: number;
  innerRadius: number;
  outerRadius: number;
  percent: number;
  index: number;
  payload: CacheDetail;
}

const renderCustomizedLabel = (props: CustomizedLabelProps) => {
  const { cx, cy, midAngle, innerRadius, outerRadius, percent, payload } =
    props;
  const RADIAN = Math.PI / 180;
  const radius = 25 + innerRadius + (outerRadius - innerRadius);
  const x = cx + radius * Math.cos(-midAngle * RADIAN);
  const y = cy + radius * Math.sin(-midAngle * RADIAN);

  // Show label if space allows
  const isSpaceAvailable = percent > 0.05; // Adjust this threshold as needed

  if (isSpaceAvailable) {
    return (
      <text
        x={x}
        y={y}
        fill="white"
        textAnchor={x > cx ? "start" : "end"}
        dominantBaseline="central"
      >
        {`${payload.key}: ${formatByteSize(payload.size)}`}
      </text>
    );
  }
  return null;
};

export default function CachePieChart() {
[...]

PR #4958, opened 2024-09-03
Pie: fix duplicate key issue

ckifer:
## Description

https://github.com/recharts/recharts/issues/4858

see PR for R19 fixes https://github.com/recharts/recharts/commit/e57c4d52b8114007cb104ef597bfa6057201743d
`key` was added before spreading props. In reality, these components always had static keys. So keep them static to prevent issues liked the linked ones. In 3.x we should probably warn if key is included with the name "key" since its a reserved word

## Related Issue

https://github.com/recharts/recharts/issues/4858

## Motivation and Context

Solve this issue in master

## How Has This Been Tested?

- npm test
- manually verify the key error goes away

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -257,7 +257,7 @@ export class Area extends PureComponent<Props, State> {
     } else {
       const className = clsx('recharts-area-dot', typeof option !== 'boolean' ? option.className : '');
       const { key, ...rest } = props;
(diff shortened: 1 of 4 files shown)


ckifer:
these keys used to be static, now they are not. Making them static again (in master).

Shouldn't be a huge deal, going to merge this

Issue #4558, opened 2024-05-20
Support React 19


justinh00k:
## Do you want to request a feature or report a bug?
Bug

## What is the current behavior?
Charts do not render when using React 19, or when using NextJS 14.3 (which uses React 19).

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar (template: https://codesandbox.io/p/sandbox/simple-line-chart-v25dl4).
Install react@rc or react@beta
Observe charts do not render
https://github.com/justinh00k/recharts-react19-repro

## What is the expected behavior?
Charts render

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
2.12.7 / No

ckifer:
React 19 will be added to peerDependencies when recharts supports it. 

Please see https://github.com/recharts/recharts/pull/4542 and https://github.com/recharts/recharts/issues/3615

ckifer:
Going to edit the title if you don't mind to just "Support React 19", and use this to track support in 2.x

ckifer:
Alpha release - https://www.npmjs.com/package/recharts/v/2.13.0-alpha.0
This BREAKS things in R19, shouldn't in R18

use this to determine what is breaking/what isn't. I just don't have bandwidth to fix it myself


Things that I know are broken:

Bar
Legend
Cursor
Pie

ckifer:
https://www.npmjs.com/package/recharts/v/2.13.0-alpha.1
Should have more R19 support if not all of it (minus the need to disregard peerDeps and install react-is@19)

Please test this one as well on all React versions 16.8-19 to see if anyone runs into issues. Thanks! 🚀

ckifer:
In order to use React 19 one must override the `react-is` version. This will always be true in 2.x. 3.x we will add react-is to peerDeps or remove it entirely
```
  "overrides": {
    "react-is": "^19.0.0-beta-26f2496093-20240514"
  }
```

rauchg:
appreciate your work @ckifer

ckifer:
Appreciate next helping out a bit here @rauchg !

juliusmarminge:
Just tried on Next.js RC (React 19.0.0-rc-f994737d14-20240522) and my beautiful charts are back. Thank you 🙏

rauchg:
[...]

PR #4970, opened 2024-09-05
fix(types): return JSX in a function instead of a class component

ckifer:
… so generator type can be inferred correctly

## Description

In 3.x we return a function component from the generator function which results in the return type being `React.JSX`. In 2.x we return a class component which has a complex return type that R19 does not like. Affects all generator charts
`PieChart cannot be used as a JSX component...`, etc.

This change does nothing but return the generator class wrapped in another function, just like 3.x but without any of the additional work of 3.x

## Related Issue

https://github.com/recharts/recharts/issues/4558#issuecomment-2330578162

## Motivation and Context

- 2.x is usable when R19 comes out before 3.x

## How Has This Been Tested?

- pre-change get errror in code with R19
- post change error is resolved in code with R19
- tests pass

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1097,7 +1097,7 @@ export const generateCategoricalChart = ({
     };
   };
 
-  return class CategoricalChartWrapper extends Component<CategoricalChartProps, CategoricalChartState> {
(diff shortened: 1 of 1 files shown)


ckifer:
All of the generator entry point tests are unit testing it 😅. But yeah probably.. oh well I confirmed it's fixed both places.

ckifer:
true that

PR #4973, opened 2024-09-05
Polar axis slice and basic selector support

PavelVanecek:
## Description

Not used anywhere yet but will be needed for Radar selectors in the next PR.

## Related Issue


PavelVanecek:
The names are not great, now AngleAxis and RadiusAxis extend CartesianAxis even though they're not cartesian. I will fix later.

src/polar/PolarAngleAxis.tsx
```diff
@@ -44,12 +47,27 @@ export interface PolarAngleAxisProps extends PropsInjectedFromRedux {
   reversed: boolean;
   dataKey?: DataKey<any>;
   tick?: SVGProps<SVGTextElement> | ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>) | boolean;
+  scale: ScaleType | RechartsScale;
+  type?: 'category' | 'number'; // so there is code that checks if angleAxis.type is number but it actually never behaves as a number
```
ckifer, line 51: Since we're hardcoding this can we just remove it?

ckifer (approved):

src/polar/PolarAngleAxis.tsx
PavelVanecek: If I remove it then various other places in the ChartUtils break. I'll leave it here for now.

PR #4979, opened 2024-09-06
Polar axis refactor to use components instead of inline functions

PavelVanecek:
## Description

Use Components instead of functions so that I can put hooks in it later.

## Related Issue


ckifer (approved):
We didn't really do this for any of the other components, what hooks will you need to use in "Ticks"?

PavelVanecek:
Some of them yes but not all. I will use different selectors inside, one for tick positions another for tick values. Maybe.

PR #4981, opened 2024-09-10
Feat: add cutoff link option for sankey chart

WooJunKang:
Added an option for cutoff link for sankey chart. 
Default value of this option is false. So, if you don't turn this option on, you can use it as is.

## Description
Currently, if there is a link that is cut in the middle, it will be located at the last depth.
I would like to add an option that allows nodes to be positioned cut off in the middle rather than at the very end.

For more detail, please refer to the screenshot section below.
I updated storybook stories so you can check in storybook also.

## How Has This Been Tested?
please check in storybook.

|as-is|to-be|
|----|----|
|<image width=400 src="https://github.com/user-attachments/assets/ed3e8294-053b-4d5b-b3cc-d047943fe21b" />|<image width=400 src="https://github.com/user-attachments/assets/6636d7f4-c2c5-4dce-9cc1-688fdc9f036d" />

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/Sankey.tsx
+++ b/src/chart/Sankey.tsx
@@ -95,6 +95,7 @@ const getNodesTree = (
   { nodes, links }: SankeyData,
   width: number,
   nodeWidth: number,
+  allowCutoff: boolean,
 ): { tree: SankeyNode[]; maxDepth: number } => {
   const tree = nodes.map((entry: SankeyNode, index: number) => {
     const result = searchTargetsAndSources(links, index);
@@ -121,7 +122,7 @@ const getNodesTree = (
     for (let i = 0, len = tree.length; i < len; i++) {
       const node = tree[i];
 
-      if (!node.targetNodes.length) {
+      if (!node.targetNodes.length && !allowCutoff) {
         node.depth = maxDepth;
       }
       node.x = node.depth * childWidth;
@@ -279,6 +280,7 @@ const computeData = ({
   nodeWidth,
   nodePadding,
   sort,
+  allowCutoff,
 }: {
   data: SankeyData;
   width: number;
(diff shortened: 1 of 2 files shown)


ckifer:
Hi, thanks for the contribution!

Can you explain the use case for this a little bit? What do you mean by "cut in the middle"? What happens when there are more links and nodes?

Can you add an example to the storybook that represents this all the time? Or that represents it in a more complicated way?

Also could you add some tests? Thanks!

WooJunKang:
Hi, Sorry for late response.

_(If the name of title`allowCutoff` is confusing, please recommend a good option name. will change it.)_

I just wanted to use the logic change the depth of node to max depth optionally. ( `node.depth = maxDepth;` )
I intend to "cut in the middle" of meaning is not use this logic. so the node does not positioned at max depth.

In the example chart below, The node highlighted in orange color is not located at the second depth, but at the last depth (0->2). This is due to the logic inside the Recharts-Snakey that changes the depth of the node to the max depth of the chart.

Screenshot 2024-09-16 at 2 52 09 PM


Although this view is common in  Sankeychart, The designers on my team and I have a need to use something like the image below. 

Screenshot 2024-09-16 at 3 07 48 PM

ckifer:
@WooJunKang no problem. This makes sense, but this is not standard in Sankey diagram creation. Understood that you have requirements to fulfill, but could you walk me through a situation where this makes more sense than extending it to the end like most other sankey diagrams do? Supporting edge cases is tricky business that I'd like to avoid if possible...

I don't see any other charting libraries that support this, do you know of any and could provide an example? Maybe their API has a good name for this? imo `allowCutoff` is a confusing name that doesn't really relate to what is happening.

something like `noLinkTargetDepth: 'max' | 'min'` might work with a really good explanation of what it does in the docs? 
Just throwing things out there, but overall I'm hesitant to merge this.

Also need some unit tests if we're to consider it

ckifer:
Going to close this as stale

PR #4985, opened 2024-09-11
Separate Funnel types internal and external

ckifer:
## Description

Small change getting Funnel ready to move to the current selector pattern. Separate internal and external types, allow them all to extend eachother temporarily until selector is in place, update references

* Since all our other components call their prop types "Props" I did the same here...its just less confusing for the patterns sake.

## Motivation and Context

move forwards

## How Has This Been Tested?

run build, tests passing

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -49,6 +49,7 @@ export type BaseValue = number | 'dataMin' | 'dataMax';
  */
 interface InternalAreaProps {
   activeDot: ActiveDotType;
+  // TODO: this isn't used
   animateNewValues?: boolean;
   animationBegin: number;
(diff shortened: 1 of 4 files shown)


src/cartesian/Area.tsx
```diff
@@ -49,6 +49,7 @@ export type BaseValue = number | 'dataMin' | 'dataMax';
  */
 interface InternalAreaProps {
   activeDot: ActiveDotType;
+  // TODO: this isn't used
```
ckifer, line 52: this param is only used in `Line` - no other used references anywhere in the project (or react-smooth). Probably good to remove before 3.0

PavelVanecek (approved):

PR #5008, opened 2024-09-16
Radar reports settings to redux store

PavelVanecek:
## Description

Same as other graphical items do

## Related Issue


ckifer (approved):
Hope you're feeling better!

Issue #4551, opened 2024-05-19
Remove Customized component


PavelVanecek:
Following up from https://github.com/recharts/recharts/discussions/4437:

Delete Customized component without replacement.

Once we fix #2788, one of the two reasons for its existence will disappear! For the second one there is https://github.com/recharts/recharts/issues/4552

ckifer:
Removing customized breaks this https://github.com/recharts/recharts/blob/3.x/test/util/context.tsx#L24
Which breaks a lot of tests :/

PavelVanecek:
So https://github.com/recharts/recharts/issues/2788. `Customized` is still in 3.0 but it doesn't receive props or state anymore.

PR #5021, opened 2024-09-17
refactor: remove getComposedData from all charts that are on selectors

ckifer:
## Description

- removes all removable `getComposedData` references
- breaks `Customized`

## Related Issue

https://github.com/recharts/recharts/issues/4551

## Motivation and Context

Keep...moving...

## How Has This Been Tested?

- npm test
- customized breaks in storybook
- no chart diff

## Types of changes

## Checklist:


test/cartesian/Area.spec.tsx
```diff
@@ -892,262 +892,3 @@ describe('getBaseValue', () => {
     });
   });
 });
-
-describe('getComposedData', () => {
```
PavelVanecek, line 896: The function and the logic is still there. Not in a static method but in a function. Can we keep the tests?

test/chart/BarChart.spec.tsx
```diff
@@ -599,27 +599,15 @@ describe('<BarChart />', () => {
     });
 
     test('Stacked bars are actually stacked', () => {
-      let seriesOneBarOneValue, seriesTwoBarOneValue;
-      const Spy = (props: { formattedGraphicalItems?: any }) => {
```
PavelVanecek, line 603: Can we replace this with selectors?

test/state/selectors/scatterSelectors.spec.tsx
```diff
@@ -33,23 +33,26 @@ describe('selectScatterPoints', () => {
...
-    render(
-      <PieChart width={100} height={200}>
-        <Pie data={pageData} dataKey="uv" />
-        <Customized component={<Comp />} />
-      </PieChart>,
-    );
-    expect(scatterPointsSpy).toHaveBeenCalledWith(undefined);
-    expect(scatterPointsSpy).toHaveBeenCalledTimes(2);
-  });
+  // TODO: unsure why this does not work... does PieChart prevent rendering for some reason?
```
PavelVanecek, line 36: What happens here? The selector should continue returning undefined

test/cartesian/Area.spec.tsx
ckifer: Yeah I can do the same for the compute function I guess

test/state/selectors/scatterSelectors.spec.tsx
ckifer: The selector is never called and I don't know why

test/chart/BarChart.spec.tsx
ckifer: Maybe but I've been having trouble with the selectors in tests

ckifer: done

test/state/selectors/scatterSelectors.spec.tsx
ckifer: I added `it.fails` and we can revisit. Nothing is rendering children in Pie specifically

We should allow children to render children at the chart level I guess..

ckifer:
Added 2 of 3 groups of tests back, going to merge this

PR #5025, opened 2024-09-20
cursor: retrieve index from redux instead of context

ckifer:
## Description

- add index as a prop
- pass it down from tooltip redux selectors. Harder to call the selector directly because the selector requires a lot of props from Tooltip that cursor doesn't have....

I know we already have a selector for this but why not return more in a single selector instead of calling another

## Motivation and Context

- move away from context
- a TODO

## How Has This Been Tested?

- npm test, no visual diff
- updated tests

## Types of changes

## Checklist:


PavelVanecek (approved):
Nice! I'm just going to leave a reference here: https://github.com/recharts/recharts/issues/4549

PR #5029, opened 2024-09-21
PolarAngleAxis reads settings and ticks from Redux

PavelVanecek:
## Description

One less use of generator. No new visual diff: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1469

## Related Issue


ckifer (approved):
🚀

PR #5082, opened 2024-10-06
Polar graphical elements now send their stack props to store

PavelVanecek:
## Description

To write RadialBar selectors I need this information in Polar item store.

## Related Issue


ckifer (approved):

ckifer:
Looks like something's failing now that both the PRs are merged

PR #5090, opened 2024-10-07
Make radarSelectors stable

PavelVanecek:
## Description

So that I don't need to scroll past the super long warning. Also it improves rendering performance as a side effect.

## Related Issue


ckifer (approved):
Nice!

Issue #5169, opened 2024-10-29
Option to hide specific values from tooltip


Cameron-Asdf:
## Do you want to request a feature or report a bug?

Feature request.

## What is the current behavior?

In a tooltip's formatter there is no way to format a value out of existance.
You can return `null` for both the name and the value, but that still leaves a bit of a gap where the value should've been.

## What is the expected behavior?

It should be possible to return something like null or undefined to indicate that this value should be omitted entirely.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

I think it's all of them.

ckifer:
You can make a custom Tooltip which should allow you to return whatever content (or lack of content) that you'd like

Cameron-Asdf:
> You can make a custom Tooltip which should allow you to return whatever content (or lack of content) that you'd like

Yes, but I'd also need to recreate all of the features of the default tooltip.

ckifer:
Why is that? You can provide a component to the content prop of the Tooltip component and it should give you all of the features/data. You just need to specify how you want that data to be rendered. I'm on my phone but I'll send an example in a bit

ckifer:
https://release--63da8268a0da9970db6992aa.chromatic.com/?path=/story/examples-tooltip--custom-content-example

Pasting this from my phone but this should be the right example

ckifer:
PR for this was merged

Cameron-Asdf:
> Why is that? You can provide a component to the content prop of the Tooltip component and it should give you all of the features/data. You just need to specify how you want that data to be rendered. I'm on my phone but I'll send an example in a bit

Sorry, I got busy.
This worked, thanks.
Though it also feels like a roundabout way of doing this so I'm glad the change got merged in now.

PavelVanecek:
Hello everyone who's watching this issue! We have released an alpha version `3.0.0-alpha.2` which fixes this problem.

Here is a migration guide: https://github.com/recharts/recharts/wiki/3.0-migration-guide

As of now we have fixed all the new 3.x bugs we are aware of so, and are ready for your feedback. Please feel free to try `recharts: 3.0.0-alpha.2` in a test build and let us know how it went!

PR #5171, opened 2024-10-29
Add ability to hide individual values in tooltip via the formatter

Cameron-Asdf:
## Description

This makes it so that returning null or undefined from a `Tooltip`'s formatter makes it not show that entry.

The `finalItemStyle` bit in the code was moved down to prevent unnecessary allocation.

## Related Issue

#5169 

## Motivation and Context

This is useful for things like stacked bar charts, where, on every column, there might be a lot of values that are all 0, with only a few non-zero ones.
In such cases, anything that is 0 on a particular column can now just be omitted from the tooltip to save on space and to not display heaps of useless information.

## How Has This Been Tested?

<s>To be completely honest, it hasn't.</s>
It is assumed to be bug-free, based on similarity to [this change](https://github.com/recharts/recharts/commit/73f367f8a418e0937a9be7a26d61f471f6444bdf#diff-9106812a15872ff214861c1aee964d50e31e7fe20a8bc7e2911dbcc9855d07e1R55).

## Types of changes

## Checklist:

### DIFF
--- a/src/component/DefaultTooltipContent.tsx
+++ b/src/component/DefaultTooltipContent.tsx
@@ -82,13 +82,6 @@ export const DefaultTooltipContent = <TValue extends ValueType, TName extends Na
           return null;
         }
 
-        const finalItemStyle = {
-          display: 'block',
-          paddingTop: 4,
-          paddingBottom: 4,
-          color: entry.color || '#000',
-          ...itemStyle,
-        };
         const finalFormatter = entry.formatter || formatter || defaultFormatter;
         const { value, name } = entry;
         let finalValue: React.ReactNode = value;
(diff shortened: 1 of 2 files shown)


ckifer (changes requested):
This is a breaking change which cannot be merged to the master branch. I'm willing to consider it for 3.x but there are most likely individuals depending on the fact that this does not return null at the moment. 

Also would need some unit tests.

src/component/DefaultTooltipContent.tsx
```diff
@@ -97,11 +90,21 @@ export const DefaultTooltipContent = <TValue extends ValueType, TName extends Na
           const formatted = finalFormatter(value, name, entry, i, payload);
           if (Array.isArray(formatted)) {
             [finalValue, finalName] = formatted;
-          } else {
+          } else if (formatted === null || formatted === undefined) {
```
ckifer, line 93: nit: this can just be `formatted == null` to cover both of these

ckifer (changes requested):

ckifer:
I'm going to close this as stale due to no response, if you'd like to open the same PR for the 3.x branch I can take a look

ckifer:
https://github.com/recharts/recharts/pull/5219

PR #5185, opened 2024-10-31
Custom angled axis fix

PavelVanecek:
## Description

I noticed a regression in 3.x where custom startAngle and endAngle did not produce the expected result so here is a fix.

For comparison here is codesandbox with 2.x behaviour: https://codesandbox.io/p/sandbox/elated-leakey-vfrxvv?workspaceId=7a1754d3-b261-47b1-9339-2aca93020961

## Related Issue


ckifer (approved):
Nice. Easy fix at least!

PR #5193, opened 2024-11-02
Add more tests for RadialBar, fix radius axis offset

PavelVanecek:
## Description

This fixes a visual difference in inner and outer radius. This is the diff _when using the selector_: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1623 you will see that the domain with implicit axes is still wrong but the radius is correct.

The actual selector is not used yet (I did not commit that part) - next PR hopefully.

## Related Issue

### DIFF
--- a/test/polar/RadialBar.spec.tsx
+++ b/test/polar/RadialBar.spec.tsx
@@ -2,15 +2,19 @@ import React from 'react';
 import { render } from '@testing-library/react';
 import { describe, it, test, expect, vi } from 'vitest';
 import userEvent from '@testing-library/user-event';
-import { Customized, PolarAngleAxis, PolarRadiusAxis, RadialBar, RadialBarChart } from '../../src';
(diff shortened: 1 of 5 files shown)


test/polar/RadialBar.spec.tsx
```diff
@@ -673,6 +677,239 @@ describe('<RadialBar />', () => {
...
+        unit: undefined,
+      });
+    });
+
+    it('should select angle axis realScaleType', () => {
+      const { spy } = renderTestCase(state => selectRealScaleType(state, 'angleAxis', 0));
+      expect(spy).toHaveBeenLastCalledWith('linear');
+    });
+
+    it('should select angle axis scale', () => {
```
PavelVanecek, line 741: The test helpers made it much easier to make sense of which test is failing

ckifer (approved):

PR #5195, opened 2024-11-03
Radial chart uses different implicit angle axis

PavelVanecek:
## Description

I think this is a horrible change from an API design perspective and I would like to change that in 4.0 but for now I am trying to not change how the generator behaves too much so here we are.

RadialBar still doesn't render 100% the same but it's closer now. Next I need to look at `niceTicks`.

## Related Issue

### DIFF
--- a/src/state/selectors/polarAxisSelectors.ts
+++ b/src/state/selectors/polarAxisSelectors.ts
@@ -47,8 +47,33 @@ export const implicitRadiusAxis: RadiusAxisSettings = {
   unit: undefined,
 };
 
-export const selectAngleAxis = (state: RechartsRootState, angleAxisId: AxisId): AngleAxisSettings | undefined =>
-  state.polarAxis.angleAxis[angleAxisId] ?? implicitAngleAxis;
+const implicitRadialBarRadiusAxis: RadiusAxisSettings = {
+  allowDataOverflow: defaultPolarRadiusAxisProps.allowDataOverflow,
+  allowDecimals: false,
+  allowDuplicatedCategory: defaultPolarRadiusAxisProps.allowDuplicatedCategory,
+  dataKey: undefined,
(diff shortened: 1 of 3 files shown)


ckifer:
I'm kinda still confused on what is happening. The implicit angle axis has the wrong default settings?

ckifer (approved):

PavelVanecek:
The generator, when it computes the implicit domain, never considers the categorical axis. So all implicit domains are numerical.

Also the default type of Angle axis is category.

Both of these are IMHO bugs rather than decisions but also both are in 2.x releases and so I'm trying to stick to it. Later I will propose to fix this behaviour but that will be another breaking change. Perhaps for 4.0

PavelVanecek:
What IMHO should happen is that the chart layout should dictate both the implicit axis type and default axis type.

What we can do is to add a special "auto" axis type that behaves this way, maybe in 3.1 release. And then switch the default in 4.0

ckifer:
Ah so the type of angle axis isn't taken into effect essentially unless it is explicitly set. Fun

ckifer:
Make sense. Thanks for digging into this madness

PavelVanecek:
> the type of angle axis isn't taken into effect essentially unless it is explicitly set

It's worse:

1. `<PolarAngleAxis type='number' />` becomes `type=number` - this is the only one that's reasonable
2. `<PolarAngleAxis />` becomes `type=category` (this should be different based on the layout)
3. `(no element)` becomes `type=number` (this should be whatever the version 2 is doing)

PavelVanecek:
I suspect that whoever was changing `getAxisMap` updated one of `getAxisMapByAxes` and `getAxisMapByItems` and forgot to update the other. Not going to happen with the selectors again but until that's done we have to live with this.

ckifer:
Gross got it. Hopefully the selectors will make it more clear for future us when we want to make a change..

PR #5198, opened 2024-11-04
Recharts Hook Inspector

PavelVanecek:
## Description

I find this useful to quickly inspect the internal state of a storybook. I usually go and write a unit test but unless I know what exactly is wrong, that takes a long time to pin down. The inspector is great for quick overview of what's going on.

Also later once we make hooks part of the public API, we can use this to show the hook values. For now I put down selectors, once we have hooks we can replace that.

The visual design is ... sub-standard. I didn't pay much attention to it, we can iterate.

image

The inspector is turned off by default so that it doesn't get to visual diff, you can turn it on with the small eye icon in the toolbar:

image

## Related Issue


storybook/stories/Examples/RadialBarChart.stories.tsx
```diff
@@ -29,12 +42,13 @@ export const SimpleRadialBarChart = {
...
 
 export const RadialBarWithColors = {
-  render: (args: Record<string, any>) => {
+  render: (args: StorybookArgs, context: StoryContext) => {
     return (
       <RadialBarChart {...args}>
         <RadialBar dataKey="pv" />
         <Legend />
         <Tooltip />
+        <Customized component={<RechartsHookInspector portalRef={context.portalRef} />} />
```
PavelVanecek, line 51: I was searching for an API that will allow me to attach the inspector everywhere without changing the stories but I couldn't find a way. This appears to be the minimum possible change unless we want to rewrite the AST on the fly which I would rather avoid.

ckifer: This will break soon though right?

PavelVanecek: The breaking part is that we're going to remove the `props` and `state`, we don't need to remove the `Customized` component itself. It will be a "just render component" useless appendix but it can stay there for a while. The inspector doesn't use props or state so it should be fine.

ckifer: Hm interesting, I thing we were removing it entirely. I guess no harm in it being there but it might confuse people if it's kept in 3.0

PavelVanecek: We can do that too and migrate all the usages with codemod.

ckifer (approved):

PR #5211, opened 2024-11-06
Customized renderAsIs

PavelVanecek:
## Description

This is a breaking change because Customized and components inside it no longer receive state or props.

## Related Issue


ckifer (approved):
🚀

PR #5214, opened 2024-11-06
refactor(Sankey): move Sankey tooltip state to redux

ckifer:
## Description

move Sankey tooltip state to redux. Little more complicated than the others, not my best work

- move calculating node and link props to the getDerivedStateFromProps
- set node and link prop arrays to state and make the payload off of those
- misc changes

Note that the tooltip fallback is still in place in the Tooltip component

## Related Issue

## How Has This Been Tested?

- unit tests pass
- node and link both show tooltip when fallback is removed

image

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/Treemap.tsx
+++ b/src/chart/Treemap.tsx
@@ -18,7 +18,6 @@ import { Global } from '../util/Global';
 import { findChildByType, validateWidthHeight, filterProps } from '../util/ReactUtils';
 import { AnimationDuration, AnimationTiming, DataKey, Margin } from '../util/types';
 import { ReportChartMargin, ReportChartSize } from '../context/chartLayoutContext';
-import { TooltipContextValue } from '../context/tooltipContext';
 import { CursorPortalContext, TooltipPortalContext } from '../context/tooltipPortalContext';
 import { RechartsWrapper } from './RechartsWrapper';
 import {
@@ -507,16 +506,6 @@ function getTooltipEntrySettings({
   };
 }
 
-function createTooltipPayload(activeNode: TreemapNode) {
-  return [
-    {
-      payload: activeNode,
-      name: activeNode.name,
-      value: activeNode.value,
-    },
-  ];
-}
-
 // Why is margin not a treemap prop? No clue. Probably it should be
 const defaultTreemapMargin: Margin = {
   top: 0,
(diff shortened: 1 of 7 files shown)


src/chart/Treemap.tsx
```diff
@@ -18,7 +18,6 @@ import { Global } from '../util/Global';
 import { findChildByType, validateWidthHeight, filterProps } from '../util/ReactUtils';
 import { AnimationDuration, AnimationTiming, DataKey, Margin } from '../util/types';
 import { ReportChartMargin, ReportChartSize } from '../context/chartLayoutContext';
-import { TooltipContextValue } from '../context/tooltipContext';
```
ckifer, line 21: removing dead code in Treemap

src/state/chartDataSlice.ts
```diff
@@ -22,6 +22,11 @@ export type AppliedChartData = ReadonlyArray<{ value: unknown }>;
 
 export type ChartDataState = {
   chartData: ChartData | undefined;
+  /**
```
ckifer, line 25: bleh this could be better. Not sure if we even want it for all the charts?.... idk

PavelVanecek: This is derived data, we should not store it explicitly unless necessary. Other charts use a selector instead.
The data goes through a lifecycle:

1. When it's the first, raw version, it's `data`
2. After zoom (dataStartIndex + dataEndIndex) it becomes `displayedData`
3. After processing by charts (applied dataKey plus whatnot) it becomes `appliedValues`

Can we stick to this pipeline?

If it's necessary to save some of these to the store then we can do that but usually it's not.

If you want to inspect the results of a selector, use unit test or the inspector in storybook

PavelVanecek: Why not store derived data? Because it will get out of sync. We're seeing that in the current generator: there are so many caches and they are all inconsistent with each other.

src/chart/Treemap.tsx
PavelVanecek: It's easier to review stuff when changes like this are isolated 😬

PavelVanecek (approved):
I approve since it gets us closer to the generator-free world but I would really like to avoid storing the data again if we can avoid it.

ckifer:
Feel free to take a stab at it, I didn't want to spend any more time on it

PR #5217, opened 2024-11-07
refactor/fix(ResponsiveContainer): remove display name and isElement checks

ckifer:
## Description

A bunch of issues stemmed from adding these `isElement` checks to recharts in responisve container. Remove it all, let everything render. If you use ResponsiveContainer incorrectly, you will get incorrect results and that is your fault

## Motivation and Context

- less problems, more rendering
- do a breaking change

## How Has This Been Tested?

npm test, run storybook

still working
image

## Types of changes

## Checklist:

### DIFF
--- a/src/component/ResponsiveContainer.tsx
+++ b/src/component/ResponsiveContainer.tsx
@@ -14,7 +14,6 @@ import React, {
 import throttle from 'lodash/throttle';
 import { isPercent } from '../util/DataUtils';
 import { warn } from '../util/LogUtils';
-import { getDisplayName } from '../util/ReactUtils';
 
 export interface Props {
   aspect?: number;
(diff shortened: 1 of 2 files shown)


PavelVanecek (approved):
What would it take to remove ResponsiveContainer completely and instead allow percentage strings in the base Chart elements?

ckifer:
Errr, I didn't think about it that much. Probably not a ton - you'd just be using the same responsive container internally if you wanted to do it with without rewriting the whole thing. You still need resize observer and all that which I guess you could write as a hook. Would need tested well for sure

ckifer:
Also we've totally had people as for a 100% svg only chat with no extra containers (and the legend also being totally svg) so people can export to images easily. Something to think about if we remove the ability to separate things

PR #5235, opened 2024-11-11
Polargrid axis ids

PavelVanecek:
## Description

PolarGrid no longer uses random axes and allows setting it explicitly (with defaults).

This is a breaking change because before, in a chart with multiple axes, PolarGrid would pick one (usually the first one) and use that. Now it will use the implicit axis instead, unless it has an axisId provided.

There is some visual diff: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1680 but I think it's for the better.

## Related Issue


ckifer (approved):
Nicee

PR #5237, opened 2024-11-12
Proposing new prop for Tooltip

PavelVanecek:
## Description

This is only the prop - no implementation yet. I want to have an agreement on what the external API should look like before implementing.

This option is what I think is the simplest - we let Tooltip stay sticky to the "categorical axis" and do not let users to override it (this is the same in 2.x). If in the future we do decide to allow overriding it, we can add a new prop that goes `axisType: xAxis | yAxis | radiusAxis | angleAxis`.

What's new is we will allow users to specify which of the categorical axes exactly should be the tooltip axis - in case there are multiple.

WDYT?

Alternatively we could have a bunch of props that goes like `xAxisId: AxisId; yAxisId: AxisId; radiusAxisId: AxisId; angleAxisId: AxisId;` but I selected against it because it will allow users to specify multiple conflicting axes at the same time. But now that I think of it, that's exactly what Scatter is doing.

## Related Issue

### DIFF
--- a/src/component/Tooltip.tsx
+++ b/src/component/Tooltip.tsx
@@ -101,8 +101,8 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Omi
   useTranslate3d?: boolean;
   wrapperStyle?: CSSProperties;
   /**
-   * Tooltip always attaches itself to "categorical" axis. Which axis is categorical? Depends on the layout:
-   * - horizontal layout -> X axis is categorical
+   * Tooltip always attaches itself to the "Tooltip" axis. Which axis is it? Depends on the layout:
+   * - horizontal layout -> X axis
    * - vertical layout -> Y axis
    * - radial layout -> radial axis
    * - centric layout -> angle axis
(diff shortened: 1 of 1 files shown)


src/component/Tooltip.tsx
```diff
@@ -99,6 +100,16 @@ export type TooltipProps<TValue extends ValueType, TName extends NameType> = Omi
   trigger?: TooltipTrigger;
   useTranslate3d?: boolean;
   wrapperStyle?: CSSProperties;
+  /**
+   * Tooltip always attaches itself to "categorical" axis. Which axis is categorical? Depends on the layout:
```
ckifer, line 104: What about when both domains are numeric? It still treats one as categorical?

PavelVanecek: Yeah the type has nothing to do with it. Unfortunate naming.

https://github.com/recharts/recharts/blob/849d7b04faac8848cfbaf59957e861ed01ef4059/src/util/ChartUtils.ts#L295

ckifer: Lol

PavelVanecek: Also see https://github.com/recharts/recharts/blob/849d7b04faac8848cfbaf59957e861ed01ef4059/src/chart/generateCategoricalChart.tsx#L578

ckifer (approved):
I think this is fine, let's see how people like it and go from there. This is similar to what happens when you have multiple Axes today. Just applied to the tooltip too

src/component/Tooltip.tsx
PavelVanecek: I want to get rid of this completely later but for now I will update the comment.

ckifer: Sounds good

PavelVanecek:
Alright, I will try and start with Scatter and how this works there.

Issue #5224, opened 2024-11-08
For each pie data, provide different/variable inner radius according to the pie-data.


ajay-studio137:
## Do you want to request a feature or report a bug?
Feature

## What is the current behavior?
As of now, a static innerRadius can be provided to Pie component.

## What is the expected behavior?
I expect to provide a different innerRadius for each cell shown inside Pie.

Let me know if we need more information in here.

ckifer:
Can you give an example of what this will allow you to do? Genuinely curious

ajay-studio137:
Hi Coltin,

This would allow me to show Pie according to the % of its applicability in another dimension than the usual one.

Something similar would be fine. 

<img width="326" alt="Screenshot 2024-11-11 at 10 31 46 AM" src="https://github.com/user-attachments/assets/326e2d38-564a-4b28-9551-d5622053aea8">

For now, using multiple Pie Charts and showing only the one which is applicable for that particular Pie. With this approach, I am unable to pick up the mouse hover for Pie chart. Hope it makes sense.

ckifer:
I see, this is possible to change on hover with activeSector (might be called something else I forget). But yeah this is different. 

I labeled as a feature request. I'm not sure when this will be able to be picked up.

PR #5244, opened 2024-11-13
Add support for customizable radius per sector in Pie Chart

ilayda-cp:
When creating a pie chart, users can currently specify only a single radius (outer radius), which applies uniformly to all pie sectors. This PR proposes allowing developers to define different radius sizes for each pie sector, enabling more customized and varied visualizations.

## Description

A function type has been added to the `outerRadius` parameter. Users can now pass a function to `outerRadius` to calculate the radius dynamically. This function takes an element as a parameter and expects a number to be returned.

## Related Issue

#5224 

## How Has This Been Tested?

Run the existing unit tests and added unit tests for the new feature.

## Screenshots:

Screenshot from 2024-11-13 15-37-47

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -59,7 +59,7 @@ interface PieDef {
   /** The inner radius of sectors */
   innerRadius?: number | string;
   /** The outer radius of sectors */
-  outerRadius?: number | string | ((element: any) => number);
+  outerRadius?: number | string | ((dataPoint: any) => number);
   cornerRadius?: number | string;
 }
 
@@ -340,16 +340,27 @@ const getTextAnchor = (x: number, cx: number) => {
   return 'middle';
 };
 
+const getOuterRadius = (
+  dataPoint: any,
+  outerRadius?: number | string | ((element: any) => number),
+  maxPieRadius?: number,
+) => {
+  if (typeof outerRadius === 'function') {
+    return outerRadius(dataPoint);
+  }
+  return getPercentValue(outerRadius, maxPieRadius, maxPieRadius * 0.8);
+};
+
 const parseCoordinateOfPie = (
   item: {
     cx?: number | string;
     cy?: number | string;
     innerRadius?: number | string;
-    outerRadius?: number | string | ((element: any) => number);
+    outerRadius?: number | string | ((dataPoint: any) => number);
     maxRadius?: number;
   },
   offset: ChartOffset,
-  element: any,
+  dataPoint: any,
 ): PieCoordinate => {
   const { top, left, width, height } = offset;
   const maxPieRadius = getMaxRadius(width, height);
@@ -357,12 +368,7 @@ const parseCoordinateOfPie = (
   const cy = top + getPercentValue(item.cy, height, height / 2);
   const innerRadius = getPercentValue(item.innerRadius, maxPieRadius, 0);
 
-  let outerRadius;
-  if (typeof item.outerRadius === 'function') {
-    outerRadius = item.outerRadius(element);
-  } else {
-    outerRadius = getPercentValue(item.outerRadius, maxPieRadius, maxPieRadius * 0.8);
-  }
+  const outerRadius = getOuterRadius(dataPoint, item.outerRadius, maxPieRadius);
 
   const maxRadius = item.maxRadius || Math.sqrt(width * width + height * height) / 2;
 
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -59,7 +59,7 @@ interface PieDef {
   /** The inner radius of sectors */
   innerRadius?: number | string;
   /** The outer radius of sectors */
-  outerRadius?: number | string;
+  outerRadius?: number | string | ((element: any) => number);
   cornerRadius?: number | string;
 }
 
@@ -137,7 +137,6 @@ interface PieProps extends PieDef {
   className?: string;
   dataKey: DataKey<any>;
   nameKey?: DataKey<any>;
-  radiusKey?: DataKey<any>;
   /** Match each sector's stroke color to it's fill color */
   blendStroke?: boolean;
   /** The minimum angle for no-zero element */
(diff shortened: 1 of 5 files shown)


src/polar/Pie.tsx
```diff
@@ -479,6 +481,12 @@ export function computePieSectors({
         },
       ];
       const tooltipPosition = polarToCartesian(coordinate.cx, coordinate.cy, middleRadius, midAngle);
+      const outerRadiusParam = getValueByDataKey(entry, radiusKey, coordinate.outerRadius);
```
ckifer, line 484: Does this no longer allow you to specify outerRadius as you currently can? I can't tell. But it should prefer outerRadius if set - otherwise use radiusKey

ckifer (changes requested):
We're going to need some unit tests to make sure this does what we say it does. 

Also this might be better added to Cell rather than as a different data key. 

@PavelVanecek thoughts?

src/polar/Pie.tsx
ilayda-cp: If you use both "outerRadius" and "radiusKey", the code first calculates the outer radius by using the "outerRadius" parameter. Then the "radiusKey" parameter is checked. If its provided it will use it otherwise it fallbacks to the calculated outerRadius. But i can definitely update it to prefer outerRadius parameter first.

src/polar/Pie.tsx
```diff
@@ -137,6 +137,7 @@ interface PieProps extends PieDef {
   className?: string;
   dataKey: DataKey<any>;
   nameKey?: DataKey<any>;
+  radiusKey?: DataKey<any>;
```
ckifer, line 140: I also have a concern with the naming of this. There is currently both innerRadius and outerRadius props. Calling the new prop radiusKey might be confusing as to which radius we are affecting? And if you can do it to outerRadius why not inner as well?

Maybe instead of a new prop at all we should just let inner and outer radius take functions which give your data as an argument. Then you can do something like 

outerRadius={(data) => data.someRadius}

That seems a lot simpler

src/polar/Pie.tsx
ilayda-cp: One concern is how a user can define a reference radius for the pie chart if the sector radius is specified as a percentage. In this case, the outer radius parameter is used to calculate the sector's radius.

ckifer: See my new comment above, I don't think this is the best way we can do this. 

If we had the prop take a number or a function we wouldn't have to prefer one over the other

ckifer: See my new comment above, I don't think this is the best way we can do this. 

If we had the prop take a number or a function we wouldn't have to prefer one over the other

ckifer: Not too sure what you mean by reference radius. Can you give me a scenario?

ilayda-cp: i agree on making outerRadius a function. It would make more sense. Ill update it

ckifer: Have it allow both. A static number or percent like you can do today, or a function that returns a number. 

Thanks! And thanks for being willing to change things a bit

ilayda-cp: made the changes

src/polar/Pie.tsx
```diff
@@ -345,17 +345,25 @@ const parseCoordinateOfPie = (
...
   const cy = top + getPercentValue(item.cy, height, height / 2);
   const innerRadius = getPercentValue(item.innerRadius, maxPieRadius, 0);
-  const outerRadius = getPercentValue(item.outerRadius, maxPieRadius, maxPieRadius * 0.8);
+
+  let outerRadius;
+  if (typeof item.outerRadius === 'function') {
+    outerRadius = item.outerRadius(element);
+  } else {
+    outerRadius = getPercentValue(item.outerRadius, maxPieRadius, maxPieRadius * 0.8);
+  }
```
ckifer, line 365: nit: we could make this a small helper function so we don't have to use a `let`?

src/polar/Pie.tsx
```diff
@@ -59,7 +59,7 @@ interface PieDef {
   /** The inner radius of sectors */
   innerRadius?: number | string;
   /** The outer radius of sectors */
-  outerRadius?: number | string;
+  outerRadius?: number | string | ((element: any) => number);
```
ckifer, line 62: `element` implies "html element" or "react element". Can we name this something related to data? 

`dataItem`, `dataPoint`, even just `item` would be ok

ckifer (changes requested):
nice! Few more comments

PavelVanecek (changes requested):
I'm okay with this change, better than the Cell. I have one request though, can you make the Pie accept a DataKey? That also allows functions, and more. We have a helper for resolving it so that should be easy change.

src/polar/Pie.tsx
ilayda-cp: made the change

src/polar/Pie.tsx
ilayda-cp: made the change

ilayda-cp:
@PavelVanecek i couldn't understand what do you mean by make the Pie accept a DataKey. Could you explain?

PavelVanecek:
DataKey is this recharts-specific thing. It's either function or string or I forgot what else and we use it to allow users to decide how to pull primitives out of the data object. 

Look at what other components do when they accept a dataKey prop. Basically I want this new feature to follow the same pattern.

PavelVanecek:
I'm on phone on a train, I can paste some examples of what it should look like later when I get to the PC.

ilayda-cp:
@PavelVanecek  I checked it out, and the `Pie` component already has a `dataKey` prop. The value is extracted as follows:  
```javascript
const val = getValueByDataKey(entry, dataKey, 0);
```  
This value is then passed to the function assigned to the `outerRadius` parameter (if its a function).

ilayda-cp:
this is similar to my first solution. I can revert back to that solution with some changes. What do you think @ckifer ?

PavelVanecek:
Ah sorry for the back and forth. Maybe I should be more careful and read the history first.

Either way please don't make it use the Cell 😬 that's a blocker for us in the future and I will try to remove it in next versions.

ckifer:
The problem with dataKey is it adds a bunch of props associated to each radius type. I guess that's fine and we shouldn't be scared of new props but this feels less intrusive. 

If we want to add a dataKey down the road we still can, this essentially acts the same though

src/polar/Pie.tsx
ckifer: I may be missing something but I don't see the changes. Can you try to push again?

ilayda-cp: done

ilayda-cp:
@ckifer sorry about the delay pushed the change

src/polar/Pie.tsx
```diff
@@ -340,22 +340,36 @@ const getTextAnchor = (x: number, cx: number) => {
   return 'middle';
 };
 
+const getOuterRadius = (
+  dataPoint: any,
+  outerRadius?: number | string | ((element: any) => number),
```
ckifer, line 345: Nit: for consistent interface this should be updated too

But it's not outward facing so all good

src/state/selectors/pieSelectors.ts
```diff
@@ -28,7 +28,7 @@ export type ResolvedPieSettings = {
   paddingAngle?: number;
   minAngle?: number;
   innerRadius?: number | string;
-  outerRadius?: number | string;
+  outerRadius?: number | string | ((element: any) => number);
```
ckifer, line 31: Nit: Same here

ckifer (approved):
Two more tiny things but no worries about them. LGTM

Issue #4546, opened 2024-05-19
3.0 Bug: Brush internal state


PavelVanecek:
How to reproduce:
- Open storybook `http://localhost:6006/?path=/story/examples-linechart--with-brush` on 3.x branch
- Drag the brush handles, observe how the label ("Page A") inside the brush doesn't update - it should update
- Click outside of the chart, observe how the Brush handle jumps - it should stay where it was

Compare to 2.x behaviour:
- https://master--63da8268a0da9970db6992aa.chromatic.com/iframe.html?args=&id=examples-linechart--with-brush&viewMode=story

ckifer:
3.x link to compare to
https://63da8268a0da9970db6992aa-cfehtsenpg.chromatic.com/?path=/story/examples-linechart--with-brush

Still broke

PR #5253, opened 2024-11-18
fix(brush): don't re-run index-from-props effect every state change

ckifer:
## Description

Partial bug fix for linked issue. This fixes where the `startIndex` does not update at all because `useEffect` ran every single render and always overrode the new startIndex.

Bugs remaining:

* brush panorama has the same state as the chart, it should have its own state and not get sliced
* controlled brush (`examples-cartesian-brush--controlled-brush`) when moving the handles breaks mouse event when the index is changed. If the brush is controlled it should still allow indices to be updated by normal brush activity

## Related Issue

https://github.com/recharts/recharts/issues/4546

## Motivation and Context

fix part of the brush bug for 3.x

## How Has This Been Tested?

- npm test

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Brush.tsx
+++ b/src/cartesian/Brush.tsx
@@ -11,7 +11,6 @@ import React, {
   SVGProps,
   TouchEvent,
   useCallback,
-  useContext,
   useEffect,
 } from 'react';
 import clsx from 'clsx';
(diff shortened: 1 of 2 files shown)


ckifer:
Merging this. Not the hugest change in the world but fixes the pesky bug

Issue #5254, opened 2024-11-18
onDoubleClick and onContextMenu for AreaChart


abarreau:
Hello,

I would like to know if it's possible to have in the v2 version bindings for `onContextMenu` and `onDoubleClick` props on `AreaChart` (right now, only `onMouseDown`, `onMouseUp`, `onMouseMove`, `onMouseEnter` and `onMouseLeave` are implemented)?

My use case is that I want to be able to do separate actions on click, double click and right click. I know the `Area` component already have these props, but it doesn't suit my use case as I want to be able to click outside of drawn areas to perform said actions.

Is there maybe a workaround?

If a new feature is possible on the v2 release, I'm willing to open a PR if someone can roughly guide me through the process of implemeting this.

Thanks for your feedback!

ckifer:
Adding new click handlers can vary wildly in terms of effort and testing needed. 

A few questions:

* do you need to access data from recharts in the click handler or is the normal browser event ok?
* where exactly are you trying to click? Anywhere on the chart or exclusively on the dots/lines/shaded area?

abarreau:
> * do you need to access data from recharts in the click handler or is the normal browser event ok?

I need to access data from recharts yes. Specifically I need to know where my click position is related to both x and y axis scales.

My use case is that I'm trying to add a `ReferenceLine` at the specified mouse event position on right click or double click.

> * where exactly are you trying to click? Anywhere on the chart or exclusively on the dots/lines/shaded area?

![rechart](https://github.com/user-attachments/assets/335cac43-0a4c-4c0e-8da6-1612ba20f10a)

Anywhere in the red area.

ckifer:
Ok so that is on the top level chart. 

If you just needed the click events, that's a bit easier (just let them be passed through correctly as props). Since you need more than that. It will take a bit more + testing, etc. on mobile right now but I will link in the master branch where you would add this. 

Personally, I don't want to implement this until maybe 3.1. But if we did it here we would have to include it in 3.0 as well and there is going to be difference in how events are handled there. 


All for this change, but getting through won't be a priority (for me) for a bit longer. Appreciate the willingness to put up a PR yourself 🙏

abarreau:
I just opened a PR regarding this feature request: https://github.com/recharts/recharts/pull/5255

Any feedback is appreciated. Thanks for your help again @ckifer

ckifer:
released in 2.14.0

PR #5255, opened 2024-11-18
Adds onContextMenu and onDoubleClick props for categorical chart #5254

abarreau:
## Description

Adds `onContextMenu` and `onDoubleClick` props for handling double click and right click events on categorical charts.

## Related Issue

https://github.com/recharts/recharts/issues/5254

## Motivation and Context

I want to be able to do actions on double click and right click events on the wrapping component of lines, areas, etc.

## How Has This Been Tested?

Tested on Brave [Version 1.71.121 Chromium: 130.0.6723.91 (Official Build) (64-bit)](https://brave.com/latest/).
Then checked that behavior works as expected in storybook.

image

## Types of changes

## Checklist:


ckifer (changes requested):
That was fast thanks!

Would you be able to add some unit tests?

storybook/stories/Examples/AreaChart.stories.tsx
```diff
@@ -66,6 +66,55 @@ export const StackedAreaChart = {
...
+      console.log('On context menu', { state, event });
+    }, []);
+
+    const handleDoubleClick = useCallback((state: any, event: any) => {
+      event.preventDefault();
+      // eslint-disable-next-line no-console
+      console.log('On double click', { state, event });
+    }, []);
+
+    const handleClick = useCallback((state: any, event: any) => {
```
ckifer, line 83: Any particular reason useCallback is needed?

ckifer (commented):
I guess this wasn't too bad. Some unit tests so we don't break this later and we should be good to go

storybook/stories/Examples/AreaChart.stories.tsx
abarreau: Just a bad habit here. Will remove it.

ckifer: It's not a bad thing, just want to keep it simple for the sake of the examples

ckifer:
Ah all I need is a simple test for each event handler in one of the categorical chart test files (Area, Line,Scatter, Bar). Add the event handlers, make them fire, and assert that they've been called and the data from the generator is being sent

abarreau:
@ckifer That should be better now. Fortunatly your suggestion made me find a bug. Really appreciate your help here :)

test/chart/AreaChart.spec.tsx
```diff
@@ -237,6 +237,34 @@ describe('AreaChart', () => {
...
+
+      const surface = container.querySelector('.recharts-wrapper');
+      fireEvent.click(surface);
+      fireEvent.doubleClick(surface);
+      fireEvent.contextMenu(surface);
+
+      expect(handleClickMock).toHaveBeenCalledTimes(1);
+      expect(handleDoubleClickMock).toHaveBeenCalledTimes(1);
+      expect(handleContextMenuMock).toHaveBeenCalledTimes(1);
+    });
```
ckifer, line 267: one more thing! Can you also assert that there is something sent to the callback?
You don't have to assert every item

just something like

```
expect(handleDoubleClickMock).toHaveBeenCalledWith(expect.any(object));
expect(handleContextMenuMock).toHaveBeenCalledWith(expect.any(object));
```

ckifer (changes requested):
One more thing! Thanks for your patience

test/chart/AreaChart.spec.tsx
abarreau: Oh yeah sure, will do.
Do you also want me to replicate this test in every other chart?

ckifer: I wouldn't require it, but if its not that much more effort you could totally do something like this

https://github.com/recharts/recharts/blob/50da69ec3e2cd8633159f50b18b0bed74f9fb1ff/test/component/Tooltip/Tooltip.sync.spec.tsx#L268-L288

make test cases which you can run in a loop, render the correct test case and the assertions should be the same for all of them

abarreau: That should be better now.

ckifer:
If you'd like you could make the same PR to 3.x. But I can also take it. 

Thanks!

abarreau:
@ckifer not quite sure I still have the time to make the same work in v3. If I look into it, it won't be until a couple of days at least sorry.
Do you have any idea when would this feature be available in v2.14.0? Thanks!

ckifer:
No worries, just a suggestion. Appreciate your time here. 

I can't promise anything, I need to look over some other issues to see if anything else needs included. Maybe a few days to a week.

Issue #5266, opened 2024-11-21
Moving from lodash to es-toolkit 


KSJaay:
## Do you want to request a feature or report a bug?
Feature request

## What is the current behavior?
Currently recharts uses lodash, which in my experience takes up about 145kb within my bundle. I've recently been looking at es-toolkit, and its performance/size seems to be much better than lodash. I was planning on raising a PR that moves recharts from lodash to es-toolkit and see how this affects the performance/size.

Would you be fine with me raising that PR?

ckifer:
I would be fine with this. 

I've never heard of es toolkit or used it. Are we positive this is a drop in replacement for everything we're using in recharts? 

There is plenty of usages of lodash that can be probably changed to native alternatives at this point, the worst part is doing that work and making 1000% sure nothing has changed afterwards. 

Let's do it in the 3.x branch and leave 2.x alone.

KSJaay:
Yeah, es toolkit is pretty new and was released earlier this year. It's mainly built to be much faster than lodash and has a MUCH smaller package size. They use a lot of native features, but yeah I agree I think moving to native features would be better than moving to es toolkit.

Where it makes sense, I'll move to native features and test everything. I might leave lodash as it is for now and move to native implementation, that way it should also reduce the final bundle size. In a later PR I'll move to es-toolkit. I think if I do both now the PR is going to be too big.

es-toolkit github: https://github.com/toss/es-toolkit

KSJaay:
Hey, just finished making a few changes and replacing lodash with native code. I've added a list of things I've removed and the code I've replaced it with (basic examples), let me know if you think those are fine and I'll get the PR up later today or tomorrow.

There are some sections that I couldn't replace because the logic behind those functions is rather large/complicated. The only function that isn't complicated but I didn't wanna replace was `isNil` as it checks for both null and undefined, doing this in the code makes it way too messy as I'd have to do `value === null || typeof value === 'undefined'`. So, I was planning on replacing that using es-toolkit instead.

Another issue I ran into was typescript throwing an error when I removed isFunction, it seems like you're checking if a prop is a valid React component using `isFunction`. This can return a false positive, I was looking, and seems like react has `isValidElement` which checks if the passed value is a React/html element. Would you be fine with me using that instead?

Removed: 
- isFunction > typeof value === 'function'
- max > Math.max(...arr)
- min > Math.min(...arr)
- isNan > Number.isNaN
- isNumber > (typeof value === 'number' || value instanceof Number)
[...]

PR #5268, opened 2024-11-22
Replace lodash functions with native code

KSJaay:
## Description
When bundling an application, lodash seems to be taking a lot of space when using recharts. This is the first PR to replace lodash with either native functions or move to es-toolkit. 

This PR replaces any functions that could be replaced with either `typeof`, `isNullOrUndefined`, `upperFirst`, and `Number.isNaN`. In the future, there will be other PRs that replace other instances of lodash functions that are more complex.

## Related Issue

Issue: https://github.com/recharts/recharts/issues/5266

## Motivation and Context

Reduces the size of recharts bundle by removing lodash and replacing with internal functions.

## How Has This Been Tested?

All tests pass, and new tests have been written for  `isNullOrUndefined`, and `upperFirst` functions.

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Funnel.tsx
+++ b/src/cartesian/Funnel.tsx
@@ -11,7 +11,7 @@ import { Layer } from '../container/Layer';
 import { Props as TrapezoidProps } from '../shape/Trapezoid';
 import { LabelList } from '../component/LabelList';
 import { Global } from '../util/Global';
-import { interpolateNumber } from '../util/DataUtils';
+import { interpolateNumber, isNumber } from '../util/DataUtils';
 import { getValueByDataKey } from '../util/ChartUtils';
 import {
   LegendType,
@@ -189,7 +189,7 @@ const getRealWidthHeight = ({ customWidth }: { customWidth: number | string }, o
   const realHeight = height;
   let realWidth = width;
 
-  if (typeof customWidth === 'number') {
+  if (isNumber(customWidth)) {
     realWidth = customWidth;
   } else if (typeof customWidth === 'string') {
     realWidth = (realWidth * parseFloat(customWidth)) / 100;
--- a/src/cartesian/Funnel.tsx
+++ b/src/cartesian/Funnel.tsx
@@ -1,9 +1,6 @@
 /* eslint-disable max-classes-per-file */
 import React, { PureComponent } from 'react';
 import Animate from 'react-smooth';
-import isFunction from 'lodash/isFunction';
-import isNumber from 'lodash/isNumber';
-import isString from 'lodash/isString';
 import omit from 'lodash/omit';
 import isEqual from 'lodash/isEqual';
 
@@ -192,9 +189,9 @@ const getRealWidthHeight = ({ customWidth }: { customWidth: number | string }, o
   const realHeight = height;
   let realWidth = width;
 
-  if (isNumber(customWidth)) {
+  if (typeof customWidth === 'number') {
     realWidth = customWidth;
-  } else if (isString(customWidth)) {
+  } else if (typeof customWidth === 'string') {
     realWidth = (realWidth * parseFloat(customWidth)) / 100;
   }
 
@@ -230,7 +227,7 @@ export class FunnelWithState extends PureComponent<InternalProps, State> {
     const { onAnimationEnd } = this.props;
     this.setState({ isAnimationFinished: true });
 
-    if (isFunction(onAnimationEnd)) {
+    if (typeof onAnimationEnd === 'function') {
       onAnimationEnd();
     }
   };
@@ -239,7 +236,7 @@ export class FunnelWithState extends PureComponent<InternalProps, State> {
     const { onAnimationStart } = this.props;
     this.setState({ isAnimationFinished: false });
 
-    if (isFunction(onAnimationStart)) {
+    if (typeof onAnimationStart === 'function') {
       onAnimationStart();
     }
   };
--- a/src/chart/Treemap.tsx
+++ b/src/chart/Treemap.tsx
@@ -10,7 +10,7 @@ import { Polygon } from '../shape/Polygon';
 import { Rectangle } from '../shape/Rectangle';
 import { getValueByDataKey } from '../util/ChartUtils';
 import { COLOR_PANEL } from '../util/Constants';
-import { uniqueId } from '../util/DataUtils';
+import { isNan, uniqueId } from '../util/DataUtils';
 import { getStringSize } from '../util/DOMUtils';
 import { Global } from '../util/Global';
 import { findChildByType, validateWidthHeight, filterProps } from '../util/ReactUtils';
@@ -116,7 +116,7 @@ export const computeNode = ({
     nodeValue = computedChildren.reduce((result: any, child: TreemapNode) => result + child[NODE_VALUE_KEY], 0);
   } else {
     // TODO need to verify dataKey
-    nodeValue = Number.isNaN(node[dataKey as string]) || node[dataKey as string] <= 0 ? 0 : node[dataKey as string];
+    nodeValue = isNan(node[dataKey as string]) || node[dataKey as string] <= 0 ? 0 : node[dataKey as string];
   }
 
   return {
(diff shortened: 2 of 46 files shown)


src/component/Label.tsx
```diff
@@ -1,15 +1,20 @@
...
+  getPercentValue,
+  uniqueId,
+  mathSign,
+  isNullOrUndefined,
+} from '../util/DataUtils';
 import { polarToCartesian } from '../util/PolarUtils';
 import { ViewBox, PolarViewBox, CartesianViewBox, DataKey } from '../util/types';
 
-export type ContentType = ReactElement | ((props: Props) => ReactNode);
+export type ContentType = ReactElement | ((props: Props) => ReactNode) | Function;
```
KSJaay, line 17: Typescript didn't like this section, as I'm doing `isValidElement(label) || typeof label === 'function'` it can't pass `content={label}` as it can be a function.

I'm not the best with typescript so wasn't sure what to put here. I just pushed with `--no-verify` so I can get some feedback.

src/cartesian/Funnel.tsx
```diff
@@ -192,9 +189,9 @@ const getRealWidthHeight = ({ customWidth }: { customWidth: number | string }, o
   const realHeight = height;
   let realWidth = width;
 
-  if (isNumber(customWidth)) {
+  if (typeof customWidth === 'number') {
```
ckifer, line 192: I don't think this matters that much but these functions are not 1:1 https://stackoverflow.com/a/56664032/8050147

But since this is in a new major version this is probably fine

src/chart/Treemap.tsx
```diff
@@ -118,7 +116,7 @@ export const computeNode = ({
     nodeValue = computedChildren.reduce((result: any, child: TreemapNode) => result + child[NODE_VALUE_KEY], 0);
   } else {
     // TODO need to verify dataKey
-    nodeValue = isNan(node[dataKey as string]) || node[dataKey as string] <= 0 ? 0 : node[dataKey as string];
+    nodeValue = Number.isNaN(node[dataKey as string]) || node[dataKey as string] <= 0 ? 0 : node[dataKey as string];
```
ckifer, line 119: lodash `isNan` converts values to a number and returns true if this can be the case 

Number.isNaN does not. We might want to use es-toolkit for this for parity's sake

src/component/Label.tsx
ckifer: `((props: Props) => ReactNode)` should be able to handle this. I'd rather not add any `Function` here as that isn't specific enough... I can take a look

src/component/Label.tsx
```diff
@@ -498,11 +510,11 @@ const parseLabel = (label: unknown, viewBox: ViewBox) => {
     return <Label key="label-implicit" content={label} viewBox={viewBox} />;
   }
 
-  if (isFunction(label)) {
+  if (isValidElement(label) || typeof label === 'function') {
```
ckifer, line 513: this should just be `typeof label === 'function'`

We already handled if it is a valid element up above

src/shape/Symbols.tsx
```diff
@@ -80,7 +80,7 @@ export interface InnerSymbolsProp {
 export type SymbolsProps = SVGProps<SVGPathElement> & InnerSymbolsProp;
 
 const registerSymbol = (key: string, factory: D3SymbolType) => {
-  symbolFactories[`symbol${upperFirst(key)}`] = factory;
+  symbolFactories[`symbol${key.substring(0, 1).toUpperCase() + key.substring(1)}`] = factory;
```
ckifer, line 83: upperFirst from `DataUtils` is unused here

src/util/DataUtils.ts
```diff
@@ -159,3 +158,21 @@ export const getLinearRegression = (data: ReadonlyArray<{ cx?: number; cy?: numb
     b: (ysum - a * xsum) / len,
   };
 };
+
+/**
+ * Checks if the value is null or undefined
+ * @param {any} value The value to check
+ * @returns {boolean} true if the value is null or undefined
+ */
+export const isNullOrUndefined = (value: any): boolean => {
```
ckifer, line 167: nit: this could be `isNullish` to shorten the name.

But not a big deal

test/util/DataUtils.spec.ts
```diff
@@ -273,3 +275,25 @@ describe('getLinearRegression', () => {
...
+  });
+
+  it('should return false when value is not null or undefined', () => {
+    expect(isNullOrUndefined(0)).toBe(false);
+    expect(isNullOrUndefined('')).toBe(false);
+    expect(isNullOrUndefined('0')).toBe(false);
+  });
+});
+
+describe('upperFirst', () => {
```
ckifer, line 295: can we handle a case where the string is emtpy or nullish? just in case

ckifer (changes requested):
A few things! Thanks!

src/cartesian/Funnel.tsx
KSJaay: Can do the following instead, that'll check for the object type as well.

```js
typeof customWidth === 'number' || customWidth instanceof Number
```

ckifer: That gets lengthy when repeated a lot if you'd like to make a small utility for that too

src/component/Label.tsx
KSJaay: Yeah, I also thought that, but I get the following ts error when I remove `| Function`.

```
Type 'Function' is not assignable to type 'ContentType'.ts(2322)

Label.tsx(47, 3): The expected type comes from property 'content' which is declared here on type 'IntrinsicAttributes & Omit<SVGProps<SVGTextElement>, "viewBox"> & LabelProps'

(property) LabelProps.content?: ContentType
```

src/cartesian/Funnel.tsx
KSJaay: Would this utility not be fine? It also makes sure the value is a valid number 
https://github.com/KSJaay/recharts/blob/9856e7ab8f82771a8e157b0c09e11ee1cfa02e83/src/util/DataUtils.ts#L17

ckifer: Oh yeah that works. Hard to keep track of them sometimes 😅

src/component/Label.tsx
ckifer: You can do something like this

```tsx
  const isLabelContentAFunction = (content: unknown): content is (props: Props) => React.ReactNode => {
    return content && typeof content === 'function';
  };
  
  ...

  if (isLabelContentAFunction(label)) {
    return <Label key="label-implicit" content={label} viewBox={viewBox} />;
  }
```

ckifer: if it is a function we know already that it should match the types we've defined. It just needs asserted upon here

src/chart/Treemap.tsx
KSJaay: Looks like es-toolkit uses `Number.isNaN()`. I'll raise an issue on their repo and see if they can change that. For now I'll just create a function within `DataUtils`.

[Lodash isNan](https://github.com/lodash/lodash/blob/f299b52f39486275a9e6483b60a410e06520c538/lodash.js#L11960)
[es-toolkit isNan](https://github.com/toss/es-toolkit/blob/a46d8c4969afa5930bccfea07c14752b7f6be0d3/src/compat/predicate/isNaN.ts)

ckifer: I doubt they'll change it. But here I'm mostly concerned that we rely on the conversion in the case that someone provides data that is quoted numbers or something

KSJaay: I added some tests, let me know if I need to add any more. I also changed `isNan` with `Number.isNaN` in the tests and it all worked as expected.

Also, double checked and made sure that lodash `isNaN` returned the same results as the isNan function.

src/component/Label.tsx
```diff
@@ -66,15 +63,19 @@ export type ImplicitLabelType =
...
 
-  if (isFunction(formatter)) {
+  if (typeof formatter === 'function') {
     return formatter(label);
   }
 
   return label;
 };
 
+export const isLabelContentAFunction = (content: unknown): content is (props: Props) => React.ReactNode => {
```
ckifer, line 75: would rather not export this if possible

ckifer (approved):
LGTM thank you for the changes!

ckifer:
Lets merge this one and follow up with whatever is next

PR #5321, opened 2024-12-07
Do not return categoricalDomain if dataKey is undefined

PavelVanecek:
## Description

Made this change to better match what the generator is doing. There is a small visual diff, for the better I think: https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=6753d460c9cfa1f0559d5ad0

Also renamed the other `xCategoricalDomain` to better distinguish which is which. I think we need a better name here though.

## Related Issue


ckifer (approved):

ckifer:
Naming is hard 😅

PR #5322, opened 2024-12-07
Tooltip axis uses redux

PavelVanecek:
## Description

Instead of selecting a random axis from the generator, Tooltip selects a specific one from Redux store.

There is no new diff: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1753

The RadialBarChart tooltip is still broken (and it was before this PR too), I'll figure that out next.

## Related Issue


ckifer (approved):
🚀 I never would have been able to do this, awesome job as always

PavelVanecek:
> 🚀 I never would have been able to do this, awesome job as always

It's just TTD. Red test, green test, repeat until it's done

PR #5323, opened 2024-12-07
Replace ReactText with number|string

Philipp91:
For React v19 type compatibility, see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69022. For anyone on v18 and older, this should have no effect because the two type are equivalent.

## How Has This Been Tested?

If I apply this patch manually to my local install of recharts, the corresponding tsc errors go away.

Beyond that, I'm relying on automated testing to catch any unforeseen issues.

## Types of changes

### DIFF
--- a/src/cartesian/Brush.tsx
+++ b/src/cartesian/Brush.tsx
@@ -6,7 +6,6 @@ import React, {
   Children,
   PureComponent,
   ReactElement,
-  ReactText,
   SVGAttributes,
   SVGProps,
   TouchEvent,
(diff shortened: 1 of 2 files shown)


Philipp91:
This patch should also be applied to 2.x, i.e. the master branch. How to do that?

ckifer (approved):

ckifer:
You'll have to make another PR to the master branch. Change the branch you start from and then change the branch you PR to when you go to create it. 

Merging this into 3.x for now

ckifer:
PR made above ^

PR #5328, opened 2024-12-09
inRange args

PavelVanecek:
## Description

Turns out that inRange function was asking for more information than it needs. If we pass in a viewbox instead of an axis, it gets all the information it requires and we no longer need to figure out which axis to read.

## Related Issue


ckifer:
Oops merge conflicts

ckifer (approved):

PR #5333, opened 2024-12-10
Remove axis maps context

PavelVanecek:
## Description

These are axisMaps, created in the generator and put into state. They are still used inside the generator but no longer used through the context or state.

## Related Issue


ckifer (approved):

ckifer:
Sweet

PR #5337, opened 2024-12-11
Add syncMethod to redux store

PavelVanecek:
## Related Issue

### DIFF
--- a/scripts/snapshots/es6Files.txt
+++ b/scripts/snapshots/es6Files.txt
@@ -138,6 +138,8 @@
   "es6/state/selectors/tooltipSelectors.js",
   "es6/state/store.js",
   "es6/state/tooltipSlice.js",
+  "es6/synchronisation",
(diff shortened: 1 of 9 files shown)


src/state/tooltipSlice.ts
```diff
@@ -182,6 +194,12 @@ export const initialState: TooltipState = {
     activeClickAxisIndex: null,
     activeClickAxisDataKey: undefined,
   },
+  syncInteraction: {
+    active: false,
```
ckifer, line 198: Is this used?

ckifer (approved):

src/state/tooltipSlice.ts
PavelVanecek: I think I'll need to use it once I integrate synchronisation and isActive prop on tooltip

ckifer: Gotcha

PR #5345, opened 2024-12-11
Move Synchronised story to its own file so that we can have storybook controls for it

PavelVanecek:
## Description

I wanted to see how the synchronisation works with `syncMethod: value`, and with custom function and I discovered that ... it doesn't. Well we can only go up from here.

## Related Issue


ckifer:
It doesn't in master or it doesn't in 3.x? Pretty sure at just the custom function worked in master

ckifer (approved):

PR #5359, opened 2024-12-17
Removing generator data from Tooltip

PavelVanecek:
## Description

I figured it's time to have Tooltip independent from the generator. This will unblock synchronisation and a11y rewrite too. I had to disable some tests to get this green, will followup in next PR.

There is some visual diff. There are two reasons:
1. The payload is now correct: before, the defaultIndex caused the charts to render "axis"-tooltip-event-type payload, even though the chart was "item"-type. This is now fixed.
2. Redux now contains some of the coordinates but not all. Notably, Scatter `tooltipPosition` is missing completely - I will add that in next PR. Until then let the stories stay denied.

I added comments to each diff, feel free to go through it:
https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1792

## Related Issue

### DIFF
--- a/src/component/Tooltip.tsx
+++ b/src/component/Tooltip.tsx
@@ -139,8 +139,8 @@ function TooltipInternal<TValue extends ValueType, TName extends NameType>(props
   const dispatch = useAppDispatch();
 
   useEffect(() => {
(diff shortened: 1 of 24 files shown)


ckifer (approved):
🚀🚀🚀🚀🚀🚀🚀

PR #5371, opened 2024-12-20
Chart synchronisation from Redux

PavelVanecek:
## Description

Okay I had to rewrite a few things but it works now. Except RadialBarChart sync, I'll look at that next.

The synchronisation is not perfect. Most noticeable are cursors, and backgrounds out of sync. But I think it's already much better than the generator, and we can continue improving it in 3.x once we're out of the generator code.

## Related Issue


ckifer (approved):
Whew nice

PR #5377, opened 2024-12-22
Tooltip state refactor

PavelVanecek:
## Description

It bothered me for some time that the tooltip state is split in four distinct shapes even though it's conceptually the same thing so here I refactored it to be the same type. This means I could remove some of the logic duplication where the code repeated `if (tooltipEventType) { if (trigger) {` in several places. Now there's still some places left but let's merge something.

## Related Issue


src/state/selectors/combiners/combineTooltipInteractionState.ts
```diff
@@ -0,0 +1,62 @@
...
+  return tooltipInteractionState.index != null;
+}
+
+export const combineTooltipInteractionState = (
+  tooltipState: TooltipState,
+  tooltipEventType: TooltipEventType | undefined,
+  trigger: TooltipTrigger,
+  defaultIndex: TooltipIndex | undefined,
+): TooltipInteractionState => {
+  if (tooltipState.syncInteraction.active && tooltipState.syncInteraction.index != null) {
```
PavelVanecek, line 32: Now we can go here and update the synchronisation preference logic and all the other selectors will follow the same.

ckifer (approved):
Nice nice

PR #5379, opened 2024-12-23
Mouse activity now has higher priority than sync activity

PavelVanecek:
## Description

https://github.com/user-attachments/assets/7b4f10a2-2bad-4dc6-8161-abdeebb16841

## Related Issue

### DIFF
--- a/test/component/Tooltip/Tooltip.sync.spec.tsx
+++ b/test/component/Tooltip/Tooltip.sync.spec.tsx
@@ -754,18 +754,17 @@ describe('Tooltip synchronization', () => {
      * This test fails because Recharts decides that the chart A has active=true,
      * therefore the last synchronised value should stay on forever.
      */
-    it.fails(
-      'after switching charts from A to B, it should follow the mouse and update coordinates on both charts',
-      () => {
-        const { wrapperA, wrapperB, debug } = renderTestCase();
-        showTooltip(wrapperA, lineChartMouseHoverTooltipSelector, debug);
-        hideTooltip(wrapperA, lineChartMouseHoverTooltipSelector);
-        showTooltipOnCoordinate(wrapperB, lineChartMouseHoverTooltipSelector, { clientX: 100, clientY: 100 }, debug);
+    it('after switching charts from A to B, it should follow the mouse and update coordinates on the second chart, but remember the previous index on the first chart because it has active=true', () => {
+      const { wrapperA, wrapperB, debug } = renderTestCase();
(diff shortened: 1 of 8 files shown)


test/component/Tooltip/Tooltip.sync.spec.tsx
```diff
@@ -754,18 +754,17 @@ describe('Tooltip synchronization', () => {
...
-        const { wrapperA, wrapperB, debug } = renderTestCase();
-        showTooltip(wrapperA, lineChartMouseHoverTooltipSelector, debug);
-        hideTooltip(wrapperA, lineChartMouseHoverTooltipSelector);
-        showTooltipOnCoordinate(wrapperB, lineChartMouseHoverTooltipSelector, { clientX: 100, clientY: 100 }, debug);
-
-        expectTooltipPayload(wrapperA, 'Page B', ['BookOne : 300']);
-        expectTooltipPayload(wrapperB, 'Page B', ['BookTwo : 300']);
-      },
-    );
+    it('after switching charts from A to B, it should follow the mouse and update coordinates on the second chart, but remember the previous index on the first chart because it has active=true', () => {
```
ckifer, line 757: Why remember the index of A?

When I enter the second chart I'd expect A and B to keep syncing. It's just when you go out of A that the tooltip remains where it last was until the sync event from B reaches it and it continues moving. It's a two way sync right?

ckifer (changes requested):

test/component/Tooltip/Tooltip.sync.spec.tsx
PavelVanecek: Because the first chart has active=true and that makes it remember. I will try to update to override even the active.

ckifer: `active` should just say that the tooltip is active in general right? 

`active: true`

This is what I would expect for `active: true`:

* Shows at default index when default index is set
* Shows at hovered index on hover
* Stays at last hovered index when hovering out of the chart <-- this is the mentioned remembering
* When synchronized should
   * remain active when hovering out of either chart (sending or receiving)
   * should start moving again when `active` tooltip is no longer the sender but becomes the receiver <--- not happening right now
   * If both are active
      * Tooltips are always in the same place <----- not sure if this is happening or not

PavelVanecek:
OK I am pushing new version

https://github.com/user-attachments/assets/865e6213-c4ac-4895-820c-388a58a94728

ckifer (approved):

ckifer:
Thanks!

PR #5392, opened 2024-12-25
Brush synchronisation in redux, and removing more unused code

PavelVanecek:
## Description

This wasn't too bad

## Related Issue

Resolves https://github.com/recharts/recharts/issues/4548


storybook/stories/Examples/LineChart.stories.tsx
```diff
@@ -550,7 +550,9 @@ export const HighlightAndZoom = {
             width={800}
             height={400}
             data={data}
+            // @ts-expect-error we're changing CategoricalChartState and activeLabel is no longer going to be there - what do we do with this story?
             onMouseDown={e => setZoomGraph(prev => ({ ...prev, refAreaLeft: e.activeLabel }))}
```
ckifer, line 554: We'll have to give a way to make this possible. It doesn't have to happen here necessarily, it could happen in Line or elsewhere

ckifer (approved):
Nice, now I need to check brush again and see if I can figure out the other issues with it

PR #5393, opened 2024-12-25
Put codecov token back - uploads on 3.x branch started failing without it

PavelVanecek:
## Description

Turns out it's still required, even if the documentation says it's not.

## Related Issue

https://github.com/recharts/recharts/pull/5380


ckifer:
RIP, weird

PR #5397, opened 2024-12-26
fix(Brush): prevent resetting brush position unless start and end index are sent as props

ckifer:
## Description

Fix issue where brush snaps to points on the brush on click or on mouse out

This isn't perfect (see video), but its much better than it was. Not sure how to prevent this entirely...
But previously controlled brush barely worked so anything is an improvement.

## Related Issue

https://github.com/recharts/recharts/issues/4546

## How Has This Been Tested?

- all tests continue to pass
- tested in storybook

https://github.com/user-attachments/assets/99df72ff-4cb2-43b5-b4ec-0e73fac9b518

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Brush.tsx
+++ b/src/cartesian/Brush.tsx
@@ -399,6 +399,14 @@ interface State {
   prevX?: number;
   prevTravellerWidth?: number;
   prevUpdateId?: string | number;
+
+  /**
+   * Used to prevent re-setting of traveller position unless controlled via props.
+   * This is not perfect as mouseout events will still cause the traveller to move position
+   * but only when start/endIndex are controlled via props.
+   * */
+  prevStartIndexControlledFromProps?: number;
+  prevEndIndexControlledFromProps?: number;
 }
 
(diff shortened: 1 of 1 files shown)


PavelVanecek:
Let's write tests for Brush! What are we missing?

ckifer:
Time is always my excuse and then I end up spending more time figuring out the issue of course 😅. 

Tests needed:

* traveler does not reset after drag/mouseout/click - check when the value is changed and also when the traveler is dragged but not changed
* traveler does not reset after slide drag and mouse out/click
* controlled brush works as expected for all data indexes (0 to data.length - 1)

PavelVanecek (approved):
I'm a firm believer that `getDerivedStateFromProps` is not a good solution for anything but! If it's already there and it almost works then let's merge this. We can consider refactoring later if we find a roadblock.

ckifer:
Oh I very much agree, it's just less cognitive load than a total refactor. Which I also did and has its issues 😆

Issue #5406, opened 2024-12-31
Downgrade react-redux to v8


PavelVanecek:
react-redux 9 requires React 18: https://github.com/reduxjs/react-redux/releases/tag/v9.0.0

but do we require react-redux 9? Maybe we are just fine on version 8, and we can keep React 16 support for a little bit longer?

ckifer:
If it was 17 and up I'd say drop 16 support. The upgrade from 16 to 17 is nothing. 18 and beyond is harder so it's probably fine.  But means we'll also need to downgrade RTK. Not sure if we lose anything there

PR #5408, opened 2024-12-31
Downgrade redux toolkit and react-redux but keep reselect@5

PavelVanecek:
## Description

So that we can get React 16 support back.

~There are several failing tests and extra re-renders. Also typescript complains about something and I can't figure out what is it. We could disable that with a ts-expect-error but the infinite re-renders bother me more.~

~What do we want to do? Try and fix the rerenders? Or start requiring React 18+ too?~

Tests are green! and typescript errors are either fixed or silenced until we decide to bump again.

## Related Issue

fixes https://github.com/recharts/recharts/issues/5406 (I got the branch name wrong).


PavelVanecek:
Hm this random website I found on Google says React 16 is still the rage: https://w3techs.com/technologies/details/js-react

PavelVanecek:
Also https://w3techs.com/technologies/history_details/js-react/all/y

ckifer:
Weird

PavelVanecek:
Hm so the container didn't help. Back to the drawing board

ckifer:
Hey but I'm sure you prevented some re-renders along the way somewhere

PavelVanecek:
I just figured why the tests were failing, it was because of reselect's 4.x `lruMemoization` method which has a default cache size of 1. In reselect 5 they have a new memoization with an effectively unlimited size: https://reselect.js.org/api/weakMapMemoize fortunately for us we can bump reselect to 5 and keep react-redux at v8 and we get green tests and React 16 support.

Hello @acomanescu, welcome to #5408 ! Do you know how to pull this kind of statistics of out npm? Also where do you see which React version is supported and which not? I looked here https://react.dev/community/versioning-policy but I can't find that kind of information.

All in all I am tempted to continue supporting React 16 given that we have green tests. Next time we arrive to a blocker then sure let's bump the minimum support to 18+ but I don't see a good reason now. Correct me if I'm wrong.

ckifer (approved):

PR #5410, opened 2024-12-31
Legend state refactor

PavelVanecek:
## Description

I wrote the `LegendPayloadProvider` before we had Redux in place, and it's more convoluted than it needs to be. So I replaced it with a Redux variant of the same thing. I imagine it will help with the infinite re-renders in the other branch.

One visual diff (because I added the Legend element in the storybook): https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1843

## Related Issue

https://github.com/recharts/recharts/issues/5406


PavelVanecek:
Also happy new year! 🥳

ckifer (approved):

ckifer:
Happy new year!

PR #5416, opened 2025-01-04
Root container state

PavelVanecek:
## Description

I am looking into how we store the container dimensions because I suspect that's what's blocking https://github.com/recharts/recharts/pull/5408. In an attempt to understand what's going on I wrote a new story that shows the dimensions:

image

## Related Issue

https://github.com/recharts/recharts/issues/5406


ckifer (approved):
Dang, fancy

PR #5432, opened 2025-01-08
Allow using custom Redux components as children inside Recharts

PavelVanecek:
## Description

alpha-1 was failing in my pet project app because I used my own selectors inside the `<Tooltip component={Here} />`. This PR fixes that.

## Related Issue

I forgot to create an issue for this one. alpha-2 I suppose.


ckifer (approved):
Aha, I knew there was something that was going to mess with this. Nice find

PR #5435, opened 2025-01-09
Merge 3.x into master

PavelVanecek:
## Description

Just wondering if we want to merge 3.x into master or rather rename branches.


PavelVanecek:
... yeah I don't feel like resolving all of those conflicts. Let's just rename `master` to `2.x` and then `3.x` to `master` (or `main` or what is the latest recommendation)?

PR #5446, opened 2025-01-11
Fix syncMethod=function

PavelVanecek:
## Description

I skipped this one when porting to Redux, fixed now. It works nicely in the storybook too.

## Related Issue

### DIFF
--- a/test/chart/LineChart.spec.tsx
+++ b/test/chart/LineChart.spec.tsx
@@ -23,8 +23,11 @@ import { generateMockData } from '../helper/generateMockData';
 import { useAppSelector } from '../../src/state/hooks';
 import { pageData } from '../../storybook/stories/data';
 import { selectAxisRangeWithReverse, selectTicksOfGraphicalItem } from '../../src/state/selectors/axisSelectors';
-import { createSelectorTestCase } from '../helper/createSelectorTestCase';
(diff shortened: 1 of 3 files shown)


test/chart/LineChart.spec.tsx
```diff
@@ -1515,33 +1603,29 @@ describe('<LineChart /> - Rendering two line charts with syncId', () => {
...
 
-      fireEvent.mouseOver(firstChart, { bubbles: true, clientX: margin.left + 0.1 * dotSpacing, clientY: height / 2 });
-      vi.advanceTimersByTime(100);
+      hoverOverFirstLineItem(firstChart);
 
       // There are two tooltips - one for each LineChart as they have the same syncId
       expect(container.querySelectorAll('.recharts-tooltip-cursor')).toHaveLength(2);
 
-      // make sure tooltips display the correct values, synced by data value
-      expect(screen.queryByText('300')).toBeTruthy();
```
PavelVanecek, line 1543: The previous test was expecting `300` to be somewhere in the document. I have no clue where that came from, it doesn't match what I would expect in the Tooltip from looking at the data and expected index. I suspect that this test used to match something from other elements other than the Tooltip itself? I switched it so that now it only looks at the Tooltip payload and nothing else.

ckifer: Sounds good

ckifer (approved):

Issue #5449, opened 2025-01-11
Control the ActiveDots programmatically or provide a way to place reference dots on top of the ActiveDots.


Fefedu973:
I'm trying to do something like this: https://codesandbox.io/p/sandbox/nifty-marco-gtpxpz
However I have some issues:
- I want to put the ReferenceDots right on the corresponding datapoint but they are always off on the x axis (the wider the graph the larger the gap)
This is fixable only (I thought) if we could either manually control wich active dot is displayed (in my configuration I want an independant active dot for ecah line) but this is not possible or at least a way to put the reference dot right on the exact same position of the datapoint (or active dot) this would fix the alignment issue. Or last a way to allow the active dot to be independant for each line

PavelVanecek:
Hi @Fefedu973 , can you please show how would you like the chart to behave? I can't tell what's wrong from the codesandbox.

Fefedu973:
As you can see my reference dot are not on they given datapoint (they are miss aligned ). I used reference dot as an alternative to active dot behavior because it dosen't match what i want:

What I want the active dots to do (but rn they are miss aligned unfortunately):
https://github.com/user-attachments/assets/cdb5a69c-8440-4999-aa74-6487b27a9ed3
so here it is like if i had ALWAYS one active dot PER line and not one active dot per graph (this is what i want)

What they are doing rn (without my reference dots alternative that dosen't work 100% as I said):
https://github.com/user-attachments/assets/6ff74d98-ce5f-4c50-b2b7-b43f9336989b
here the active dot is traveling between the two lines (so the blue or red line don't always has an active dot active)

PavelVanecek:
Interesting, on my computer it looks perfectly aligned:

https://github.com/user-attachments/assets/f6cce763-b02f-4f1c-bf8e-6ec64ee24793

Oh I see if I make the chart a lot wider then it shows the difference:

<img width="479" alt="image" src="https://github.com/user-attachments/assets/c0f8dd45-0643-44a7-82af-1c498ada10f4" />

PavelVanecek:
And the whole chart looks completely wrong on the 3.x branch: 

<img width="1469" alt="image" src="https://github.com/user-attachments/assets/10e0e23e-d9ee-4a1e-a0ef-2aae8537df8e" />

I will take a look at what's going on there.

PavelVanecek:
Okay so good news and bad news.

Good news first.

[...]

PR #5451, opened 2025-01-12
Add story for closest data point and dual ReferenceDots

PavelVanecek:
Add example from https://github.com/recharts/recharts/issues/5449 to the storybook so that we can see how it works with 3.0 and the hook

### DIFF
--- a/src/synchronisation/types.ts
+++ b/src/synchronisation/types.ts
@@ -1,6 +1,6 @@
 import { Coordinate, DataKey, TickItem } from '../util/types';
 
-export type CustomSyncMethodDataParam = {
+export type MouseHandlerDataParam = {
   activeTooltipIndex: number | undefined;
   isTooltipActive: boolean;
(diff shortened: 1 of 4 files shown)


ckifer (approved):

ckifer:
Nit: there's other info people may want about the active index (index itself mostly. 

Should we make every hook individual or is it ok to combine them?

PavelVanecek:
I think individual hooks give us more opportunity for fewer re-renders.

On the other hand the tooltip activeX all change together anyway. So no strong feelings I suppose. Do you have a preference?

ckifer:
No strong preference really. My reasoning is that the selector inputs and the selector itself are mostly the same for this part of the state. If createSelector is actually memoizing things there shouldn't be any extra renders at all right? 

Also activeLabel doesn't seem that useful on its own (although it is here)

Anyways doesn't matter too much, just giving thoughts

Issue #5472, opened 2025-01-16
useSyncExternalStore error


PavelVanecek:
https://codesandbox.io/p/sandbox/recharts-brush-workaround-forked-x5fq3y?file=%2Fpackage.json%3A24%2C1

PR #5473, opened 2025-01-16
Use use-sync-external-store shim, also add scripts for testing react16 and react19 installations

PavelVanecek:
## Description

Turns out that `import from 'use-sync-external-store'` will only work on React18+ and to get backwards compatibility (which I thought was the only point of existence of this package but I digress) we need to `import from 'use-sync-external-store/shim'`.

Also I added two scripts so that I can more easily get a React@16 and React@19 setup going.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5472 and React 16 compatibility

### DIFF
--- a/scripts/install-react-16.sh
+++ b/scripts/install-react-16.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+# Useful for testing the package with React 16
+
+set -o errexit
+set -o pipefail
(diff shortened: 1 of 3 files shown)


ckifer (approved):
Ahh I thought this was the case but I couldn't find the with-selector version of the shim

Issue #5477, opened 2025-01-16
3.0.0-alpha.2 / chart tooltip not using container offset


karimbeyrouti:
## Do you want to request a feature or report a bug?
Bug

## What is the current behavior?

```tsx
          <AreaChart
            data={data}
            height={200}
            margin={{ top: 0, right: 0, left: 0, bottom: 0 }}
            width={900}
          >
            <Area
              dataKey="data"
              isAnimationActive={false}
              opacity={0.8}
              type="monotone"
            />
            <Tooltip />
          </AreaChart>
```

The tooltip does not use the container offset to calculate it's position:

![Image](https://github.com/user-attachments/assets/4f6acee1-1aba-450d-ba2f-7286e6a51db9)

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
3.0.0-alpha.2 / MacOS / Chrome

ckifer:
Odd. Will look into it

karimbeyrouti:
Thank you!

ckifer:
Hi - I can't reproduce this https://codesandbox.io/p/sandbox/line-chart-width-xaxis-padding-forked-p98hmg

karimbeyrouti:
Will have a look, see if I can make a simple example which shows the bug & post it back.

karimbeyrouti:
There you go: https://codesandbox.io/p/sandbox/area-chart-alpha-tooltip-issue-forked-6khrc5 

Pretty sure it worked in 2.0 with that layout / nesting, hopefully that helps.

ckifer:
Thanks - unsure what causes this other than flexbox. 

We didn't really change anything with our containers. Labeled for investigation

ckifer:
https://github.com/recharts/recharts/blob/3.x/src/state/selectors/containerSelectors.ts#L24-L25

Looks like offset left and top are always 0 in this case

ckifer:
released in alpha 4

PR #5478, opened 2025-01-17
Use getBoundingClientRect() instead of offsetLeft + offsetTop

PavelVanecek:
## Description

because that better matches mouse event coordinates when the chart is wrapped in a flexbox and position: absolute parents.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5477


ckifer (approved):
🚀

Issue #5476, opened 2025-01-16
3.0.0-alpha.2 /  chart mouse events / missing data


karimbeyrouti:
## Do you want to request a feature or report a bug?

Bug

## What is the current behavior?

```tsx
          <AreaChart
            data={data}
            height={200}
            margin={{ top: 0, right: 0, left: 0, bottom: 0 }}
            width={200}
            onMouseDown={(e) => {
              console.log('onMouseDown', { e });
            }}
            onMouseMove={(e) => {
              console.log('onMouseMove', { e });
            }}
            onMouseUp={(e) => {
              console.log('onMouseUp', { e });
            }}
            onMouseLeave={(e) => {
              console.log('onMouseLeave', { e });
            }}
          >
            <Area
              dataKey="insertions_no_ws"
              isAnimationActive={false}
              opacity={0.8}
              type="monotone"
            />
          </AreaChart>
```

Mouse events do not return any data (at least on the AreaChart). We used to have `activeState` on that object (amongst other things) - however it's now empty...

<img width="343" alt="Image" src="https://github.com/user-attachments/assets/529e467d-53d6-4785-97b2-cb411ca77b23" />

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
3.0.0-alpha.2 / MacOS / Chrome

ckifer:
This is expected - we can no longer pass back everything that used to be there in the event. 

Can you describe what you used the data in the mouse event for?

ckifer:
@PavelVanecek

PavelVanecek:
Yeah this is expected to happen. Instead of the mouse event callback parameters we now have a hook `useActiveTooltipLabel`, here is an example: https://github.com/recharts/recharts/blob/3.x/storybook/stories/Examples/DualLineChart.stories.tsx#L143

ckifer:
@PavelVanecek it looks like we need some sort of Provider wrapper above the chart itself 

https://codesandbox.io/p/sandbox/area-chart-alpha-tooltip-issue-forked-5t4c4t

Its impossible to get the hook to work correctly with custom components within `LineChart` because we need the chart wrapper event handlers

karimbeyrouti:
Thank you for looking at this. 

It was much simpler with states & events, than the party of effects. It's good we can encapsulate that in 3.0 if that is going to be the way forward, but having the data on the events would be a lot nicer.

ckifer:
added back in alpha 4

PR #5483, opened 2025-01-18
Add active tooltip state in external mouse events

PavelVanecek:
## Description

This information is accessible through hooks so in a way it's not strictly necessary to do that but some people prefer callbacks. Also Recharts2 used to pass this through callbacks. So let's allow for some flexibility.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5476


ckifer (approved):
Even though it works with hooks the Provider being the chart itself it a pretty limiting factor. It's much easier like this. I think this is a good compromise as long as it works half decently, thanks for doing this

Issue #5487, opened 2025-01-20
[Alpha.3 Bug] Non-shared Bar chart tooltip resets position to 0,0 onMouseLeave


ckifer:
## Do you want to request a feature or report a bug?

Alpha bug

## What is the current behavior?

Tooltip resets position to 0,0 every mouse out and flys back in on every mouse in

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar (template: https://codesandbox.io/p/sandbox/simple-line-chart-v25dl4).

https://codesandbox.io/p/sandbox/stacked-bar-chart-forked-mzn4fp?file=%2Fsrc%2FApp.tsx%3A76%2C18-76%2C33&workspaceId=ws_G5Rnj5KjeM4fNQbHpKaCKS

## What is the expected behavior?

The tooltip will not fly in from the corner of the chart except on the first render. Tooltip position is preserved.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
3.0.0 alpha 3

ckifer:
fixed in linked PR

ckifer:
released in alpha 5

PR #5490, opened 2025-01-22
Fix tooltip coordinate reset onMouseLeave

ckifer:
## Description

Fix tooltip coordinate reset.

This affects non-item charts too but I think for the better. Instead of always flying in from 0,0 after chart leave we now always come from the last known coordinate

## Related Issue

https://github.com/recharts/recharts/issues/5487

## Motivation and Context

Fixes some buggy behavior in charts, especially those with `item` type or `shared={false}`

## How Has This Been Tested?

`npm test`, update tests

Before https://63da8268a0da9970db6992aa-yujhlvsjyn.chromatic.com/?path=/story/examples-barchart--stacked

After https://63da8268a0da9970db6992aa-ynarlpziva.chromatic.com/?path=/story/examples-barchart--stacked

## Types of changes

## Checklist:

### DIFF
--- a/src/state/selectors/combiners/combineTooltipInteractionState.ts
+++ b/src/state/selectors/combiners/combineTooltipInteractionState.ts
@@ -61,5 +61,8 @@ export const combineTooltipInteractionState = (
     };
   }
 
-  return noInteraction;
+  return {
+    ...noInteraction,
+    coordinate: appropriateMouseInteraction.coordinate,
+  };
 };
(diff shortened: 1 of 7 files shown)


ckifer (commented):
fix is here, the rest is tests

src/state/selectors/combiners/combineTooltipInteractionState.ts
```diff
@@ -61,5 +61,8 @@ export const combineTooltipInteractionState = (
     };
   }
 
-  return noInteraction;
+  return {
```
ckifer, line 64: fix is here, the rest is test updates

PavelVanecek (approved):
Nice!

Issue #5489, opened 2025-01-21
[Alpha Bug] Infinite re-renders when container content changes with delay


ckifer:
> I keep running into 
> `Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.`
> not sure what's causing it. 
> 
> Is this a known issue? Else I would ttry to create a reproduction? 

 _Originally posted by @Makisuo in [#5445](https://github.com/recharts/recharts/issues/5445#issuecomment-2601061685)_

Referencing here as a new issue

 https://codesandbox.io/p/sandbox/line-chart-width-xaxis-padding-forked-6wvmkd

ckifer:
[@Makisuo](https://github.com/Makisuo) fixed in alpha 5

Issue #5445, opened 2025-01-11
3.0.0-beta.2 available for testing


PavelVanecek:
Hello everyone! We have released an alpha version `3.0.0-alpha.9`. In 3.0 we rewrote much of the state management, added 4000 unit and integration tests, and 220 visual regression tests.

Here is a migration guide: https://github.com/recharts/recharts/wiki/3.0-migration-guide

As of now we have fixed all the new 3.x bugs we are aware of, and we are ready for your feedback. Please feel free to try `recharts@alpha` in a test build and let us know how it went!

Edit: April 20, 2025 - `recharts@beta` up to date with `v3.0.0-beta.2` https://github.com/recharts/recharts/releases/tag/v3.0.0-beta.2
May 12 - beta.1 available
June 18 - beta.2 available

snelsi:
Do you still need to override `react-is` version to match `react`/`react-dom` in v3?

PavelVanecek:
We removed the need for react-is everywhere except Cell I think. So for using Cell, yes you still need a matching version. I'm not certain how it behaves when you don't use it - I will try it out.

My goal is to get rid of react-is completely but that will be another breaking change in the API. I will steer it towards the 4.0 version

PavelVanecek:
Ok so looking at the code more closely, it appears that these components are still using `react-is`:
- `Label`
- `LabelList`
- `Cell` (as a child in Bar, Funnel, Scatter, Pie, RadialBar)
- `Tooltip` (as a child of Sankey + Treemap)

So if you are not using any of these you should be fine without `react-is`.

I will continue removing react-is from everywhere except Cell - that one will require breaking change in an API so I will wait until 4.0 with that one.

philipplentzen:
@PavelVanecek I'm currently working on a project that requires the updated type defs from #5450. Is it possible to release a new alpha? Installing this project directly from GitHub doesn't work.

ckifer:
I'll release later today when I get a chance

ckifer:
alpha 3 published

Fefedu973:
Hey ! How can I install the beta versions in my project ?

ckifer:
@Fefedu973 
`npm i recharts@alpha -S`

or just replace your current recharts version with

"3.0.0-alpha.3"

Fefedu973:
Thanks !

Fefedu973:
Thanks !

Makisuo:
I keep running into 
[...]

PR #5492, opened 2025-01-22
Update hook dependencies

PavelVanecek:
## Description

I think I forgot this one when updating the container sizing. It was only a problem when the position changed but dimensions stayed the same which matches what we're seeing from the report. Unfortunately I couldn't figure out how to write a test for it.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5489

Originally reported in https://github.com/recharts/recharts/issues/5445#issuecomment-2601061685 so let's leave a reference too!


ckifer (approved):
Oh whaaaat, I totally tried this and it didn't work :o lol

Issue #5516, opened 2025-01-30
[Alpha Bug] Tooltips lose events and coordinates when a line chart is rendered lower on the page


gperl27:
I've encountered a bug where line chart tooltips seem to lose their state, position, or something seemingly internal when the chart is rendered further down on the page. The expectation is that the library should handle dynamic data sets and render at any place in the page regardless of its position This is a major blocker for my team in migrating us to use this library; it's also ideal for us to use 3.0 to meet certain UX requirements due to how flexible the components can be rendered vs 2.0.
 
I'm able to do a minimal reproduction found below. Any path forward or help would be amazing, thank you!

[Codesandbox here](https://codesandbox.io/p/sandbox/8hfjc5)

Developing on Firefox; using latest release of 3.0

ckifer:
Hi - thanks for the report. 3.0 is in alpha so things like this are expected and 3.0-alpha will include breaking changes until 3.0 is released. Just FYI.

Can you explain a bit more? Is this when the chart is rendered outside of the users current viewport?

ckifer:
I see the chart no matter what - can you post a screenshot or a video of what happens?

gperl27:
Ah, sorry about that. Please click the toggle button. The tooltips stop working when you do something like that.

ckifer:
Got it thanks, labelled as a bug

FYI @PavelVanecek

ckifer:
https://63da8268a0da9970db6992aa-caazrjvxcf.chromatic.com/?path=/story/examples-tooltip--recharts-alpha-tooltip-bug-5516-repro 

Added a reproduction to the storybook for easier debugging

PavelVanecek:
This is because https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect is relative to the viewport which changes on scroll. Re-render forces recharts to read and persist getBoundingClientRect() which is why it's working before click - the chart remembers the last getBoundingClientRect from when the scroll was up at 0,0.

Mouse events come relative to 0,0 instead of viewport so we need to read offsetTop and offsetLeft instead of getBoundingClientRect. I will fix.

PR #5523, opened 2025-01-31
Use event.currentTarget instead of redux state

PavelVanecek:
## Description

Reading the dimensions from the event is more reliable than attempting to synchronize them to the state.

Now I wonder how this works with item-based tooltips, we should test that too.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5516

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -23,6 +23,7 @@ import { ReportMainChartProps } from '../state/ReportMainChartProps';
 export interface MousePointer {
   pageX: number;
   pageY: number;
+  currentTarget: Pick<HTMLElement, 'offsetTop' | 'offsetLeft'>;
 }
 
 export interface ChartPointer {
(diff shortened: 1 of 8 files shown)


PavelVanecek:
Okay I added a story for Tooltip shared=false, it looks okay.

Now the `state.layout.offset` is completely unused. If you @ckifer verify that this PR works fine then I can remove that state, and one hook call with it.

ckifer (approved):

ckifer:
I'll test a bit later and confirm. Nice find!

ckifer:
This works as far as I can tell, though do we not need `offset` other places? I totally thought we did

PavelVanecek:
We use it in one of the storybooks and that's it.

ckifer:
I guess more importantly is it a potentially useful hook for others to have? I think I've seen offset used for custom positioning. 

But either way can't have this bug so it's probably fine to remove and we can deal with any requests for something similar later

PR #5530, opened 2025-02-02
Remove ref prop from XAxis, YAxis types

PavelVanecek:
## Description

We don't pass the ref anywhere anyway so let's remove it from the props too.

## Related Issue

Reported in https://github.com/recharts/recharts/issues/5445#issuecomment-2628711736


PavelVanecek:
@milessorce I couldn't reproduce the same problem with BarChart. If that still happens for you can you please share your code in codesandbox or a github example? Thanks!

milessorce:
> @milessorce I couldn't reproduce the same problem with BarChart. If that still happens for you can you please share your code in codesandbox or a github example? Thanks!

My mistake, it is `Bar` that has the issue, not `BarChart`.

ckifer (approved):

PR #5543, opened 2025-02-05
refactor: rm pie sector navigation, add test for accessibilityLayer

ckifer:
## Description

Pie had some custom keyboard handlers which interfered with accessibilityLayer

## Related Issue

N/A

## Motivation and Context

Remove custom focus handling

## How Has This Been Tested?

- ehhh, it needs some work but so do all Polar charts

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] I have added a storybook story or extended an existing story to show my changes

PR #5544, opened 2025-02-05
feat: allow index wraparound for keyboard control

ckifer:
## Description

allow wrap around in accessibilityLayer/keyboard navigation. Polar charts (and all charts) should wrap back to 0 when the end of the chart is reached. For polar charts this enables pie to navigate "normally" as the last index is in the position that I would generally consider 2nd

## Related Issue

N/A

## Motivation and Context

- anticipate issues reported for accessibilityLayer in polar charts

## How Has This Been Tested?

https://github.com/user-attachments/assets/5db77cce-9feb-4bc1-bfbe-0a27d13d5705

https://github.com/recharts/recharts/pull/5543 interferes with this right now but it will be removed in that PR

## Types of changes

## Checklist:

### DIFF
--- a/src/state/keyboardEventsMiddleware.ts
+++ b/src/state/keyboardEventsMiddleware.ts
@@ -49,9 +49,11 @@ keyboardEventsMiddleware.startListening({
     const direction = selectChartDirection(state);
     const directionMultiplier = direction === 'left-to-right' ? 1 : -1;
     const movement = key === 'ArrowRight' ? 1 : -1;
-    const nextIndex = currentIndex + movement * directionMultiplier;
-    if (nextIndex >= tooltipTicks.length || nextIndex < 0) {
-      return;
+    let nextIndex = currentIndex + movement * directionMultiplier;
+    if (nextIndex >= tooltipTicks.length) {
+      nextIndex = 0;
+    } else if (nextIndex < 0) {
+      nextIndex = tooltipTicks.length;
     }
     const coordinate = selectCoordinateForDefaultIndex(state, 'axis', 'hover', String(nextIndex));
 
--- a/test/chart/AccessibilityLayer.spec.tsx
+++ b/test/chart/AccessibilityLayer.spec.tsx
@@ -147,39 +147,39 @@ describe.each([true, undefined])('AccessibilityLayer with accessibilityLayer=%s'
       const cursor = container.querySelector('.recharts-tooltip-cursor');
       expect(cursor).not.toBeNull();
 
-      // Ignore left arrow when you're already at the left
+      // Wrap around when already at the left
       arrowLeft(svg);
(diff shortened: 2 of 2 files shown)


PavelVanecek:
So this might work but I would like to go in different direction. And that is: each chart should have its own logic for what "Next" and "Previous" means.

The cartesian charts will keep what is there right now. So, go all the way to the right, stop there. All the way to the left, stop there.

The polar charts will wrap around because they go around in a circle so it feels more natural.

Treemap, Sankey, Sunburts - they use their own special format for tooltip index so they won't use array-based loops anyway.

We have this pattern already in place with `tooltipPayloadSearcher`. Can we add `getNextTooltipIndex` and `getPreviousTooltipIndex` next to it?

ckifer:
Sure, this was just the least effort stab at something. 

I'm not sure I agree that Cartesian charts shouldn't wrap though. Why not? I get they don't go in a circle but it's a long way back to the beginning from the end if you're a keyboard only user

PavelVanecek:
@julianna-langston what is your view on the question of wrapping around the tooltip index?

julianna-langston:
It would be my personal preference for values to *not* wrap. While I understand that it makes more intuitive sense for keyboard-only users, especially in the case of charts that are literally circles, my reasoning is focused more on screen reader users. How does the screen reader user know that the chart has "wrapped around"? Let's say the user presses RIGHT, and learns there are values 1, 2, 3, 4, 1. Are there 4 slices, and they've wrapped around, or are there 5+ slices? I think it's better for a blind user, who has to figure out the layout of the space at the same time as they figure out the data values, if we're consistent with both cartesian and polar charts.

That said, there are significantly more keyboard-only users than screen reader users. If we anticipate this being a popular request among sighted, keyboard-only users, I won't oppose the change.

Note that I assume HOME and END still work in a polar chart. It doesn't make as much sense as in the context of a cartesian plot, but it's the best I've come up with.

ckifer:
Thanks @julianna-langston I trust you more than myself here - while I could see it as a useful feature to wrap in either scenario for a non-screen reader user I see your point. We could make the screen reader announce that the chart is back to the beginning? But thats more work to maintain I guess.

Let's disregard wrapping for now and revisit if we get feature requests, etc. its not worth of being anticipatory

Right now in 3.x we have a few issues then:

1. Polar charts do not go the correct direction when pressing RIGHT or LEFT (LEFT goes right, RIGHT goes left). This is because the charts are rendered the opposite direction you'd expect in terms of which index you are on (RTL instead of LTR)
2. `HOME` and `END` do not work at all on any charts. @julianna-langston did this work previously? I don't think it works in 2.x at the latest and I don't see code for it

ckifer:
Will close this PR and follow up with the ability to modify how an index is selected in each chart to give us the flexibility to do some cool things with this later

Issue #5542, opened 2025-02-04
[Alpha.6 Bug] offset not calculated correctly chart has positioning styles applied


ckifer:
> Seeing a positioning issue with Tooltip in alpha 6 that I didn't see in alpha 5 or in 2.x. If the chart is a child of a grid (not the first child), and also a child of a relatively positioned element, the tooltip does not position itself correctly.
> 
> Reproduced in [CodeSandbox](https://codesandbox.io/p/sandbox/tooltip-issue-alpha-6-dmd375) 

 _Originally posted by @milessorce in [#5445](https://github.com/recharts/recharts/issues/5445#issuecomment-2634964383)_

ckifer:
@PavelVanecek updating the offset to be event only caused this I believe https://github.com/recharts/recharts/pull/5527

PavelVanecek:
Interesting. I will have a look. I haven't tried a chart in a Grid yet

PR #5548, opened 2025-02-06
Fix mouse position coordinates in charts with an offsetParent

PavelVanecek:
## Description

Okay I think this time it's bulletproof.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5542


src/util/ChartUtils.ts
```diff
@@ -734,25 +734,19 @@ export function getTooltipNameProp(
 export function inRange(
   x: number,
   y: number,
-  scale: number,
```
ckifer, line 737: This might break the calculations when someone is zoomed in?

ckifer: Nvm you do this later

ckifer (approved):
Very nice

src/util/ChartUtils.ts
PavelVanecek: Technically earlier 😅 the coordinates arrive to this function already scaled. The storybook with zoom in/out was useful.

Issue #4547, opened 2024-05-19
Tooltip and Legend payload sorting


PavelVanecek:
In Recharts 2.x, Legend and Tooltip values are sorted by order of components in DOM. Once we solve #2788, we are no longer going to have access to the DOM element order (because there might be components inside of components inside of components and we do not want to parse the whole world).

Right now, in 3.x branch, the Legend and Tooltip values arbitrarily swap, based on how React decides to render things. Sometimes that happens on mouse hover and it's quite horrible frankly.

We're going to need to:
1. Decide on some default sort that will replace DOM element order
2. Add new props to Tooltip and Legend to allow overwriting the default sort

milessorce:
Has anyone thought about what the interface for sorting labels will be? I ran into this issue while testing 3.x. Currently handling it by using custom content components.

`labelSorter` prop added to Tooltip and Legend content components:
```
labelSorter={(payloadItem) => {
  return dataKeyList.indexOf(item.dataKey)
}}
```


In content components:
```
return labelSorter.payload.sort((a, b) => sorter(a) - sorter(b))
```

ckifer:
I don't think anyone has thought of anything quite yet but good to bring this up.

I think your idea is good, but we might want to give the user a little more flexibility. Like passing the comparison function itself.
In the legend we would do:

`payload.sort(sortingComparitor)`

and the user would do something like

`legendSortingComparitor={(a, b) => labelSorter(a) - labelSorter(b)}`

`legendSortingComparitor` is a bit long winded though 😅

PavelVanecek:
We could make it accept a `DataKey` prop so that it's easy to pass in a single string value. Something like

```
<Legend sortBy="name" />
```

PavelVanecek:
Ha but the Tooltip.itemSorter is a good point. We could do the same for Legend. Or even make it so that the Legend uses Tooltip sorter, if one is provided but the other one is not?

Aaaand add a default itemSorter so that the components are not jumping around by default.

ckifer:
Yeah that sounds good to me

PR #5555, opened 2025-02-09
Add tests for Tooltip.itemSorter, allow `dataKey` and `value` as inputs

PavelVanecek:
## Description

Testing the existing behavior

## Related Issue

### DIFF
--- a/src/component/DefaultTooltipContent.tsx
+++ b/src/component/DefaultTooltipContent.tsx
@@ -53,7 +53,7 @@ export interface Props<TValue extends ValueType, TName extends NameType> {
   labelFormatter?: (label: any, payload: ReadonlyArray<Payload<TValue, TName>>) => ReactNode;
   label?: any;
   payload?: ReadonlyArray<Payload<TValue, TName>>;
-  itemSorter?: (item: Payload<TValue, TName>) => number | string;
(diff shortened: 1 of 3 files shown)


ckifer (approved):

src/component/DefaultTooltipContent.tsx
```diff
@@ -53,7 +53,7 @@ export interface Props<TValue extends ValueType, TName extends NameType> {
   labelFormatter?: (label: any, payload: ReadonlyArray<Payload<TValue, TName>>) => ReactNode;
   label?: any;
   payload?: ReadonlyArray<Payload<TValue, TName>>;
-  itemSorter?: (item: Payload<TValue, TName>) => number | string;
+  itemSorter?: 'dataKey' | 'value' | ((item: Payload<TValue, TName>) => number | string);
```
ckifer, line 56: Were these always usable options? 😅

PavelVanecek: Anything in the payload works. I documented these two as they're the only two that are always populated and make sense to sort by.

PavelVanecek: Array of strings works too since we just feed it to lodash.

Issue #5458, opened 2025-01-14
[Tooltip] Negative `offset` causes the tooltip escape x viewBox even disallowed by `allowEscapeViewBox`


awx-doma-lei:
## Do you want to request a feature or report a bug?

Bug

## What is the current behavior?

I'm adding a negative y-offset to my tooltip, to make the tooltip appear above my cursor. But the tooltip bleeds out the right boundary when I hover the right-most data item.

Looks like it's due to the inability to specify offset for x, y direction separately, and recharts assumes the offset should be applied to both x, y directions.

## Repro

https://codesandbox.io/p/sandbox/simple-line-chart-forked-nnxnsf?file=%2Fsrc%2FApp.tsx%3A68%2C2

<img width="393" alt="image" src="https://github.com/user-attachments/assets/053a7602-0c4d-42e8-a65d-e25f8613a48e" />

## What is the expected behavior?

When `allowEscapeViewBox.x` is not true, the tooltip should not escape the right-hand boundary of the view box.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

2.15.0, Chrome, macOS

ckifer:
https://github.com/recharts/recharts/blob/3.x/src/util/tooltip/translate.ts#L58-L62

Probably `positive` and `negative` need switched when `offset` is negative. Looks like offset wasn't intended to be negative

awx-doma-lei:
Thanks @ckifer. Would this be fixed in 2.x? I noticed 3.x is around, but it still looks like an early alpha.

ckifer:
I can't promise anything but if the fix is easy it can be backported easy enough. This still needs a little more research though

awx-doma-lei:
Understood. I'll keep an eye on it, thanks @ckifer

phogel:
Have quickly recorded a video demonstrating the behaviour in the branch I linked above:
https://github.com/user-attachments/assets/7c7a99ba-2b2b-4620-bf83-be59fe9317c3

@awx-doma-lei Does this look like a valid fix for your issue?

ckifer:
I thought you both were on the same team or something hah 😅

phogel:
Nope, the proximity was coincidence. I just found this issue when searching for a similar one to mine, as suggested by the pull request template.

ckifer:
Sweet, appreciate you doing that

awx-doma-lei:
[...]

PR #5571, opened 2025-02-13
#5458 Keep Tooltip within viewBox for negative offsets

phogel:
## Description

When setting the `offset` prop on the Tooltip to a negative number, the Tooltip would escape the viewBox area even when `allowEscapeViewBox` was false. Solution is to ignore a negative offset when calculating the value for `negative`. 

## Related Issue

#5458 

## Motivation and Context

Using a negative offset prop on the Tooltip to center it and having a chart that extends all the way to the edges of the container, parts of the Tooltip would run outside of the positive ends of the container. 

## How Has This Been Tested?

Tested in the storybook for Tooltip, see new story named "Tooltip with negative offset". 
This change does not affect other areas of the code.

**Before**
Bildschirmfoto 2025-02-13 um 16 18 09

**After**
Bildschirmfoto 2025-02-13 um 16 17 55

## Types of changes

## Checklist:

### DIFF
--- a/src/util/tooltip/translate.ts
+++ b/src/util/tooltip/translate.ts
@@ -55,8 +55,9 @@ export function getTooltipTranslateXY({
     return position[key];
   }
 
-  const negative = coordinate[key] - tooltipDimension - offsetTopLeft;
+  const negative = coordinate[key] - tooltipDimension - (offsetTopLeft > 0 ? offsetTopLeft : 0);
   const positive = coordinate[key] + offsetTopLeft;
+
   if (allowEscapeViewBox[key]) {
     return reverseDirection[key] ? negative : positive;
   }
--- a/storybook/stories/Examples/Tooltip.stories.tsx
+++ b/storybook/stories/Examples/Tooltip.stories.tsx
@@ -512,3 +512,23 @@ export const RechartsTooltipBug5542Repro = {
     data: pageData,
   },
 };
+
+export const TooltipWithNegativeOffset = {
+  render: () => {
+    return (
+      <ResponsiveContainer width="100%" height={400}>
+        <ComposedChart data={pageData}>
+          <XAxis dataKey="name" />
+          <YAxis />
+          <Line dataKey="uv" />
+          <Tooltip
+            offset={-50}
+            wrapperStyle={{
+              width: 100,
+            }}
+          />
+        </ComposedChart>
+      </ResponsiveContainer>
+    );
+  },
+};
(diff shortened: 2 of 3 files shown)


ckifer:
Thanks for this.

a question first. Why are you using a negative offset in the first place? From the original issue its so that the Y value changes? 

Just want to make sure I understand

phogel:
Hey @ckifer thanks for getting back.

My intention is to show a tooltip of 100px width centered on the x axis of the cursor. Achieved this ostensibly by adding `offset={-50}`. But that offset is then added, instead of substracted, to the viewBox area, making the tooltip slide outside of the visible area.

Might there be another way to achieve this? As a temporary fix I've added the offset as extra width to the wrapper, but that is not very clean.

ckifer:
You can do your own positioning with `coordinates` but I believe that doesn't respect allowEscapeViewbox. 

I'm fine with the change as long as it works in all cases. What I don't want is this to break something that was previously working with a negative offset. 

And it's hard to test all cases 😅. Appreciate the tests you added though

ckifer:
This is going into 3.x so a breaking change is fine. I haven't had to chance to play with it yet so I might wait until I can. 

Also maybe @PavelVanecek can take a Quick Look

PavelVanecek:
> the Tooltip would escape the viewBox area even when allowEscapeViewBox was true

when `allowEscapeViewBox=true` then the Tooltip should be able to escape the viewbox? Is this a typo or does the library behave contrary to the prop name?

src/util/tooltip/translate.ts
```diff
@@ -55,8 +55,9 @@ export function getTooltipTranslateXY({
     return position[key];
   }
 
-  const negative = coordinate[key] - tooltipDimension - offsetTopLeft;
+  const negative = coordinate[key] - tooltipDimension - (offsetTopLeft > 0 ? offsetTopLeft : 0);
```
PavelVanecek, line 58: I am trying to understand why is this zero. My first intuition says it should be `Math.abs(offsetTopLeft)`.

phogel:
> > the Tooltip would escape the viewBox area even when allowEscapeViewBox was true
> 
> when `allowEscapeViewBox=true` then the Tooltip should be able to escape the viewbox? Is this a typo or does the library behave contrary to the prop name?

Sorry, it was a typo. I'll update the description.

src/util/tooltip/translate.ts
phogel: Negative offsets with `allowEscapeViewBox=false` result in the Tooltip stopping at the minimum edge of the viewbox at the moment, at `0`. Using `Math.abs(offsetTopLeft)` here results in the Tooltip being offset from the boundary on the maximum edge by the given value. So the behaviour for min and max edge would be inconsistent with `Math.abs(offsetTopLeft)`

PavelVanecek (approved):

ckifer:
Going to merge then, let's make sure we confirm behavior is as expected after. Thanks @phogel

phogel:
Thank you very much!

Issue #209, opened 2016-09-06
Stacked chart legend has the wrong order


SachaG:
It seems like stack charts legends get generated with the wrong order. Assuming you have:

```
        <BarChart width={600} height={300} data={data}
            margin={{top: 20, right: 30, left: 20, bottom: 5}}>
       <Bar dataKey="pv" stackId="a" fill="#8884d8" />
       <Bar dataKey="uv" stackId="a" fill="#82ca9d" />
      </BarChart>
```

Bar `uv` will appear on top, and `pv` on bottom (see http://recharts.org/examples/#StackedBarChart). 

The problem is that the legend has `pv` first, and `uv` second. Which is confusing especially if you chose to use a vertical layout.

xile611:
@SachaG I think this's reasonable. The bars are stacked from the first group to the second group

SachaG:
OK, here's what I mean. If you chose a vertical layout for the legend, it would be:

- pv
- uv

On the other hand, the bar are stacked in this order:

- uv
- pv

What I'm saying is that the legend and the bars should be stacked in the same order.

xile611:
![image](https://cloud.githubusercontent.com/assets/6111424/18409771/a871228c-7782-11e6-80d2-01c53975e7cb.png)
![image](https://cloud.githubusercontent.com/assets/6111424/18409775/b2002dca-7782-11e6-962f-ecadb05ee30e.png)

For vertical chart, the bars stacked from the left (the orientation of YAxis) to right.
For horizontal chart, the bars stacked form the bottom (the orientation of XAxis)  to top.

SachaG:
I don't know how I can make it any clearer than that: http://jsfiddle.net/y9bxhbkj/ (I don't know why the legend is overlayed on the chart but that's another matter).

Is it not clear that the bars and the legend are in opposite orders?

SachaG:
Any thoughts? Would you accept a PR to add an `order` prop to `Legend`?

SachaG:
By the way, in the meantime I'll use:

```
    display: flex;
    flex-direction: column-reverse;
```

xile611:
@SachaG The order of Legend is consistent with the order of stacked bars. So I'm considering whether to adjust the order of stacked bars.

SachaG:
Well it's true that the order of stacked bar is the opposite of the markup order. So you're right, maybe the bars order should be changed.

xile611:
Fixed in version 0.17.1.

[...]

PR #5572, opened 2025-02-14
RadialBar itemSorter tests

PavelVanecek:
## Description

Okay this concludes the tests. My next plan is to:

1. Add a default `itemSorter` to Tooltip
2. Have Legend reuse the same `itemSorter`
3. Add its own independent `itemSorter` prop to the Legend (and also have Tooltip share that)

Sounds good?

## Related Issue

### DIFF
--- a/test/component/Tooltip/itemSorter.spec.tsx
+++ b/test/component/Tooltip/itemSorter.spec.tsx
@@ -1619,3 +1619,5 @@ describe('itemSorter in RadialBarChart', () => {
     });
   });
 });
+
+// Funnel, Treemap, Sankey - always only render one payload item so there is nothing to sort either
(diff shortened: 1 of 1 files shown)


ckifer:
Sounds good to me. This also mostly solves https://github.com/recharts/recharts/issues/209

(Having both be default the same direction that is)

ckifer (approved):

PavelVanecek:
Interesting, I didn't consider stacked Bar. I will add tests for that too

ckifer:
Yeah stacked bar is the contentious one. I personally don't have a strong opinion on what is the "wrong" or "right" default direction but some do. 

Stacked Area too.

PR #5581, opened 2025-02-14
Make dependencies more flexible

PavelVanecek:
## Description

Okay so couple of things going on here:

1. It was a mistake to have a specific major `@reduxjs/toolkit` and `react-redux` versions in `dependencies`. The specific versions should have been in `devDependencies` so that our tests are stable, the `dependencies` should have a range. So I fixed that. We already do this with `react` and `react-dom` and `react-is` versions.
2. I removed `peerDependencies` and moved the ranges to `dependencies`. I think technically it works just the same - npm dedupes - but spiritually and semantically, `peerDependencies` are for other packages that we don't use directly but we play well with them. So an eslint plugin would specify eslint as a peerDependency - it doesn't import it directly, but it relies on a specific version of it. `react` and `react-redux` we use directly so they should have been in `dependencies.`

This change, I hope, should allow both React16 projects (as it will pull `@reduxjs/toolkit` version 1) and React19 projects (it would pull `@reduxjs/toolkit` version 2).

## Related Issue

I'm not sure where it was reported but we're seeing errors when installing in a new React@19 projects


ckifer:
> peerDependencies are for other packages that we don't use directly but we play well with them

Why wouldn't peerDeps be used for both? They are used for packages we use directly. I don't think moving things to dependencies allows us to ever remove "support" for anything. 

In 4.0 if we drop react 16 i don't think the user running npm i will be prevented from using something that will break their app, it'll just install two versions (one that satisfies recharts and one that doesn't). 

Feel like id need to test this in new projects a bit

ckifer (approved):
I'll merge this and test before a new release but a little skeptical

PavelVanecek:
Removing support from `peerDependencies` would be a breaking change, same as it would be when removing from `dependencies`. Tbf I couldn't find much differences between these two other than the semantics.

PR #5583, opened 2025-02-15
Add react 16 + 18 + 19 integration tests

PavelVanecek:
## Description

I am quite happy with how the dependencies turned out:

- React 18 and 19 projects work out of the box
- React 16 project will install reduxjs/toolkit 2 which in turn installs React 19 so now there are two Reacts in the project, not great. The `overrides` in package.json fixes it. So projects using React 16 + Recharts 3 will need a bit of extra config. Not perfect but better than not supporting React 16 at all

I will play around with what it looks like when either typescript or dependencies fail so that we know what it looks like in GitHub actions.

There are not many tests (just one now) but we can use this structure to add more.


PavelVanecek:
It works on my machine! I will debug

PavelVanecek:
I am starting to think that a separate project would be easier here

PavelVanecek:
Okay I give up. Let's just have separate repositories.

ckifer (commented):
Oops. Let me know what you want to make - just a recharts-integ repo?

PavelVanecek:
Yes please

PR #5584, opened 2025-02-15
Add itemSorter=name as a default prop in Tooltip

PavelVanecek:
## Description

I think this is the most reasonable default.

Visual diff: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=1998

## Related Issue


ckifer (approved):

ckifer:
Will make sure to add this to the breaking change list

PR #5585, opened 2025-02-16
Sort Legend payload by `name`, update tests

PavelVanecek:
## Description

Sort Legend alphabetically, instead of render order. I will make this configurable in the next PR.

Visual diff: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=2000

## Related Issue

### DIFF
--- a/src/component/DefaultLegendContent.tsx
+++ b/src/component/DefaultLegendContent.tsx
@@ -4,7 +4,6 @@
 import React, { PureComponent, ReactNode, MouseEvent, ReactElement } from 'react';
 
 import clsx from 'clsx';
-import { warn } from '../util/LogUtils';
 import { Surface } from '../container/Surface';
 import { Symbols } from '../shape/Symbols';
 import {
@@ -23,7 +22,10 @@ export type VerticalAlignmentType = 'top' | 'bottom' | 'middle';
 export type Formatter = (value: any, entry: LegendPayload, index: number) => ReactNode;
 
 export interface LegendPayload {
-  value: any;
+  /**
+   * This is the text that will be displayed in the legend in the DOM.
+   */
+  value: string;
   type?: LegendType;
   color?: string;
   payload?: {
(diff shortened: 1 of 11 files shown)


src/component/DefaultLegendContent.tsx
```diff
@@ -23,7 +22,10 @@ export type VerticalAlignmentType = 'top' | 'bottom' | 'middle';
 export type Formatter = (value: any, entry: LegendPayload, index: number) => ReactNode;
 
 export interface LegendPayload {
-  value: any;
+  /**
+   * This is the text that will be displayed in the legend in the DOM.
+   */
+  value: string;
```
ckifer, line 28: Is the value always a string?

src/state/selectors/legendSelectors.ts
```diff
@@ -10,5 +11,5 @@ const selectAllLegendPayload2DArray = (state: RechartsRootState): ReadonlyArray<
 
 export const selectLegendPayload: (state: RechartsRootState) => ReadonlyArray<LegendPayload> = createSelector(
   [selectAllLegendPayload2DArray],
-  payloads => payloads.flat(1),
+  payloads => sortBy(payloads.flat(1), 'value'),
```
ckifer, line 14: Maybe we should start using native sort functions? But.. maybe not. Just throwing it out there. Or maybe es-toolkit

People get upset about lodash for whatever silly reason.

ckifer (approved):

src/component/DefaultLegendContent.tsx
PavelVanecek: I made it a string now. It used to always pass the `dataKey` through and that could be a function.

src/state/selectors/legendSelectors.ts
ckifer: Meh. As long as lodash is in recharts people will complain about bundle size. I have no preference so all good

Issue #5582, opened 2025-02-14
[Alpha] Tooltip cursor is rendered beneath CartesianGrid


jasongerbes:
## Do you want to request a feature or report a bug?

Report a bug

## What is the current behavior?

In the 3.0.0 alpha, the `<Tooltip>` cursor is now rendered beneath the `<CartesianGrid>`.

![Image](https://github.com/user-attachments/assets/16e251c9-c68a-4e82-aa12-778804d36491)

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar

https://codesandbox.io/p/sandbox/simple-line-chart-forked-266qsh

## What is the expected behavior?

The `<Tooltip>` cursor should be rendered above the `<CartesianGrid>`.

![Image](https://github.com/user-attachments/assets/20ddc243-a0ca-4ba9-9498-afffba4ddbdb)

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

3.0.0-alpha.7, all browsers, previously worked as expected in 2.15.1.

Originally posted in #5445

ckifer:
Yep we can use this to track, thanks

PavelVanecek:
Yep so here's what it looks like now in 3.x

<img width="339" alt="Image" src="https://github.com/user-attachments/assets/a92e197a-923c-49f4-8780-566f843ef13b" />

If I undo the cursor portal change and have it render where the Tooltip is in the DOM, it looks like this

<img width="264" alt="Image" src="https://github.com/user-attachments/assets/8a7a2524-269e-4aca-bb61-eb9b70e23bb2" />

So yes on top of the grid but also on top of the active dots. I suspect that's not great either. Where do we want the cursor? Should it be drawing on top of the graphical items, or below? Should it render on top of the active dots or below?

If we introduce something like z-index to Recharts, can we get away with predefined set of depths and layers? Or does the order need to be configurable?

ckifer:
SVG has no concept of z index so I feel like implementing our own will be tough. 

I'm going to assume most people want things rendered the same way.

Grid - Cursor - reference elements - Graph elements

Or something along those lines.

jasongerbes:
> I'm going to assume most people want things rendered the same way.

I think that's a reasonable assumption, especially if nobody complained about the V2 behaviour.

If the layer ordering needs to be customisable, I think the most straightforward API would be to respect the render order, e.g:

```tsx
export default function App() {
[...]

PR #5586, opened 2025-02-16
Render cursor in JSX order, instead of a portal

PavelVanecek:
## Description

Allow the position of `<Tooltip />` element to decide on the z-index of the cursor

Visual diff: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=2002

## Related Issue

This doesn't completely fix https://github.com/recharts/recharts/issues/5582 but it should somewhat improve the situation


ckifer (approved):

ckifer:
So I think this caused https://63da8268a0da9970db6992aa-bihqlzsqfj.chromatic.com/?path=/story/examples-linechart--dashed

now we render the cursor above everything else again? Including the graphical element?

PavelVanecek:
Now we render the cursor where the `<Tooltip />` is. If that's the last in the JSX then it will be on top of other elements.

PR #5593, opened 2025-02-18
Replace lodash isEqual with dequal

alitas:
`dequal` is an order of magnitude faster than `_.isEqual` while being much smaller (500B gzipped vs almost 5KB gzipped). 

## Description

All instances of `_.isEqual` are replaced with `dequal`.

## Related Issue
#5266

## Motivation and Context

Reducing the bundle size while improving the performance of recharts.

## How Has This Been Tested?

All existing tests pass, since the functionality is the same.

## Types of changes

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@reduxjs/toolkit": "1.x.x || 2.x.x",
         "clsx": "^2.1.1",
         "decimal.js-light": "^2.5.1",
+        "dequal": "^2.0.3",
         "eventemitter3": "^5.0.1",
         "immer": "^10.1.1",
         "lodash": "^4.17.21",
@@ -10515,7 +10516,7 @@
       "version": "2.0.3",
       "resolved": "https://registry.npmjs.org/dequal/-/dequal-2.0.3.tgz",
       "integrity": "sha512-0je+qPKHEMohvfRTCEo3CrPG6cAzAYgmzKyxRiYSSDkS6eGJdyVJm7WaYA5ECaAD9wLB2T4EEeymA5aFVcYXCA==",
-      "dev": true,
+      "license": "MIT",
       "engines": {
         "node": ">=6"
       }
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "@reduxjs/toolkit": "1.x.x || 2.x.x",
     "clsx": "^2.1.1",
     "decimal.js-light": "^2.5.1",
+    "dequal": "^2.0.3",
     "eventemitter3": "^5.0.1",
     "immer": "^10.1.1",
     "lodash": "^4.17.21",
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -3,7 +3,7 @@ import React, { PureComponent } from 'react';
 import clsx from 'clsx';
 import Animate from 'react-smooth';
 import max from 'lodash/max';
-import isEqual from 'lodash/isEqual';
+import { dequal } from 'dequal';
 import { Curve, CurveType, Point as CurvePoint, Props as CurveProps } from '../shape/Curve';
 import { Dot } from '../shape/Dot';
 import { Layer } from '../container/Layer';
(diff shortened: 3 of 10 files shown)


ckifer:
replacing this without replacing all of lodash won't remove the dependency on lodash.

But regardless of that, why `dequal` over any of the other deep equal libraries? `fast-deep-equal`?

https://npmtrends.com/deep-equal-vs-dequal-vs-fast-deep-equal-vs-fast-equals-vs-lodash-vs-react-fast-compare

alitas:
> replacing this without replacing all of lodash won't remove the dependency on lodash.
> 
> But regardless of that, why `dequal` over any of the other deep equal libraries? `fast-deep-equal`?
> 
> https://npmtrends.com/deep-equal-vs-dequal-vs-fast-deep-equal-vs-fast-equals-vs-lodash-vs-react-fast-compare

I'm working on replacing lodash piecewise instead of going at it all at once. That being said while the entire lodash is installed in `node_modules`, this should still reduce the bundle size being sent to the client, since the `_.isEqual` will not be in the final bundles in a react project that's using recharts.

I don't have a particular preference for any of the libraries. `fast-equals` has a slightly larger bundle size, `dequal` seems to be the fastest one according to their own benchmark (https://github.com/lukeed/dequal?tab=readme-ov-file#benchmarks) and with 15M weekly downloads seems popular enough. But I'm definitely open to changing it to any other library that does a similar job.

ckifer:
> I'm working on replacing lodash

Can I ask why?

I want to understand why there is so much hate against lodash. I get that its not great for bundling and it has a large install size. Is that it? Whats the goal here?

ckifer:
Going to wait for @PavelVanecek 's thoughts before merging

alitas:
> > I'm working on replacing lodash
> 
> 
> 
> Can I ask why?
> 
> 
> 
> I want to understand why there is so much hate against lodash. I get that its not great for bundling and it has a large install size. Is that it? Whats the goal here?

It's not hatred, it's just a low hanging fruit because it's easy to replace with native code / smaller libraries and the changes are easier to test.

PavelVanecek:
I will have a look at the code later today.
I'd also like to explore why we need deep equality in the first place. With an immutable state we should be fine with triple equals everywhere

PavelVanecek:
Okay I'm on a laptop.

I can see this is adding a new dependency but not removing the old one. So it's not 5000B vs 500B kind of situation; it's 5000B vs 5500B. So that's not a bonus. Can we remove the old part completely to reduce the bundle size?

Magnitude of performance improvement sounds impressive. That would be worth trading for the 500B. Can you share how you tested it in recharts? Is there a storybook where we can see the improvement, or a codesandbox? I looked at the links in this PR but all I see is popularity, no perf tests in recharts.

And before we do all that. What if we replace both with a triple equals check? We have stable selectors now, with immutable data structures, so a simple `===` should be enough to check for deep equality. That would be the fastest, and with 0B of dependencies.

alitas:
> Okay I'm on a laptop.
> 
> I can see this is adding a new dependency but not removing the old one. So it's not 5000B vs 500B kind of situation; it's 5000B vs 5500B. So that's not a bonus. Can we remove the old part completely to reduce the bundle size?
> 
> Magnitude of performance improvement sounds impressive. That would be worth trading for the 500B. Can you share how you tested it in recharts? Is there a storybook where we can see the improvement, or a codesandbox? I looked at the links in this PR but all I see is popularity, no perf tests in recharts.
> 
> And before we do all that. What if we replace both with a triple equals check? We have stable selectors now, with immutable data structures, so a simple `===` should be enough to check for deep equality. That would be the fastest, and with 0B of dependencies.

It's adding a dependency but it's removing the line `import isEqual from 'lodash/isEqual';` from the files. When building a project, webpack will pack that 5000B `lodash/isEqual` into one of the JS bundles. This change does not make the `node_modules` folder on the developer's machine smaller, it just shrinks the code being shipped to the client at the end of the day.

The benchmark for `dequal` is here: https://github.com/lukeed/dequal?tab=readme-ov-file&rgh-link-date=2025-02-18T23%3A38%3A13Z#benchmarks. The performance improvement is not recharts specific and is probably not much considering the data being compared is usually not that big.

I'm not sure if using `===` makes sense, since most of these variables are stored in React state, and might not be referentially equal to their previous state even when they are deep equal.

PavelVanecek:
They're broken well before it even gets to the point where it compares. So in 3.x branch too, with lodash. I'll fix that first and once that's done then we can start testing performance difference.

Issue #5567, opened 2025-02-12
[Alpha] Line activeDot props vs ActiveDotType mismatch


ckifer:
> I'm sorry if this is a bit of a spam, and I can't tell what's the same as 2.x, but I never used recharts before 3-alpha and I really wanted to give it a go.
> 
> I've come across a type mismatch in the dot and activeDot props of `Line` compared to `ActiveDotType`.
> 
> Example
> ```tsx
> <Line
>       activeDot={({ cx, cy, r, strokeWidth, ...props }) => {
>         console.log(props);
>         if (props.value === null) return null;
> 
>         return (
>           <circle
>             cx={cx}
>             cy={cy}
>             r={r}
>             strokeWidth={strokeWidth}
>           />
>         );
>       }}
>     />
> ```
> 
> When I log spread props I get the following result:
> ![Image](https://github.com/user-attachments/assets/ea3e3c3a-90c2-4baf-a340-7b1fa82485cc)
> 
> Inside `ActiveDotType` the `((props: DotProps) => ReactElement<SVGElement>)` does not contain `index`, `value`, `payload` etc. but suggests that props are `SVGCircleElement` attributes. This is especially not good if spread props are used and each dot has an attribute payload of [object, object] because of this. 

 _Originally posted by @timheerwagen in [#5445](https://github.com/recharts/recharts/issues/5445#issuecomment-2654958420)_

ckifer:
@timheerwagen for things that you report we'll have to determine if they are new issues to 3.x or if they are old and have existed for a while. Either way, a bug is a bug but it'll help us determine severity, etc.

Recharts is rather flexible (to our detriment) with types. SVG attributes are implicitly on a lot of the components so that people can throw SVG attributes that apply down to the root component. This can be a good thing and a bad thing. Sometimes there's cases like this one where the types are just wrong and its totally unrelated to SVG props. Sometimes there's a subset of attributes that are correct and some that are just implicitly there. Technically not _wrong_ but most likely won't ever be used.

Anyways I'll take a look at this and get back as to what the situation is

PR #5595, opened 2025-02-19
Fix activeDot props type

PavelVanecek:
## Description

`@ts-expect-error` was hiding a problem.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5567


ckifer (approved):
Thanks

PR #5605, opened 2025-02-24
Make Tooltip a function component

PavelVanecek:
## Description

I am trying to debug the React 19 typescript failure and I can't figure out why. The error says that the `shouldComponentUpdate` signature doesn't match. But we don't even use `shouldComponentUpdate`.

One of the errors is coming from Tooltip. Maybe if I switch it to functional component it will help? If this works then I will switch the other components too.

I still can't reproduce the error anywhere outside of recharts-integ.

### DIFF
--- a/src/component/Tooltip.tsx
+++ b/src/component/Tooltip.tsx
@@ -1,4 +1,4 @@
-import React, { CSSProperties, PureComponent, ReactElement, ReactNode, useEffect } from 'react';
+import React, { CSSProperties, ReactElement, ReactNode, useEffect } from 'react';
 import { createPortal } from 'react-dom';
 import {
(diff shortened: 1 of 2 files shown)


ckifer:
I'll wait on this until Treemap is done

ckifer (approved):
But LGTM

ckifer:
does this need rebased?

ckifer:
Dependabot updated storybook and that expect error stopped complaining

PavelVanecek:
Why didn't the error stop in the dependabot PR?

ckifer:
It did, I'm used to seeing one error on dependabot PRs cause chromatic fails to deploy. But I merged on my phone and didn't see the other

Issue #5565, opened 2025-02-11
[Alpha] Tooltip doesn't change onTouchDrag


ckifer:
> I've noticed another change in the tooltip behaviour in 3.0:
> 
> In 2.15.1, you could tap and drag left/right on touchscreen devices to move the tooltip cursor.
> This interaction no longer works in 3.0.0-alpha.7 

 _Originally posted by @jasongerbes in [#5445](https://github.com/recharts/recharts/issues/5445#issuecomment-2649966687)_

ckifer:
@PavelVanecek I think we missed `onTouchMove`

https://github.com/recharts/recharts/blob/master/src/chart/generateCategoricalChart.tsx#L1705-L1709

PavelVanecek:
Should be an easy fix

ckifer:
https://github.com/recharts/recharts/blob/3.x/src/chart/RechartsWrapper.tsx#L122 it looks like its here? but yeah hopefully

Fefedu973:
Have you been able to find the issue?

ckifer:
Haven't had time yet sorry

Fefedu973:
> Haven't had time yet sorry

Np, I can try on my end

ckifer:
This is resolved in alpha 8, please help us test if you could @Fefedu973

Fefedu973:
I'm updating...

Fefedu973:
It is working!

ckifer:
This task is done then, but I'm going to rename this one because this still doesn't work for a few charts (Pie, Funnel, Treemap, etc.) which we knew when merging the fix for this

PavelVanecek:
This won't be as trivial fix as I thought because there is no onTouchEnter + onTouchLeave :/ I will figure something out

PavelVanecek:
Good news is 😬 this doesn't work well on 2.x either. So we're not making it worse with 3.x

PR #5611, opened 2025-02-25
fix: add onTouchMove touchEvent middleware

ckifer:
## Description

- we forgot a touch middleware, add it back

## Related Issue

https://github.com/recharts/recharts/issues/5565

## Motivation and Context

https://github.com/recharts/recharts/issues/5565

## How Has This Been Tested?

- Ran the storybook in both Chrome and Firefox's phone simulations
- I need to add unit tests but 🙃 

## Types of changes

## Checklist:

### DIFF
--- a/src/chart/RechartsWrapper.tsx
+++ b/src/chart/RechartsWrapper.tsx
@@ -8,6 +8,7 @@ import { focusAction, keyDownAction } from '../state/keyboardEventsMiddleware';
 import { useReportScale } from '../util/useReportScale';
 import { ExternalMouseEvents } from './types';
 import { externalEventAction } from '../state/externalEventsMiddleware';
+import { touchEventAction } from '../state/touchEventsMiddleware';
 
 type Nullable<T> = {
   [P in keyof T]: T[P] | undefined;
@@ -96,6 +97,7 @@ export const RechartsWrapper = forwardRef(
       dispatch(externalEventAction({ handler: onTouchStart, reactEvent: e }));
     };
     const myOnTouchMove = (e: React.TouchEvent<HTMLDivElement>) => {
+      dispatch(touchEventAction(e));
       dispatch(externalEventAction({ handler: onTouchMove, reactEvent: e }));
     };
     const myOnTouchEnd = (e: React.TouchEvent<HTMLDivElement>) => {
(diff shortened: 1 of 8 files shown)


ckifer:
Its probably a bit more complicated than this because we have to consider if the chart is within axis or item event type...
Would be nice just to be able to dispatch the mouseMove middleware and be done with it..

src/chart/RechartsWrapper.tsx
```diff
@@ -96,6 +97,7 @@ export const RechartsWrapper = forwardRef(
       dispatch(externalEventAction({ handler: onTouchStart, reactEvent: e }));
     };
     const myOnTouchMove = (e: React.TouchEvent<HTMLDivElement>) => {
+      dispatch(touchEventAction(e));
```
PavelVanecek, line 100: Do we also need touchStart and touchEnd? Or are they behaving like clicks?

test/component/Tooltip/Tooltip.visibility.spec.tsx
```diff
@@ -409,6 +409,50 @@ describe('Tooltip visibility', () => {
       expect(tooltip2.getAttribute('style')).toContain(expectedTransform);
     });
 
+    it('should move onTouchMove', async context => {
+      // TODO: these charts currently do not work onTouchMove. Did they before?
```
PavelVanecek, line 413: Charts with `item` eventtype will dispatch their events from individual graphical items' event handlers. We might need to add `onTouchMove` there too. Example: https://github.com/recharts/recharts/blob/3.x/src/polar/Pie.tsx#L307

PavelVanecek (approved):
Do you need help testing? I have various touch devices with and without pencils to try out.

test/component/Tooltip/Tooltip.visibility.spec.tsx
ckifer: Yeah I forgot about that. So not quite done but close then 👍

src/chart/RechartsWrapper.tsx
ckifer: It's a little flaky. Sometimes it seems to behave fine like a click, sometimes it does not. Let's test after this merge

ckifer:
Yes please!

PR #5612, opened 2025-02-25
Remove Tooltip state from Treemap, and unused function findChildByType

PavelVanecek:
## Description

This state is no longer necessary now that we have Tooltip state in redux and middleware.

### DIFF
--- a/src/util/ReactUtils.ts
+++ b/src/util/ReactUtils.ts
@@ -104,25 +104,6 @@ export function findAllByType<
 }
 
 /**
- * @deprecated instead find another approach that does not require reading React Elements from DOM.
- *
- * Return the first matched child by type, return null otherwise.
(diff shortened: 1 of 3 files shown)


src/util/ReactUtils.ts
```diff
@@ -103,25 +103,6 @@ export function findAllByType<
...
- * @deprecated instead find another approach that does not require reading React Elements from DOM.
- *
- * Return the first matched child by type, return null otherwise.
- * `type` must be a React.ComponentType
- *
- * @param children do not use
- * @param type do not use
- * @return deprecated do not use
- */
-export function findChildByType<ComponentType extends React.ComponentType>(
```
ckifer, line 116: Sweeeeet

ckifer (approved):

PR #5615, opened 2025-02-25
Remove some test exclusion - all charts now observe Tooltip.active the same, no more exceptions

PavelVanecek:
## Description

This one I fixed when I removed the special bespoke Tooltip handling but forgot to re-enable the tests so here we are.

I couldn't find any issue for it. But `<Tooltip active />` now works in Sankey  and Treemap and Sunburst. Yay.


ckifer (approved):
🚀

Issue #5439, opened 2025-01-10
XAxis (interval="equidistantPreserveStart") is incorrectly calculated in width  and the values on the sides are clipped. 


AlexRafrey:
`<AreaChart>
    <XAxis
        interval="preserveStartEnd"
      />
</AreaChart>`
  Auto calculation of width works incorrectly. It cuts off the values at the edges.
  You can check it here https://codesandbox.io/p/sandbox/wswgrl
  
<img width="731" alt="image" src="https://github.com/user-attachments/assets/1b875db1-14bb-467d-a4df-1e54c92e2722" />

ckifer:
Hi, you just need more margin left. If there is no room in the svg there will be things that get cut off

ckifer:
https://codesandbox.io/p/sandbox/recharts-auto-width-issue-forked-wmcmns
<img width="689" alt="image" src="https://github.com/user-attachments/assets/2b96fb82-fc80-487b-a2e3-1d66c786e5dd" />

This isn't linked to equidistantPreserveStart or any specific interval, it just only takes into account colliding with other ticks at the moment not escaping the bounds of the SVG.

Feel free to open another issue as a feature request (I think there may be some pre-existing ones regarding this though)

Going to close this for now though

PR #5617, opened 2025-02-27
Add new interval type `equidistantPreserveZeroStart`

c0xa:
Add new interval type `equidistantPreserveZeroStart`

## Description

This PR introduces a new axis interval type called equidistantPreserveZeroStart. This interval type ensures that:

1. The first tick always starts at 0.
3. The first tick's label is fully visible and does not overflow the axis boundary.
5. Ticks are evenly spaced with a step N, avoiding collisions between labels.

This is particularly useful for axes where zero is a critical reference point, such as in financial charts or scientific data visualizations.

## Related Issue

https://github.com/recharts/recharts/issues/5439

## Motivation and Context

Currently, the axis interval types in Recharts do not guarantee that the text label for the first tick is fully visible. This can lead to clipped or overlapping labels, which makes the visualization harder to interpret. This change solves this problem by introducing a new interval type that ensures the first tick's label is always fully visible, even if it requires shifting the tick's position slightly. This is particularly important when 0 is a critical reference point, as it ensures clarity and accuracy in the visualization.

## How Has This Been Tested?
```
const data = [
      {
        date: 1704067200,
        value: 1,
        dynamic: 0,
      },
      {
        date: 1706745600,
        value: 1,
        dynamic: 0,
      },
      {
        date: 1709251200,
        value: 1,
        dynamic: 0,
      },
      {
        date: 1711929600,
        value: -2,
        dynamic: -3,
      },
      {
        date: 1714521600,
        value: 0,
        dynamic: 2,
      },
      {
        date: 1717200000,
        value: 0,
        dynamic: 0,
      },
      {
        date: 1719792000,
        value: 0,
        dynamic: 0,
      },
      {
        date: 1722470400,
        value: 0,
        dynamic: 0,
      },
      {
        date: 1725148800,
        value: 0,
        dynamic: 0,
      },
    ];
    return (
      <div className="App" style={{ height: 400 }}>
        <ResponsiveContainer>
          <AreaChart data={data ?? []}>
            <XAxis
              tickLine
              dataKey="date"
              minTickGap={16}
              padding="no-gap"
              interval="equidistantPreserveZeroStart"
              tickMargin={12}
              shiftTicks
            />
            <YAxis
              allowDecimals={false}
              tickCount={4}
              tickLine={false}
              orientation="right"
              axisLine={false}
              tickMargin={8}
            />
            <CartesianGrid vertical={false} />

            <Area dataKey='value' type="linear" strokeWidth={2} />
            <Tooltip />
          </AreaChart>
        </ResponsiveContainer>
      </div>
```

Снимок экрана 2025-02-27 в 17 14 44

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] My change requires a change to the documentation.
- [x ] I have updated the documentation accordingly.
- [ x] I have added a storybook story or extended an existing story to show my changes

### DIFF
--- a/src/cartesian/getEquidistantTicks.ts
+++ b/src/cartesian/getEquidistantTicks.ts
@@ -1,6 +1,6 @@
 import { isVisible } from '../util/TickUtils';
 import { CartesianTickItem } from '../util/types';
-import { getEveryNthWithCondition } from '../util/getEveryNthWithCondition';
+import { getEveryNthWithCondition, getEveryNthWithShiftingFirst } from '../util/getEveryNthWithCondition';
 import { Sign } from './getTicks';
 
 export function getEquidistantTicks(
@@ -9,6 +9,7 @@ export function getEquidistantTicks(
   getTickSize: (tick: CartesianTickItem, index: number) => number,
   ticks: ReadonlyArray<CartesianTickItem>,
   minTickGap: number,
+  shiftFirstTick?: boolean,
 ): ReadonlyArray<CartesianTickItem> {
   // If the ticks are readonly, then the slice might not be necessary
   const result = (ticks || []).slice();
@@ -28,6 +29,7 @@ export function getEquidistantTicks(
 
     // Break condition - If we have evaluated all the ticks, then we are done.
     if (entry === undefined) {
+      if (shiftFirstTick) return getEveryNthWithShiftingFirst(ticks, stepsize, getTickSize);
       return getEveryNthWithCondition(ticks, stepsize);
     }
 
@@ -55,7 +57,11 @@ export function getEquidistantTicks(
 
     if (isShow) {
       // If it can be shown, update the start
-      start = tickCoord + sign * (getSize() / 2 + minTickGap);
+      if (index === 0 && shiftFirstTick) {
+        start = tickCoord + sign * (getSize() + minTickGap);
+      } else {
+        start = tickCoord + sign * (getSize() / 2 + minTickGap);
+      }
       index += stepsize;
     }
   }
--- a/src/cartesian/getTicks.ts
+++ b/src/cartesian/getTicks.ts
@@ -162,8 +162,15 @@ export function getTicks(
   const sign = ticks.length >= 2 ? mathSign(ticks[1].coordinate - ticks[0].coordinate) : 1;
   const boundaries = getTickBoundaries(viewBox, sign, sizeKey);
 
-  if (interval === 'equidistantPreserveStart') {
-    return getEquidistantTicks(sign, boundaries, getTickSize, ticks, minTickGap);
+  if (interval === 'equidistantPreserveStart' || interval === 'equidistantPreserveZeroStart') {
+    return getEquidistantTicks(
+      sign,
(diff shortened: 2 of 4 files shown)


ckifer:
Thanks for the contribution! I'm not 100% sure we should make a new type of interval for this. Why not just make it part of equidistantPreserveStart if it's what's desired?

And if this is the case should we not do it with the last one as well?

What happens if the ticks are angled? 

We're going to need some unit tests that at least cover these code paths. It's a little tough since all of the getSize functions need to be mocked but we should at least be able to assert that the position differs in this mode vs the other.

c0xa:
Thank you for your feedback! I agree that integrating this functionality into equidistantPreserveStart makes sense. I'll update the PR to include this change.

I have a few questions to clarify:

- Handling the last tick:
Is it okay to shift the last tick to the left to ensure its label is fully visible? Are there any cases where the last tick's label gets clipped that I should be aware of?
last tick


- New parameter:
 Would it be acceptable to add a new parameter (e.g., preserveZero: boolean) to control this behavior? This would allow users to opt into ensuring the first tick's label is fully visible.

- Unit tests:
Could you provide more details about the unit tests you'd like to see? Are there any existing test examples in the codebase that I can refer to? Specifically, how would you like the tests to be structured, and what edge cases should they cover?

I'll proceed with the updates once I have your guidance on these points. Thank you for your time and support!

ckifer:
Thinking about this more I'm not as sure about it.

* `equidistantPreserveStart` is not `equidistantPreserveStartEnd` so yes the end tick could get clipped and many times does, especially when the screen size is smaller. So I guess we should leave the last tick alone.
* New parameter: I generally dislike adding new params if we can help it, especially booleans as it makes for more "magic" combinations in order to get certain behavior. And not everyone's chart starts at 0. "preserveStart" is supposed to be that; always preserve the first tick - and it does, its just doesn't shift the view around so that it doesn't get cut off by the SVG viewBox. 
* unit tests: Here are the tests for [getEquidistantTicks](https://github.com/recharts/recharts/blob/3.x/test/cartesian/getEquidistantTicks.spec.ts) and [getTicks](https://github.com/recharts/recharts/blob/3.x/test/cartesian/getTicks.spec.ts). Feel free to follow the same format, but generally we just want assersions on all code paths you added. You can see the reduction in coverage from codecov above.


fyi this can also be accomplished by using a custom tick. Or you can add `XAxis` `padding` so that the tick has more room to display. It seems many charting libraries do this as opposed to a specific interval or mode ([highcharts example](https://jsfiddle.net/oc7yq9ev/3/)). I would rather have ticks adjust when you change these parameters with a custom tick as then you have complete control of everything... but we calculate tick positions in advance so its a bit tough
```tsx
            tick={({payload, ...props}) => {
              return (
                <Text
                  {...props}
                  textAnchor={props.index === 0 ? "start" : "middle"}
                >
                  {payload.value}
                </Text>
              );
            }}
```

or 

`XAxis padding={{ left : 20 }}`

ckifer:
Closing this as stale

Issue #5625, opened 2025-03-02
[3.0 alpha bug] animations are broken


PavelVanecek:
How to reproduce: have a button that swaps dataKey in an Area. The chart animates in 2.x but doesn't in 3.x.

PavelVanecek:
I suspect it's all other animations that are broken too, I haven't tested the others yet.

PavelVanecek:
https://github.com/user-attachments/assets/96de584f-9c41-408b-9692-e4dbbb23e46e

Line animations work okay but do not support interrupts - I will add that too while I'm here.

Also I suspect that https://github.com/recharts/recharts/issues/4935 is a symptom of the same problem

PavelVanecek:
New complication: the initial animation breaks when Legend is present (because Legend moves the whole chart up immediately, so it cancels the first "fly in" animation, and starts a second "move the chart up 20px" animation).

PavelVanecek:
Ha I think I fixed

https://github.com/user-attachments/assets/709983ad-b8b9-48b6-ba99-3b89b34d04f9

remi2257:
Yo Pavel

I see that you are making progress on it. I just want to let you know I also face the problem when changing the data => it is maybe because I go from 2 Areas to 3 Areas, but I always have a "jump" effect. (like if the animation was expecting to last 500ms and then, jump straight to the end

I didn't find any way to retrigger the initial drawing animation when changing the data
- I tried creating 2 different chart components
- I tried using `key` basically everywhere
- I tried to use the `hide` param
- Using different stackids
- I tried to play with the animation params

Is it linked to this bug ? Or do you have any way to fully restart the animation / force the re-render ?


You can [test it here](https://remi-lux.dev/en/bourse/calculateur_interets_composes), you will see the bug when toggling "Consider Inflation"

_Chart config_
```
  const chartConfig = {
    // Classic
    contribution: {
      label: tBourseMisc("contribution"),
      color: "hsl(var(--chart-1))",
    },
    interests: {
      label: tBourseMisc("interests"),
      color: "hsl(var(--chart-3))",
    },

    // With Inflation
    contributionAdjusted: {
      label: t("contributionAdjusted"),
      color: "hsl(var(--chart-1))",
    },
    interestsAdjusted: {
      label: t("interestsAdjusted"),
      color: "hsl(var(--chart-3))",
    },
    totalLost: {
      label: t("totalLost"),
      color: "hsl(var(--chart-2))",
    },
[...]

PR #5627, opened 2025-03-02
Add custom matcher for scale

PavelVanecek:
## Description

So that we have more visibility into the assertions, better than `expect.any(Function)`.

## Related Issue


test/state/selectors/radarSelectors.spec.tsx
```diff
@@ -196,7 +196,10 @@ describe('selectRadiusAxisForBandSize', () => {
       includeHidden: false,
       name: undefined,
       reversed: false,
-      scale: expect.any(Function),
+      scale: expect.toBeRechartsScale({
```
PavelVanecek, line 199: Much better now

ckifer (approved):

PR #5643, opened 2025-03-06
Area animations

PavelVanecek:
## Description

Okay this was more difficult than I thought but I got it working. I tried to keep it as simple as possible but it's still not super intuitive. If you have any ideas how to approach the animation state better I'm all ears.

I added comments throughout the code trying to explain to myself what's going on. The problem was that since we got rid of `updateId` counter in the generator, the animations stopped working completely. This was fairly simple to fix but then I figured while I'm here I might just as well fix the animations interrupting so that now we have animations that can smoothly transition from an in-between state to a new state.

Try it yourself here: https://63da8268a0da9970db6992aa-bvykiscyja.chromatic.com/?path=/story/examples-areachart--with-changing-data-key-and-animations

Compare to 2.x behaviour here: https://codesandbox.io/p/sandbox/areachart-with-changing-datakey-mj9yqz?file=%2Fsrc%2FApp.tsx%3A148%2C28 now that I look at it again, the animations didn't work on 2.x either? Maybe it wasn't alpha bug.

Oh and bonus, this doesn't use lodash anymore. Or any deep equals at all.

## Related Issue

https://github.com/recharts/recharts/pull/5593

## Video

https://github.com/user-attachments/assets/e920ac13-02cf-4b43-852e-f18f5f6d5d64


ckifer (approved):
I've tired to do the animation refactoring a few times each with failure. This is awesome, thank you!

Issue #4480, opened 2024-05-03
Points property not generating for Area Chart in v2.12.6


Millian-22:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/simple-area-chart-forked-mqrs6z?file=%2Fsrc%2FApp.tsx%3A59%2C33-59%2C36)

### Steps to reproduce
1. Navigate to provided [link](https://codesandbox.io/p/sandbox/simple-area-chart-forked-mqrs6z?file=%2Fsrc%2FApp.tsx%3A59%2C33-59%2C36)

2. Open console and view a "props" log 
3. See that no [points property](https://recharts.org/en-US/api/Area#points) is being generated

### What is expected?
Points property is being generated by Area chart. Customized Dots appear on Area Chart

### What is actually happening?
No points property is being generated by area chart. No dots appear.

| Environment | Info |
|---|---|
| Recharts | v2.12.6 |
| React | 18.2.0 |
| System | MacOS |
| Browser | Chrome Version 124.0.6367.119 (Official Build) (arm64) |

---
This works in earlier releases of Recharts (tested up to version 2.1.15) but it stopped working sometime after that.

ckifer:
Looks like between 2.2.0 and 2.3.0 to me.

https://github.com/recharts/recharts/compare/v2.2.0...v2.3.0
A lot to compare there

Millian-22:
@ckifer thanks for the quick response! Looking forward to the update

ckifer:
Cool, it'll be fixed in the next patch (undetermined eta, but at least a week or so)

ckifer:
Fixed in v2.12.7

PR #5657, opened 2025-03-10
Backport Line renderDots to include points object in the dotProps

brodriguezmilla:
## Description

`points` was an object accessible through the Line dots object, and I am using that object to build a graph. That was some versions ago, and I just updated my recharts version and `points` is no longer there, though you just put it back 2 days ago in the 3.x branch.

The issue is the same as the Area fix in https://github.com/recharts/recharts/pull/4481/files?diff=split&w=1, https://github.com/recharts/recharts/issues/4480, https://github.com/recharts/recharts/blob/3.x/src/cartesian/Line.tsx#L257, and https://github.com/recharts/recharts/discussions/5656.

## Related Issue

The issue is the same as the Area fix in https://github.com/recharts/recharts/pull/4481/files?diff=split&w=1, https://github.com/recharts/recharts/issues/4480, https://github.com/recharts/recharts/blob/3.x/src/cartesian/Line.tsx#L257, and https://github.com/recharts/recharts/discussions/5656.

## Motivation and Context

I am using the points object to build a custom graph, I am way behind in the recharts version, and I want to bring my code up to the latest version.

## How Has This Been Tested?

I added almost same test you added for area, and the test passes, not the pipeline because of a different file that fails with and without my change.

Screenshot 2025-03-10 at 3 32 17 PM

## Types of changes

## Checklist:


brodriguezmilla:
Not sure what to do about yarn.lock. Please advise.

ckifer:
recharts does not use yarn, please run npm install before making a PR rather than yarn install

ckifer (approved):
this should be fine. As mentioned in the issue, I'm not sure when this will be released

PR #5661, opened 2025-03-11
Prevent rerenders in selectChartOffset

PavelVanecek:
## Description

selectChartOffset was accepting more properties than it needed. The payload which is extra is updating and that triggers extra re-renders and those extra re-renders break animations in Pie. Because selectChartOffset does not use the payload, we can fix this by using more targeted selectors which is what I did.

Pie animations are still broken because there are more problems but this is a step forward.

## Related Issue


ckifer (approved):
Thanks

Issue #5644, opened 2025-03-07
[Alpha Bug] Axis tick string size calculation does not update with fontSize changes in 3.x


ckifer:
## Do you want to request a feature or report a bug?
Bug

## What is the current behavior?
In 3.x, if you use a different fontSize than the default the fontSize and therefore the tickSize cache does not update

https://codesandbox.io/p/devbox/line-chart-width-xaxis-padding-forked-ftph5c

## What is the expected behavior?

Recharts would behave like 2.x and perform string size calculation based on fontSizes.

We need to fix the code here https://github.com/recharts/recharts/blob/3.x/src/cartesian/CartesianAxis.tsx#L124-L134

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

3.0-alpha

PR #5681, opened 2025-03-16
Set tick fontSize in ref callback instead of onMount

PavelVanecek:
## Description

CartesianAxis no longer rerenders after every update so `componentDidMount` no longer runs as often as it used to in 2.x

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5644


ckifer (approved):
Dang you're good at this. I didn't spend much time on it but I certainly didn't figure to do this lol

PR #5712, opened 2025-03-25
Add SVGElements to payload type

ally1002:
## Description

Added SVGProps<SVGElement> into the Payload type to reference all SVG possible fields at the payload

## Related Issue

#5711

## Motivation and Context

It solves problem when trying to access SVG props at the payload content inside a Tooltip component

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] I have added a storybook story or extended an existing story to show my changes

Issue #5711, opened 2025-03-25
Tooltip Props type does not declare "opacity" for payload


ally1002:
## Do you want to request a feature or report a bug?
bug

## What is the current behavior?
TooltipProps type is not declaring "opacity" field into the payload

![Image](https://github.com/user-attachments/assets/2da0b1e0-8ddf-4558-a5e4-7c8b4ab08be4)

## Code Sandbox of the bug:
`App.tsx:91`

https://codesandbox.io/p/sandbox/line-chart-width-xaxis-padding-forked-jf5qrg

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

I tried using 2.15.1 and 2.15.0

ckifer:
Hi sorry I'm out on vacation until next Wednesday and won't be able to take a look until then. 

opacity is an SVG prop so it is implicit and just passed down the the SVG elements rather than something that is explicitly supported by recharts which is why it isn't in the type. 

@PavelVanecek may or may not be able to look while I'm out

ally1002:
No problem 😄 , I can create a PR to resolve that if you want to

ckifer:
Going to close this as resolved, released in 2.15.2

PR #5715, opened 2025-03-26
Add SVGElements to payload type

ally1002:
#5711
#5712

### DIFF
--- a/src/component/DefaultTooltipContent.tsx
+++ b/src/component/DefaultTooltipContent.tsx
@@ -2,7 +2,7 @@
  * @fileOverview Default Tooltip Content
  */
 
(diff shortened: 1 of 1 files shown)


PavelVanecek (approved):

ckifer:
So I kind of feel like adding all of the SVGProps in the world is a little misleading and it makes the important props (those that recharts actually passes back to the tooltip on purpose) hard to find/discern. There may not be a perfect solution here but just voicing the concern before we merge

PavelVanecek:
What would be a better alternative?

ckifer:
Right, I don't have one at the moment. Adding specific properties as people want them? But that's more work every time

Issue #5717, opened 2025-03-26
Bug: activeBar Prop Not Working Correctly when tooltip shared is false


nizans:
https://github.com/user-attachments/assets/10e11376-d62b-497c-a21f-7a9f5292b2ee

The activeBar prop doesn’t work correctly when the tooltip `shared` is set to `false`. If I understand the prop correctly, it should behave similarly regardless of whether `shared` is true or false, and this definitely looks like a bug.
 
I've tested this on `2.15.1` and `2.12`

This issue was previously mentioned in [#4101](https://github.com/recharts/recharts/issues/4101) and marked as fixed, but it is still reproducible.

Steps to Reproduce:

Set `shared={false}` on any BarChart example on the documentation site.
And hover on bars, the bars marked as active are the wrong ones.

You can also reproduce it using this [CodeSandbox](https://codesandbox.io/p/sandbox/simple-bar-chart-forked-kkz248?file=%2Fsrc%2FApp.tsx).

I've opened a [PR](https://github.com/recharts/recharts/pull/5718) with a fix.

ckifer:
Released in 2.15.2

PR #5718, opened 2025-03-26
Fix for activeBar Prop Not Working with tooltip  shared is false

nizans:
Fixes the activeBar prop not working when tooltip `shared=false`.

## Description

Seems like tooltip data was missing on the `getMouseInfo` returned value when `tooltipEventType !== 'axis'` (shared is false)

## Related Issue

https://github.com/recharts/recharts/issues/5717
https://github.com/recharts/recharts/issues/4101
## Motivation and Context

This fixes a bug that makes the chart unusable when tooltip shared is false and bars has activeBars 

## How Has This Been Tested?
Tested locally using examples in storybook and ran all tests.

## Types of changes

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1421,6 +1421,7 @@ export const generateCategoricalChart = ({
 
       const { xAxisMap, yAxisMap } = this.state;
       const tooltipEventType = this.getTooltipEventType();
+      const toolTipData = getTooltipData(this.state, this.props.data, this.props.layout, rangeObj);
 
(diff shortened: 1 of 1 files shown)


PavelVanecek:
Thank you @nizans for the fix. Have you had a chance to try out the same bug in the `3.x` branch?

ckifer:
I think this is probably already solved in 3.x

nizans:
@PavelVanecek Thanks for the quick review! I didn’t test this on 3.x, but it looks like it’s already fixed there based on @ckifer’s comment.

I'm not sure how releases work here — any expectations on when this will be released as a patch?

PR #5722, opened 2025-03-27
Fix: Attach Tooltip events when shared prop is false

nizans:
## Description

This change relates to the fix in https://github.com/recharts/recharts/pull/5718, and was missed in the previous PR but is required to fully resolve the issue.

## Related Issue

https://github.com/recharts/recharts/issues/5717
https://github.com/recharts/recharts/issues/4101

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1471,11 +1471,11 @@ export const generateCategoricalChart = ({
 
     parseEventsOfWrapper() {
       const { children } = this.props;
-      const tooltipEventType = this.getTooltipEventType();
+
       const tooltipItem = findChildByType(children, Tooltip);
       let tooltipEvents: any = {};
 
-      if (tooltipItem && tooltipEventType === 'axis') {
+      if (tooltipItem) {
         if (tooltipItem.props.trigger === 'click') {
           tooltipEvents = {
             onClick: this.handleClick,


src/chart/generateCategoricalChart.tsx
```diff
@@ -1471,11 +1471,10 @@ export const generateCategoricalChart = ({
 
     parseEventsOfWrapper() {
       const { children } = this.props;
-      const tooltipEventType = this.getTooltipEventType();
       const tooltipItem = findChildByType(children, Tooltip);
       let tooltipEvents: any = {};
 
-      if (tooltipItem && tooltipEventType === 'axis') {
+      if (tooltipItem) {
```
PavelVanecek, line 1477: This change is likely to break more than it fixes - these handlers are not meant to be attached with eventtype item. Which charts and combinations have you tested?

nizans: Hi @PavelVanecek,

Yes, I agree, I noticed it later after opening the PR.

Let me explain the issue, maybe you could point me in the right direction. On the previous [PR](https://github.com/recharts/recharts/pull/5718), I mistakenly thought the change was enough; I probably tested it with other unstaged changes.

After checking again, I noticed that when the tooltip shared prop is false (which results in `tooltipEventType !== 'axis'`), the `parseEventsOfWrapper` function doesn't attach the tooltip events, which causes the bug to persist.

When making the change in the function, it attaches the events, but to the entire bar category and not just to the bar itself, which is also a bug. What I'm trying to do is make the tooltip work correctly with the `shared` prop turned off. The expected correct behavior is for the tooltip to be active when the mouse is hovering over one of the bars itself, not the background of the bar category.

Here is the original issue describing the bug: https://github.com/recharts/recharts/issues/5717

This bug still persist because of the missing events attachment, and the discussed change in this PR (simply attaching the same events when `shared` prop is on or off) makes the tooltip behave the same like the shared prop is on.

PavelVanecek: Have you tried the 3.0.0 alpha version? https://github.com/recharts/recharts/issues/5445 there's a better chance it will work there.

nizans: We're waiting for a stable version before migrating to 3.
I was hoping to resolve this issue for now.
I'll try to take another look at it, find how to fix it without attaching all events.

ckifer: I also have some thoughts here. Will review when I'm back from vacation but agreed that this will break things.

ckifer: From what I can tell everything is working without this change (at least how it was before `shared` broke in 2.x), I don't think making this change is needed.

What are you trying to achieve that you can't without these events?

nizans: @ckifer  
Seems to be that when the `shared` is false the `activeBar` does not work correctly on the bar chart.
In this [PR](https://github.com/recharts/recharts/pull/5718) I thought adding the tooltip data is enough but I noticed that without attaching the event handlers this that function never gets called.

Added a video of the bug


https://github.com/user-attachments/assets/4277d82f-9e6f-4f34-9baf-3a5029b5eeeb

ckifer: Both of these look broken to me which tells me that this isn't the right change to fix the issue

PR #5743, opened 2025-04-06
touchMove handlers for Scatter

PavelVanecek:
## Description

So the approach I selected before wasn't doing well for Scatter - user had to _start_ the touch on a Scatter element for it to work. Scatters are single points so that was a pain. Now I changed it so that I can start the touch anywhere in the chart and then move my finger and the tooltip still appears.

There are multiple problems still, I left it in a comment. We will probably need a synchronizable ID before we can solve it properly. If you have a better idea I am all ears.

## Related Issue

https://github.com/recharts/recharts/issues/5565

### DIFF
--- a/src/util/Constants.ts
+++ b/src/util/Constants.ts
@@ -24,3 +24,33 @@ export const COLOR_PANEL = [
   '#FA7D92',
   '#D598D9',
 ];
+
+/**
+ * We use this attribute to identify which element is the one that the user is touching.
+ * The index is the position of the element in the data array.
+ * This can be either a number (for array-based charts) or a string (for the charts that have a matrix-shaped data).
+ */
+export const DATA_ITEM_INDEX_ATTRIBUTE_NAME = 'data-recharts-item-index';
+/**
+ * We use this attribute to identify which element is the one that the user is touching.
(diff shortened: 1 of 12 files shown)


src/util/Constants.ts
```diff
@@ -24,3 +24,33 @@ export const COLOR_PANEL = [
...
+ * We use this attribute to identify which element is the one that the user is touching.
+ * The index is the position of the element in the data array.
+ * This can be either a number (for array-based charts) or a string (for the charts that have a matrix-shaped data).
+ */
+export const DATA_ITEM_INDEX_ATTRIBUTE_NAME = 'data-recharts-item-index';
+/**
+ * We use this attribute to identify which element is the one that the user is touching.
+ * DataKey works here as a kind of identifier for the element. It's not a perfect identifier for ~two~ three reasons:
+ *
+ * 1. There can be two different elements with the same dataKey; we won't know which is it
```
ckifer, line 38: whew this is a lot of caveats. I guess _most_ of the time its fine so it'll be better than what it is in 3.x today.. but hmmm. 

But yeah I got no solutions for now...

ckifer (approved):

src/util/Constants.ts
PavelVanecek: It's better than 2.x too

Issue #5740, opened 2025-04-03
[3.x Alpha Bug] syncMethod="value" does not sync charts


ckifer:
> this is working in 3.x alpha but syncMethod="value" is not 

 _Originally posted by @ckifer in [#4869](https://github.com/recharts/recharts/issues/4869#issuecomment-2776752705)_

https://codesandbox.io/p/sandbox/line-chart-width-multi-series-forked-kd2jmw

```tsx

      <LineChart width={500} height={300} syncId="anyId" syncMethod={"value"}>
        <XAxis dataKey="x" type="number" domain={[0, 3]} />
        <YAxis dataKey="y" />
        <Tooltip cursor={{ stroke: "red" }} />
        {series.map((s) => (
          <Line dataKey="y" data={s.data} name={s.name} key={s.name} />
        ))}
      </LineChart>
      <LineChart width={500} height={300} syncId="anyId" syncMethod={"value"}>
        <XAxis dataKey="x" type="number" domain={[0, 3]} />
        <YAxis dataKey="y" />
        <Tooltip cursor={{ stroke: "red", strokeWidth: "5px" }} />
        {series2.map((s) => (
          <Line dataKey="y" data={s.data} name={s.name} key={s.name} />
        ))}
        <Brush />
      </LineChart>
```

PavelVanecek:
Interesting thing is - we already have a test case for this: https://github.com/recharts/recharts/blob/3.x/test/chart/LineChart.spec.tsx#L1570

PavelVanecek:
Could it be that the difference is data defined on chart vs data defined on item?

PR #5744, opened 2025-04-06
Fix Tooltip synchronization with numerical domain

PavelVanecek:
## Description

Type mismatch.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5740


ckifer (approved):

ckifer:
Nice find!

Issue #5752, opened 2025-04-11
Animations break/skips when data is updated faster than animation completion


nik-webdevelop:
Example: [codesandbox](https://codesandbox.io/p/sandbox/recharts-bug-skipped-animation-n2dy4z)

Actual behavior: Animations break/skips when data is updated faster than animation completion

Expected Behavior: Smoothly animate to new data

I'm not sure if it's expected behavior. But in cases where data changes rapidly, it looks jumpy. And in cases where only the reference to the data object has changed, but values didn't, there is no animation at all

nik-webdevelop:
Could be related to #1431

ckifer:
Can you try the alpha version? A lot of this should be resolved

ckifer:
If you make the difference in your update noticeable (more than one value) the transition is smooth. The jumpiness is just because its hard to make a very small change (50 to 51) look smooth

> And in cases where only the reference to the data object has changed, but values didn't, there is no animation at all

Why would there be animation if the value didn't change?

nik-webdevelop:
I think the jumpiness is not due to the small difference in values but because the previous animation stopped instantly. 


> And in cases where only the reference to the data object has changed, but values didn't, there is no animation at all

> Why would there be animation if the value didn't change?
```ts
const handleActual = () => {
    setData(([{ number }]) => [{ number: 50 }]);

    setTimeout(() => {
      setData(([{ number }]) => [{ number: 51 }]); <- change this to 50. No animation at all
    }, 0);
  };
```

There is a case when new data can be recalculated multiple times in the component tree 'above' and passed several times to the chart. 

Example: new data has been fetched from BE, was transformed by the hook, and passed to the chart. At the same time, state change triggered hook that transforms data, and a new instance of chart data (with same values) has been passed to chart.

I guess, in this particular example, this could be avoided with useMemo/useCallback. But it wouldn't be necessary if the previous animation wouldn't break.

PavelVanecek:
I tried it on the 3.0 alpha and it's even worse there: https://codesandbox.io/p/devbox/recharts-bug-skipped-animation-forked-lqpy4j?file=%2Fsrc%2FApp.tsx&workspaceId=ws_G5Rnj5KjeM4fNQbHpKaCKS I will have a look

nik-webdevelop:
@PavelVanecek @ckifer thank you guys!

PR #5753, opened 2025-04-13
Return of animationId

PavelVanecek:
## Description

Okay so changing the dataKey has the middle step where selectors return `undefined` which marks a reset for the animation component. Changing the data itself doesn't do that so we need a manual restart. animationId will do that.

https://github.com/user-attachments/assets/1fd29d5b-5fa7-4931-8810-7e2cbe4994b0

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5752


ckifer (approved):
Nice not too bad. And hopefully not as confusing as how animationId was used in the past

PR #5762, opened 2025-04-16
Fix XAxis padding calculation

### LINKED ISSUE #5758: Axis padding calculation for continuous axes is incorrect for number domains

Report a bug.

Axis `padding="gap"` and `padding="no-gap"` (added in #2457) misbehave with numerical values:

Image

https://codesandbox.io/p/sandbox/recharts-recharts-gap-issue-reproduction-47tlmk

Detailed in #2457.

It is broken in both `recharts@2.15.1` and `recharts@3.0.0-alpha.9` with React 19 (and presumably earlier versions of React) on MacOS. I believe this has been a bug since the feature was introduced.

jackfletch:
## Description

This PR fixes the XAxis padding calculation similarly to #5759 except this time for `recharts@3`. The only differences to #5759 are (1) that the domain values are already constrained to numbers (so can use a number-only sort comparator) and (2) that there was another bug in the calculation of the smallest distance between consecutive values in an array.

## Related Issue

#5758

## Motivation and Context

It fixes #5758 for `recharts@3`. 

## How Has This Been Tested?

- Created a storybook story for the XAxis calculated padding feature to demonstrate the change (seen below).
- Reproduced the bug with existing unit tests to confirm it was an issue.
  - Added a datum to XAxis test data to cover numerical inputs with heterogeneous digit counts.
  - Updated the XAxis tests' expectations to match the new test data. Notably, the asserted values for `x` locations in the `padding="gap"` or `padding="no-gap"` unit tests were often off-screen.
- Fixed the bug.
- Reran the unit tests to confirm the fix.
  - The only tests that were affected were the two tests related to `padding="gap"` and `padding="no-gap"`.
  - The new asserted values for `x` locations in these tests were now on-screen in a similar location as before changing the test data.

Storybook story before this PR:
Screenshot 2025-04-16 at 4 14 35 PM

Storybook story after this PR:
Screenshot 2025-04-16 at 4 13 53 PM

## Types of changes

   - (*technically* this is a breaking change but I doubt anyone is depending on this broken behavior)

## Checklist:

### DIFF
--- a/src/state/selectors/axisSelectors.ts
+++ b/src/state/selectors/axisSelectors.ts
@@ -1096,7 +1096,7 @@ export const selectSmallestDistanceBetweenValues: (
       return undefined;
     }
     let smallestDistanceBetweenValues = Infinity;
-    const sortedValues = Array.from(onlyAllowNumbers(allDataSquished.map(d => d.value))).sort();
+    const sortedValues = Array.from(onlyAllowNumbers(allDataSquished.map(d => d.value))).sort((a, b) => a - b);
     if (sortedValues.length < 2) {
       return Infinity;
     }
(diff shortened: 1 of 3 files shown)


ckifer (approved):
lgtm, thank you

PR #5771, opened 2025-04-20
Move resolveDefaultProps to a separate file, add docs and tests

PavelVanecek:
## Description

`resolveDefaultProps` is useful for moving all the class components into functional components so let's have some docs and tests. The types could be better so I will do that next.

To confirm what is the class defaultProps behaviour I wrote this thing in codesandbox, I figured I don't need to add these as tests:

```
import { Component } from "react";

function stringify(val) {
  if (val === undefined) {
    return "undefined";
  }
  if (Number.isNaN(val)) {
    return "NaN";
  }
  return JSON.stringify(val);
}

class Foo extends Component {
  static defaultProps = {
    val: 1,
  };

  render() {
    return stringify(this.props.val);
  }
}

export default function App() {
  return (
    <table>
      <thead>
        <tr>
          <th>Input</th>
          <th>Resolved prop</th>
        </tr>
      </thead>
      <tr>
        <td>No prop at all:</td>
        <td>
          <Foo />
        </td>
      </tr>
      {[undefined, null, false, 0, "0", NaN, "", "undefined", []].map((val) => (
        <tr>
          <td>{stringify(val)}:</td>
          <td>
            <Foo val={val} />
          </td>
        </tr>
      ))}
    </table>
  );
}
```

This shows that only `undefined` gets filled with a default:

image

### DIFF
--- a/src/util/resolveDefaultProps.tsx
+++ b/src/util/resolveDefaultProps.tsx
@@ -1,0 +1,24 @@
+/**
+ * This function mimics the behavior of the `defaultProps` static property in React.
+ * Functional components do not have a defaultProps property, so this function is useful to resolve default props.
+ *
+ * The common recommendation is to use ES6 destructuring with default values in the function signature
+ * but you need to be careful there and make sure you destructure all the individual properties
+ * and not the whole object. See the test file for example.
+ *
+ * And because destructuring all properties one by one is a faff and it's easy to miss one property,
+ * this function exists.
+ *
+ * @param realProps - the props object passed to the component by the user
+ * @param defaultProps - the default props object defined in the component by Recharts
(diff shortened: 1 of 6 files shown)


src/util/resolveDefaultProps.tsx
```diff
@@ -0,0 +1,24 @@
+/**
+ * This function mimics the behavior of the `defaultProps` static property in React.
+ * Functional components do not have a defaultProps property, so this function is useful to resolve default props.
+ *
+ * The common recommendation is to use ES6 destructuring with default values in the function signature
```
ckifer, line 5: This was more of a problem when there was element cloning imo. I guess there still is quite a bit of that

ckifer (approved):

PR #5777, opened 2025-04-21
Add Draggable Pie example

PavelVanecek:
## Description

It needs a bit more polishing but the idea is there. I should figure out the mouse event type and `getChartPointer` looks useful so we should document it.

## Related Issue

https://github.com/recharts/recharts/discussions/5775


ckifer (approved):
Noice

Issue #5768, opened 2025-04-19
enable strictNullChecks in tsconfig.json


PavelVanecek:
The codebase is severely lacking in null checks. Typescript has a setting for it ->> `strictNullChecks`. When I enable it now I get 1266 errors (in `src` folder alone).

Let's bring it down to 0 errors (or as close to 0 as we can) and then enable `strictNullChecks`

PavelVanecek:
I'm putting this on pause until https://github.com/recharts/recharts/issues/5914 is done - it turns out that adding null checks can and does break animations so I don't want to touch it until we have more coverage and confidence.

ckifer:
Got it

PR #5782, opened 2025-04-22
Strict null checks

PavelVanecek:
## Description

1220 -> 1108

## Related Issue

### DIFF
--- a/src/state/selectors/polarAxisSelectors.ts
+++ b/src/state/selectors/polarAxisSelectors.ts
@@ -37,7 +37,7 @@ export const implicitRadiusAxis: RadiusAxisSettings = {
   allowDecimals: false,
   allowDuplicatedCategory: defaultPolarRadiusAxisProps.allowDuplicatedCategory,
   dataKey: undefined,
-  domain: defaultPolarRadiusAxisProps.domain,
+  domain: undefined,
   id: undefined,
   includeHidden: false,
   name: undefined,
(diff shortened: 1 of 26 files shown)


src/state/selectors/polarAxisSelectors.ts
```diff
@@ -19,14 +19,14 @@ export const implicitAngleAxis: AngleAxisSettings = {
...
-  domain: defaultPolarAngleAxisProps.domain,
+  domain: undefined,
   id: undefined,
   includeHidden: false,
   name: undefined,
   reversed: defaultPolarAngleAxisProps.reversed,
   scale: defaultPolarAngleAxisProps.scale,
   tick: defaultPolarAngleAxisProps.tick,
-  tickCount: defaultPolarAngleAxisProps.tickCount,
+  tickCount: undefined,
```
ckifer, line 29: Why make all of these undefined?

src/util/ReactUtils.ts
```diff
@@ -94,6 +94,7 @@ export function findAllByType<
 
   toArray(children).forEach(child => {
     const childType = get(child, 'type.displayName') || get(child, 'type.name');
+    // ts-expect-error toArray and lodash.get are not compatible. Let's get rid of the whole findAllByType function
```
ckifer, line 97: Doesn't this need an @ sign?

ckifer (approved):

src/state/selectors/polarAxisSelectors.ts
PavelVanecek: They are already undefined in the `defaultPolarAngleAxisProps` object, typescript doesn't like that with the strict settings

src/util/ReactUtils.ts
ckifer: Sounds good

Issue #5783, opened 2025-04-22
Expose chart's dimensions, offset, and margin with a hook


volodymyr-kushnir:
Before `v3` the `Customized` component exposed the internal state including chart's dimensions, offset, and margin, which were all useful when positioning custom components. To restore access to this data `recharts` must export a hook that returns a slice of the state. Theoretically should be as simple as `export const useSelectChartOffset = () => useAppSelector(selectChartOffset)` (see https://github.com/recharts/recharts/issues/5445#issuecomment-2819657554 for details) but perhaps the slice could be bigger or smaller and the name shouldn't necessarily match the internals (just in case these eventually change).

ckifer:
Thanks!

PavelVanecek:
Hi @volodymyr-kushnir, thanks for the suggestion. Agreed here these are useful things that we should export and document them.

I would like to use this opportunity to redefine the `offset`. Right now it's a mishmash of everything together. left right top width brush all together and everything optional. Instead of `useOffset` I think we should export `useChartDimensions` and `useAbsolutePagePosition` or something like that.

Which properties exactly from the offset are you using and how?

volodymyr-kushnir:
I'm building a quick and somewhat dirty POC and for the reasons I no longer remember I've opted into rendering "brush links" as part of the actual SVG (thus absolute positioning and no flexbox) and below the `<Brush />` component instead of rendering them "outside" like the `<Legend />` component is laid out. The elements in question are just a bunch of `text` nodes (e.g. "First 10", "Last 10%", "Q3", etc.) with an `onClick` handler that helps the user to navigate/resize the brush in one click instead of dragging. So to snap some of these exactly below the brush and align them left-center-right I had to calculate the `"y"` as `height` + (height of the X axis) + (height of the brush) and `"x"` as either `left` (which is the width of the Y axis, I suppose) or `left + width` or `left + width / 2` with the `"textAnchor"` of `"start"`, `"end"`, `"middle"` accordingly.

[...]

PR #5785, opened 2025-04-23
Export and document useChartWidth, useChartHeight

PavelVanecek:
## Description

These two hooks I think everyone will agree on so let's export from the main and document and agree that they are now a public API?

## Related Issue

https://github.com/recharts/recharts/issues/5783

image


src/context/chartLayoutContext.tsx
```diff
@@ -62,10 +62,48 @@ export const useOffset = (): ChartOffsetRequired => {
...
+ * If the chart has any axes or legend, the `width` will be the size of the chart
+ * including the axes and legend. Meaning: adding axes and legend will not change the width.
+ *
+ * The dimensions do not scale, meaning as user zoom in and out, the width number will not change
+ * as the chart gets visually larger or smaller.
+ *
+ * Returns `undefined` if used outside a chart context.
+ *
+ * @returns {number | undefined} The width of the chart in pixels, or `undefined` if not in a chart context.
+ */
```
ckifer, line 83: I think these are fine but is a combined "useChartDimentions" nicer to consume? 1 hook call and two properties rather than 2 hook calls?

Not a big deal but curious

ckifer (approved):
Sweet, LGTM

src/context/chartLayoutContext.tsx
PavelVanecek: I don't have a strong preference. If it's two different hooks then we avoid the memoization discussion. But that isn't a difficult problem to solve either.

ckifer: Eh let's ship it like this. If there's someone in the aether who cares we can accommodate later but access to the things is what most care about, no matter how

Issue #2804, opened 2022-03-17
Area stacked chart is broken when it's used with log scale


abdulrahmancz:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/stacked-area-chart-forked-eoky0k)

### Steps to reproduce
1. have a basic stacked area chart (I used for the repro the area chart example from the docks)
2. change the yaxis to use log scale, also adjust the domain to start from 1 and enable allowDataOverflow, otherwise the chart won't work at all
3. observe broken are chart, the area of the most bottom series doesn't start form the yaxis as it should, instead it is weirdly above the line

### What is expected?
The area part of the series should start at the yaxis and should end with the actual line of the series.

### What is actually happening?
The area is rendered weirdly above the series line and it is straight between first and second point of the line.

| Environment | Info |
|---|---|
| Recharts | v2.1.6 |
| React | 17.0.1 |
| System | Mac OS 12.3 |
| Browser | Chrome 99.0.4844.74 (Official Build) (x86_64) |

ckifer:
Use `<YAxis scale="symlog" />`

None of the said extra parameters are needed when you do this and it doesn't have the 0 problem. its 99% of the time the log scale you want to use for charts

https://d3js.org/d3-scale/symlog

vschoener:
Hi, we had this issue in the team but our linter raised an error since `symlog` is not available in the type (union) when we ant to use it. It's not referred in the [documentation](https://recharts.org/en-US/api/YAxis#scale) as well.

Do you think these parts could be improved?

PR #5789, opened 2025-04-25
Use d3-scale export for scaleType

torztomasz:
## Related Issue

#2804 

## Motivation and Context

I have stumbled upon issue #2804 where @ckifer suggested using `<YAxis scale="symlog" />`. However when I wanted to make the change in my codebase I immediately got a Typescript error saying that `symlog` is not valid for the type `ScaleType` although the chart was using the symlog scale. I dug down and it turned out that in `parseScale` function there is this fragment of code:

```ts
  if (isString(scale)) {
    const name = `scale${upperFirst(scale)}`;

    return {
      scale: ((d3Scales as Record<string, any>)[name] || d3Scales.scalePoint)(),
      realScaleType: (d3Scales as Record<string, any>)[name] ? name : 'point',
    };
  }
```

I figured out that the type is outdated, so I've modified it so that it can accept any scale that d3-scale exports.

## Types of changes

## Checklist:

### DIFF
--- a/src/util/types.ts
+++ b/src/util/types.ts
@@ -23,7 +23,7 @@ import {
   JSX,
 } from 'react';
 import isObject from 'lodash/isObject';
-import { ScaleContinuousNumeric as D3ScaleContinuousNumeric } from 'victory-vendor/d3-scale';
+import * as d3Scale from 'victory-vendor/d3-scale';
 import type { Props as XAxisProps } from '../cartesian/XAxis';
 import type { Props as YAxisProps } from '../cartesian/YAxis';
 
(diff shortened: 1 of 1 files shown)


PavelVanecek:
Hi @torztomasz, have you tested that the charts actually work with all the available scales?

ckifer:
Same concern here

torztomasz:
> Hi @torztomasz, have you tested that the charts actually work with all the available scales?

Have tested some of them but definitely not all, will do and raport back once I will have a spare moment.

src/util/types.ts
```diff
@@ -92,22 +92,11 @@ export interface ChartCoordinate extends Coordinate {
...
-  | 'time'
-  | 'band'
-  | 'point'
-  | 'ordinal'
-  | 'quantile'
-  | 'quantize'
-  | 'utc'
-  | 'sequential'
-  | 'threshold';
+type D3ScaleType = {
```
ckifer, line 95: so really all we gain here is `scaleSymlog` which we could add to the list manually

The other scales added are 
- `radial` (which is confusing because its a variation of a linear scale)
- sequential variants of scales https://d3js.org/d3-scale/sequential
- diverging variants of scales https://d3js.org/d3-scale/diverging
- `implicit` - which results in an error because it isn't a function and also isn't useful to us https://d3js.org/d3-scale/ordinal#scaleImplicit

We risk adding more breaky scale types when we define it like this.

Can we just add `symlog`? At the very least we need to `Exclude` `implicit`

ckifer (changes requested):

ckifer:
Going to closet this due to inactivity, I will make a PR to add symlog explicitly

ckifer:
https://63da8268a0da9970db6992aa-txzidbsqyf.chromatic.com/?path=/story/examples-linechart--logarithmic-y-axis

torztomasz:
Sorry, a lot has been going on in my life, and I honestly had zero time to test the changes. I'm glad you added `symlog` in the other PR cuz that’s actually what I was originally planning to do.

ckifer:
No worries at all! Thanks for doing this in the first place

Issue #5809, opened 2025-05-03
Some components have type errors if you don't enable allowSyntheticDefaultImports


tfaller:
## Do you want to request a feature or report a bug?
Bug report.

## What is the current behavior?
If your apps tsconfig.json contains the following:

```json
{
    "compilerOptions": {
        "allowSyntheticDefaultImports": false
    }
}
```

some components usage like `<XAxis/>`, `<YAxis/>` and `<ZAxis/>` have type errors like:

> JSX element class does not support attributes because it does not have a 'props' property.

This is because TS won't resolve `import React from 'react'` in the d.ts of theses components.

## What is the expected behavior?

The provided d.ts types of these components should work regardless how your apps `allowSyntheticDefaultImports` is configured.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

I've noticed it in 2.x versions.

ckifer:
Ok now TS should complain if we disobey the rules of disallowing synthetic default imports in 3.x.

Going to close this, will be included in next release(s)

ckifer:
released in 2.15.4

PR #5810, opened 2025-05-03
Import react with namespace import

tfaller:
This fixes components that will have type errors like

> JSX element class does not support attributes because it does not have a 'props' property.

If your app has disabled `allowSyntheticDefaultImports`

## Description

Use the namespace import of react for these components.

## Related Issue

#5809

## Motivation and Context

The provided d.ts types of these components should work regardless how your apps `allowSyntheticDefaultImports` is configured.

## How Has This Been Tested?

Yes, this fixes mentioned type error.

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/XAxis.tsx
+++ b/src/cartesian/XAxis.tsx
@@ -1,7 +1,7 @@
 /**
  * @fileOverview X Axis
  */
-import React from 'react';
+import * as React from 'react';
 import clsx from 'clsx';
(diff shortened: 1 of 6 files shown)


ckifer:
I can't believe the fix is as simple as that 😅

Thank you

ckifer (approved):

PavelVanecek:
Thanks @tfaller . May I ask why only the 6 components and not all? Also is it possible to write a unit or integration test so that this problem won't come back again later?

PR #5811, opened 2025-05-03
Replace lodash with native methods and es-toolkit

### LINKED ISSUE #5266: Moving from lodash to es-toolkit

dayongkr:
## Description

- Removed dependencies on `lodash` and `@types/lodash`
- Added `es-toolkit` as a dependecy
- Replaced with native methods where applicable:
  - `max(array)` -> `Math.max(...array)`
  - `min(array)` -> `Math.min(...array)`
  - `some(array, fn)` -> `array.some(fn)`
- Replaced with `es-toolkit` functions:
  - `range`, `omit`, `maxBy`, `sumBy`, `last`, `throttle`, `isPlainObject`
- Replaced with `es-toolkit/compat` for `lodash` compatibility (for functions not provided or not fully covered by `es-toolkit` itself):
  - `get`, `sortBy`, `mapValues`, `every`
- Update ESLint config to ignore resolver issues specifically for `es-toolkit/compat`

While doing the migration, I noticed that `recharts` doesn't rely heavily on `lodash`'s more complex features, which aligns well with the philosophy of the `es-toolkit` project - so I decied to open this PR.

And I hope the v3 release goes really well 😄 

## Related Issue

close #5266

## Motivation and Context

I saw the issue about migrating from `lodash` to `es-toolkit` and native methods(#5266), and after reviewing the `recharts` code, I thought it was feasible to drop `lodash` entirely! 

In my view, one major issue with `lodash` is that it sacrifices performance for edge cases, impacting the majority of use cases. It also reimplements many features already available via native methods, which adds unnecessary overhead. On top of that, it's no longer actively maintained, making it harder to address issues.

To address these problems,`es-toolkit` focuses on simplicity - covering the 90% most common cases, avoiding overly complex functions, and leveraging native methods for better performance. It also bundles types, ESM, and CJS in a single pakcage, so there's no need to install multiple packages like `lodash-es` or `@types/lodash`, and bundle size is smaller. We also maintain over 99% test coverage to ensure stability.

If you're hesitating because it's a newer library, I totally understand. But I want to share that Storybook already uses es-toolkit in its core, and more packages are gradually starting to adopt it as well.

Thank you!!

## How Has This Been Tested?

I ran the existing Storybook, Vitest, and ESLint tests.

### Bundle Size(KB)

| AS-IS | TO-BE |
| --- | --- |
|Screenshot 2025-05-03 at 9 55 04 PM|Screenshot 2025-05-03 at 10 43 13 PM|

### Benchmarks

Screenshot 2025-05-03 at 10 49 01 PM
Screenshot 2025-05-03 at 10 49 36 PM
Screenshot 2025-05-03 at 10 50 01 PM
Screenshot 2025-05-03 at 10 50 13 PM

In most cases, you can see that `es-tookit` is faster.

## Types of changes

## Checklist:


ckifer (approved):
Hey @dayongkr thanks for doing this!

To be honest there is a lot less usage than I thought there was, I guess we did clean up quite a bit in the other PRs linked in that issue. 

Will let @PavelVanecek review as well but LGTM I'm definitely for this. Might as well do it before the release too

PavelVanecek:
Thanks for the PR. I wonder how difficult would it be to get rid of the whole dependency? So all vanilla instead of es-toolkit. We would need to reimplement `get` for sorting, what else?

ckifer:
Why not just do this for now and go from there? It checks all the boxes, makes tree-shaking, happy, and fixes the synthetic imports issue along the way

ckifer:
Looks like the compat functions are 

`mapValues`, `every`, `get`, `sortBy`, `first`, and `uniqueId`

---

I'm going to merge this as it is progress, but we can totally work to remove any dependency should we want to

dayongkr:
Sorry for the late reply. I also strongly agree with the opinions of maintainers who actively use native methods!

However, implementing all utility functions from scratch takes too much time when considering types, performance, and so on. That’s why I wanted es-toolkit to support things that can’t be easily done with native methods, and I’ve replaced some functions with es-toolkit dependencies for now.

Thank you for reviewing it, and I’ll be back again! Thanks a lot!

PavelVanecek:
This fails to run in an integration test, @dayongkr do you know what could be the problem? https://github.com/recharts/recharts-integ/pull/60

kien-ngo:
thanks so much for this PR!

PR #5812, opened 2025-05-03
Disable allowSyntheticDefaultImports

tfaller:
Disables the tsconfig.json `allowSyntheticDefaultImports`.

## Description

As mentioned in #5809 apps can have type issues if they don't have `allowSyntheticDefaultImports` enabled. PR #5810 fixes some of these issues for 2.x. But as @PavelVanecek comments, this is not a permanent fix. The actual fix is to disable `allowSyntheticDefaultImports` (disabled is also the TS default). This PR does all necessary changes to make this possible.

* Update all violating imports
* ~~Migrate from lodash to lodash-es~~
* Disable allowSyntheticDefaultImports
* Remove esModuleInterop (not needed anymore as well because of these changes)

~~We could still use lodash instead of lodash-es, but this results in larger bundles, because tree shaking is not possible without a Babel or Webpack plugin. You can't do `import get from 'lodash/get'` anymore with `allowSyntheticDefaultImports` disabled, only `import { get } from 'lodash'`.~~

~~The UMD bundle `Recharts.js` decreased with this PR from 494K to 487K. While a bundle which still uses lodash (without working tree shaking) would have 528K.~~

## Related Issue

#5809 and #5810 

## Motivation and Context

Improved compatibility and alignd settings of `allowSyntheticDefaultImports` and `esModuleInterop` according the TS default.

## How Has This Been Tested?

Yes. All tests succeed as before, plus type issues are resolved.

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -1,6 +1,7 @@
 // eslint-disable-next-line max-classes-per-file
-import React, { MutableRefObject, PureComponent, useCallback, useMemo, useRef, useState } from 'react';
-import clsx from 'clsx';
+import * as React from 'react';
+import { MutableRefObject, PureComponent, useCallback, useMemo, useRef, useState } from 'react';
+import { clsx } from 'clsx';
 import Animate from 'react-smooth';
 import { Curve, CurveType, Point as CurvePoint, Props as CurveProps } from '../shape/Curve';
 import { Dot } from '../shape/Dot';
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -2,17 +2,9 @@
  * @fileOverview Render a group of bar
  */
 // eslint-disable-next-line max-classes-per-file
-import React, {
-  Key,
-  MutableRefObject,
-  PureComponent,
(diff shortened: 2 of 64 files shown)


ckifer:
Maybe we just want to move to es-toolkit? 

This PR is already pending

https://github.com/recharts/recharts/pull/5811

--- 

Either way though, wouldn't this break the fact that we publish as commonjs? I guess babel should take care of that?

tfaller:
Yes, I'll rebase against #5811 when it lands.

> Either way though, wouldn't this break the fact that we publish as commonjs? I guess babel should take care of that?

Yes. Babel should take care regarding lodash-es. On the other hand, `allowSyntheticDefaultImports` is just TS type checking and won't change the output. `esModuleInterop`is a bit more complex, but affects generated imports, but not exports. So no breaking there as well.

ckifer:
@tfaller es-toolkit changes merged

ckifer (approved):
LGTM, thanks!

ckifer:
Updated the description to remove the stuff about lodash in case anyone looks back at this

PR #5815, opened 2025-05-05
Don't use categoricalDomain for time scale or custom scale ticks

sfishel18:
## Description

First off, Recharts is an excellent charting library, thank you for all of the hard work that goes into it!

I'm trying to use it to build a chart with a time-based x-axis like this:

```js
<LineChart data={...}>
  [...]
  <XAxis type="number" scale="time" {...}  />
</LineChart>
```

What I'm finding is that the x-axis ticks do not snap to "round" time values and time intervals the way I would want them to. I tracked this down to the code I'm changing here, which uses the `categoricalDomain` instead of letting the d3 scale's `ticks` method choose the ticks.

## Related Issue

https://github.com/recharts/recharts/issues/5816

## Motivation and Context

Without this change, library consumers have to compute the ticks in user-land code and pass those tick values explicitly to `XAxis` component. The composability/ease-of-use of the `XAxis` component would be improved if the default behavior for a `"time"` scale could be relied upon to intelligently choose axis ticks.

## How Has This Been Tested?

I tested this by creating a Storybook example where different time-series data sets can be generated and ensuring that the tick selection looks correct

Screenshot 2025-05-05 at 10 21 55 AM

## Types of changes

## Checklist:

### DIFF
--- a/src/state/selectors/axisSelectors.ts
+++ b/src/state/selectors/axisSelectors.ts
@@ -1683,8 +1683,9 @@ export const combineAxisTicks = (
     return result.filter((row: TickItem) => !isNan(row.coordinate));
   }
 
-  // When axis is a categorical axis, but the type of axis is number or the scale of axis is not "auto"
-  if (isCategorical && categoricalDomain) {
+  // When axis is a categorical axis, but the type of axis is number or the scale of axis is not "auto".
+  // Unless the scale is a time scale or a custom scale, in which case defer to the scale.ticks function below.
+  if (isCategorical && categoricalDomain && axis.scale !== 'time' && typeof axis.scale !== 'function') {
     return categoricalDomain.map(
       (entry: any, index: number): TickItem => ({
         coordinate: scale(entry) + offset,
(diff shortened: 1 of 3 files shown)


src/state/selectors/axisSelectors.ts
```diff
@@ -1683,8 +1683,9 @@ export const combineAxisTicks = (
     return result.filter((row: TickItem) => !isNan(row.coordinate));
   }
 
-  // When axis is a categorical axis, but the type of axis is number or the scale of axis is not "auto"
-  if (isCategorical && categoricalDomain) {
+  // When axis is a categorical axis, but the type of axis is number or the scale of axis is not "auto".
+  // Unless the scale is a time scale or a custom scale, in which case defer to the scale.ticks function below.
+  if (isCategorical && categoricalDomain && axis.scale !== 'time' && typeof axis.scale !== 'function') {
```
ckifer, line 1688: I think this is a _good_ change, but I don't want this particular function to become this place where start adding more conditions in order to avoid the below block of code.

I'm not sure if there are more scale types within d3-scale that would also fall into this bucket, but if there are I'd rather not list them out here.

Will let @PavelVanecek review and see if he has any ideas

ckifer:
going to close this as stale, please comment back or reopen this Pr if you'd like

Issue #5788, opened 2025-04-24
[beta] TooltipContentProps - label can be undefined or numeric


ckifer:
Something small that I just saw: `TooltipContentProps` sets `label: string;`, but it seems that it can also be undefined.

<img width="215" alt="Image" src="https://github.com/user-attachments/assets/540c139a-0259-4e81-81a2-fdc53428a20a" />

~~This happens when I use `hide` on `<XAxis/>`. But I actually rely on the `label` being available for some logic inside `<Tooltip content={}/>`.~~

This happens when I use `<XAxis dataKey="index"/>`. Are there reserved data keys that should not be used? It worked for v2. Something else then `"index"` works.

Also `label` can actually be a number!

<img width="172" alt="Image" src="https://github.com/user-attachments/assets/075ca347-6641-4ea5-a813-62ee55a16b7b" />

_Originally posted by @donaldpipowitch in https://github.com/recharts/recharts/issues/5445#issuecomment-2826942144_

ckifer:
@donaldpipowitch would you be able to reproduce this for me? (the dataKey="index" part)

I can't seem to https://codesandbox.io/p/devbox/line-chart-width-xaxis-padding-forked-qcchcg?file=%2Fsrc%2FApp.tsx%3A86%2C9&workspaceId=ws_G5Rnj5KjeM4fNQbHpKaCKS

donaldpipowitch:
@ckifer This is weird. I cannot reproduce it anymore even with `3.0.0-beta.0`. Could a dependency of recharts have caused it...? But maybe this is due to a new side effect: when I remove `dataKey="index"` completely it seems to fallback to some internal "index" which coincidentally matches my index values. You can see this also in your example. If I use `dataKey="index"` (and remove `hide` from the x axis) you can see 10, 20, 30... If I delete `dataKey="index"` you see 0, 1, 2... (which matches my local values). Is this new behaviour or was it always like this?

Anyway I guess this issue is just about fixing the types for `label` now. Thanks for having a look!

ckifer:
@donaldpipowitch I think this is just a coincidence. If recharts has no data specified to work with but it has an `XAxis` it will just make its own serial numbers for as many datapoints as there are. This was the same in 2.x

---

The types should be fixed in 3.00.-beta.1 :)

PR #5817, opened 2025-05-05
Update tooltip `label` type

ckifer:
## Description

Update `TooltipContent` `label` prop to be undefined, string, or number.

This might not always be 100% true because this is based on the users data but it should be true most of the time

## Related Issue

#5788

## Motivation and Context

#5788

be more accurate in terms of types

## Types of changes

    - Anyone who currently references `label` as a value that is required or as solely a `string` will get a type error

## Checklist:


ckifer:
Going to ship this

Issue #5804, opened 2025-04-30
[3.0 Beta] PieChart animation interruption onMouseOver


ckifer:
## Do you want to request a feature or report a bug?

Bug

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar (template: https://codesandbox.io/p/sandbox/simple-line-chart-v25dl4).

https://63da8268a0da9970db6992aa-tcfvxoomhy.chromatic.com/?path=/story/examples-pie-piewithtooltip--pie-with-tooltip

Mouse over immediately after load of the page and before animation completes. Animation pauses until you mouse out of the chart.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Animation continues while mouse over event occurs

https://github.com/user-attachments/assets/f136e176-e72d-4078-9148-d6ef71addc40

ckifer:
Seems to not be an issue in codesandbox https://codesandbox.io/p/sandbox/pie-chart-with-customized-label-forked-hmd8sr

ckifer:
Essentially anything that causes a re-render causes this. Which is kinda gross. 

If you set state in event handlers it will interrupt animation when those handlers are fired, even if said state isn't passed back to the chart. Still this doesn't seem to be an issue for other charts.

ckifer:
This can be reproduced in any chart. It seems that when state is set in an event handler such as `onMouseMove` the `Animation` component unmounts and calls `onAnimationEnd`. This gets called hundreds of times in a second in this case.

Attempted to remove `animationId` which makes it worse. Attempted to try and not do any of the other tooltip actions when `position` is sent in. None of these help much.

ckifer:
linking details from PR https://github.com/recharts/recharts/pull/5839#issuecomment-2876987676

ckifer:
For anyone who runs into this it isn't really "fixed", but there isn't much we can do about it. All user functions need to be memoized in order to prevent the rerenders that this causes

PR #5839, opened 2025-05-13
fix: remove pie animation id

ckifer:
## Description

Removing Pie `animationId` as it creates a regression from 2.x where the Pie animation pauses upon state being set in its event handlers. Removing the id doesn't cause any differences in instances where there is no custom state being set or passed

## Related Issue

https://github.com/recharts/recharts/issues/5804

## Motivation and Context

solve issues before release

## How Has This Been Tested?

- test pass. See difference in storybooks
  - Current, bad - https://63da8268a0da9970db6992aa-tcfvxoomhy.chromatic.com/?path=/story/examples-pie-piewithtooltip--pie-with-tooltip

## Types of changes

## Checklist:

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -51,7 +51,6 @@ import { SetTooltipEntrySettings } from '../state/SetTooltipEntrySettings';
 import { selectActiveTooltipIndex } from '../state/selectors/tooltipSelectors';
 import { SetPolarLegendPayload } from '../state/SetLegendPayload';
 import { DATA_ITEM_DATAKEY_ATTRIBUTE_NAME, DATA_ITEM_INDEX_ATTRIBUTE_NAME } from '../util/Constants';
-import { useAnimationId } from '../util/useAnimationId';
 import { resolveDefaultProps } from '../util/resolveDefaultProps';
 
 interface PieDef {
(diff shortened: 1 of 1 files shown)


PavelVanecek:
The animationId is there to fix this thing: https://github.com/recharts/recharts/pull/5753 would you like to add a story that changes the data in a Pie and see how that animates?

How difficult would it be to not re-render the Pie on mouse over? That doesn't seem like it should be related. Why would Pie need to rerender when tooltip shows up.

ckifer:
Ok so correct without animationId there is no re-animation when data changes.

---

Minimally, this happens as soon as you set _any_ state inside _any_ of the event handlers, which I think it certainly a regression. Reducing that story to this 

```tsx
onMouseEnter={() => {
    setActive(true);
}}
```

Without passing `active` anywhere or doing anything else still results in the same issue. So its within Pie.

Adding something like
```ts
  Object.keys(input).forEach(key => {
    // @ts-expect-error bleh
    if (prevProps.current[key] !== input[key]) console.log('Changed', key, prevProps.current[key], input[key]);
  });

```

to `useAnimationId` results in:

```
Changed onMouseEnter
Changed children 
Changed sectors 
Changed onMouseEnter 
Changed children 
Changed sectors 
Changed onMouseEnter 
```

on any mouse enter movement whereas on initial render we only have `Changed sectors`

If I go to any `Pie` without event handlers, this doesn't happen and `sectors` change on initial render alone.

ckifer:
I don't really think there is a way around it - if you `setState` you are causing a re-render. Using a ref "works" but it doesn't cause anything to update and is therefore useless. We also can't tell our consumers to do that and them be ok with it even if it did work.

Issue #5738, opened 2025-04-03
X-Axis texts are cut when custom tick is used and body font-size is not default


AryamnovEugeniy:
Hi everyone!

## Do you want to request a feature or report a bug?

Bug report.

## What is the current behavior?

Sometimes X-Axis tick texts are measured incorrectly by recharts and get cut (or overlap each other) if custom ticks are used and body font-size is set to `8px`.

## Steps to reproduce:

1. Open this demo: https://codesandbox.io/p/sandbox/line-chart-with-customized-label-forked-7xhv4d
2. Take a look at the very first X-Axis tick. It displays 7.03 instead of 07.03:

<img width="229" alt="Image" src="https://github.com/user-attachments/assets/1d7b482a-cbe4-4d56-9069-ebc288c8a93a" />

3. Notice that body has font-size of 8px: https://codesandbox.io/p/sandbox/line-chart-with-customized-label-forked-7xhv4d?file=%2Fpublic%2Findex.html%3A29%2C1-29%2C24 - that is how my frontend is configured and I cannot change it.
4. If [this line](https://codesandbox.io/p/sandbox/line-chart-with-customized-label-forked-7xhv4d?file=%2Fsrc%2FApp.tsx%3A79%2C1-80%2C1) is uncommented and the page with chart is reloaded - everything starts to work correctly.

Possible cause of the issue: if we take a look [here](https://github.com/recharts/recharts/blob/3.x/src/cartesian/CartesianAxis.tsx#L342) we can see that ticks are queried by their classnames. And className `recharts-cartesian-axis-tick-value` is not passed in props to custom tick. So, the `fontSize` is not set and further calculations are somehow affected.

## What is the expected behavior?

Tick texts should be measured correctly by recharts and placed according to their actual sizes.
## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
recharts: 2.15.1, in 2.7.2 the problem is also present
OS: MacOS
Browsers: Chrome, Safari.

Is this behaviour expected and I should do something on my side to get charts working correctly or is this a bug?

ckifer:
For now your workaround would be to set the className on your custom tick component. 

But where you linked the code from can be changed to always pass the className regardless if it's custom or not. Labeled as a bug

AryamnovEugeniy:
[...]

PR #5859, opened 2025-05-25
fix: combine className in CartesianAxis, and edit assertion in xaxis spec test

MyungAe:
## Description

1. When a custom tick component is rendered, ensure that the 'recharts-cartesian-axis-tick-value' class name is included.
(along with any existing className specified by the user if needed)

2. The expected values in the xaxis spec test vary depending on the user's locale.

## Related Issue

#5738 

## Motivation and Context

1. When using a custom X-axis tick component (a user-defined component passed to the tick prop of XAxis), the CSS class (recharts-cartesian-axis-tick-value) that Recharts internally uses to measure text size is not properly applied (passed) to this custom tick component, resulting in text being cut off due to text size calculation errors, especially when the font-size of the body is different from the default value.

2. I'm a Korean, and I encountered a locale-related issue in an unchanged part (xaxis.spec.tsx). To resolve this problem, I tried setting vitest environment variables like LC_ALL, LC_MESSAGE, LANG, LANGUAGE and referencing Node.js type definitions, but the issue wasn't resolved. So I ended up modifying part of the test content.

## How Has This Been Tested?

After modifying the library, I tested it in a local environment using npm pack.

enviroments
1. Chrome, Mac m1 Sequoia (15.4.1)
2. Safari, Mac m1 Sequoia (15.4.1)

Screenshot 2025-05-25 at 9 41 29 PM

Screenshot 2025-05-25 at 9 41 48 PM

## Types of changes

## Checklist:

### DIFF
--- a/test/cartesian/XAxis.spec.tsx
+++ b/test/cartesian/XAxis.spec.tsx
@@ -351,6 +351,21 @@ describe('<XAxis />', () => {
         scaleTypeSpy(useAppSelector(state => selectRealScaleType(state, 'xAxis', 0)));
         return null;
       };
+
+      const tickFormatter = (value: number) =>
+        new Date(value).toLocaleString('en-US', {
+          weekday: 'short',
+          year: 'numeric',
+          month: 'short',
+          day: '2-digit',
+          hour: '2-digit',
+          minute: '2-digit',
+          second: '2-digit',
+          timeZone: 'UTC',
+          timeZoneName: 'long',
+          hourCycle: 'h23',
+        });
+
       const { container } = render(
         <LineChart width={600} height={300} data={timeData} margin={{ top: 5, right: 30, left: 20, bottom: 5 }}>
           <XAxis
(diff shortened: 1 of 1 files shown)


PavelVanecek:
Deleting content in brackets will likely lead to more errors later, let's not do that.

There's a better solution for the xaxis test mismatch. I've done it in the storybook already, can you copy it out of there? I think it was tick formatter or something similar.

MyungAe:
The solution seems to be setting 'en-us' in Date.toLocaleDateString within the tick formatter, but this doesn't appear to be an appropriate solution.

The issue is that the '(Coordinate Universal Time)' part gets translated into the user's language based on their locale, and even when using toLocaleDateString, you would still need to use replace or similar methods to add the 'GMT+0000 (Coordinate Universal Time)' part.

Instead, how about replacing the user-language-translated portion in the return value of the Date.toString method with '(Coordinate Universal Time)' in tick formatter? It's a method of adding a tick formatter to one of the tests and including the replace part in that tick formatter.

code example :
`const tickFormatter = (value: number) => {
  const date = new Date(value);
  return date.toString().replace(/GMT.*$/, 'GMT+0000 (Coordinated Universal Time)');
};`

ckifer:
Setting en-US seems like a fine solution to me. We don't need it to change based on someone's locale, it's just a test that we want to always pass regardless of locale. At the end of the day, it doesn't really matter how it's done I don't think

MyungAe:
I'll modify the commit to create the tickFormatter using the 'en-US' option : )

MyungAe:
I modified some expected values in the tests due to the different formats returned by Date.toString() and Date.toLocaleDateString(), and designed this change to not affect other tests.

ckifer (approved):

PR #5871, opened 2025-05-28
codecov-bundle-size

PavelVanecek:
## Description

... I missed that the template has dryRun: true. No wonder it didn't upload anything.


PavelVanecek:
It shows quite different numbers from what my filesystem shows (codecov: 945 kB, my folder: 1360 kB), not sure why is that. I suppose we're more interested in the trend than the exact number.

PR #5878, opened 2025-05-29
Add lodash.get as a dependency

PavelVanecek:
I want to see how many kB does this add in the bundle


PavelVanecek:
It's 1 out of 2

ckifer:
It looks like there's quite a bit no?

Get, uniqBy, every, mapValues, first?

---

And you're just talking about removing es-toolkit/compat? Not the lib itself?

PavelVanecek:
I was aiming at es-toolkit/compat first. But tbh there isn't much that we use.

ckifer (approved):

PR #5879, opened 2025-05-29
Fix interrupting Pie animation

PavelVanecek:
## Description

the solution, I'm afraid, is firmly in the userland. We can't do much if the upstream component changes props on every render.

The `Cell` API is not helping. I have a plan to get rid of it but I will save it for Recharts@4.

## Related Issue

Closes https://github.com/recharts/recharts/issues/5804

Alternative to https://github.com/recharts/recharts/pull/5839

Bonus: Data changing animation still works

https://github.com/user-attachments/assets/b8692425-65ca-4322-b448-6e80f6d79c39


ckifer:
Oof, I spent so much time trying to do this inside the lib 🙃 

Welp, I guess something is better than nothing. Why is this so slow btw?

ckifer (approved):

Issue #5861, opened 2025-05-26
y-axis width should be dynamic


saurabhraj123:
## Do you want to request a feature or report a bug?
I want to work on this issue. 

## What is the current behavior?
A fixed width of 60px is given to the y-axis by default. If the label size increases, it overflows. We can calculate the width of the maximum label instead and use it as the width of y-axis.

## What is the expected behavior?
The width should be calculated dynamically based on the maximum value.

I have explored the codebase a bit, I see there is this selector `selectYAxisSize` which gives us the width. I was thinking of doing something like this. Is it the correct approach?

```javascript
export const selectYAxisSize: (
  state: RechartsRootState,
  axisType: XorYType,
  yAxisId: AxisId,
  isPanorama?: boolean,
) => Size | undefined = createSelector(
  selectChartOffset,
  selectYAxisSettings,
  selectTicksOfAxis,
  (offset: ChartOffset, axisSettings: YAxisSettings, ticks: ReadonlyArray<CartesianTickItem>): Size | undefined => {
    if (axisSettings == null) {
      return undefined;
    }
    const measureText = getTextWidth(16);

    const maxValue = ticks?.[ticks.length - 1]?.value;

    const maxValueWidth = maxValue ? measureText(maxValue.toString()) : 0;

    return {
      width: maxValueWidth || axisSettings.width,
      height: offset.height,
    };
  },
);
```

ckifer:
Hi, thanks for wanting to work on this. 

We're trying to get 3.0 out the door and then looking into enhancements like this one after that. If you'd like to make a PR, feel free. I think I agree with the general premise of this, though there are other ways to take care of the issue too such as formatting your numbers with abbreviations, wrapping text, or angling ticks

---

Measuring string size is an expensive operation so just keep that in mind

PavelVanecek:
I think we can close this as done. YAxis now allows `width='auto'`. Thx @saurabhraj123

saurabhraj123:
Hi @PavelVanecek, I will pick the up the follow up task on it soon. Had a few things in my plate recently, but it's on my radar.

Issue #2027, opened 2020-02-12
How to remove gap between yAxis label start and recharges surface.


Gunner17:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[http://recharts.org/en-US/examples](http://recharts.org/en-US/examples)

### Steps to reproduce
Render any line, bar or area chart and go to inspect element to see that there is always some gap between the yAxis label start and the chart surface. 

### What is expected?
Is there a way to make sure that whatever be the yAxis label width, there is no gap/space between the longest yAxis label and the chart surface. Setting chart left margin to 0 does not eliminate this issue.

### What is actually happening?
Seems like since the default yAxis width is 60, there is always a gap between the surface edge and the yAxis label. 

| Environment | Info |
|---|---|
| Recharts | v1.8.5 |
| React | 16.8.5 |
| System | Mac |
| Browser | Safari/Chrome |

Screenshot:
![Screen Shot 2020-02-12 at 11 18 50 AM](https://user-images.githubusercontent.com/11354728/74369746-c0946e00-4d8a-11ea-8e61-b6ade4c41b8c.png)

mkaraula:
+1

zhangwins:
+1

petertenhoor:
+1

refrigerator:
+1

vlad-buzea-ullink:
+1

GideonCS:
+1 This is still happening

gfgabrielfranca:
+1

xentobias:
+1

ajoabraham:
+1

zhengissov:
+1

viniciusdacal:
+1

tactycadmin:
+1

nerdmax:
+1 any updates?

sgwanlee:
Use `width`
```
<YAxis width={20} ... />
```

teemu47:
**Workaround 1**
Calculate the length of the longest label and set `YAxis` `width` according to the result. For example:
```
const longestLabelLength = data
  .map(c => c.displayName)
  .reduce((acc, cur) => (cur.length > acc ? cur.length : acc), 0)

<YAxis
  type={'category'}
  dataKey={'displayName'}
  width={longestLabelLength * 10} // adjust the number depending on your font
/>
```

**Workaround 2:**
Set a constant `width` and shorten the labels. For example:
[...]

Issue #947, opened 2017-10-10
[Feature]: Is it possible to make y-Axis width auto calculated?


IvanKalinin:
Hi there. So, everything is in the header.
Currently we can only adjust fixed width for y-Axis. It would be nice to have an option for auto width. That  could remove extra white space on the container left edge.
Design 
![image](https://user-images.githubusercontent.com/12299212/31399364-e569c450-ada0-11e7-9c74-2542651f1a2b.png)

vs Real world

![image](https://user-images.githubusercontent.com/12299212/31399397-fda238f4-ada0-11e7-8a4b-5307b033ce7a.png)

itwasmattgregg:
+1

xile611:
Duplicated with #2027

saroj-nf:
i am looking for same

marluscw:
+1

is there any hack to fix this?

uixmat:
still looking for a way to do this 🙏🏼

PR #5880, opened 2025-05-29
Add support for responsive y-axis width

saurabhraj123:
## Summary

A fixed width of 60px is given to the y-axis by default. If the label size increases, it overflows. Added a prop named `isWidthResponsive`, which dynamically calculates the **yAxis width** based on the width of the ticks.

Refer to this issue: #5861

## Description
I have added a new prop named `isWidthResponsive`. When this is made true, it dynamically calculates the width of the y-axis when:
* The `label` prop is either a string/number or an object.
* If the `label` prop is a function or a react element, default width of 60px is used as the old flow.

The following logic is used to achieve the same:
* Calculate width of each y-axis label and find out the maximum width.
* If valid `label` prop is present, it calculates its width as well along with `tickWidth` and `tickMargin`.
* Now the final `calculated width = maxWidth + tickWidth + tickMargin + 5px`. A 5px gap is for gap between the tick and the y-axis label. 

This PR rotates the y-axis label to 270 degrees by default as its the most commonly used layout.

## Related Issue
#5861 #2027

## Motivation and Context
This feature is long awaited and being asked for since 2017 - Refer #947. When I worked on it, I had to do a lot of customisations to achieve this but I feel this feature should be present by default.

## How Has This Been Tested?
I have done manual testing by examining different charts in the storybook. Also, the automated test cases are used.

**Before:**

Screenshot 2025-05-29 at 11 54 04 PM

**After:**
Screenshot 2025-05-29 at 11 55 12 PM

**Without labels:**
Screenshot 2025-05-29 at 11 56 34 PM

## Types of changes

## Checklist:


ckifer:
Can you add some unit tests

PavelVanecek:
Before looking into code, is it possible to have the API as `width="auto"`? If we add a new prop then we have to define what happens when people pass in both `width` and `isWidthResponsive`. Discoverability suffers.

src/cartesian/CartesianAxis.tsx
```diff
@@ -343,7 +348,10 @@ export class CartesianAxis extends Component<Props, IState> {
         className={clsx('recharts-cartesian-axis', className)}
         ref={ref => {
           if (ref) {
-            const tick: Element | undefined = ref.getElementsByClassName('recharts-cartesian-axis-tick-value')[0];
+            const tickNodes = ref.getElementsByClassName('recharts-cartesian-axis-tick-value');
+            this.tickRefs.current = Array.from(tickNodes) as SVGTextElement[];
```
PavelVanecek, line 352: `as` is unsafe, try to avoid it unless necessary. Why does this need to be typed as `SvgTextElement[]` if the line immediately below types it as `Element` anyway?

src/cartesian/CartesianAxis.tsx
```diff
@@ -105,6 +108,8 @@ export class CartesianAxis extends Component<Props, IState> {
 
   constructor(props: Props) {
     super(props);
+    this.tickRefs = React.createRef() as React.MutableRefObject<SVGTextElement[]>;
```
PavelVanecek, line 111: Avoid `as`

src/util/YAxisUtils.tsx
```diff
@@ -0,0 +1,93 @@
...
+ * @returns {Object} An object containing the modified label
+ */
+export const getModifiedYAxisProps = ({
+  label,
+}: {
+  label: string | number | ReactElement | object;
+}): { label: object } => {
+  let modifiedLabel = label;
+
+  const commonLabelProps = { angle: 270, position: 'insideLeft', textAnchor: 'middle', verticalAnchor: 'start' };
```
PavelVanecek, line 21: This is new.

How is this related to the automated responsive width?

Why is this inside a props modifier function? If we want to add new default behaviour, why not put this in default props?

If we change the default behaviour we also need update to the documentation.

PavelVanecek (changes requested):
Okay the "after" variant looks much better. Let's have that as a new breaking change in 3.x @ckifer?

@saurabhraj123 thank you for the PR. Would you mind to fix the conflicts? Then we can have a look at what the CI build looks like and what is the visual diff.

src/cartesian/CartesianAxis.tsx
saurabhraj123: I think it is not needed. Will fix it.

src/util/YAxisUtils.tsx
saurabhraj123: This is to rotate the y-axis label to 270 degrees. By default, `getAttrsOfCartesianLabel` in `Label.tsx` uses **center** position. But we need `insideLeft` position for the y-axis label. I can not directly modify that function as it is used in other cases as well.

The user can pass a **string/object** in the label prop but we still want these additional attributes to rotate and position it appropriately. 

If you have any suggestions on simplifying it or somehow putting it in default props, please let me know.

saurabhraj123:
> Can you add some unit tests

I will work on it @ckifer.

src/util/YAxisUtils.tsx
saurabhraj123: If we keep the old behaviour when `label` prop is received, wouldn't it be inconsistent? I also checked **highcharts** and by default their y-axis label is rotated by -90 degrees only.

Also, just passing `label="My label` seems more natural to me. So, I think, we should handle this as well. What do you think?

Regarding, using `<Label />` as children of `YAxis`, thanks for pointing it out. I wasn't aware of that bit. Will make changes to support that as well.

ckifer: > Also, just passing label="My label seems more natural to me. So, I think, we should handle this as well. What do you think?

Maybe but it should be outside the scope of this change

src/component/Text.tsx
```diff
@@ -198,83 +198,91 @@ export const getWordsByLines = ({ width, scaleToFit, children, style, breakAll,
...
-          // eslint-disable-next-line react/no-array-index-key
-          <tspan x={x} dy={index === 0 ? startDy : lineHeight} key={`${words}-${index}`}>
-            {words}
-          </tspan>
-        );
-      })}
-    </text>
-  );
-};
+Text.displayName = 'Text';
```
ckifer, line 288: why was this added?

PavelVanecek:
> > Run `npm run test-build-output -- --update` to get new bundle file snapshots, that will fix the build
> 
> The command runs successfully in local but produces no change. Do I need to do something else?

Ah sorry I forgot. First `npm run build` that will generate the bundles, then `npm run test-build-output -- --update` that will update the bundle snapshot. Then you can run `npm run test-build-output` to verify that it worked, commit and push.

src/util/YAxisUtils.tsx
PavelVanecek: Good point about the `label="My label`, I haven't considered that.

And +1 to @ckifer, let's keep this PR focused on the responsive width, and then have second PR for changing the default props? One thing at a time.

src/component/Text.tsx
saurabhraj123: `displayName`?

Now that I am using `forwardRef` here, the `displayName` appears in the React DevTools instead of just `forwardRef` or `Anonymous`. Also if some errors occurs, we see the `displayName` in the stack trace.

src/util/YAxisUtils.tsx
saurabhraj123: Acknowledged. Will only keep code related to y-axis responsive width in this PR.

saurabhraj123: If I don't modify the default props, then it's output won't be good if a `label` is passed. It will overlap because, it is `center` position by default. What should I do? Ignore the label in the calculation for now or keep it like this but should not modify the props?

Screenshot 2025-06-02 at 12 31 41 AM

ckifer: Ignore the label for now but follow up to include it

saurabhraj123:
> Can you add some unit tests

I am done with all other changes and was trying to add some test cases but I don't know why, but the render method is not rendering the y-axis properly when `width='auto'` is passed. It gives `null` when I try to get it using `querySelector`. 

This is the test code I am using:
```javascript
 it('should render y-axis with dynamically calculated width', async () => {
    const ticks = [0, 400, 1200, 1600];

    const { container } = render(
      <BarChart width={100} height={100}>
        <YAxis width="auto" ticks={ticks} />
        <CartesianGrid />
      </BarChart>,
    );

    const yAxis = container.querySelector('.yAxis');
    const yAxisWidth = yAxis.getBoundingClientRect().width;
    const yAxisLine = yAxis.querySelector('line');

    expect(yAxis).toBeVisible();
    expect(yAxisLine).toHaveAttribute('width', String(Math.round(yAxisWidth)));
  });
```

ckifer:
You're going to need to mock getBoundingClientRect. We should have a utility to do that somewhere

saurabhraj123:
> getBoundingClientRect

But **YAxis** itself is not being rendered when `width="auto"` is passed. I wanted to get the width of the axis that is being rendered and then use the util function get the width and compare them. In my current code, I get `yAxis = null`.

I also used `debug`. I found that **YAxis** doesn't even exist in the DOM.

Am I missing something or is it some limitation of the testing library?

PavelVanecek:
Do you have any insight as of why is it not being rendered? If you add `console.log({updatedYAxisWidth})` what do you see? If you place a debugger what do you see?

We have tests that show that YAxis does in fact render, for example in `test/cartesian/YAxis.spec.tsx`, so the testing library can do that.

Coltin was recommending to use `mockGetBoundingClientRect({ width: 80, height: 30 });`, have you tried that? Look at `test/component/Legend.spec.tsx` for example. jsdom by default returns all zeroes from `getBoundingClientRect()` (because it doesn't layout) so we have to mock the results in jsdom environment.

saurabhraj123:
> Do you have any insight as of why is it not being rendered? If you add `console.log({updatedYAxisWidth})` what do you see? If you place a debugger what do you see?
> 
> We have tests that show that YAxis does in fact render, for example in `test/cartesian/YAxis.spec.tsx`, so the testing library can do that.
> 
> Coltin was recommending to use `mockGetBoundingClientRect({ width: 80, height: 30 });`, have you tried that? Look at `test/component/Legend.spec.tsx` for example. jsdom by default returns all zeroes from `getBoundingClientRect()` (because it doesn't layout) so we have to mock the results in jsdom environment.

Thank you @ckifer and @PavelVanecek. That was some mistake from my end. I was not rendering the `Bar`, so the ticks were not being rendered. Fixed and pushed the changes.

You guys can review it now.

PavelVanecek:
@saurabhraj123 the build failures are unrelated to your changes. I have fixed that on `3.x` branch. May I please ask you to merge or rebase?

src/util/YAxisUtils.tsx
```diff
@@ -0,0 +1,56 @@
...
+  labelRef?: React.RefObject<Element>;
+  labelGapWithTick?: number;
+}): number => {
+  if (!cartesianAxisRef || !cartesianAxisRef.current) {
+    return 0; // Return 0 if the ref is not set or current is null
+  }
+
+  // Access the tickRefs from the CartesianAxis component
+  const axisComponent = cartesianAxisRef.current;
+  const tickNodes = axisComponent?.tickRefs?.current;
```
PavelVanecek: Let's not parse the internal structure of a component from outside of that component. Can we pass in the required data as props instead?

```
type IGetBoundingClient = Pick<Element, 'getBoundingClientRect'>

export const getCalculatedYAxisWidth = ({
  ticks,
  labelRef,
  labelGapWithTick = 5, // Default gap between label and tick in pixels
  tickSize = 0,
  tickMargin = 0
}: {
  ticks: ReadonlyArray<IGetBoundingClient> | undefined;
  labe: IGetBoundingClient | undefined;
  labelGapWithTick: number | undefined;
  tickSize: number | undefined;
  tickMargin: number | undefined;
}): number => {
```

That way we don't depend on React, we don't depend on other component internal structure, and it's way easier to mock in tests.

Also, please prefer `x | undefined` instead of `prop?:`. The question mark syntax makes it easy to forget that the function expects a particular input. The `| undefined` forces people to always pass in the parameter - and it handles undefined just the same. Impossible to forget.

src/util/YAxisUtils.tsx
```diff
@@ -0,0 +1,56 @@
...
+        }
+      }
+    });
+
+    // calculate width of the axis label
+    const labelNode = labelRef?.current;
+    const labelWidth = labelNode ? labelNode.getBoundingClientRect().width : 0;
+
+    // calculate tick width: tickSize + tickMargin
+    const { tickSize = 0, tickMargin = 0 } = axisComponent.props || {};
```
PavelVanecek: Same here, please let's pass these as parameters to the function instead of reading props of another component.

saurabhraj123:
> @saurabhraj123 the build failures are unrelated to your changes. I have fixed that on `3.x` branch. May I please ask you to merge or rebase?

My branch is already up to date with 3.x @PavelVanecek. Ran build commands as well, no changes are reflected.

ckifer:
Looks like the vitest downgrade hasn't been merged into your branch. Make sure to do a fresh npm install

src/util/YAxisUtils.tsx
saurabhraj123: Agreed - passing in the data keeps the function decoupled and easier to test. I have modified the function as suggested.

ckifer (approved):

PavelVanecek:
Would you like to add the new property to the storybook and docs too, to make it easier to find for other devs?

image

image

PavelVanecek:
No visual diff in the storybook: https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=2358

saurabhraj123:
> Would you like to add the new property to the storybook and docs too, to make it easier to find for other devs?

I will @PavelVanecek. Also, I will create a follow up PR to add support for the labels.

PavelVanecek:
Thanks!

hc-12:
Hi, i need this feature. Has this been released in v3 beta?

ckifer:
Not yet but looking at releasing 3.0 shortly

ckifer:
released beta.2 for the time being

src/component/Label.tsx
```diff
@@ -483,47 +490,53 @@ const parseViewBox = (props: any): ViewBox => {
...
   }
 
   if (isValidElement(label)) {
     if (label.type === Label) {
-      return cloneElement<LabelProps>(label, { key: 'label-implicit', viewBox });
+      return cloneElement<LabelProps>(label, { key: 'label-implicit', ...commonProps });
     }
 
-    return <Label key="label-implicit" content={label} viewBox={viewBox} />;
+    return <Label key="label-implicit" content={label} {...commonProps} />;
```
ckifer, line 513: when `label` is an HTML or SVG element `label` gets passed to `content` and then `content` is cloned with `props`. `props` includes `labelRef` which isn't a valid prop in the DOM

PR #5891, opened 2025-06-02
LineChart no longer uses generateCategoricalChart

PavelVanecek:
## Description

If we agree we want to go this direction, here is what I will do next (followup PR):
1. Introduce PolarChart (as a sibling to CartesianChart)
2. Move all charts to use either CartesianChart or PolarChart
3. Create a separate state type for polar charts
4. remove cartesian data from polar state, and remove polar data from cartesian state
5. strict null checks!

## Related Issue

### DIFF
--- a/src/chart/CartesianChart.tsx
+++ b/src/chart/CartesianChart.tsx
@@ -1,0 +1,146 @@
+import { forwardRef } from 'react';
+import * as React from 'react';
+import { filterProps, validateWidthHeight } from '../util/ReactUtils';
+import { ChartOptions } from '../state/optionsSlice';
+import { PolarChartOptions } from '../state/polarOptionsSlice';
+import { RechartsStoreProvider } from '../state/RechartsStoreProvider';
+import { ChartDataContextProvider } from '../context/chartDataContext';
+import { ReportMainChartProps } from '../state/ReportMainChartProps';
+import { ReportChartProps } from '../state/ReportChartProps';
+import { ReportPolarOptions } from '../state/ReportPolarOptions';
+import { CategoricalChartProps } from './generateCategoricalChart';
+import { Margin, TooltipEventType } from '../util/types';
+import { TooltipPayloadSearcher } from '../state/tooltipSlice';
+import { RootSurface } from '../container/RootSurface';
+import { RechartsWrapper } from './RechartsWrapper';
+import { ClipPathProvider } from '../container/ClipPathProvider';
+
+const defaultMargin: Margin = { top: 5, right: 5, bottom: 5, left: 5 };
+
+const defaultProps: Partial<CategoricalChartProps> = {
+  accessibilityLayer: true,
+  layout: 'horizontal',
+  stackOffset: 'none',
+  barCategoryGap: '10%',
+  barGap: 4,
+  margin: defaultMargin,
+  reverseStackOrder: false,
+  syncMethod: 'index',
+};
+
+const CategoricalChart = forwardRef<SVGSVGElement, CategoricalChartProps>((props: CategoricalChartProps, ref) => {
+  const { children, className, width, height, style, compact, title, desc, ...others } = props;
+  const attrs = filterProps(others, false);
+
+  // The "compact" mode is used as the panorama within Brush
+  if (compact) {
+    return (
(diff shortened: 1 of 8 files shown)


test/chart/LineChart.spec.tsx
```diff
@@ -857,11 +857,13 @@ describe('<LineChart />', () => {
...
 
   describe('LineChart - test ref access', () => {
-    test('should allow access to the CategoricalChartWrapper through the ref prop forwarded from CategoricalChart', () => {
-      // TODO we need to fix the ref typing in generateCategoricalChart, this class is created dynamically and cannot be typed
-      // @ts-expect-error TODO remove in followup PR
-      // eslint-disable-next-line no-undef
-      let refNode: CategoricalChartWrapper | null = null;
+    test('should allow access to the main SVG through the ref prop forwarded from CategoricalChart', () => {
+      /*
+       * This is a breaking change from v2.0.0, where the ref was not forwarded
```
PavelVanecek, line 862: @ckifer where should we point the `ref`? It can't be the class instance anymore (because there is no class to point to). Should it reference the div wrapper, or the SVG?

src/chart/LineChart.tsx
```diff
@@ -1,10 +1,18 @@
...
 
-export const LineChart = generateCategoricalChart({
-  chartName: 'LineChart',
-  tooltipPayloadSearcher: arrayTooltipSearcher,
+export const LineChart = forwardRef<SVGSVGElement, CategoricalChartProps>((props: CategoricalChartProps, ref) => {
+  return (
+    <CartesianChart
+      chartName="LineChart"
+      defaultTooltipEventType="axis"
+      validateTooltipEventTypes={['axis']}
```
PavelVanecek: I should cache this array instance

src/chart/CartesianChart.tsx
```diff
@@ -0,0 +1,146 @@
...
+  reverseStackOrder: false,
+  syncMethod: 'index',
+};
+
+const CategoricalChart = forwardRef<SVGSVGElement, CategoricalChartProps>((props: CategoricalChartProps, ref) => {
+  const { children, className, width, height, style, compact, title, desc, ...others } = props;
+  const attrs = filterProps(others, false);
+
+  // The "compact" mode is used as the panorama within Brush
+  if (compact) {
```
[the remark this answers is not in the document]
PavelVanecek: Yeah this is on purpose, we don't want refs to the panorama.

src/chart/CartesianChart.tsx
```diff
@@ -0,0 +1,146 @@
...
+    eventEmitter: undefined,
+  };
+  let polarOptions: PolarChartOptions;
+  const { innerRadius = defaultProps.innerRadius, outerRadius = defaultProps.outerRadius } = categoricalChartProps;
+  if (defaultProps.startAngle != null) {
+    polarOptions = {
+      cx: categoricalChartProps.cx,
+      cy: categoricalChartProps.cy,
+      startAngle: defaultProps.startAngle,
+      endAngle: defaultProps.endAngle,
```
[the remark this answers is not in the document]
PavelVanecek: This too is on purpose, I will remove the polar setter from CartesianChart later

test/chart/LineChart.spec.tsx
ckifer: I guess what would be more useful? I think the svg makes more sense/is more useful

src/chart/LineChart.tsx
```diff
@@ -1,10 +1,21 @@
...
+import { CartesianChart } from './CartesianChart';
+import { CategoricalChartProps } from './generateCategoricalChart';
+import { TooltipEventType } from '../util/types';
 
-export const LineChart = generateCategoricalChart({
-  chartName: 'LineChart',
-  tooltipPayloadSearcher: arrayTooltipSearcher,
+const allowedTooltipTypes: ReadonlyArray<TooltipEventType> = ['axis'];
+
+export const LineChart = forwardRef<SVGSVGElement, CategoricalChartProps>((props: CategoricalChartProps, ref) => {
```
ckifer, line 10: Wonder how dramatically this changes the type. Good thing we're doing it as breaking

ckifer (approved):

ckifer:
Sweet

PR #5900, opened 2025-06-03
Splitting props and components into Cartesian and Polar variants

PavelVanecek:
## Description

Some properties only make sense for Polar charts so I am separating these to different types and props so that we can have more strict types and don't have to repeat the same null checks thousands of times.

782 errors

## Related Issue


ckifer (approved):
sweet

PR #5902, opened 2025-06-04
Downgrade vitest to resolve typescript error

PavelVanecek:
https://github.com/vitest-dev/vitest/pull/7016


PavelVanecek:
Why does change in the test dependencies change bundle size?

PR #5912, opened 2025-06-05
Add null check for Area points type

PavelVanecek:
## Description

711 errors

## Related Issue

image


PavelVanecek:
We need tests for the animations.

Issue #5914, opened 2025-06-05
Add tests for animations


PavelVanecek:
Recharts animations are fairly complicated. There are multiple charts, various edge cases, it's tedious to test everything by hand, tedious to write automated tests (so far we have 0 tests as far as I know), and easy to add mistakes (I created a bug just yesterday myself).

Let's have unit tests for animations. I propose:

1. Inline react-smooth into the main repo so that we don't have to deal with bumping two libraries - if you look at react-smooth dependants, the only usages are recharts and rechart forks anyway: https://www.npmjs.com/browse/depended/react-smooth
2. Migrate react-smooth to typescript and add a few more tests
3. Add a utility that allows to replace animations manager on the fly so that we can do things like create a mock and then set animation progress in a unit tests without faffing around with requestAnimationFrame
4. And finally add unit tests for recharts animations using the mock animation manager

What do you think?

ckifer:
I'm fine with that, I have a branch in the react-smooth repo that has the whole thing (or enough of it) refactored to TS already. I just broke things when I released it so I reverted

PavelVanecek:
Do you have a note of what broke? I will pick it up from there

ckifer:
It was just negligence on the build output. I did a release but only some of the files were in the correct dist folder. 

Just making sure the files get added to the build snapshots should be enough for that

ckifer:
https://github.com/recharts/react-smooth/tree/refactor-to-ts

PR #5924, opened 2025-06-06
Inline react-smooth

PavelVanecek:
## Description

Inline the library, remove dependency, add some basic types

## Related Issue

### DIFF
--- a/src/animation/Animate.tsx
+++ b/src/animation/Animate.tsx
@@ -1,0 +1,351 @@
+import * as React from 'react';
+import { PureComponent, cloneElement, Children } from 'react';
+import { isEqual } from 'es-toolkit';
+import { AnimationManager, createAnimateManager } from './AnimationManager';
+import { configEasing } from './easing';
+import configUpdate from './configUpdate';
+import { getTransitionVal, identity } from './util';
+
+type AnimateStep = {
+  duration: number;
+  style: Record<string, any>;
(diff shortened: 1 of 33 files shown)


src/animation/easing.ts
```diff
@@ -0,0 +1,168 @@
...
+          //     "oneOf 'linear', 'ease', 'ease-in', 'ease-out', " +
+          //     "'ease-in-out','cubic-bezier(x1,y1,x2,y2)', instead received %s",
+          //   args,
+          // );
+        }
+      }
+    }
+  }
+
+  // TODO replace with a typescript error
```
ckifer, line 63: Can remove this I think

src/animation/Animate.tsx
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  runStepAnimation(props: AnimateProps) {
+    const { steps, begin, onAnimationStart } = props;
+    const { style: initialStyle, duration: initialTime = 0 } = steps[0];
+
+    const addStyle = (
+      sequence: any,
+      nextItem: {
+        duration: any;
```
ckifer, line 215: Don't we know the types of a lot of these todos?

PavelVanecek: I'll work on that next, first I wanted to just copy paste the thing without changing any behaviour. Next come tests, then I update the types.

PavelVanecek:
Weird that the bundle size goes up so much.

ckifer:
Wonder if it measures dependencies as their minified size?

PR #5925, opened 2025-06-06
Add width 'auto' support in storybook

saurabhraj123:
## Description
Adding `width='auto'` support in storybook. Refer to this: #5880 

**Implementation summary:**
Changed control type to `text` for YAxis. Then inside `YAxis.stories.tsx`, wrote a function to convert entered text to number or 'auto' and passed it to YAxis.

## Related Issue
#5880 

## How Has This Been Tested?
By entering different values in the input field for width.

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/API/cartesian/YAxis.stories.tsx
+++ b/storybook/stories/API/cartesian/YAxis.stories.tsx
@@ -9,14 +9,25 @@ export default {
   argTypes: YAxisProps,
 };
 
+const getWidth = (width: string | number) => {
+  if (width === 'auto' || typeof width === 'number') {
+    return width;
+  }
+
+  const num = parseInt(width, 10);
+  return Number.isNaN(num) ? 120 : num;
+};
+
 export const API = {
   render: (args: Record<string, any>) => {
+    const width = getWidth(args.width);
+
     return (
       <ResponsiveContainer width="100%" height={500}>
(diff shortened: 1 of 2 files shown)


saurabhraj123:
We'll have to update [recharts.org](https://recharts.org/en-US/api/YAxis) API section as well, right?

ckifer:
Yeah, but not yet because this is only in 2.x

recharts.org is the 2.x docs still and isn't versioned

storybook/stories/API/cartesian/YAxis.stories.tsx
```diff
@@ -32,7 +43,7 @@ export const API = {
     allowDataOverflow: true,
     tickMargin: 20,
     angle: 45,
-    width: 120,
+    width: '120',
```
ckifer, line 46: why can't it be a number?

ckifer (approved):

storybook/stories/API/cartesian/YAxis.stories.tsx
saurabhraj123: On **docs** page, since the input type is changed to `text`, it was showing a blank input box despite passing number `120`. That's why changed it to a string.

ckifer: hm interesting, ok thanks

PR #5926, opened 2025-06-07
Update es-toolkit to support CJS tree-shaking and non-exports field environments

dayongkr:
## Description

As mentioned, we've addressed one key issue and additionally modified part of the build structure to better support CommonJS (CJS) environments by adjusting the import paths. The changes are as follows:

1. For environments that do not support the exports field, we changed the build structure to allow folder-based access to es-toolkit/compat.
2. Since tree-shaking does not work well in CJS environments, we've adjusted the structure so that specific utility files can be imported directly (e.g., es-toolkit/compat/get).

While es-toolkit is originally ESM-first and does not officially deprecated environments, we’ve made an exception for the compat layer, which exists for Lodash compatibility. With this principle update, we've implemented the above changes to fully accommodate recharts’ needs.

Thank you for your patience!

## Related Issue

https://github.com/recharts/recharts-integ/pull/60#issuecomment-2944300841

## Types of changes

## Checklist:


src/chart/Sankey.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react';
 import { MouseEvent, PureComponent, ReactElement, SVGProps } from 'react';
-import { maxBy, sumBy } from 'es-toolkit';
-// @ts-expect-error not installing types for lodash.get, I just want to see the bundle size difference
-import get from 'lodash.get';
+import maxBy from 'es-toolkit/compat/maxBy';
```
PavelVanecek, line 3: Why switch these two to the compat version?

PavelVanecek:
image

src/chart/Sankey.tsx
dayongkr: recharts is a CJS, so when it's imported the previous way, it brings in all functions, which increases the bundle size at the places where recharts is used.

However, the recent change imports only the specific function file directly, which prevents unnecessary code from being included—even in a CJS environment.

I hadn't considered this in my previous implementation, so I made sure to fix it as part of this update.

ckifer:
Interesting no drop in bundle size but no more failures so 🚢

PR #5927, opened 2025-06-07
Add tests and types for AnimationManager

PavelVanecek:
## Description

some of the functionality is never used outside of tests (like, unsubscribing, and the object styles). I will remove that once the whole animation thing has tests.

## Related Issue

### DIFF
--- a/src/animation/AnimationManager.ts
+++ b/src/animation/AnimationManager.ts
@@ -1,21 +1,34 @@
 import { setRafTimeout } from './setRafTimeout';
 
-type ReactSmoothStyleType = string | string[] | Record<string, unknown> | (() => void);
+type ReactSmoothStyle = object;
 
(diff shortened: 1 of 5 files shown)


src/animation/AnimationManager.ts
```diff
@@ -1,21 +1,34 @@
 import { setRafTimeout } from './setRafTimeout';
 
-type ReactSmoothStyleType = string | string[] | Record<string, unknown> | (() => void);
+type ReactSmoothStyle = object;
```
ckifer, line 3: Why just an object and not something more specific? Record even?

ckifer (approved):

src/animation/AnimationManager.ts
PavelVanecek: I will delete this option later, we never use it anyway.

PR #5945, opened 2025-06-11
Use same timeout mechanism for both animationmanager and configupdate

PavelVanecek:
## Description

AnimationManager and configUpdate, for some reason, used to use two separate timeout mechanism. So now I made it just one mechanism which will make it easier to mock it in the `Animate` component tests.

## Related Issue


ckifer:
Gotta update snaps

PavelVanecek:
I forgot the build snapshot update again. I'll fix that tomorrow

ckifer (approved):

Issue #5944, opened 2025-06-11
Allow the minPointSize function to receive null/undefined values


eino:
## Do you want to request a feature or report a bug?

Request a feature

## What is the current behavior?

For Bars, the `minPointSize` attribute can be a function. But currently, such functions can only receive numbers. If it receives a `null`/`undefined` value, the `minPointSizeCallback` wrapper will throw, and the graph will crash.

## What is the expected behavior?

Nullish values are valid values in graphs, representing the absence of a datapoint (as opposed to 0 representing the value 0). `minPointSizeCallback` should allow null and undefined values, and pass them to `minPointSize`. This will allow to give a different min size for `null` vs `0` values.

## Recharts version

I'm using the 2.15.1 version

ckifer:
Hi! Can you make a codesandbox reproduction?

minPointSize should be able to take in a number or a function that returns a number. You have control over that function so you should be able to guard against nullish values. 

We should guard against the app crashing but at the end of the day I would just treat a nullish minPointSize as 0. I don't think the idea was ever to return the current value from the minPointSize function since they represent different things

eino:
Hi!

Indeed we can control this function, but the issue is that this function is never called at all, when the parameter is not a number, because the test in [minPointSizeCallback](https://github.com/recharts/recharts/blob/5d67a5f1b66cbb9f3d9289ab947a11f6436451c5/src/util/BarUtils.tsx#L74).
The exception is raised in rechart's code, so there's no type guard or other possibility to handle the code from the calling app.

The aim is not to return a null minPointSize (which wouldn't make much sense indeed). It's the opposite: if `value` is null, allow  `minPointSize` to be called, so it can return a value (be it 0 or any other value).

I opened a PR #5946, hope it will suit,

Regards,

ckifer:
Ah sorry got it. 

Let me look back at the history and see why this was restricted to number values only. It's amazing what you forget about things you've written

Thanks for the PR!

eino:
implemented in [v3.0.0-beta.2](https://github.com/recharts/recharts/releases/tag/v3.0.0-beta.2)

ckifer:
released in 2.15.4 as well

PR #5946, opened 2025-06-11
feat: allow minPointSize function to receive null and undefined values (2.x)

eino:
## Description

A Bar's `minPointSize` attribute can be a function. This PR is aimed at allowing to pass `null` or `undefined` values to this function (rather than throwing)

## Related Issue

#5944 

## Motivation and Context

Null or undefined are valid values for datapoints, and are distinct from `0` (i.e. the absence of data). It's interesting to allow to use those values even if using a `minPointSize` function.

## How Has This Been Tested?

I tested the fork with a local dataset with null values, and a minPointSize function returning different values for null non-null values:

```
const minPointSize = useCallback((value: number | null, _index: number) => isNil(value) ? 0 : 2, []);
```

Other than that, there is little risk of side-effects:
- for existing `minPointSize` functions taking number values, the function will continue working as is.
- if someone has a codebase sending null values to `minPointSize`, rechart currently throws.
- minPointSizeCallback is only called at one place in the rechart code, so there's no other impact
 
## Types of changes

## Checklist:

### DIFF
--- a/src/util/BarUtils.tsx
+++ b/src/util/BarUtils.tsx
@@ -4,6 +4,7 @@ import { ActiveShape } from './types';
 import { Props as RectangleProps } from '../shape/Rectangle';
 import { BarProps } from '../cartesian/Bar';
 import { Shape } from './ActiveShapeUtils';
+import { isNullish, isNumber } from './DataUtils';
 
 // Rectangle props is expecting x, y, height, width as numbers, name as a string, and radius as a custom type
 // When props are being spread in from a user defined component in Bar,
@@ -50,7 +51,7 @@ export function BarRectangle(props: BarRectangleProps) {
   );
 }
 
-export type MinPointSize = number | ((value: number, index: number) => number);
+export type MinPointSize = number | ((value: number | undefined | null, index: number) => number);
 
 /**
  * Safely gets minPointSize from from the minPointSize prop if it is a function
@@ -62,14 +63,14 @@ export const minPointSizeCallback =
   (minPointSize: MinPointSize, defaultValue = 0) =>
   (value: unknown, index: number): number => {
     if (typeof minPointSize === 'number') return minPointSize;
-    const isValueNumber = typeof value === 'number';
-    if (isValueNumber) {
-      return minPointSize(value, index);
+    const isValueNumberOrNil = isNumber(value) || isNullish(value);
+    if (isValueNumberOrNil) {
+      return minPointSize(value as number | undefined | null, index);
     }
 
     invariant(
-      isValueNumber,
-      `minPointSize callback function received a value with type of ${typeof value}. Currently only numbers are supported.`,
+      isValueNumberOrNil,
+      `minPointSize callback function received a value with type of ${typeof value}. Currently only numbers or null/undefined are supported.`,
     );
     return defaultValue;
   };
--- a/src/util/DataUtils.ts
+++ b/src/util/DataUtils.ts
@@ -2,6 +2,7 @@ import isString from 'lodash/isString';
 import isNan from 'lodash/isNaN';
 import get from 'lodash/get';
 import lodashIsNumber from 'lodash/isNumber';
+import isNil from 'lodash/isNil';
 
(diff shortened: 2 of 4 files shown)


PavelVanecek:
Please raise the same against 3.x branch, we're about to release the new version

ckifer:
Oh +1 to the above. Also this is a breaking change so I don't think I'd be comfortable releasing it in 2.x at all. 

If someone relies on the fact that null and undefined arent passed and then we do we will break apps that do number operations on the value without any extra checks

src/util/BarUtils.tsx
```diff
@@ -62,9 +62,9 @@ export const minPointSizeCallback =
   (minPointSize: MinPointSize, defaultValue = 0) =>
   (value: unknown, index: number): number => {
     if (typeof minPointSize === 'number') return minPointSize;
-    const isValueNumber = typeof value === 'number';
+    const isValueNumber = typeof value === 'number' || typeof value === 'undefined' || value === null;
```
ckifer: Can we just move the other two checks into the if statement? If not then change the variable name since it's no longer true that this is always a number

ckifer (changes requested):

src/util/BarUtils.tsx
eino: You mean to the if statement line 64? Not really as the point is to call the function with `value`. But I will indeed rename the variable :+1:

ckifer: No I mean line 66 - 

`isValueNumber || typeof value === ...etc` (sorry I'm on mobile)

ckifer: Also `value == null` would cover both null and undefined without needing the typeof check

eino:
> Oh +1 to the above. Also this is a breaking change so I don't think I'd be comfortable releasing it in 2.x at all.
> If someone relies on the fact that null and undefined arent passed and then we do we will break apps that do number operations on the value without any extra checks

As you prefer, but with the current implementation, rechart throws (when receiving a null value), so the chart is not displayed at all. It's hard to see how someone would rely on the chart crashing, but technically it's true.

src/util/BarUtils.tsx
eino: Ah yes indeed. I will. :+1: 

I'll open a distinct PR against the 3.x branch.

ckifer:
lol you're right, I keep forgetting that it's throwing. I think I added this when I thought invariant didn't throw in production (but it just strips out the message)

All good to leave this (but it may not get released soon) but also need to do in 3.x. Thanks!

src/util/BarUtils.tsx
```diff
@@ -62,9 +62,9 @@ export const minPointSizeCallback =
...
     if (isValueNumber) {
-      return minPointSize(value, index);
+      return minPointSize(value as number | undefined | null, index);
```
ckifer, line 68: The change in the condition should also allow you to remove this I think? Prefer not to cast if possible

eino:
> All good to leave this (but it may not get released soon) but also need to do in 3.x. Thanks!

Haha ok no problem :slightly_smiling_face: I'll finalize both PRs (and switch to the 3.x for my project if it's merged first).

src/util/BarUtils.tsx
eino: Actually it doesn't, as `null` is of type object in js, so a test `== null` will not type the value well. But I just saw there's a `isNil` typeguard in `es-toolkit`, I'll change to use it!

ckifer: Instead of importing `isNil` you can use our `isNullish` - same thing

https://github.com/recharts/recharts/blob/5d67a5f1b66cbb9f3d9289ab947a11f6436451c5/src/util/DataUtils.ts#L184

Sorry, just got to the computer so can check all the things easier 😅

eino: Yes indeed :slightly_smiling_face: I found it on the 3.x branch, however it doesn't exist on master so I recreated it. Pushing a new version.

PR #5947, opened 2025-06-11
feat: allow minPointSize function to receive null and undefined values (3.x)

eino:
## Description

A Bar's `minPointSize` attribute can be a function. This PR is aimed at allowing rechart v 3.x to call `minPointSize` with `null` or `undefined` values, rather than throwing.

## Related Issue

#5944 

## Motivation and Context

Null or undefined are valid values for datapoints, and are distinct from `0` (i.e. the absence of data). It's interesting to allow to use those values even if using a `minPointSize` function.

## How Has This Been Tested?

I tested the fork with a local dataset with null values, and a minPointSize function returning different values for null non-null values:

```
const minPointSize = useCallback((value: number | null, _index: number) => isNil(value) ? 0 : 2, []);
```

Other than that, there is little risk of side-effects:
- for existing `minPointSize` functions taking number values, the function will continue working as is.
- if someone has a codebase sending null values to `minPointSize`, rechart currently throws.
- minPointSizeCallback is only called at one place in the rechart code, so there's no other impact

## Types of changes

## Checklist:

### DIFF
--- a/src/util/BarUtils.tsx
+++ b/src/util/BarUtils.tsx
@@ -5,7 +5,7 @@ import { ActiveShape } from './types';
 import { Props as RectangleProps } from '../shape/Rectangle';
 import { BarProps } from '../cartesian/Bar';
 import { Shape } from './ActiveShapeUtils';
-import { isNumber } from './DataUtils';
+import { isNullish, isNumber } from './DataUtils';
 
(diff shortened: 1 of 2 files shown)


ckifer (approved):
thank you

PR #5955, opened 2025-06-15
Add tests for line animations

PavelVanecek:
## Related Issue

### DIFF
--- a/src/animation/Animate.tsx
+++ b/src/animation/Animate.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { PureComponent, cloneElement, Children } from 'react';
+import { PureComponent, cloneElement, Children, createContext, useContext } from 'react';
 import isEqual from 'es-toolkit/compat/isEqual';
 import { AnimationManager, createAnimateManager } from './AnimationManager';
 import { configEasing, EasingInput } from './easing';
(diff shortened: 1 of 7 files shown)


test/cartesian/Line.animation.spec.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+      const line = getLine(container);
+      expect(line).toBeInTheDocument();
+      // the path is fully rendered
+      expect(line).toHaveAttribute('d', 'M5,5L23,27.5L41,27.5L59,50L77,32.45L95,52.475');
+      // but the strokeDasharray is 0 so the line is not visible
+      expect(line).toHaveAttribute('stroke-dasharray', '0px 0px');
+    });
+
+    it('should animate line left-to-right by continually extending the stroke-dasharray', async () => {
```
PavelVanecek, line 119: This test is my favourite. Once we add tests like this one for all graphical components, that should be the end of animation regressions.

src/animation/Animate.tsx
```diff
@@ -273,3 +273,16 @@ export class Animate extends PureComponent<AnimateProps, AnimateState> {
     return <div>{Children.map(children, child => cloneContainer(child))}</div>;
   }
 }
+
+export const AnimationManagerContext = createContext<AnimationManager | null>(null);
+
+export function Animate(props: AnimateProps) {
+  const contextAnimationManager = useContext(AnimationManagerContext);
```
PavelVanecek, line 280: > 3. Add a utility that allows to replace animations manager on the fly so that we can do things like create a mock and then set animation progress in a unit tests without faffing around with requestAnimationFrame

test/cartesian/Line.animation.spec.tsx
ckifer: Quite beautiful actually, 🙌

PR #5983, opened 2025-06-22
Remove semicolon from Sankey

PavelVanecek:
## Description

Somehow Sankey was adding semicolon to the render output for a few months and nobody noticed.


ckifer (approved):
Oops

PR #6002, opened 2025-06-24
Remove @storybook/addon-storysource plugin and replace with codePanel: true

PavelVanecek:
## Description

I noticed the "Code" panel in storybook renders compiled code and I didn't like it:

image

so I went and discovered that the plugin we use is deprecated: https://storybook.js.org/docs/writing-docs/code-panel

> Code Panel is a replacement for the [Storysource addon](https://storybook.js.org/addons/@storybook/addon-storysource), which was [discontinued in Storybook 9](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#storysource-addon-removed).

so I uninstalled the plugin and turned on the built-in code panel and turns out it looks way better:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,6 @@
         "@storybook/addon-interactions": "^8.6.13",
         "@storybook/addon-links": "^8.6.13",
         "@storybook/addon-mdx-gfm": "^8.6.13",
-        "@storybook/addon-storysource": "^8.6.13",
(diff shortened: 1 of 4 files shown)


ckifer:
Nice!

What's up with our integ tests? Unrelated to this

PavelVanecek:
https://github.com/recharts/recharts-integ/pull/72

ckifer (approved):

Issue #3573, opened 2023-05-09
Close Tooltip with trigger=click by clicking outside


khuongtp:
- [x] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Right now if I use trigger='click' for Tooltip, there's no way I could close it

### What does the proposed API look like?
I notice that with Chrome dev tool when I resize the screen using responsive mode or just select an arbitrary phone view port, it works as expected, the tooltip being shown when I tap on the chart and closed when I tap outside

ckifer:
for now you can use the Escape key to dismiss the tooltip, but agreed this needs to be enhanced

JoaquinPalacios:
Hi, I noticed this feature request has been open for a while and I’m also interested in having this functionality.
Is there any update on whether this is being considered or if any work has started on it?
Thanks  🙌

ckifer:
No update but it could be done relatively easily I think

JoaquinPalacios:
Thanks for the quick reply! If it’s not too complicated, I’d be happy to give it a go and submit a PR. I’d just need a bit of guidance on where to start and what parts of the codebase I’d need to look at. Let me know if that’s alright

PR #6007, opened 2025-06-25
Feature/tooltip closing

JoaquinPalacios:
# Implement tooltip close on outside click for trigger="click"

## Description

This PR implements the ability to close click-triggered tooltips by clicking outside the chart area. Previously, tooltips with trigger="click" would remain open until the user clicked on another chart element or manually closed them. Now, clicking anywhere outside the chart will close the tooltip.

## Related Issue

#3573 

## Motivation and Context

Users expect click-triggered tooltips to behave like standard UI tooltips that close when clicking outside their area. This is a common UX pattern that improves usability and provides a more intuitive interaction model for charts with click-triggered tooltips.

## How Has This Been Tested?

- Unit Tests: Added comprehensive tests in tooltipEventType.spec.tsx to verify outside click behavior for both axis-based and item-based tooltips
- Integration Tests: Verified that the feature works across all chart types (Bar, Pie, Scatter, Funnel, RadialBar, Treemap)
- Regression Testing: Ensured existing tooltip functionality remains intact for hover-triggered tooltips
- Edge Cases: Tested scenarios with multiple charts, external event handlers, and various tooltip configurations

### Test Environment:

- Node.js environment with Jest test runner
- All existing tooltip tests continue to pass
- New outside click tests specifically validate the new functionality

## Types of changes

## Checklist:

## Note: The implementation includes:

- Redux actions for tooltip closing (tooltipCloseStart, tooltipCloseEnd, clearClickTooltip)
- Outside click detection in RechartsWrapper component
- Updated event handler logic in chart components to properly forward external handlers
- Fixed Treemap rendering issue that was discovered during testing
- Comprehensive test coverage for the new functionality
- The changes are backward compatible and only affect click-triggered tooltips, leaving hover-triggered tooltips unchanged.


src/cartesian/Bar.tsx
```diff
@@ -242,28 +256,73 @@ function BarBackground(props: BarBackgroundProps) {
...
         }
 
-        // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
-        const onMouseEnter = onMouseEnterFromContext(entry, i);
-        // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
-        const onMouseLeave = onMouseLeaveFromContext(entry, i);
-        // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
-        const onClick = onClickFromContext(entry, i);
+        // Always attach external event handlers (passed as props)
+        const eventHandlers: any = {
```
ckifer: can this be typed?

src/cartesian/Bar.tsx
```diff
@@ -242,28 +256,73 @@ function BarBackground(props: BarBackgroundProps) {
...
+          // For hover mode, also ensure external onClick is called if provided
+          const externalOnClick = eventHandlers.onClick;
+          if (externalOnClick) {
+            eventHandlers.onClick = (e: React.MouseEvent<SVGElement, MouseEvent>) => {
+              externalOnClick(entry, i, e);
+            };
+          }
+        }
+
+        const noop = () => {}; // function to prevent errors when event handlers are undefined
```
ckifer: I think we have a noop function defined that can be imported

src/cartesian/Bar.tsx
```diff
@@ -329,16 +397,62 @@ function BarRectangles({
           dataKey,
         };
 
+        // Always attach external event handlers (passed as props)
+        const eventHandlers: any = {
+          ...adaptEventsOfChild(restOfAllOtherProps, entry, i),
+        };
+
+        // Conditionally attach internal tooltip handlers based on tooltip trigger
+        if (tooltipTrigger === 'click') {
```
ckifer: I'm confused at what all of this is doing that what was here wasn't doing?

these three events are already combined internal and external using `useMouseClickItemDispatch`, etc.

src/chart/RechartsWrapper.tsx
```diff
@@ -194,7 +226,7 @@ export const RechartsWrapper = forwardRef(
             onTouchStart={myOnTouchStart}
             ref={innerRef}
           >
-            {children}
+            {React.isValidElement(children) ? React.cloneElement(children) : children}
```
ckifer: why is this needed? We're trying to avoid adding more cloning

src/chart/Treemap.tsx
```diff
@@ -382,6 +382,10 @@ type ContentItemProps = {
   onClick?: (e: React.MouseEvent<SVGPathElement, MouseEvent>) => void;
   onMouseEnter?: (e: React.MouseEvent<SVGPathElement, MouseEvent>) => void;
   onMouseLeave?: (e: React.MouseEvent<SVGPathElement, MouseEvent>) => void;
+  externalOnClick?: (node: TreemapNode, e: React.MouseEvent<SVGPathElement, MouseEvent>) => void;
+  externalOnMouseEnter?: (node: TreemapNode, e: React.MouseEvent<SVGPathElement, MouseEvent>) => void;
+  externalOnMouseLeave?: (node: TreemapNode, e: React.MouseEvent<SVGPathElement, MouseEvent>) => void;
+  internalOnClick?: (node: TreemapNode) => void;
```
ckifer, line 388: why do we need these? There are already events here?

src/polar/Pie.tsx
```diff
@@ -456,17 +465,63 @@ function PieSectors(props: PieSectorsProps) {
           [DATA_ITEM_DATAKEY_ATTRIBUTE_NAME]: allOtherPieProps.dataKey,
         };
 
+        // Always attach external event handlers (passed as props)
+        const eventHandlers: any = {
```
ckifer: this same logic is duplicated in 6-ish places, it can probably moved to a single function

src/polar/Pie.tsx
```diff
@@ -456,17 +465,63 @@ function PieSectors(props: PieSectorsProps) {
...
+          ...adaptEventsOfChild(restOfAllOtherProps, entry, i),
+        };
+
+        // Conditionally attach internal tooltip handlers based on tooltip trigger
+        if (tooltipTrigger === 'click') {
+          const triggerInfo = {
+            tooltipPayload: entry.tooltipPayload as any,
```
ckifer: why do we need to cast and why to any?

src/state/tooltipSlice.ts
```diff
@@ -219,6 +219,11 @@ export type TooltipState = {
    * we have to change this to an array - and update all the places reading this state too.
    */
   settings: TooltipSettingsState;
+  /**
+   * This is a flag that is set when the tooltip is closing.
+   * This is used to prevent the tooltip from being closed again.
+   */
```
ckifer: Can't we just check state and derive that the tooltip is not showing?

src/chart/RechartsWrapper.tsx
```diff
@@ -51,11 +51,42 @@ export const RechartsWrapper = forwardRef(
...
     useSynchronisedEventsFromOtherCharts();
 
     const setScaleRef = useReportScale();
 
+    const handleRequestTooltipClose = useCallback(() => {
+      dispatch(tooltipCloseStart());
+      setTimeout(() => {
+        dispatch(tooltipCloseEnd());
+      }, 100);
+    }, [dispatch]);
```
ckifer: I don't think we want to introduce timeouts if we can help it, this seems like something that might cause some funky bugs in the long run

src/chart/RechartsWrapper.tsx
```diff
@@ -51,11 +51,42 @@ export const RechartsWrapper = forwardRef(
...
+        if (chartRef.current && !chartRef.current.contains(event.target as Node)) {
+          handleRequestTooltipClose();
+        }
+      };
+
+      document.addEventListener('mousedown', handleDocumentClick, true);
+      return () => {
+        document.removeEventListener('mousedown', handleDocumentClick, true);
+      };
+    }, [hasClickTooltipActive, handleRequestTooltipClose]);
```
ckifer: I think if this is the only thing we did in this PR the concern would be solved right?

we can request the tooltip to be closed if its open by calling `dispatch` and updating tooltip state directly? No need to update all of the event handlers

src/state/tooltipSlice.ts
```diff
@@ -325,6 +333,32 @@ const tooltipSlice = createSlice({
...
+       */
+      state.isClosing = true;
+      state.itemInteraction.click.active = false;
+      state.axisInteraction.click.active = false;
+    },
+    tooltipCloseEnd: state => {
+      /*
+       * End the tooltip closing process.
+       * This action resets the isClosing flag to false, allowing new tooltip interactions.
+       * Should be called after a short delay to prevent race conditions.
```
ckifer: Its adding a race condition I think

test/chart/Treemap.spec.tsx
```diff
@@ -4,11 +4,188 @@ import { render, fireEvent } from '@testing-library/react';
...
+    const { container } = render(
+      <svg>
+        <Rectangle x={0} y={0} width={100} height={100} />
+      </svg>,
+    );
+
+    expect(container.querySelector('.recharts-rectangle')).toBeInTheDocument();
+  });
+
+  test('should render Animate component directly', () => {
```
ckifer: there's a bunch of random tests in here that don't apply to Treemap at all?

src/cartesian/Bar.tsx
JoaquinPalacios: Thanks for the feedback @ckifer you are right, the code was redundant. I have just committed a refactor of it.

src/chart/RechartsWrapper.tsx
JoaquinPalacios: I have removed it. At the beginning of this work, I initially thought the React.cloneElement(children) was necessary for some edge cases, but I didn't go back to verify if it was still needed after completing the implementation. My bad - I should have double-checked this before submitting the PR.

src/cartesian/Bar.tsx
JoaquinPalacios: Thanks, I have now import it from `es-toolkit`

src/chart/Treemap.tsx
JoaquinPalacios: Hi @ckifer  these event handler props are necessary to make sure that the user-provided event handlers are actually invoked when the relevant DOM events occur on the rendered nodes, while still allowing the component to run its own internal logic for those events. I had to come up with this solution otherwise the test "navigates through nested nodes correctly" in line 220 of Treemap.spec.tsx fails because no click handler exist (internalOnClick) and it had no navigation logic (the Treemap doesn't change its view when clicked).

What would you advise me to do in this case? Keep the external event handlers to maintain the component's public API and navigation functionality or remove the navigation test? Perhaps I am missing something else. Let me know, thanks.

src/polar/Pie.tsx
JoaquinPalacios: Hi @ckifer thanks for your feedback, I've refactored this duplicated event handler logic into a reusable createEventHandlers utility function in ReactUtils.ts. I have just done a commit with the updates.

src/polar/Pie.tsx
JoaquinPalacios: This has been resolved on the commit 35036e2

src/cartesian/Bar.tsx
JoaquinPalacios: This has been resolved on the commit https://github.com/recharts/recharts/commit/35036e2756626c55f08a8c24b0eba18e00637b3f

src/state/tooltipSlice.ts
JoaquinPalacios: The flag prevents race conditions during the brief window when a tooltip is being closed but the middleware hasn't finished processing the close event. I have updated it the comment on a new commit https://github.com/recharts/recharts/pull/6007/commits/c6d62443b219e8872a663331f09dde1919daea0b . Please let me know if you consider that I should seek a different solution.

src/chart/RechartsWrapper.tsx
JoaquinPalacios: Thanks for the feedback, I completely agree about the bugs it could introduce in the long term.

I've updated the PR with a new approach that removes the timer entirely. Now, the mouseMove event immediately following a close action is caught by the middleware, which then "consumes" a gate flag and resets the state. This is now on this new commit [5b5258a](https://github.com/recharts/recharts/pull/6007/commits/5b5258acfd6cf374f9aac95053db9277e2e79504)

src/chart/RechartsWrapper.tsx
JoaquinPalacios: Thanks for the suggestion! I have just made a commit [3787ce4](https://github.com/recharts/recharts/pull/6007/commits/3787ce4b94ab5e9b855a15397406b095fec2389d) with a more simplify approach.

ckifer:
Will take a look at the changes tomorrow my time, thanks for making them!

src/state/tooltipSlice.ts
JoaquinPalacios: Now we the addition of the previous work, I've pushed a new commit that removes the isClosing state and the tooltipCloseStart/End actions entirely. The fix now exclusively uses the existing clearClickTooltip action, which is dispatched directly by the middleware for "tap away" and the useEffect for "outside clicks."

This makes the solution much simpler and more direct.

src/chart/RechartsWrapper.tsx
```diff
@@ -51,11 +51,42 @@ export const RechartsWrapper = forwardRef(
...
 
     useSynchronisedEventsFromOtherCharts();
 
     const setScaleRef = useReportScale();
 
+    const handleRequestTooltipClose = useCallback(() => {
+      dispatch(tooltipCloseStart());
+      setTimeout(() => {
+        dispatch(tooltipCloseEnd());
+      }, 100);
```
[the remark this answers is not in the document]
JoaquinPalacios: This has been addressed on a commit.

test/chart/Treemap.spec.tsx
JoaquinPalacios: Sorry, my bad. On one of first commits it failed to cover the percentage of code added, I added some tests and because eas still failing I let cursor to add more and clearly didn't help much. I have removed them entirely from this file.

src/chart/RechartsWrapper.tsx
```diff
@@ -51,11 +51,42 @@ export const RechartsWrapper = forwardRef(
...
+
+      const handleDocumentClick = (event: MouseEvent) => {
+        if (chartRef.current && !chartRef.current.contains(event.target as Node)) {
+          handleRequestTooltipClose();
+        }
+      };
+
+      document.addEventListener('mousedown', handleDocumentClick, true);
+      return () => {
+        document.removeEventListener('mousedown', handleDocumentClick, true);
```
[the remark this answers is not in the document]
JoaquinPalacios: Added 'pointerdown' on the latest commit

JoaquinPalacios:
Hi @ckifer thanks for all your feedback and thorough check of the PR. I ran into some original errors while trying to implement this feature that led to the addition of unnecessary code and a bit of overuse of Cursor. I’ve now addressed most of the comments and cleaned up the repeated code and tests. There is one question for you on one of the comments. Please let me know about it and if there is anything else I should look into it.

src/cartesian/Bar.tsx
```diff
@@ -243,28 +251,34 @@ function BarBackground(props: BarBackgroundProps) {
...
-          onMouseEnter,
-          onMouseLeave,
-          onClick,
+          ...eventHandlers,
           dataKey,
           index: i,
           className: 'recharts-bar-background-rectangle',
+          onClick: eventHandlers.onClick || noop,
+          onMouseEnter: eventHandlers.onMouseEnter || noop,
+          onMouseLeave: eventHandlers.onMouseLeave || noop,
```
ckifer: nit: can you default these to `noop` in `createEventHandlers` so that we don't have to import it in every file?

src/cartesian/Bar.tsx
```diff
@@ -243,28 +251,34 @@ function BarBackground(props: BarBackgroundProps) {
           return null;
         }
 
-        // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
-        const onMouseEnter = onMouseEnterFromContext(entry, i);
-        // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
-        const onMouseLeave = onMouseLeaveFromContext(entry, i);
-        // @ts-expect-error BarRectangleItem type definition says it's missing properties, but I can see them present in debugger!
-        const onClick = onClickFromContext(entry, i);
+        // Use the new utility function
```
ckifer: nit nit: this won't always be new, can probably remove these comments

src/chart/Treemap.tsx
```diff
@@ -793,7 +820,7 @@ class TreemapWithState extends PureComponent<InternalTreemapProps, State> {
     const { formatRoot } = this.state;
 
     if (!formatRoot) {
-      return null;
+      return <g />;
```
ckifer: why change what is returned here? Can we change this back to `null`?

src/chart/Treemap.tsx
```diff
@@ -676,22 +701,17 @@ class TreemapWithState extends PureComponent<InternalTreemapProps, State> {
...
     const { width, height, x, y, depth } = nodeProps;
     const translateX = parseInt(`${(Math.random() * 2 - 1) * width}`, 10);
-    let event = {} as any;
-    if (isLeaf || type === 'nest') {
-      event = {
-        onMouseEnter: this.handleMouseEnter.bind(this, nodeProps),
-        onMouseLeave: this.handleMouseLeave.bind(this, nodeProps),
-        onClick: this.handleClick.bind(this, nodeProps),
-      };
-    }
```
ckifer, line 690: doesn't this break nesting functionality?

src/chart/Treemap.tsx
```diff
@@ -433,14 +437,11 @@ function ContentItem({
 
   const colors = colorPanel || COLOR_PANEL;
   return (
-    <g>
+    <g onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave} onClick={onClick}>
```
ckifer, line 440: why move these to the group?

src/chart/Treemap.tsx
ckifer: trying to understand if these are being exposed in the public interface or not. If so, we can't do this

src/polar/Pie.tsx
```diff
@@ -743,46 +758,54 @@ const defaultPieProps = {
 function PieImpl(props: Props) {
   const propsWithDefaults: InternalProps = resolveDefaultProps(props, defaultPieProps);
 
+  // Merge event handlers into propsWithDefaults
+  const mergedProps: InternalProps = {
+    ...propsWithDefaults,
+    onClick: props.onClick,
+    onMouseEnter: props.onMouseEnter,
+    onMouseLeave: props.onMouseLeave,
```
ckifer: do these need included here? I think they just get filtered out below?

src/util/ReactUtils.ts
```diff
@@ -176,3 +187,34 @@ export const filterProps = (
...
+ */
+export const createEventHandlers = (
+  restOfAllOtherProps: any,
+  entry: Entry,
+  itemIndex: number,
+  triggerInfo: TriggerInfo,
+  contextHandlers: ContextHandlers,
+): any => {
+  // Always attach external event handlers (passed as props)
+  const eventHandlers: any = {
```
ckifer, line 212: Can this be typed?

test/component/Tooltip/Tooltip.visibility.spec.tsx
```diff
@@ -227,7 +227,7 @@ const FunnelChartTestCase: TooltipVisibilityTestCase = {
     </FunnelChart>
   ),
   mouseHoverSelector: funnelChartMouseHoverTooltipSelector,
-  expectedTransform: 'transform: translate(355px, 55px);',
+  expectedTransform: 'transform: translate(40px, 15px);',
```
ckifer: why did this change?

ckifer (commented):
Looking better, thank you. I still want to pull this down and test it a bit. Added a few nits and questions

src/cartesian/Bar.tsx
ckifer: looks like this is still there, just in a different place

src/chart/Treemap.tsx
PavelVanecek: Also why is this change here? How is it related to tooltip closing?

src/state/mouseEventsMiddleware.ts
```diff
@@ -25,6 +28,24 @@ mouseClickMiddleware.startListening({
           activeCoordinate: activeProps.activeCoordinate,
         }),
       );
+    } else {
```
PavelVanecek, line 31: This looks like the only relevant change to the PR title.

src/util/ReactUtils.ts
PavelVanecek: Also can this be in separate PR?

test/component/Tooltip/Tooltip.visibility.spec.tsx
PavelVanecek: This looks like a bug to me. Why did this coordinate change? I do not see how it's related to the PR target.

test/state/mouseEventsMiddleware.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      const state = store.getState().tooltip;
+      expect(state.itemInteraction.click.active).toBe(false);
+      expect(state.axisInteraction.click.active).toBe(false);
+    });
+
+    it('should activate axis tooltip when clicking on valid axis area', async () => {
+      store = createMinimalStore();
+
+      // Mock the selector to return valid active props
+      const mockSelectActivePropsFromChartPointer = vi.mocked(
```
PavelVanecek: Why so many mocks? Can we replace the mocks with actual interactions? That way we will not depend on the selector/state implementation.

PavelVanecek (changes requested):
Little bit larger than I would like but okay. Can you please figure out the failing test, and mocks, before merging?

src/cartesian/Bar.tsx
JoaquinPalacios: I have just done this. See commit d80d13d

src/cartesian/Bar.tsx
JoaquinPalacios: removed it.

src/chart/Treemap.tsx
JoaquinPalacios: Previously I had errors on the Treemap. It seems some of the changes I did a the beginning made me modify unnecessary code. I had put null back and works as expect it e7c3906

src/chart/Treemap.tsx
JoaquinPalacios: No on my end, is there anything in particular I should test to dig further?

src/chart/Treemap.tsx
JoaquinPalacios: Screenshot 2025-07-04 at 4 07 19 pm

If not, this test fails.

JoaquinPalacios: Please let me know if I should seek a different solution.

JoaquinPalacios:
Hi @ckifer this is in response to [this](https://github.com/recharts/recharts/pull/6007#discussion_r2178719541) comment that I cannot reply below it. 

As per my understanding of the code the externalOnClick, externalOnMouseEnter, externalOnMouseLeave, and internalOnClick props are not part of the public API. They are internal props within the ContentItemWithEvents component.

src/polar/Pie.tsx
JoaquinPalacios: You're right, those lines are redundant. The event handlers are passed down through the rest of the props and picked up by createEventHandlers. I'd removed them on the commit [bbcc682](https://github.com/recharts/recharts/pull/6007/commits/bbcc682e20e1b08cf696543e42548dc71b2217ee)

src/util/ReactUtils.ts
JoaquinPalacios: Hi @ckifer and @PavelVanecek I really tried to typed properly but I couldn't as the Bar.tsx expects one type and the rest another, and both won't accept to have both types on it. Happy to create another PR, question about how should I integrate this with my current PR as well?

test/component/Tooltip/Tooltip.visibility.spec.tsx
JoaquinPalacios: I had reverted back he changes and updated the trigerInfo function in the Funnel.tsx and solved the issue. The tooltip was being positioned at the top-left corner of the funnel segment instead of its center. I've corrected the logic to use the proper tooltipPosition, and the test now passes with the original expected value. See this commit 
1969150 about it.

test/state/mouseEventsMiddleware.spec.ts
JoaquinPalacios: I have simplified and updated the test file with actual interactions on this commit [fd439be](https://github.com/recharts/recharts/pull/6007/commits/fd439bea49b33934c879d535ce727b52c3a2b0c4)

ckifer:
I'm going to close this as stale, I think we can do this in a more elegant way without so many changes to events, etc.

Issue #6004, opened 2025-06-24
Bug: Reference components label prop fails to render a custom component


davesnx:
- React 19.1
- recharts 3.0.0

## What is the current behavior?

```
React does not recognize the `labelRef` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `labelref` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

https://codesandbox.io/p/sandbox/reference-line-ifoverflow-extenddomain-forked-5qnk83

## What is the expected behavior?

Render a valid element

ckifer:
Thanks!

for now prefer to render with a function `() => <text>OK</text>`

or to render it visibly
```jsx
        label={({ viewBox }) => (
          <Text x={viewBox.x - 5} y={viewBox.y + 15} viewBox={viewBox}>
            Hello
          </Text>
        )}
```
https://codesandbox.io/p/sandbox/reference-line-ifoverflow-extenddomain-forked-lzld43

Note that `<p>` or other regular HTML tags will never render correctly (without foreignObject) since we're within an SVG

ckifer:
Just to note that this didn't render in 2.x either, but it didn't throw the error in the console so we can address that

ckifer:
Happened in https://github.com/recharts/recharts/pull/5880

@saurabhraj123  can you take a look?

https://github.com/recharts/recharts/pull/5880/files#r2165224459

saurabhraj123:
> Happened in [#5880](https://github.com/recharts/recharts/pull/5880)
> 
> [@saurabhraj123](https://github.com/saurabhraj123) can you take a look?
> 
> https://github.com/recharts/recharts/pull/5880/files#r2165224459

Checking

ckifer:
console warning should be resolved next patch

ckifer:
released in 3.0.1

PR #6008, opened 2025-06-25
fix: labelRef being passed to DOM node causing warning

saurabhraj123:
## Description
`labelRef` was being passed to a DOM node. It is not a valid node prop and was causing warning.

## Related Issue
#6004 
#5880 

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Label.tsx
+++ b/src/component/Label.tsx
@@ -409,7 +409,7 @@ export function Label({ offset = 5, ...restProps }: Props) {
   }
 
   if (isValidElement(content)) {
-    return cloneElement(content, props);
+    return cloneElement(content, filterProps(props, true));
   }
 
   let label: ReactNode;


src/component/Label.tsx
```diff
@@ -409,7 +409,7 @@ export function Label({ offset = 5, ...restProps }: Props) {
   }
 
   if (isValidElement(content)) {
-    return cloneElement(content, props);
+    return cloneElement(content, filterProps(props, true));
```
ckifer, line 412: doesn't this mean it'll also be filtered out if the validElement is `Label` itself?

ckifer (approved):
Good enough for now thanks for the fix

PavelVanecek:
There is now no label being rendered, can it be related to this PR? https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=685bd07ec377767b042a0408

Edit: false alarm, the labels are still there - but delayed. Looks like a VR snapshot problem.

ckifer:
@PavelVanecek from what I can see you were right - this is the storybook link from that snapshot https://63da8268a0da9970db6992aa-tzmlwrsuhc.chromatic.com/?path=/story/examples-linechart--with-customized-label

no labels no matter the delay @saurabhraj123 can you look?

saurabhraj123:
> @PavelVanecek from what I can see you were right - this is the storybook link from that snapshot https://63da8268a0da9970db6992aa-tzmlwrsuhc.chromatic.com/?path=/story/examples-linechart--with-customized-label
> 
> no labels no matter the delay @saurabhraj123 can you look?

checking

src/component/Label.tsx
saurabhraj123: When type is `Label`, it is being handled inside `parseLabel` itself.

```javascript

if (isValidElement(label)) {
    if (label.type === Label) {
      return cloneElement<LabelProps>(label, { key: 'label-implicit', ...commonProps });
    }

    return <Label key="label-implicit" content={label} {...commonProps} />;
  }

```

Issue #5992, opened 2025-06-23
Keep order from the payload/data after hide/un-hide an element


noxify:
## Do you want to request a feature or report a bug?

First, congrats to the V3 release!

I'm currently testing some use cases and decided to use the shadcn charts wrapper.
After updating the relevant parts to make it work with recharts v3, i started to test one of my use cases:

> Note: I'm not sure if this is really a bug or a "works as designed"

* Hiding chart data on legend click

This is similar to to the charts examples:

* v2: https://release--63da8268a0da9970db6992aa.chromatic.com/?path=/story/examples-linechart--hide-on-legend-click
* v3: https://main--63da8268a0da9970db6992aa.chromatic.com/iframe.html?globals=&args=&id=examples-linechart--hide-on-legend-click

## What is the current behavior?

Currently, it seems that the `payload` is reordered based on the `inactive` flag or something else.
Based on this "reordering" the overlap ( or just the color? ) is different, based on the last choosen element.

here an short video to show what i mean:

https://github.com/user-attachments/assets/9acffd33-6a9b-4d2b-afc2-4f9d0de6524f

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar (template: https://codesandbox.io/p/sandbox/simple-line-chart-v25dl4).

* Stackblitz: https://stackblitz.com/~/github.com/noxify/recharts-repro
* Stackblitz Preview: https://rechartsrepro-gf3z--3000--b980e309.local-credentialless.webcontainer.io
* Github Repo: https://github.com/noxify/recharts-repro

or

```sh
git clone git@github.com:noxify/recharts-repro.git
cd recharts-repro
nvm use
pnpm i
pnpm dev
# open the dev instance in your browser
# and click on the legend
```

> Note: I think this behaviour is also reproducible in the mentioned v3 example 

## What is the expected behavior?

I would assume the the order keeps as is ( in my example, desktop than mobile )

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

```
> next info

Operating System:
  Platform: darwin
  Arch: arm64
  Version: macos 15.5
Binaries:
  Node: 22.16.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: 10.11.0
Relevant Packages:
  next: 15.3.4 // Latest available version is detected (15.3.4).
  eslint-config-next: 15.3.4
  react: 19.1.0
  react-dom: 19.1.0
[...]

PR #6012, opened 2025-06-25
Apply Legend itemSorter for custom content too

PavelVanecek:
## Description

This was a journey to reproduce.

## Related Issue

Reported in https://github.com/recharts/recharts/issues/5992.

This PR solves only the Legend part. I do realize @noxify you also asked for consistent stacked item order, I have an idea how to fix that but that will come in next PR.

https://github.com/user-attachments/assets/3b66de52-367b-4cba-81c7-4efe8566e867

### DIFF
--- a/src/state/legendSlice.ts
+++ b/src/state/legendSlice.ts
@@ -2,11 +2,13 @@ import { createSlice, current, PayloadAction } from '@reduxjs/toolkit';
 import { castDraft } from 'immer';
 import { LayoutType, Size } from '../util/types';
 import { HorizontalAlignmentType, LegendPayload, VerticalAlignmentType } from '../component/DefaultLegendContent';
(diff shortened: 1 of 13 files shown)


src/state/legendSlice.ts
```diff
@@ -51,6 +54,7 @@ const legendSlice = createSlice({
       state.settings.align = action.payload.align;
       state.settings.layout = action.payload.layout;
       state.settings.verticalAlign = action.payload.verticalAlign;
+      state.settings.itemSorter = action.payload.itemSorter;
```
ckifer, line 57: This is non serializable state when it's a function right

PavelVanecek: Indeed. Should be fine.

PR #6017, opened 2025-06-25
After hiding and showing stacked elements, keep them in the same order as before

PavelVanecek:
## Description

I changed the state update hooks to preserve order during update.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5992

https://github.com/user-attachments/assets/796b6c94-8241-49c8-9a13-67803f91c31e


ckifer (approved):
nice

ckifer:
this broke things, reverting

src/state/SetGraphicalItem.ts
```diff
@@ -16,16 +17,30 @@ type SetCartesianGraphicalItemProps = Omit<CartesianGraphicalItemSettings, 'stac
...
+
   useEffect(() => {
     const settings: CartesianGraphicalItemSettings = {
       ...props,
       stackId: getNormalizedStackId(props.stackId),
     };
-    dispatch(addCartesianGraphicalItem(settings));
+
+    if (prevPropsRef.current === null) {
+      dispatch(addCartesianGraphicalItem(settings));
```
ckifer, line 29: the bug here when reproducing seems to run this first -> then cleanup `removeCartesianGraphicalItem` -> then `replaceCartesianGraphicalItem` ->

PavelVanecek: Turns out that this happens in strict mode.

Which would be fine if it just did the usual mount + unmount + mount again, but strict mode has bonus! It will not clear refs that were set by the first render. So in browser, the ref would revert back to `null` after unmount, in strict mode it stays.

I have a repro + test for it: github.com/recharts/recharts-integ/pull/76 and I will push the same PR with a fix.

Issue #6022, opened 2025-06-26
[3.0.1][Bug] Charts no longer render correctly


ckifer:
Looks like we broke something in 3.0.1

~~Still investigating~~... Sorry!

ckifer:
Something before or at this 74bb50ac950178449232bddea521b316da97ca81 and after this abe67cccda1a9c446d584c240bbd1ac0446ad953

ckifer:
~~https://github.com/recharts/recharts/pull/6019~~

https://github.com/recharts/recharts/pull/6017

ckifer:
3.0.2 releasing

ckifer:
resolved in 3.0.2

PR #6026, opened 2025-06-27
After hiding and showing stacked elements, keep them in the same order as before PLUS tanstack rendering fix

PavelVanecek:
## Description

Second try on https://github.com/recharts/recharts/pull/6017

The first attempt caused https://github.com/recharts/recharts/issues/6022, I found the root cause, added a reproduction here: https://github.com/recharts/recharts-integ/pull/76 and unit test in this main repo too. Let's hope it will be fine this time.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/5992


test/helper/renderWithStrictMode.ts
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { render, RenderOptions } from '@testing-library/react';
+
+export function renderWithStrictMode<T>(
+  ui: React.ReactElement<T>,
+  options?: Omit<RenderOptions, 'wrapper'>,
+): ReturnType<typeof render> {
+  return render(ui, { wrapper: React.StrictMode, ...options });
```
ckifer, line 8: https://github.com/testing-library/react-testing-library/issues/338#issuecomment-1671842912

There might even be a Boolean option in testing library. Not that we need to use it

PR #6035, opened 2025-06-30
Update hook inspector

PavelVanecek:
## Description

I made some updates to the hook inspector. There is still a long way to go but it's getting better. I will use it to visualize what the hooks are doing and hopefully it will be useful for other devs too.

image

image

image

### DIFF
--- a/storybook/stories/Examples/BarChart/BarChart.stories.tsx
+++ b/storybook/stories/Examples/BarChart/BarChart.stories.tsx
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import { Args } from '@storybook/react';
 import { pageData, rangeData } from '../../data';
 import {
   ResponsiveContainer,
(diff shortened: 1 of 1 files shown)


storybook/storybook-addon-recharts/HookInspectorWrapper.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+function getStyles(position: Position): CSSProperties {
+  switch (position) {
+    case 'WEST':
+      return { flexDirection: 'row', alignItems: 'center' };
+    case 'EAST':
+      return { flexDirection: 'row-reverse', alignItems: 'center' };
+    case 'NORTH':
+      return { flexDirection: 'column', alignItems: 'flex-start' };
+    case 'SOUTH':
```
ckifer, line 21: Not top, bottom, left, right? 😅😝

ckifer:
Conflicts

storybook/storybook-addon-recharts/HookInspectorWrapper.tsx
PavelVanecek: Where's the fun in that

Issue #6037, opened 2025-06-30
Uncaught TypeError: Cannot read properties of null (reading 'props') in ResponsiveContainer


jderusse:
## Do you want to request a feature or report a bug?

Bug report after upgrading to 3.0

## What is the current behavior?

When rendering a chart with "optional" components, rechart trigger an exception

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar (template: https://codesandbox.io/p/sandbox/simple-line-chart-v25dl4).

```
  <ResponsiveContainer minWidth={200} minHeight={100}>
    {displayLimits && <Line />}

    <Line />
  </ResponsiveContainer>,
```

This have been introduced in https://github.com/recharts/recharts/pull/5217

## What is the expected behavior?

Expected to render the chart

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Version 3.0, all browsers

ckifer:
While this shouldn't happen, responsive container is only supposed to have one child (chart) component. Can you describe what your use case is? Why can't you conditionally render all of responsive container, or the elements within the chart (rather than the chart itself)

Typescript should also prevent more than one child in responsive container

jderusse:
Moving everything in a single `<chart>` child fixes the issue.

ckifer:
Ok! Will still try to fix the loud failure. Thank you!

PR #6038, opened 2025-06-30
Allow falsy values in Responsive container

jderusse:
This PR partially reverts #5217 to allow users to render `null` or `false` children inside the `<ResponsiveContainer />` component.

It allows code like
```
  <ResponsiveContainer minWidth={200} minHeight={100}>
    {displayLimits && <Line ... />}

    <OtherComponent />
  </ResponsiveContainer>,
```

## Description

This reintroduces the check on `React.isValidElement(child)` before cloning the children

## Related Issue

#6037

## Motivation and Context

Allows user to render conditionnaly some components

## How Has This Been Tested?

Test added

## Types of changes

## Checklist:

### DIFF
--- a/src/component/ResponsiveContainer.tsx
+++ b/src/component/ResponsiveContainer.tsx
@@ -162,19 +162,21 @@ export const ResponsiveContainer = forwardRef<HTMLDivElement, Props>(
       );
 
       return React.Children.map(children, child => {
-        return cloneElement(child, {
-          width: calculatedWidth,
-          height: calculatedHeight,
-          // calculate the actual size and override it.
-          style: {
-            height: '100%',
-            width: '100%',
-            maxHeight: calculatedHeight,
-            maxWidth: calculatedWidth,
-            // keep components style
-            ...child.props.style,
-          },
-        });
+        if (React.isValidElement<any>(child)) {
+          return cloneElement(child, {
+            width: calculatedWidth,
+            height: calculatedHeight,
+            // calculate the actual size and override it.
+            style: {
+              height: '100%',
+              width: '100%',
+              maxHeight: calculatedHeight,
+              maxWidth: calculatedWidth,
+              // keep components style
+              ...child.props.style,
+            },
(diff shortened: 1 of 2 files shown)


test/component/ResponsiveContainer.spec.tsx
```diff
@@ -83,6 +83,20 @@ describe('<ResponsiveContainer />', () => {
     expect(screen.getByTestId('inside')).toBeTruthy();
   });
 
+  it('Renders omponent when part of the children are falsy inside', () => {
```
ckifer, line 86: Nit: typo, component

src/component/ResponsiveContainer.tsx
```diff
@@ -162,19 +162,21 @@ export const ResponsiveContainer = forwardRef<HTMLDivElement, Props>(
...
+            style: {
+              height: '100%',
+              width: '100%',
+              maxHeight: calculatedHeight,
+              maxWidth: calculatedWidth,
+              // keep components style
+              ...child.props.style,
+            },
+          });
+        }
```
ckifer, line 179: in the falsy case this is now returning `undefined` in `.map`, which I think works for your case but might not work for everyone

src/component/ResponsiveContainer.tsx
```diff
@@ -162,19 +162,21 @@ export const ResponsiveContainer = forwardRef<HTMLDivElement, Props>(
...
-          style: {
-            height: '100%',
-            width: '100%',
-            maxHeight: calculatedHeight,
-            maxWidth: calculatedWidth,
-            // keep components style
-            ...child.props.style,
-          },
-        });
+        if (React.isValidElement<any>(child)) {
```
ckifer, line 165: in React 19 this check requires that your `react-is` version exactly match your `react/react-dom` versions. Not everyones always does so this will cause issues

Maybe we can just do the following for now

```tsx
 if (child == null) {
    return null;
 }
```

PR #6059, opened 2025-07-04
Add hook inspector documentation, put in all stories

PavelVanecek:
## Description

Let's put the hook inspector more forward so that people can start using it to understand what the hooks are returning.

Also updated the Polar inspector so that it allows hiding hooks


PavelVanecek:
Fixed the conflicts, and fixed the visual diff from https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=6867e6a631c09efc70292032 too

PavelVanecek:
Always something to improve!

PR #6089, opened 2025-07-12
Add IDs to graphical items

PavelVanecek:
## Description

Every potentially-stackable graphical item now has unique ID in the store.

Recharts 2.x used to identify elements by their DOM object reference. Since we no longer read DOM objects directly we needed a new identifier. I assumed that `dataKey` is going to be good enough but that turns out to be false. So let's have an ID.

## Related Issue


src/polar/Pie.tsx
```diff
@@ -802,12 +802,11 @@ export class Pie extends PureComponent<Props, State> {
 
   static defaultProps = defaultPieProps;
 
-  id = uniqueId('recharts-pie-');
-
   render() {
     return (
       <>
         <SetPolarGraphicalItem
+          id={this.props.id}
```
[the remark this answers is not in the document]
PavelVanecek: Hmmm true, this is using the wrong context. I will fix

PavelVanecek:
🎖️ for the integration tests

src/cartesian/Bar.tsx
```diff
@@ -556,14 +556,15 @@ function BarImpl(props: Props) {
 
   const barSettings: BarSettings = useMemo(
     (): BarSettings => ({
+      id: props.id,
```
PavelVanecek, line 559: This should be the autogenerated ID, not just the optional props variant - I'll fix that in next PR.

src/util/useId.ts
```diff
@@ -0,0 +1,19 @@
...
+export const useIdFallback = (): string => {
+  const [id] = React.useState(() => uniqueId('uid-'));
+  return id;
+};
+
+/*
+ * This weird syntax is used to avoid a build-time error in React 17 and earlier when building with Webpack.
+ * See https://github.com/webpack/webpack/issues/14814
+ */
+export const useId: () => string = ((React as any)['useId'.toString()] as () => string) ?? useIdFallback;
```
ckifer, line 19: weird indeed

test/cartesian/Area.spec.tsx
```diff
@@ -440,8 +440,8 @@ describe.each(chartsThatSupportArea)('<Area /> as a child of $testName', ({ Char
 
       const { rerender } = render(
         <ChartElement data={data}>
-          <Area dataKey="value" data={data2} xAxisId={7} yAxisId={9} stackId="q" hide />
-          <Customized component={<Comp />} />
+          <Area dataKey="value" data={data2} xAxisId={7} yAxisId={9} stackId="q" hide id="my-custom-area-id" />
```
ckifer, line 443: slight concern that `id` will be conflated with DOM element `id` in props even though this doesn't get passed down to the DOM element.

Maybe a different name would make sense for the public interface? `chartId?`

ckifer (approved):
are we overloading "id"? maybe maybe not, but external perception of "id" is always going to be Element.id

PavelVanecek:
The id was in props already. If it's provided then the state will use that, otherwise it will generate one. I think it's a fit here.

PR #6090, opened 2025-07-13
Add assertUniqueHtmlIds

PavelVanecek:
## Description

As I am making changes to the ID I have discovered that some of my changes made the DOM contain duplicate IDs. Tests were green so let's fix that before we have a regression.

## Related Issue


ckifer (approved):
If there were duplicates shouldn't CI be red after adding this test? Or it was already fixed?

PavelVanecek:
Duplicates are on my local changes that I haven't pushed yet.

PR #6107, opened 2025-07-15
Cartesian items use their state ID instead of generating a new one

PavelVanecek:
## Description

I will use this ID to fetch data later so first it has to be available inside of the component.

And since it's there and it's available, there is no point generating another one for the clip path.

## Related Issue

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -7,7 +7,7 @@ import { Dot } from '../shape/Dot';
 import { Layer } from '../container/Layer';
 import { LabelList } from '../component/LabelList';
 import { Global } from '../util/Global';
-import { interpolate, isNan, isNullish, isNumber, uniqueId } from '../util/DataUtils';
+import { interpolate, isNan, isNullish, isNumber } from '../util/DataUtils';
 import { getCateCoordinateOfLine, getTooltipNameProp, getValueByDataKey, StackId } from '../util/ChartUtils';
 import {
   ActiveDotType,
@@ -39,6 +39,7 @@ import { resolveDefaultProps } from '../util/resolveDefaultProps';
 import { isWellBehavedNumber } from '../util/isWellBehavedNumber';
 import { Animate } from '../animation/Animate';
 import { usePlotArea } from '../hooks';
+import { WithIdRequired, WithoutId } from '../util/useUniqueId';
 
 export type BaseValue = number | 'dataMin' | 'dataMax';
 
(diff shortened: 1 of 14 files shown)


test/container/ClipPath.spec.tsx
```diff
@@ -78,7 +78,7 @@ describe('clip paths', () => {
       });
       const clipPath = line.querySelector('clipPath');
       const clipPathId = clipPath.getAttribute('id');
-      expect(clipPathId).toEqual(expect.stringMatching(/clipPath-recharts-line-\d+/));
+      expect(clipPathId).toEqual(expect.stringMatching(/clipPath-recharts-line-[:a-z]+/));
```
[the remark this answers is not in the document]
PavelVanecek: I want to match literal colon.

src/cartesian/Area.tsx
```diff
@@ -709,14 +715,16 @@ function AreaImpl(props: Props) {
   const isPanorama = useIsPanorama();
 
   const areaSettings: AreaSettings = useMemo(
-    () => ({
+    (): AreaSettings => ({
```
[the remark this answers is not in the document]
PavelVanecek: No harm in a little bit of redundancy here

src/cartesian/Line.tsx
```diff
@@ -527,9 +537,8 @@ const errorBarDataPointFormatter: ErrorBarDataPointFormatter = (
   };
 };
 
+// eslint-disable-next-line react/prefer-stateless-function
```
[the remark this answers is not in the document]
PavelVanecek: This is correct, but for some other PR

src/cartesian/Area.tsx
```diff
@@ -257,6 +261,10 @@ function StaticArea({
 }) {
   const { layout, type, stroke, connectNulls, isRange, ...others } = props;
 
+  const { id, ...propsWithoutId } = props;
```
ckifer: Doesn't this mean that id won't make it to the dom? Not great when its user specified

PavelVanecek: The id was applied to repeat elements (like dots) so I have to filter it away. It should still remain on the top level.

PavelVanecek: Not sure if we have tests for that. I suppose I could add those first

ckifer (approved):

src/cartesian/Area.tsx
PavelVanecek: https://github.com/recharts/recharts/pull/6109

PavelVanecek: And https://github.com/recharts/recharts/issues/6110 - this problem is already there.

ckifer (approved):

ckifer (approved):
Good here?

Issue #6110, opened 2025-07-16
Area + Bar + Line + Scatter applies id on multiple elements


PavelVanecek:
## What is the current behavior?

Pass explicit `id` prop to Scatter, or Bar. Or Line. Or Area.

Run `document.querySelectorAll('[id="my-example-id"]')` in console, there should be only single element with an ID.

## What is the expected behavior?

Each Scatter point gets the same ID.

## Please provide a demo of the problem in a sandbox

https://stackblitz.com/edit/react-sskghdsd?file=src%2FExample.jsx&showSidebar=0

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

3.1.0

Issue #6098, opened 2025-07-15
Get path from the Area to use as mask/clipPath


LukaszGrela:
**Is your feature request related to a problem? Please describe.**
I'd like to mask custom dot shape using the Area shape path.

**Describe the solution you'd like**
A way to get the `path` alone from the `Area` component e.g.

```

      <AreaChart
        width={150}
        height={40}
        data={data}
        margin={{
          left: 10,
          right: 10,
          top: 0,
          bottom: 0,
        }}
      >
        <defs>
          <clipPath id="areaMask">
            <Area
              pathOnly
              type="natural"
              dataKey="uv"
              stroke={'none'}
              dot={false}
              activeDot={false}
            />
          </clipPath>
        </defs>
        <XAxis
          orientation="top"
          axisLine={false}
          tickLine={false}
          dataKey={'name'}
        />
        <Area
          type="natural"
          dataKey="uv"
          stroke={'none'}
          fillOpacity={1}
          fill="url(#colorUv)"
          dot={false}
          activeDot={dot}
        />
      </AreaChart>
```

**Describe alternatives you've considered**
Still searching for an alternative, maybe my lack of knowledge of the API prevents me from achieving this:)

**Additional context**
The idea is to use the `dot` renderer as a way of implementing gradient based activeFill functionality. I can create a custom dot, but I need that dot shape to be cropped to the shape of the area, like on the picture (where I've hardcoded the shape for the clipping path)

<img width="241" height="121" alt="Image" src="https://github.com/user-attachments/assets/27dd3177-83df-4c44-9934-de56541e76ef" />

LukaszGrela:
I could use the `Curve` but am missing the functionality to convert the data to points required by the component. The `AreaImpl` is providing this by pulling this from the state

```
  const { points, isRange, baseLine } =
    useAppSelector(state => selectArea(state, xAxisId, yAxisId, isPanorama, areaSettings)) ?? {};
```

But I can't get this in my components:
```
import { useAppSelector } from 'recharts/es6/state/hooks';
import { selectArea } from 'recharts/es6/state/selectors/areaSelectors';


const AreaMask = ({ data }: { data?: unknown[] }) => {
  const area =
    useAppSelector((state) =>
      selectArea(state, 0, 0, false, {
        dataKey: 'uv',
        connectNulls: false,
        baseValue: 120,
        data,
        stackId: undefined,
      }),
    ) ?? {};
  console.log('AreaMask', area); // This falls back to {}
[...]

PR #6111, opened 2025-07-16
Don't apply duplicate IDs in DOM

PavelVanecek:
## Description

Apply ID to element that is only rendered once instead of element that is rendered multiple times because IDs should be unique.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6110

Adds tests that cover https://github.com/recharts/recharts/issues/6098

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -287,7 +287,7 @@ function BarRectangles({
   props: BarRectanglesProps;
   showLabels: boolean;
 }) {
-  const baseProps = filterProps(props, false);
+  const { id, ...baseProps } = filterProps(props, false);
   const { shape, dataKey, activeBar } = props;
 
   const activeIndex = useAppSelector(selectActiveTooltipIndex);
@@ -503,7 +503,7 @@ class BarWithState extends PureComponent<InternalProps> {
     const clipPathId = isNullish(id) ? this.id : id;
 
     return (
-      <Layer className={layerClass}>
+      <Layer className={layerClass} id={id}>
         {needClip && (
           <defs>
             <GraphicalItemClipPath clipPathId={clipPathId} xAxisId={xAxisId} yAxisId={yAxisId} />
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -255,14 +255,17 @@ function StaticArea({
   props: InternalProps;
   showLabels: boolean;
 }) {
-  const { layout, type, stroke, connectNulls, isRange, ...others } = props;
+  const { layout, type, stroke, connectNulls, isRange } = props;
+
+  const { id, ...propsWithoutId } = props;
+  const allOtherProps = filterProps(propsWithoutId, false);
 
   return (
     <>
       {points?.length > 1 && (
-        <Layer clipPath={needClip ? `url(#clipPath-${clipPathId})` : undefined}>
+        <Layer clipPath={needClip ? `url(#clipPath-${clipPathId})` : undefined} id={id}>
           <Curve
-            {...filterProps(others, true)}
+            {...allOtherProps}
             points={points}
             connectNulls={connectNulls}
             type={type}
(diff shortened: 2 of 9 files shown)


PavelVanecek:
Hm but this will break the trick that they found in https://github.com/recharts/recharts/issues/6098. Perhaps we could make Area an exception, and apply ID to one of the curves. wdyt @ckifer ?

Or will it. Perhaps the trick works with id on the `g` element too. Let's reproduce it?

src/cartesian/Bar.tsx
```diff
@@ -503,7 +503,7 @@ class BarWithState extends PureComponent<InternalProps> {
     const clipPathId = isNullish(id) ? this.id : id;
 
     return (
-      <Layer className={layerClass}>
```
ckifer, line 506: Feel like this is a breaking change. Was the id not always going to some base element?

Oh well hopefully not too many are relying on access via id

ckifer:
Yeah let's see if it works either way first

src/cartesian/Bar.tsx
PavelVanecek: I tried on 3.1.0 and it's going to multiple elements. Because of the habit of spreading everything everywhere.

PavelVanecek:
Yep so when I try without clipPath:

```jsx
<rect width={surfaceWidth} height={surfaceHeight} fill="blue" />
```

image

When I add a clipPath (on this branch):

```jsx
<Area dataKey="uv" {...args} id="area-id" />
<defs>
  <clipPath id="areaMask">
    <use href="#area-id" />
  </clipPath>
</defs>
<rect width={surfaceWidth} height={surfaceHeight} fill="blue" clipPath="url(#areaMask)" />
```

image

When I do the same thing but switch the `id` to the Curve instead of Layer:

```jsx
<Layer clipPath={needClip ? `url(#clipPath-${clipPathId})` : undefined}>
  <Curve
    {...allOtherProps}
    id={id}
```

image

Sooo yep looks like id on the curve would be better. Now I wonder how Bars behave.

src/cartesian/Bar.tsx
ckifer: Ok. Technically breaking but breaking to fix something unreliable anyways

PavelVanecek:
Hm so for Bars this works only on the first element: https://stackblitz.com/edit/react-uugiddaz?file=src%2FExample.tsx&showSidebar=0

image

image

ckifer:
Makes sense, it's going to get the first id it finds since there should be only one. Will certainly break the area trick

PavelVanecek:
Turns out, this PR enables tricks like this:

```jsx
<BarChart {...args}>
  <defs>
    <Bar dataKey="uv" fill="#8884d8" id="my-bar" />
    <clipPath id="mask">
      <rect width={550} height={550} x={0} y={0} fill="blue" />
    </clipPath>
  </defs>
  <use href="#my-bar" clipPath="url(#mask)" fill="blue" />
</BarChart>
```

image

Which wasn't possible when the ID was on all bars.

(I learned this trick from https://stackoverflow.com/a/66418398)

Edit: or alternatively with the overlay:

```jsx
<BarChart {...args}>
  <defs>
    <Bar dataKey="uv" id="my-bar" xAxisId="1" fill="blue" />
    <clipPath id="mask">
      <rect width={550} height={550} x={0} y={0} fill="blue" />
    </clipPath>
  </defs>
  <Bar dataKey="uv" fill="#8884d8" id="my-bar2" xAxisId="2" />
  <use href="#my-bar" clipPath="url(#mask)" fill="blue" />
</BarChart>
```

image

PavelVanecek:
I'm pushing the area change because that's easy and convenient to use, and I will leave the rest unchanged because I think it's fixing a bug and we have no better place to put it anyway.

ckifer:
Ok. Probably want aim to remove it in 4.x and replace with something else? Maybe

Issue #6080, opened 2025-07-09
When hovering over ranged Area, only one of the points is active


PavelVanecek:
## What is the current behavior?

<img width="834" height="787" alt="Image" src="https://github.com/user-attachments/assets/65afc84b-d442-4649-a7f8-6ead2b394803" />

## What is the expected behavior?

<img width="837" height="790" alt="Image" src="https://github.com/user-attachments/assets/ee4fc7bf-da53-4fcc-b203-799ae0e4cfe6" />

## Please provide a demo of the problem in a sandbox

https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=686dbc68af15ee9b01e64b56

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Introduced in 3.1.0

sroy8091:
This is fixed, should we close this @PavelVanecek ?

PR #6116, opened 2025-07-17
#6080: Fix the ranged area chart hovering issue

sroy8091:
Fixed the issue in ranged area chart where after hovering mouse on it one of the points was not active

## Description
The lower bound (baseLine) points did not have the correct payload, so they were never considered "active" by the tooltip logic.

## Related Issue
https://github.com/recharts/recharts/issues/6080

## How Has This Been Tested?
- Added a unit test for this change
- Did a manual storybook test
- All unit tests are passing

Screenshot 2025-07-18 at 1 16 29 AM

## Types of changes

## Checklist:


test/chart/AreaChart.spec.tsx
```diff
@@ -614,4 +615,36 @@ describe('AreaChart', () => {
...
+    fireEvent.mouseMove(chart, { clientX: 100, clientY: 100 });
+    // There should be two active dots for the range
+    const dots = container.querySelectorAll('.customized-active-dot');
+    // Log for debugging
+    // eslint-disable-next-line no-console
+    console.log('Active dots found:', dots.length);
+    // Check if tooltip is present
+    const tooltip = container.querySelector('.recharts-tooltip-wrapper');
+    // eslint-disable-next-line no-console
+    console.log('Tooltip present:', !!tooltip);
```
ckifer: thanks for the test! Can you remove these logs?

src/cartesian/Area.tsx
```diff
@@ -661,7 +661,7 @@ class AreaWithState extends PureComponent<InternalProps> {
         />
         {this.props.isRange && Array.isArray(baseLine) && (
           <ActivePoints
-            points={baseLine}
+            points={baseLine.map((pt, idx) => ({ ...pt, payload: points[idx]?.payload }))}
```
ckifer: why do we need to add `payload` here? Looks like we didn't have to previously

ckifer (commented):
Thanks for the PR

src/cartesian/Area.tsx
sroy8091: The baseline here does have the coordinates but no payload and that's why activepoints are not being rendered. There is another fix which is to add the payload to the baseline in `computeArea` function as well. But before changing a function used everywhere, wanted a safer route and discuss. wdyt?

ckifer: Do other charts add their payload in their compute function? If so let's do it. If not I'm okay with this for now

PavelVanecek: I too prefer to have this done in the selector, because:

1. It's easier to test the selector
2. Keeps the React component (which is effectively a view layer) small and focused
3. We will eventually want to publish the selector as a hook, and we will want external consumers to have access to same information as internal

I almost wrote "consistent" but in Recharts everything is inconsistent with everything so I won't use that as an argument

PavelVanecek (changes requested):
Hello Roy, and welcome! Thanks for the PR, I'll just ask to please move the points map to selector, remove the console.log and also if you don't mind, pls merge (or rebase) latest main too. TIA!

PavelVanecek (approved):

PR #6121, opened 2025-07-19
Run test-vr in CI

PavelVanecek:
Runs the VR tests in CI. If there is a failure it will upload the report to github pages and write a comment with links:

image

Main trouble now is that the github pages upload will overwrite whatever is there now. We could hack around it with a branch and merging and things but I decided against that because:

1. we don't have that many PRs to begin with
3. if they fail, they fail on unit tests or typescript (so the VR test never starts executing)
4. if a VR test passes then it does not publish the report (to reduce the risk of overwriting another, failing, and therefore more interesting report)
5. if two concurrent PRs fail on VR tests, then yes they will overwrite each others' reports, in that case we will have to use the artifact link to download to local and examine from there (or rerun the PR build)

I think I'm fine with this limitation. If it turns out to be a problem then we can refine further.

Closes https://github.com/recharts/recharts/discussions/6031

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,7 +233,7 @@ jobs:
         if: steps.vr_test_step.outcome == 'failure' && github.event_name == 'pull_request'
         run: |
           REPORT_URL=${{ steps.deployment.outputs.page_url }}/vr-report/$SANITIZED_BRANCH_NAME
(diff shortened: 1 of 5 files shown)


PavelVanecek:
Okay I think I'm happy with the workflow now, wdyt @ckifer ?

ckifer (approved):
Yep LGTM. 

GH pages is able to host multiple index files at different paths if we were to have to move the website?

PavelVanecek:
No this is independent from the website.

PR #6126, opened 2025-07-20
Fix polarViewBox in Label, add storybooks

PavelVanecek:
## Description

I think we broke labels in polar charts in 3.0 so here is a fix.

### DIFF
--- a/src/component/Label.tsx
+++ b/src/component/Label.tsx
@@ -7,6 +7,8 @@ import { isNumOrStr, isNumber, isPercent, getPercentValue, uniqueId, mathSign, i
 import { polarToCartesian } from '../util/PolarUtils';
 import { ViewBox, PolarViewBox, CartesianViewBox, DataKey } from '../util/types';
 import { useViewBox } from '../context/chartLayoutContext';
+import { useAppSelector } from '../state/hooks';
+import { selectPolarViewBox } from '../state/selectors/polarAxisSelectors';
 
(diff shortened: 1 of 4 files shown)


ckifer (approved):

unav4ila8le:
This PR now is preventing access to the Pie viewbox. So now we cannot center a label to the Pie chart anymore. Any solutions?

The following code doesn't seem to work anymore as "viewbox" is now undefined.

```
<Label
                content={({ viewBox }) => {
                      if (viewBox && "cx" in viewBox && "cy" in viewBox) {
                        return (
                          <text
                            x={viewBox.cx}
                            y={viewBox.cy}
                            textAnchor="middle"
                            dominantBaseline="middle"
                          >
                            <tspan
                              x={viewBox.cx}
                              y={(viewBox.cy || 0) - 5}
                              style={{
                                fontSize: "16px",
                                fontWeight: "bolder",
                                fill: "var(--foreground)",
                              }}
                            >
                              {formatCompactCurrency(netWorth, currency)}
                            </tspan>
                            <tspan
                              x={viewBox.cx}
                              y={(viewBox.cy || 0) + 15}
                              style={{
                                fontSize: "12px",
                                fill: "var(--muted-foreground)",
                              }}
                            >
                              Net Worth
                            </tspan>
                          </text>
                        );
                      }
                      return null;
                    }}
                  />
```

PR #6145, opened 2025-07-27
Add more BarChart assertions

PavelVanecek:
## Description

Tests only

## Related Issue


PavelVanecek:
These are tests only, I will merge without review if you don't mind @ckifer

PR #6146, opened 2025-07-28
Add expectLastCalledWith type helper

PavelVanecek:
## Description

Vitest out of the box does not check the return type so I added a helper for it.

## Related Issue


ckifer (approved):
Well that's helpful, nice

PR #6148, opened 2025-07-28
New state types for graphical items

PavelVanecek:
## Description

I have inverted the type definition for graphical items. Before, there was a state type, and each item extended that. Now, each item defines their types and then the state has union.

The main benefit is that graphical items no longer need to declare irrelevant properties - i.e. no `stackId` on Pie and no `barSize` on Line.

631 errors

## Related Issue


ckifer (approved):
😍

PR #6150, opened 2025-07-29
Add svgOnlyNoEvents utility

PavelVanecek:
## Description

The `filterProps` utility returns `any`, we need to replace it if we want strict types. Also it is overloaded, supports like 6 different use-cases in a single function so let's split it a little.

This PR introduces `svgOnlyNoEvents` which replaces one of the `filterProps` use-cases.

634 errors -> 619

## Related Issue


src/util/svgOnlyNoEvents.ts
```diff
@@ -0,0 +1,327 @@
...
+
+export function isSvgElementPropKey(key: PropertyKey): boolean {
+  if (typeof key !== 'string') {
+    return false;
+  }
+  const allowedSvgKeys: ReadonlyArray<string> = SVGElementPropKeys;
+  return allowedSvgKeys.includes(key);
+}
+
+export function svgOnlyNoEvents<T extends Record<string, any>>(
```
ckifer: I feel like this deserves a little bit longer of a name. svgPropertiesNoEvents or something? Otherwise it doesn't really seem like a filtering function

PavelVanecek: Renamed

ckifer (approved):

src/util/svgOnlyNoEvents.ts
PavelVanecek: Ah of course I forgot to run build

Issue #6078, opened 2025-07-09
Stacked charts breaks with type="number" on xAxis


aivarassimkus:
## What is the current behavior?

Stacked charts (Bar and/or Area) does not work, when xAxis has prop type='number'.
If this prop is removed, chart appears, but ticks and bars gets misaligned
If stack IDs are removed, chart works as expected but then its not stacked.

## What is the expected behavior?

Stacked charts should be able to work when axis type is number in a same manner as it works when type is category.

## Please provide a demo of the problem in a sandbox

https://stackblitz.com/edit/react-dn1ds6tn-eds5mckj?file=src%2FExample.tsx

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Issue is reproducible with versions 3.1.0, 3.0.2

ckifer:
This is fixed, closing

PR #6162, opened 2025-07-31
fix: Exclude stacked axis domain when not relevant for axis

rinkstiekema:
Exclude stacked domain for irrelevant axes. Also added test coverage and stories.

Co-authored-by: aivarassimkus

## Description

When using an axis of type "number", the stacked domain was always taken into account, even when the stacking didn't happen on that specific axis. For example, when we are in a horizontal layout, stacking happens on the y-axis, so should only be taken into account for those axes. The opposite holds for vertical layout. 

## Related Issue

[6078](https://github.com/recharts/recharts/issues/6078)

## Motivation and Context

In my horizontal area chart, the stacked domain would be included. I'm using timestamp data (so large numbers like `1753949804`), but the stacked domain was between 0 and 100. This caused the x-axis domain to become `[0,  1753949804]`, making the chart display incorrectly.

With this fix, the stacked domain is excluded for the x-axis and only affects the y-axes.

## How Has This Been Tested?

I've tested this on storybook by adding a story. Also added tests for area and bar charts (mostly based on the work of @aivarassimkus but also included one for vertical bar charts). 

## Types of changes

## Checklist:

### DIFF
--- a/src/state/RechartsReduxContext.tsx
+++ b/src/state/RechartsReduxContext.tsx
@@ -1,7 +1,7 @@
 import { createContext } from 'react';
 import type { Action, Store } from '@reduxjs/toolkit';
 import type { Subscription } from 'react-redux';
-
+import type { CheckFrequency } from 'react-redux/es/hooks/useSelector';
 import type { RechartsRootState } from './store';
 
 /*
@@ -14,6 +14,8 @@ import type { RechartsRootState } from './store';
 export type RechartsReduxContextValue = {
   store: Store<RechartsRootState, Action>;
   subscription: Subscription;
+  stabilityCheck: CheckFrequency;
+  noopCheck: CheckFrequency;
 };
 
 /**
--- a/src/state/store.ts
+++ b/src/state/store.ts
@@ -42,8 +42,8 @@ export const createRechartsStore = (
     reducer: rootReducer,
     // redux-toolkit v1 types are unhappy with the preloadedState type. Remove the `as any` when bumping to v2
     preloadedState: preloadedState as any,
+    // @ts-expect-error redux-toolkit v1 types are unhappy with the middleware array. Remove this comment when bumping to v2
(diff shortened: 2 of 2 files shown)


src/state/selectors/axisSelectors.ts
```diff
@@ -894,11 +896,14 @@ export const combineNumericalDomain = (
...
 
-  return parseNumericalUserDomain(
-    domainDefinition,
-    mergeDomains(domainOfStackGroups, referenceElementsDomain, computeNumericalDomain(allDataWithErrorDomains)),
-    axisSettings.allowDataOverflow,
-  );
+  const shouldIncludeDomainOfStackGroups =
+    (layout === 'vertical' && axisType === 'xAxis') || (layout === 'horizontal' && axisType === 'yAxis');
+
+  const mergedDomains = shouldIncludeDomainOfStackGroups
```
ckifer, line 902: I'm confused, this works with the latest code on main? 

I pulled down the previous PR and this change did not resolve the issue anymore. Weird

rinkstiekema: The change in the PR of aivarassimkus did take a different approach. I suggested the change I've applied here in the thread, but the PR itself was never updated. Could that be the reason for confusion? 

I've tested my branch just now, and the storybook shows a correctly working stacked area chart with a number type x-axis, so yes the code works with the latest code on main. I tried to rebase with main and that gives me:
```
git rebase main
Current branch fix-only-include-stack-domain-when-relevant is up to date.
```

Let me know if I misunderstand your question though!

ckifer: I tried it both ways and neither worked. But oh well if this works it works, I'll pull it down and check it out

ckifer: checked it out and looks good

PR #6182, opened 2025-08-05
Switch Rectangle to new animation components

PavelVanecek:
## Description

Just refactor, no functional change

## Related Issue


src/shape/Rectangle.tsx
```diff
@@ -131,37 +141,56 @@ export const Rectangle: React.FC<Props> = rectangleProps => {
     );
   }
 
+  const prevWidth = prevWidthRef.current;
```
ckifer, line 144: are we doing anything with these refs?

PavelVanecek: Yep these are the previous states for interpolation below.

ckifer: lol I'm blind

Issue #6072, opened 2025-07-07
Barchart LabelList creates unnecessary line breaks


eino:
## What is the current behavior?

Hi,

When using LabelList on a vertical BarChart, the first label is well displayed, but for some reason the next ones are unnecessarily cut into multiple lines

See for instance:
![Image](https://github.com/user-attachments/assets/924f3286-a725-4353-a2e8-bbe8fc8b91bd)

## What is the expected behavior?

As the two lower labels could comfortably fit on their line, they could be displayed on a single line.

## Please provide a demo of the problem in a sandbox

Here's a codesandbox example.
https://codesandbox.io/p/sandbox/recharts-bar-chart-forked-7cf59r?file=%2Fsrc%2FComponents%2FBarChartWithLabels.jsx%3A30%2C10-30%2C18

Did I miss something, or is there a way to push the labels to stay more on a single line?

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

3.0.2, same behavior with the 2.x versions as well

Thanks in advance!

ckifer:
thanks for the repro - since this was the same in 2.x it probably won't get prioritized for a while but if anyone wants to take a look the logic is in here https://github.com/recharts/recharts/blob/main/src/component/Text.tsx#L30-L32

eino:
Hi, I investigated a bit, it seems to be because the squeezed Label components, despite being positioned `right`, are created with a `width` prop equivalent to the width of the bar (e.g. with a run of my example, first label is created with a width=950, second 90, third, 64)... same width than the respective labels that are positioned `inside`.

So there's somewhere where, depending on the `labelPosition`, the width should be inverted, with something like `width={totalWidth-width}`. Seems it should be either in the [LabelList](https://github.com/recharts/recharts/blob/main/src/component/LabelList.tsx#L43) component that create the `Labels`, but I'm not sure this component should handle such logic. So it seems the right place should be in Bar, [where LabelList is created](https://github.com/recharts/recharts/blob/main/src/cartesian/Bar.tsx#L368)? WDYT?

eino:
Hi,
I investigated some more, and actually it was simpler: the code to calculate was in place but the necessary `parentViewBox` parameter was just never passed, hence the calculation was wrong. 

[...]

PR #6214, opened 2025-08-14
fix: unnecessary line breaks in Bar labels

eino:
## Description

When creating Labels for a bar, pass the parent's viewbox, so that the label sizes can be well calculated.

## Related Issue

#6072

## Motivation and Context

There was an issue where bar labels either overflowed the chart's bounds, or on the opposite were constrained to a too small space. This was because as the parentViewBox was not available, the labels bounds were wrongly calculated (expecially for labels outside the bars) 

## How Has This Been Tested?

Tested Those changes with different BarChart configurations; in all cases the labels seem to take the space that's expected (see example screenshots).
Screenshot from 2025-08-14 10-07-01
Screenshot from 2025-08-14 10-06-49
Screenshot from 2025-08-14 10-06-37
Screenshot from 2025-08-14 10-03-59
Screenshot from 2025-08-14 09-59-02

## Types of changes

## Checklist:

### DIFF
--- a/storybook/stories/Examples/BarChart/BarChart.stories.tsx
+++ b/storybook/stories/Examples/BarChart/BarChart.stories.tsx
@@ -16,6 +16,7 @@ import {
   ErrorBar,
   Rectangle,
   LegendPayload,
+  LabelList,
 } from '../../../../src';
 import { getStoryArgsFromArgsTypesObject } from '../../API/props/utils';
 import { BarChartProps } from '../../API/props/BarChartProps';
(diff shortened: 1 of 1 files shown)


PavelVanecek:
Hi @ei, thanks for the PR. I have another change in progress that has a potential to break this again. May I please ask you to add a VR test or storybook that shows this behaviour? Or more direct unit tests? The coverage now only expects that the viewbox is provided, doesn't assert the size of the label. Thanks

eino:
Hi @PavelVanecek ,
yes I updated the storybook to show the behavior (FYI joined are the screenshot of the story before and after the fix). I'll try to create a unit test to cover the case.

Regards,
Screenshot from 2025-08-14 14-49-53
Screenshot from 2025-08-14 14-49-48
.

PavelVanecek:
Thanks! We have visual regression tests on storybooks but we usually get rate limited on those. Unit or playwright test won't have this problem.

eino:
It's not specific to BarChart, so using a Bar in a ComposedChart would give the same issue.

ckifer:
Right, but how about the other types of charts - Line, Area, Scatter, Pie, etc.?

eino:
Haven't noticed the issue with other types I've used

ckifer:
Sounds good thanks!

ckifer (approved):

Issue #6215, opened 2025-08-14
LabelList doesn't work inside Pie since v3


F1na:
## What is the current behavior?

Since the v3, LabelList inside a pie chart display nothing

<img width="256" height="304" alt="Image" src="https://github.com/user-attachments/assets/3ba0a706-3928-4671-aca5-6e009571b858" />

<img width="239" height="304" alt="Image" src="https://github.com/user-attachments/assets/2d93ceec-3a0f-4750-a1d9-55f23733f4e0" />

## What is the expected behavior?

Display labels inside Pie

## Please provide a demo of the problem in a sandbox

V3 : https://stackblitz.com/edit/react-5sr6mo79-jvlfhrwb?file=src%2FExample.tsx
V2 : https://stackblitz.com/edit/react-5sr6mo79-jvlfhrwb?file=src%2FExample.tsx

Please try to keep the demo as small as possible while still showing the problem.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Tried in version 3.&.2 and not reproduced in version 2.15.4

ckifer:
Thanks will take a look

ckifer:
released in 3.2

PR #6232, opened 2025-08-19
Simplify rendering components with and without animation

PavelVanecek:
## Description

We do not need separate component to check for `isAnimationActive` because `<JavascriptAnimation>` component does that internally already so we can just use that straight.

This change will make fixing https://github.com/recharts/recharts/issues/6215 easier because it means half the code now needs changing.

Refactor only, no functional change.

## Related Issue

https://github.com/recharts/recharts/issues/6215


ckifer (approved):
Huh, well that's nice

PR #6246, opened 2025-08-24
LabelList uses React context instead of reading DOM

PavelVanecek:
## Description

Switched LabelList from renderCallByParent (which was reading DOM nodes and parsing props from them) into React context.

+ LabelList can now be nested arbitrarily, no need to be direct child
+ Few bugfixes in the animation flow, and in passing props

Some visual diff in Storybook:
- Funnel labels moved a few pixels - looks alright to me? Should I dig into it?
- The other Funnel story had explicit fill=black in it, the labels were grey, now they are black, so I consider that a bugfix.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6215

## Motivation and Context

Removed one more usage of findChildrenByType - now only Cell is remaining, and then we can delete react-is dependency completely.

## Types of changes

## Checklist:


test-vr/__snapshots__/tests/LabelList.spec-vr.tsx-snapshots/Funnel-explicit-LabelList-1-chromium-linux.png
PavelVanecek, line 1: Now uses fill color instead of stroke, to be consistent with other chart behaviour. We can change this back if we want.

test-vr/__snapshots__/tests/LabelList.spec-vr.tsx-snapshots/Line-explicit-LabelList-1-chromium-linux.png
PavelVanecek, line 1: Order changed, now labels are below the dot. We can change this with Portals but perhaps this is fine - I assume everyone will choose different positions to prevent overlapping labels anyway?

test-vr/__snapshots__/tests/LabelList.spec-vr.tsx-snapshots/Pie-explicit-LabelList-1-chromium-linux.png
PavelVanecek, line 1: https://github.com/recharts/recharts/issues/6215 is fixed!

test-vr/__snapshots__/tests/LabelList.spec-vr.tsx-snapshots/Pie-implicit-LabelList-1-chromium-linux.png
PavelVanecek, line 1: Pie label has special behaviour, with the little label line on each label. This is the same in this PR (and recorded in multiple storybook snapshots).

If a label however has explicit `position` set then the LabelList takes over which is what we are seeing here.

Some labels render inside but because they inherit the same `fill` color as the Pie we can't see them.

test-vr/tests/LabelListComponents.tsx
```diff
@@ -97,7 +97,6 @@ export const BarImplicitLabelListTest = () => (
         dataKey="uv"
         fill="none"
         stroke="green"
-        // @ts-expect-error typescript type says that valueAccessor is not a valid prop, but it does work
```
PavelVanecek, line 100: Fixed type bug

test/cartesian/Area.animation.spec.tsx
```diff
@@ -27,55 +27,106 @@ function getAreaCurveD(container: Element): string {
 
 const expectedUvLabels: ReadonlyArray<ExpectedLabel> = [
   {
-    height: null,
```
PavelVanecek, line 30: width and height don't do anything in `text` SVG element, we should remove them.

test/cartesian/Area.animation.spec.tsx
```diff
@@ -294,7 +329,10 @@ describe('Area animation', () => {
     it('should hide labels during the animation', async () => {
       const { container, animationManager } = renderTestCase();
 
-      return expectLabelsRemainVisible(container, animationManager);
```
PavelVanecek, line 297: This test was wrong - expects labels to hide, but tests that they are visible? Fixed.

test/cartesian/Area.spec.tsx
```diff
@@ -58,12 +58,14 @@ describe.each(chartsThatSupportArea)('<Area /> as a child of $testName', ({ Char
         );
       };
 
-      render(
+      const { debug } = render(
```
PavelVanecek, line 61: I should remove this.

test/cartesian/Line.animation.spec.tsx
```diff
@@ -1153,10 +1098,9 @@ describe('Line animation', () => {
         /*
          * The labels should still be hidden! But all labels now appear again.
          * Why? Because the animation is not started yet, but they swap to new label values immediately.
-         * Unfortunately the labels still show at the old position so we have a flash of new labels at the old position.
```
PavelVanecek, line 1156: Fixed a bug

src/cartesian/Line.tsx
```diff
@@ -288,20 +292,56 @@ function Dots({
...
+        width: 0,
+        height: 0,
+      };
+      return {
+        ...viewBox,
+        value: point.value,
+        payload: point.payload,
+        viewBox,
+        /*
+         * Line is not passing parentViewBox to the LabelList so the labels can escape - looks like a bug, should we pass parentViewBox?
```
ckifer, line 318: Does anything bad happen if we do?

src/cartesian/Scatter.tsx
```diff
@@ -207,8 +240,57 @@ function ScatterLine({ points, props }: { points: ReadonlyArray<ScatterPointItem
   );
 }
 
+function ScatterLabelListProvider({
```
ckifer, line 243: I'm not sure I like that labels are now below points for scatter. Feel like words should always render on top even if both are bad. People like to render large `z` scatter items and then put text in those

Also why is it below for scatter but not for other charts?

src/polar/Radar.tsx
```diff
@@ -251,16 +256,69 @@ function Dots({ points, props }: { points: ReadonlyArray<RadarPoint>; props: Pro
...
+}: {
+  showLabels: boolean;
+  points: ReadonlyArray<RadarPoint>;
+  children: ReactNode;
+}) {
+  /*
+   * Radar provides a Cartesian label list context. Do we want to also provide a polar label list context?
+   * That way, users can choose to use polar positions for the Radar labels.
+   */
+  // const labelListEntries: ReadonlyArray<PolarLabelListEntry> = points.map(
```
ckifer, line 272: delete?

src/state/errorBarSlice.ts
```diff
@@ -40,6 +40,17 @@ const errorBarSlice = createSlice({
       }
       state[itemId].push(errorBar);
     },
+    replaceErrorBar: (
```
ckifer, line 43: did adding this do something for labels or fix something else?

ckifer (approved):
nice

src/cartesian/Line.tsx
PavelVanecek: If we add parent view box then that should push labels back into the viewbox if they are peeking out. I haven't tested it though.

src/cartesian/Scatter.tsx
PavelVanecek: Because they are rendered together with error bars and error bars are better looking under points.

Proper fix would be portals, then we could have error bars below and labels on top.

Different default for label positions would help too.

src/state/errorBarSlice.ts
PavelVanecek: Fixed infinite rerender loop. Error bars and label lists are in the same {children}

src/polar/Radar.tsx
PavelVanecek: I think that Radar being a polar chart should also support polar label positions, so I propose to pass both viewboxes and then use one or the other depending on the position? I didn't do this here because the PR is already large enough.

src/cartesian/Scatter.tsx
ckifer: Ah yeah. Another thing I suppose

src/polar/Radar.tsx
ckifer: Sounds good

PR #6260, opened 2025-08-29
Strict null checks

PavelVanecek:
## Description

536 errors -> 443

## Related Issue

### DIFF
--- a/src/component/Label.tsx
+++ b/src/component/Label.tsx
@@ -15,21 +15,15 @@ import { Text } from './Text';
 import { filterProps } from '../util/ReactUtils';
 import { isNumOrStr, isNumber, isPercent, getPercentValue, uniqueId, mathSign, isNullish } from '../util/DataUtils';
 import { polarToCartesian } from '../util/PolarUtils';
-import {
-  ViewBox,
-  PolarViewBox,
-  CartesianViewBox,
-  DataKey,
-  CartesianViewBoxRequired,
-  PolarViewBoxRequired,
-} from '../util/types';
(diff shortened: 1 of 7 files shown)


src/component/Label.tsx
```diff
@@ -174,27 +176,34 @@ const getDeltaAngle = (startAngle: number, endAngle: number) => {
...
+    case 'insideEnd':
+      labelAngle = endAngle - sign * offset;
+      direction = !clockWise;
+      break;
+    case 'end':
+      labelAngle = endAngle + sign * offset;
+      direction = clockWise;
+      break;
+    default:
+      throw new Error(`Unsupported position ${position}`);
```
ckifer, line 206: Technically a breaking change but if someone is using wrong values that's their problem

test/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "noEmit": true,
     "noImplicitAny": true,
     "skipLibCheck": true,
-    "target": "es6",
+    "target": "es2023",
```
ckifer, line 10: This is a breaking change no?

PavelVanecek: Not in the test folder

src/component/Label.tsx
PavelVanecek: There's `if` above checking for these three positions, this change is for typescript only

test/tsconfig.json
ckifer: lol 🫠🙃✅

ckifer (approved):

Issue #6257, opened 2025-08-28
LegendEffectOpacity story causes X-Axis to 'blink'


Bryhn-Bjolgerud:
I have recently upgraded from Recharts version 2.x -> 3.1.2. I had a custom legend implementation that was inspired by this [storybook example](https://recharts.org/en-US/examples/LegendEffectOpacity). After upgrading the behaviour has changed. 

At first i asumed it had something to do with my custom implementation. But after testing with the exact implementation from the Story, it still happens.

![Image](https://github.com/user-attachments/assets/a90dd594-49ad-412d-b6d7-d865ad4235a1)

## What is the expected behavior?

![Image](https://github.com/user-attachments/assets/45fcc227-b722-45e3-b996-a7673cf064b9)

## Please provide a demo of the problem in a sandbox

![Image](https://github.com/user-attachments/assets/22b171cf-f052-4f31-aafb-99b8328498a6)

In your own [story in stackblitz](https://stackblitz.com/edit/react-6f3pwhpd?file=src%2FExample.tsx) it happens. 

## Browsers
The blinking is most frequent in Google Chrome (i have tested with v.139.0.7258.155). I was rarely able to reproduce it in Firefox, but never in Safari. Didnt test any other browsers.

ckifer:
Hi, I'm confused. The stackblitz link you sent and the example link both look good to me. Can you provide the steps to reproduce?

Bryhn-Bjolgerud:
> Hi, I'm confused. The stackblitz link you sent and the example link both look good to me. Can you provide the steps to reproduce?

I have not checked out the `recharts` code to try and identify what causes this, it was just some behaviour that I noticed. But i expanded on the previous stackblitz demo by adding 2 more lines and that made the _blinking_ happen way more frequently. From what i have gathered it happens less frequently in FireFox, but is very noticeable in Google Chrome. 

#### Steps to reproduce:
- Open [demo](https://stackblitz.com/edit/react-6f3pwhpd-mtt4ktpl?file=src%2FExample.tsx) in Google Chrome. 
- Wiggle mouse back and forth across the legend and you will see the X-Axis blinking. Like it is re-rendered.

![Image](https://github.com/user-attachments/assets/c8d2f539-2e52-419b-b027-5bddf7adfe56)

[...]

PR #6262, opened 2025-08-30
Fix Axis blinking when changing props, and YAxis width=auto wrapping

PavelVanecek:
## Description

Remove the return null that makes the XAxis disappear in Chrome on weak hardware (or 4x CPU slowdown)

## More findings

Okay so once I fixed the blink, the text started wrapping with width=auto. I can fix that by excluding the axis width from tick props if width=auto (because in that case we want the Text to calculate its own width, and push it up to the axis). That creates visual diff elsewhere but that looks more like a bugfix to me so I will do that.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6257


ckifer:
Huh, that's pretty wild

ckifer (approved):

PavelVanecek:
There's something wrong with Text wrapping. Let's wait with merging this before I figure it out.

ckifer:
Sounds good

src/cartesian/XAxis.tsx
```diff
@@ -50,20 +50,14 @@ function SetXAxisSettings(props: SetXAxisSettingsProps): ReactNode {
...
     dispatch(addXAxis(settings));
     return () => {
       dispatch(removeXAxis(settings));
     };
   }, [settings, dispatch]);
-
-  if (settingsAreSynchronized) {
-    return props.children;
-  }
-  return null;
```
PavelVanecek, line 66: This was the "blink"

src/cartesian/YAxis.tsx
```diff
@@ -111,20 +128,23 @@ const YAxisImpl: FunctionComponent<Props> = (props: Props) => {
...
   return (
     <CartesianAxis
       {...allOtherProps}
+      {...restSynchronizedSettings}
       ref={cartesianAxisRef}
       labelRef={labelRef}
       scale={scale}
       x={position.x}
       y={position.y}
+      tickTextProps={width === 'auto' ? { width: undefined } : { width }}
```
PavelVanecek, line 147: If width is auto then we need to let Text calculate its own width, so we overwrite whatever width is coming from the axisProps

test-vr/__snapshots__/tests/YAxis.spec-vr.tsx-snapshots/YAxis-with-width-auto-and-long-ticks-1-chromium-linux.png
PavelVanecek, line 1: We have some visual diff in the PR.

This source code is:

```jsx
<YAxis width="auto" tickFormatter={value => `Long Long Y tick: ${value}`} />
```

I think it's an improvement - I would expect width=auto to show ticks unwrapped. Or should we prefer one word per line?

test-vr/__snapshots__/tests/YAxis.spec-vr.tsx-snapshots/YAxis-with-width-auto-and-long-ticks-and-tick-object-1-chromium-linux.png
PavelVanecek, line 1:
```jsx
<YAxis width="auto" tick={{ fontSize: 9 }} tickFormatter={value => `Long Long Y tick: ${value}`} />
```

Same as above, I consider this an improvement

PavelVanecek:
What still doesn't work is `maxLines` on the ticks. The prop is right there but doesn't do anything. I don't recall anyone complaining about it though.

ckifer:
It doesn't work pre this either?

test-vr/__snapshots__/tests/YAxis.spec-vr.tsx-snapshots/YAxis-with-width-auto-and-long-ticks-1-chromium-linux.png
ckifer: I think unwrapped by default but there were some people talking about this I think. It makes width auto hard to use when you have long ticks

ckifer (approved):

PavelVanecek:
No it ignores the prop always. Before this PR it always wraps, after it never wraps. (With width auto)

PR #6264, opened 2025-08-30
Make devTools registration opt-in (fixes #6250).

### LINKED ISSUE #6250: Redux devtools polluted since 3.0

Since the Rechart version 3.0 Redux Devtools get polluted with every chart without being able to disabled the redux devtools. This gets really messing when having multiple charts on one page.

Being able to disable redux devtools.

> 3.0

uncaught:
## Description

Makes devTools-registration opt-in via global flag.

## Related Issue

#6250

## Motivation and Context

The [Redux DevTools](https://github.com/reduxjs/redux-devtools) browser extension does not handle multiple tabs with multiple registered stores well. They are all garbaged together in every instance of the opened dev tools.

You might have a production page open in one tab, showing a few recharts-charts. And then in another tab you have your own development software running, which uses redux for example, but doesn't even have any charts in it. But still you will get all the recharts store entries there.

For recharts developers using this, I would recommend using something like:

```
window.RECHARTS_DEV_TOOLS_ENABLED = process.env.NODE_ENV === 'development';
```

## How Has This Been Tested?

Used `pnpm patch` to patch recharts and see the effects in our app in the dev tools.

## Types of changes

## Checklist:

This change only affects recharts-developers, not any chart implementation in the user-space.

### DIFF
--- a/src/state/store.ts
+++ b/src/state/store.ts
@@ -53,7 +53,7 @@ export const createRechartsStore = (
         externalEventsMiddleware.middleware,
         touchEventMiddleware.middleware,
       ]),
-    devTools: {
+    devTools: !!window.RECHARTS_DEV_TOOLS_ENABLED && {
       serialize: {
         replacer: reduxDevtoolsJsonStringifyReplacer,
       },
@@ -64,3 +64,9 @@ export const createRechartsStore = (
 
 export type RechartsRootState = ReturnType<typeof rootReducer>;
 export type AppDispatch = Dispatch<Action>;
+
+declare global {
+  interface Window {
+    RECHARTS_DEV_TOOLS_ENABLED?: boolean;
+  }
+}
--- a/storybook/preview.ts
+++ b/storybook/preview.ts
@@ -36,3 +36,6 @@ const preview: Preview = {
 export default preview;
 
 export const decorators: Decorator[] = [];
+
+// Enable dev tools in the storybook:
+window.RECHARTS_DEV_TOOLS_ENABLED = true;


src/state/store.ts
```diff
@@ -53,7 +53,7 @@ export const createRechartsStore = (
         externalEventsMiddleware.middleware,
         touchEventMiddleware.middleware,
       ]),
-    devTools: {
+    devTools: !!window.RECHARTS_DEV_TOOLS_ENABLED && {
```
PavelVanecek, line 56: Can we enable this by default in storybooks?

uncaught: Yes, I've added the flag to the storybook. I hope `preview.ts` is an appropriate place, it looks like config - I haven't worked with storybooks before.

ckifer (approved):

ckifer:
@uncaught this breaks some Next JS builds

```
.634 ReferenceError: window is not defined
5.634     at b.createRechartsStore (.next/server/chunks/284.js:1:115145)
5.634     at b.RechartsStoreProvider (.next/server/chunks/284.js:1:213462)
```

src/state/store.ts
ckifer, line 56: we need to check if window is defined

uncaught: Oh dang, please see https://github.com/recharts/recharts/pull/6309

I suppose using that `Global`-module is cleaner than using window anyway.

ckifer: Thank you 🙌

Issue #6190, opened 2025-08-07
scaleToFit={true} in Text component causes TypeError: Cannot read properties of undefined (reading 'width')


j-utunen:
This happened when I updated Recharts from 2.x to latest 3.x version in my project.

The exact point of crashing can be seen here:

<img width="1375" height="294" alt="Image" src="https://github.com/user-attachments/assets/bf2a586f-668c-4bc3-ac30-b95c19bde88a" />

ckifer:
Thanks for the report, labeled

ethphan:
@ckifer can i take this one?

ckifer:
Yeah for sure

ckifer:
@j-utunen can you give more steps to reproduce? What does your chart look like?

j-utunen:
```javascript 
<ResponsiveContainer {...props}>
                <BarChart
                    data={chartsData}
                    margin={{
                        top: 5,
                        right: 5,
                        left: calculateMargin(maxValue, store.globalZoomLevel),
                        bottom: 10,
                    }}
                    stackOffset="sign"
                >
                    <Text scaleToFit={true} />
                    <CartesianGrid strokeDasharray="3 3" />
                    <XAxis
                        dataKey="year"
                        tick={{ fontSize: "1.3rem" }}
                        interval={0}
                        style={{ fill: "black" }}
                    />
                    <YAxis tickFormatter={tickFormatter} style={{ fill: "black" }} />
                    <Tooltip
                        itemStyle={{ color: "black" }}
                        itemSorter={itemSorter}
                        formatter={itemFormatter}
                        allowEscapeViewBox={{ x: true, y: false }}
                    />
                    <Legend verticalAlign="bottom" content={renderLegend} />
                    {sortedCategories.map((item) => (
                        <Bar
                            key={item.name}
                            name={fixSvAlv(item.name, store.language)}
                            dataKey={item.name}
                            stackId="a"
                            fill={item.color}
                            maxBarSize={200}
                        />
                    ))}
                </BarChart>
</ResponsiveContainer>
```

ckifer:
Thank you

ckifer:
released in 3.2

PR #6278, opened 2025-09-04
Add docs and tests for Text component

PavelVanecek:
I think there is a bug where vertical position is calculated wrong when scaleToFit is set.

I tried to reproduce https://github.com/recharts/recharts/issues/6190 but couldn't.


PavelVanecek:
Sigh different font on the screenshot.

Issue #6267, opened 2025-09-03
Performance degradation for large datasets using bar charts - v3.0


sudo-ds:
## What is the current behavior?
Version 3.0 has significantly degraded performance on larger datasets when bar charts are present. In the provided example if you remove the bar chart, the performance is mostly the same between versions.

The provided example is a real-world scenario from our app. I'm including both charts (which have synced tooltips) so you have the full picture

_I didn't know if this would be duplicated issue but Ive came a cross #6207 which seems to have also problem with performance but we are not dealing with the `debounce`._

https://github.com/user-attachments/assets/ad568d52-297e-46b9-9a22-dc1861a05c2d

## What is the expected behavior?
- Keeping same performance or at least not significantly degraded

https://github.com/user-attachments/assets/bad07cb4-aa92-4e7e-b4b0-dbbf84569b0e

## Please provide a demo of the problem in a sandbox
- Recharts 2: https://codesandbox.io/p/sandbox/5sz5lv
- Recharts 3: https://codesandbox.io/p/devbox/recharts-2-forked-gcyhxy

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
- Version 3.0 - degraded
- Version 2.0 - works OK
- Browser/OS - Agnostic

ckifer:
well we aren't going back to 2.x so that isn't the way forwards. During refactoring we had to do a lot in the name of keeping roughly the same recharts API that everyone is familiar with. There isn't a whole lot of ways to do that without hackery (2.x) or a whole lot of state updates in a global store per chart (3.x)

> when bar charts are present

So this isn't the case when other types of chart are present? Is it the case when bar charts are rendered alone without another chart? 

From the sandbox it looks like this is the case but just checking

at what number of data points does the slowdown start?

@PavelVanecek any reason why bar charts would be slower than others?

PavelVanecek:
Not on top of my mind no.

~I looked at the videos and I can't tell the difference. Where do I look?~ Edit: it's easier to see from the sandbox than the video.

ckifer:
Yeah easier to tell from the sandbox for sure. I wouldn't necessarily call it significant either, but I'm on good hardware.

Any bit of throttled CPU and its pretty bad, especially in comparison to the same in 2.x

ckifer:
[...]

PR #6290, opened 2025-09-08
Optimize Bar component rendering when activeBar is falsy

PavelVanecek:
## Description

Do not read which index is active if we know we will never use this information anyway.

This optimizes the chart when `activeBar` is falsy. It remains slow when `activeBar` is set! But it is the same slow in 2.x.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6267

### Exhibit A, before any changes

Mouse hover - 177m jank

Lot of time spent in rendering Rectangle - updateFunctionComponent 41% (256ms)

Chart takes about 4 seconds to load before it's interactive.

### Exhibit B: with optimized animation ID dependencies

Only generate new animation IDs when props that actually animate change.

Mouse hover - 50-110ms jank

updateFunctionComponent still high 52% (539ms)

Feels about the same in the browser, no subjective improvement.

4.5s to interactive.

### Exhibit C: with hooks extracted to child component

introduce `BarRectangleWithActiveState` that calls the selectActiveTooltipIndex selector in a child, not in a loop.

Mouse hover - 50-110ms jank

updateFunctionComponent still high - 46% 416ms

This subjectively feels faster but profile says it's about the same.

3.9s to interactive.

### Exhibit D: do not call hooks if active prop is not set

Adds BarRectangleNeverActive so that we never call any hooks if `activeBar` prop is falsy.

Looks like a jackpot, the Tooltip is as smooth as it gets on my 60fps screen.

Mouse hover - no jank

updateFunctionComponent Rectangle went away - it never updates at all. The Tooltip is still there, that has to stay.

3.8s to interactive, this didn't get any better.

Okay I'm happy with this. The chart remains slow with `activeBar` prop but that's the same in 2.x so it's not a regression anymore.

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { Key, MutableRefObject, PureComponent, ReactElement, ReactNode, useCallback, useRef, useState } from 'react';
 import { clsx } from 'clsx';
 import { Series } from 'victory-vendor/d3-shape';
-import { Props as RectangleProps } from '../shape/Rectangle';
+import { Rectangle, Props as RectangleProps } from '../shape/Rectangle';
 import { Layer } from '../container/Layer';
 import { ErrorBarDataItem, ErrorBarDataPointFormatter } from './ErrorBar';
 import { Cell } from '../component/Cell';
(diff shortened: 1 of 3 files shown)


PavelVanecek:
Perhaps we could use https://react.dev/reference/react/memo for the activeBar case, let me try

ckifer:
will merge this as is

patrikholcak:
Legend

PR #6298, opened 2025-09-09
type: add extensible type for LegendPayload interface

kaleb110:
## Description

changes on LegendPayload interface of payload property object. before: payload?: {
    strokeDasharray?: number | string;
    value?: any;
  };. this type doesn't accept custom properties.

now: payload?: {
    strokeDasharray?: number | string;
    value?: any;
    [key: string]: any
  };. this allows to pass any property setted without type errors.

## Motivation and Context

type error

## Types of changes

- type errors fixed

### DIFF
--- a/src/component/DefaultLegendContent.tsx
+++ b/src/component/DefaultLegendContent.tsx
@@ -33,6 +33,7 @@ export interface LegendPayload {
   payload?: {
     strokeDasharray?: number | string;
     value?: any;
+    [key: string]: any
   };
   formatter?: Formatter;
   inactive?: boolean;


kristiandueholm:
Sorry I am new, but what is the purpose of adding this to the interface? What kind of type errors are you trying to fix?

PavelVanecek:
I have the same question

kaleb110:
> payload

i have passed custom key name to "payload", which caused type error, not a big deal but ... it kinda solved my issue

ckifer:
Can your type just be `LegendPayload & { payload: { myCustomKey} }`?

ckifer:
Going to close this as stale, I don't think this is the best way of accomplishing the goal here. We should look at refactoring to generics eventually but that is a much larger change

PR #6305, opened 2025-09-10
Remove redundant `Rectangle` import (and fix linter warning in the same file)

guscost:
## Description

The src/cartesian/Bar.tsx module has a redundant import of the <Rectangle /> component, that gets overwritten by the module-scoped Rectangle data type. This doesn't break any code, but e.g. tsup fails to parse the project without removing it.

## Motivation and Context

No change to the package behavior, this is just a quality-of-life improvement to allow tsup to compile typedefs.

## How Has This Been Tested?

Unit tests and linter run and pass

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { Key, MutableRefObject, PureComponent, ReactElement, ReactNode, useCallback, useRef, useState } from 'react';
 import { clsx } from 'clsx';
 import { Series } from 'victory-vendor/d3-shape';
-import { Rectangle, Props as RectangleProps } from '../shape/Rectangle';
+import { Props as RectangleProps } from '../shape/Rectangle';
 import { Layer } from '../container/Layer';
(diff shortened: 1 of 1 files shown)


src/cartesian/Bar.tsx
```diff
@@ -286,7 +286,7 @@ function BarBackground(props: BarBackgroundProps) {
           className: 'recharts-bar-background-rectangle',
         };
 
-        return <BarRectangle key={`background-bar-${i}`} {...barRectangleProps} />;
+        return <BarRectangle key={`background-bar-${barRectangleProps.index}`} {...barRectangleProps} />;
```
ckifer, line 289: These are the same no?

ckifer: ah, you fixed the linter error by tricking the linter I see. It isn't really fixed per se but 😅

guscost: Assigning to `barRectangleProps.index` helps signal the intent that that value is indeed intended to be a stable key suffix, and/or could be assigned w/ a different scheme, if the array index isn't stable in some mutable-data use case 🤷

Issue #6307, opened 2025-09-11
Blinking line


sysoev-dev:
## What is the current behavior?

When changing the scale of the graph with the left mouse button pressed, the graph lines disappear when moving the mouse. Maybe the docs have an old example code for a graph with scaling. I would appreciate any hints.

https://recharts.org/en-US/examples/HighlightAndZoomLineChart

## What is the expected behavior?
https://github.com/user-attachments/assets/5ff2b728-c233-4517-9d45-64be7953bb9c

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Recharts v3.2.0
Chrome, Safari and Firefox
macOS, Linux

ckifer:
seems like a bug, labeled

@PavelVanecek I wonder if there's anything we can do to address these memoization type issues more systematically. probably not

roman7722:
I am seeing the same problem

sysoev-dev:
Thanks for the prompt fix! @PavelVanecek @ckifer 
And when can we expect this functionality?

ckifer:
I will release a patch within a week probably - there is one more bug I'd like to get fixed before the next patch release

PR #6315, opened 2025-09-15
Reduce re-renders to fix blinking line in Zoom&Highlight story

PavelVanecek:
## Description

So the story in question https://main--63da8268a0da9970db6992aa.chromatic.com/iframe.html?globals=&args=&id=examples-linechart--highlight-and-zoom&viewMode=story was doing some pretty silly things:

- slice the data array on every render (like wtf, all perf goes out of the window if the core of the whole thing is forcefully updated on mouse move) - there is no point having the data array in the local state anyway so fixed
- define helper functions inside of the render method - this is React 101, fixed
- define mouse handlers without useCallback - fixed

So this helped a lot, but it's still doing some legit things like:

- define domain inline - yes this is React 101 anti-pattern but we also have that in many of our examples so let's optimize for this
- renders dynamic component (the Axes) as siblings of the static component (the Lines) - it would help if these were separated in the tree but yet again this is pattern that our docs recommend (and indeed 2.x forced this) so let's optimize for it.

I fixed these by adding React.memo with basic comparator for graphical items, and custom comparator for axes (to catch the domain case). Plus tests.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6307

## Motivation and Context

Perf perf

### Before

Mouse hover renders YAxis labels:
https://github.com/user-attachments/assets/a7ff70c2-018a-4b5e-a73b-8ad6186f109d

Mouse click renders the whole thing (but nothing visually changes):
https://github.com/user-attachments/assets/c681e3f8-935d-4a49-8257-baf874d715d3

### After

Mouse hover renders only cursor + tooltip + active dots (= only the components that visually change, can't get better than this):
https://github.com/user-attachments/assets/efa99aab-c481-4f92-98f2-8f293e85aaae

Mouse click renders some outer layers of the example (we could improve this further). The most interesting part - the line itself - no longer re-renders.
https://github.com/user-attachments/assets/6d3a435f-2d54-4193-8c78-23c81d0a19cc

## Types of changes

## Checklist:


ckifer (approved):
Wish things didn't have to be this complicated in order to keep everting working correctly 

But so it goes with the API that is defined

Issue #6266, opened 2025-09-02
Stackblitz links on website broken with dependency issues


ckifer:
## What is the current behavior?

<img width="980" height="186" alt="Image" src="https://github.com/user-attachments/assets/7f3da731-6685-40c5-80dd-d57d9a396f05" />

https://stackblitz.com/run?file=src%2FExample.tsx&showSidebar=0

## What is the expected behavior?

Screenshots or videos are welcome.

## Please provide a demo of the problem in a sandbox

https://stackblitz.com/run?file=src%2FExample.tsx&showSidebar=0

Please try to keep the demo as small as possible while still showing the problem.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Worked previously, all 3.x versions affected

ckifer:
cc @PavelVanecek looks like `immer` is breaking stackblitz links via redux toolkit

PavelVanecek:
Should we open this with immer? Or with stackblitz?

ckifer:
From what I dug up this has happened before with `immer` and `rtk` https://github.com/reduxjs/redux-toolkit/issues/2315

but its unclear what happened this time based on what I can see in their changelogs...

ckifer:
this https://github.com/immerjs/immer/commit/b2fcc6608773f0886ccc673d62deb1ca7f4a0e83 seems sus but it looks fine

ckifer:
Most likely we won't get any support from stackblitz unless we become paying customers - their github issues are a big mess

PavelVanecek:
Hm when I try to download the stackblitz project on my laptop, it works fine.

Upon closer inspection, looks like by using the project template, we opt-in to an older environment with some custom stackblitz package manager: https://developer.stackblitz.com/guides/user-guide/available-environments#webcontainers so this might be a path forward, I will give it a try.

PavelVanecek:
https://publint.dev/recharts@3.2.1 this might be relevant too

PavelVanecek:
https://arethetypeswrong.github.io/?p=recharts%403.2.1 this looks useful too. Shows all green for Recharts though

PR #6332, opened 2025-09-17
Open stackblitz projects in webcontainers environment

PavelVanecek:
## Description

Also fix transformRechartsStory so that it will work with our Brush stories.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6266


ckifer (approved):
Sweet

Issue #6353, opened 2025-09-20
Enhance ESLint jsx-key rule configuration for better IDE visibility


daiboom:
**Is your feature request related to a problem? Please describe.**
Currently, developers' IDEs (VSCode) do not display errors or warnings for missing keys in .map() functions, creating a risk that mistakes could be deployed to production.

**Describe the solution you'd like**
The eslint.config.mjs file needs to be modified.
For strict error enforcement regarding .map() keys:
```javascript
'react/jsx-key': ['error', {
  'checkFragmentShorthand': true,
  'checkKeyMustBeforeSpread': true,
  'warnOnDuplicates': true
}]
```

The current ESLint configuration appears to be set to warn, which suggests the maintainer's intention to allow flexible handling by developers. In such cases, setting it as a warning like below would alert developers:
```javascript
'react/jsx-key': 'warn'
```

PavelVanecek:
Sounds good to me, would you like to open a PR?

PR #6354, opened 2025-09-20
config: enhance jsx-key ESLint rule for better IDE visibility

daiboom:
## Description

Enhanced react/jsx-key ESLint rule configuration to display warnings in IDE for better visibility of missing key props in .map() functions.

## Related Issue

#6353 

## Motivation and Context

Current ESLint configuration doesn't show jsx-key warnings in IDE, allowing missing key props to reach production. This change improves developer awareness during coding.

## How Has This Been Tested?

- Tested ESLint configuration with missing key props
- Verified warnings appear in VSCode IDE
- Confirmed existing tests still pass
- No impact on build process

VSCode warning display when key is missing:
image

Terminal warning display when running npm run lint with missing key:
image

## Types of changes

## Checklist:

Updated react/jsx-key rule to 'warn' for better IDE visibility while maintaining flexibility for developers.

Note: This can be upgraded to 'error' in the future to enforce strict key requirements and prevent builds with missing keys.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -121,6 +121,15 @@ const reactConfig = [
   {
     name: 'x/react',
     rules: {
+      // 'react/jsx-key': [
+      //   'error',
+      //   {
+      //     checkFragmentShorthand: true,
+      //     checkKeyMustBeforeSpread: true,
+      //     warnOnDuplicates: true,
+      //   },
+      // ],
+      'react/jsx-key': 'warn',
       'react/no-did-update-set-state': 'warn',
       'react/jsx-no-constructed-context-values': 'warn',
       'react/no-unused-class-component-methods': 'off',
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   ],
   "scripts": {
     "prepare": "husky",
+    "prebuild": "npm run lint",
     "build": "npm run build-types && npm run build-cjs && npm run build-es6 && npm run build-umd && npm run test-build-output",
     "build-cjs": "rimraf lib && cross-env NODE_ENV=commonjs babel ./src -d lib --extensions '.js,.ts,.tsx'",
     "build-es6": "rimraf es6 && cross-env NODE_ENV=es6 babel ./src -d es6 --extensions '.js,.ts,.tsx'",


eslint.config.mjs
```diff
@@ -121,6 +121,15 @@ const reactConfig = [
   {
     name: 'x/react',
     rules: {
+      // 'react/jsx-key': [
```
PavelVanecek: Why is this commented out?

ckifer: That's the error version, below it's on as a warning. 

Either is better than now but I think let's turn it on as an error

daiboom: @PavelVanecek I included both configurations (error and warning) because I wasn't sure which approach would be preferred.

daiboom: @ckifer Will change to error. Thanks for the feedback!

src/polar/RadialBar.tsx
```diff
@@ -149,15 +149,19 @@ function RadialBarSectors({ sectors, allOtherRadialBarProps, showLabels }: Radia
...
           forceCornerRadius: others.forceCornerRadius,
           cornerIsExternal: others.cornerIsExternal,
           isActive,
           option: isActive ? activeShape : shape,
         };
 
-        return <RadialBarSector {...radialBarSectorProps} />;
+        return (
+          <RadialBarSector
+            key={`sector-${entry.cx}-${entry.cy}-${entry.innerRadius}-${entry.outerRadius}-${entry.startAngle}-${entry.endAngle}`}
```
ckifer: this won't be unique, please add the index back to the key

src/polar/RadialBar.tsx
```diff
@@ -342,13 +346,17 @@ class RadialBarWithState extends PureComponent<RadialBarProps> {
...
-        key: `sector-${i}`,
         className: clsx('recharts-radial-bar-background-sector', backgroundProps?.className),
         option: background,
         isActive: false,
       };
 
-      return <RadialBarSector {...props} />;
+      return (
+        <RadialBarSector
+          key={`background-${rest.cx}-${rest.cy}-${rest.innerRadius}-${rest.outerRadius}-${rest.startAngle}-${rest.endAngle}`}
```
ckifer: same here, this won't be unique

src/polar/RadialBar.tsx
daiboom: You're absolutely right! I'll add the index to both locations to ensure uniqueness. Thank you!

src/polar/RadialBar.tsx
daiboom: Fixed the same issue here!

ckifer (approved):

ckifer:
Thanks

PR #6366, opened 2025-09-22
Mobile nav menu

PavelVanecek:
## Description

Refactored navigation and added a mobile nav variant


ckifer:
Why the right side? Usually hamburger menus/navs would come from the left?

ckifer (approved):
🙏 much better no matter which side it's on

PavelVanecek:
Errr coz we already have logo on the left so I put the menu on the right and then the sidebar slides from where the button is 😬 and just so happens that it's closer to my thumb anyway. Why left?

Issue #6245, opened 2025-08-22
ResponsiveContainer now has an zero height


eino:
## What is the current behavior?

Hi,
since a recent change in the way ResponsiveContainer works, this container doesn't have a height anymore (if not explicitely given). So although it does appear on the screen, it does so via overflow, so any design relying on it to have a height (content coming after it, border, etc.) doesn't work anymore.

For instance, 
```
<div style={{ border: "2px solid red", padding: "8px" }}>
        <ResponsiveContainer width="100%" aspect={5 / 2}>
       // ...
```
until the 3.0.2 appeared as we would expect
<img width="633" height="303" alt="Image" src="https://github.com/user-attachments/assets/d2593386-64c6-41ab-ab65-6d070207be5b" />

But now doesn't anymore

<img width="633" height="303" alt="Image" src="https://github.com/user-attachments/assets/a9de00e5-282f-4945-b4f1-894e29f0da5c" />

## What is the expected behavior?

It would be expected ResponsiveContainer to have a height so it can fit in the other elements flow.

## Please provide a demo of the problem in a sandbox

https://codesandbox.io/p/devbox/recharts-bar-chart-forked-wzj9yr

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

It's versions of recharts from 3.1.0 up. In think it's caused by [this change](https://github.com/recharts/recharts/commit/c3818b79822bc27d6ac060d08481d0578df2cd37#diff-5fc91769582139b75d929f305194dd06c468324d29c9037763228599996e1515R189-R196) which fixed other issues, so I'm not sure how to reconcile the two issues.

Personally for now I need to apply a `&>div { height: auto}` on the ResponsiveContainer; not sure if there is another way to use when we don't know its height in advance (like we want to apply a specific aspect ratio).

Regards,

ckifer:
Can't look at the example just yet but can you just put the border on the chart itself? It should be the height that the responsive container was. 

Cc @PavelVanecek 

Sounds like we fixed some resize issues but created a few new ones. This probably should've been a breaking change since it modifies positioning of the wrapper that consumers are rendering things around. I think my preference would be to revert this and improve responsive container on the whole or make it implicit next major release.

eino:
Thanks for your answer. If it was just a border, indeed, but if someone mean to display content below the chart, it will now be displayed overlapping it (cf the "Some other content" text, from my screenshot, though I could have made it more visible in the screenshot).

[...]

PR #6367, opened 2025-09-22
ResponsiveContainer only shrinks in the dimensions where it needs to shrink

PavelVanecek:
## Description

Only apply the overflow when necessary, not always.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6245 I think

Hope it doesn't cause more trouble than it solves!

## Types of changes

## Checklist:


PavelVanecek:
Hm okay so this makes things worse than before. Back to the drawing board

PavelVanecek:
Okay much better now

src/util/types.ts
```diff
@@ -1053,3 +1053,5 @@ export interface PolarChartProps extends Partial<ExternalMouseEvents> {
   title?: string;
   width?: number;
 }
+
+export type Percent = `${number}%`;
```
ckifer, line 1057: this is technically a breaking change.

But I think since things were either checked for percentage or casted to a number we should be ok

PavelVanecek: Or a bugfix - typescript should have flagged strings other than percentage strings, because those won't work anyway

ckifer: Yep

ckifer (approved):

PR #6384, opened 2025-09-25
Self-healing playwright versions

PavelVanecek:
## Description

I don't like how dependabot keeps breaking the playwright setup so I automated it.


ckifer (approved):
Must've really been bothering you 😆

PavelVanecek:
Very!

ckifer:
huh

> This regular expression has an unescaped '.' before 'microsoft.com', so it might match more hosts than expected.

PR #6388, opened 2025-09-27
ResponsiveContainer is built-in to all charts

PavelVanecek:
## Description

I never really understood why does this need to be its own element so let's have these props directly on the main chart shall we.

This also means that now charts without any dimensions at all will default to 'width=100% height=100%'. Before this PR, chart without dimensions is just quietly invisible which I think is a disaster. In a way this is a change of default behaviour - but I prefer to think of it as a bugfix.

The ResponsiveContainer itself is not as important as before but it still has some use I suppose - for example in the Bubble story it allows setting the dimensions from one place. Although that could be a variable too.


ckifer (commented):
My only qualm with this PR is that I think this makes SSR even more broken than it already is?

Pre-3.0 people would SSR charts without responsive container because otherwise resize observer wouldn't be available. 

I suppose we can add a conditional to just return children at some point easily enough

src/component/ResponsiveContainer.tsx
```diff
@@ -240,3 +223,62 @@ export const ResponsiveContainer = forwardRef<HTMLDivElement, Props>(
...
+     * If it just so happens that the combination of width, height, and aspect ratio
+     * results in fixed dimensions, then we don't need to monitor the container's size.
+     * We can just provide these fixed dimensions to the context.
+     *
+     * Note that here we are not checking for positive numbers;
+     * if the user provides a zero or negative width/height, we will just pass that along
+     * as whatever size we detect won't be helping anyway.
+     */
+    return (
+      <ResponsiveContainerContextProvider width={calculatedWidth} height={calculatedHeight}>
```
PavelVanecek, line 274: I did add this shortcut @ckifer - in case we are able to calculate specific size from the given arguments then the helper div doesn't render at all.

I haven't tested SSR though, and I am not sure what is the problem in SSR in the first place, so can't tell if this makes it better or worse.

src/chart/Treemap.tsx
```diff
@@ -327,6 +328,11 @@ export interface Props {
 
   style?: React.CSSProperties;
 
+  /**
+   * This is aspect ratio of both the chart, and the individual treemap rectangles.
+   * If you define both width and height on the treemap, this prop will be ignored for the main chart,
+   * but will still be used for the rectangles.
+   */
   aspectRatio?: number;
```
PavelVanecek, line 336: Also callout on Treemap.aspectRatio - this prop was there before, but only applied on the individual nodes. I think it makes sense to have it apply on the chart as a whole as well.

src/component/ResponsiveContainer.tsx
ckifer: Yeah just need to make sure the things before this don't crash things out and actually make it here I guess

ckifer (approved):

ckifer:
We probably should update the docs when this one goes out

ckifer:
storybook test timeout hmmm, will re-run once

PR #6389, opened 2025-09-27
Add unit tests for website examples, update Label docs

PavelVanecek:
## Description

I found that the Label docs didn't actually use recharts to demonstrate recharts capabilities which I find embarassing so I updated it.

## Before

image

## After

image

image

image

### DIFF
--- a/www/test/docs/api/api.spec.ts
+++ b/www/test/docs/api/api.spec.ts
@@ -1,0 +1,34 @@
+import { describe, it, expect } from 'vitest';
+import allExamples from '../../../src/docs/api';
+
+describe('API docs', () => {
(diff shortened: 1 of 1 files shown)


www/src/docs/apiExamples/Label.tsx
```diff
@@ -48,125 +63,61 @@ const data = [
   },
 ];
 
-const example = (locale: SupportedLocale) => (
-  <div>
+const cartesianPositions = (locale: SupportedLocale) => (
+  <>
     <p style={{ fontSize: 18 }}>{localeGet(locale, 'label', 'cartesian-title')}</p>
-    <div className="cartesian-label-position">
```
ckifer, line 54: LOL

ckifer (approved):

Issue #6112, opened 2025-07-17
Since 3.1.0, charts are empty in print view


nickredmark:
I'm investigating this further, will write when I know more.

nickredmark:
I suspect it's this change.

<img width="2523" height="687" alt="Image" src="https://github.com/user-attachments/assets/0aa6d4a6-520d-4e04-8824-ee04f3b3e98a" />

nickredmark:
yeah, using ResponsiveContainer doesn't show in print

```
  return (
    <ResponsiveContainer width="100%" height="100%">
      <LineChartRecharts data={data} margin={{ top: 0, right: isPrint ? 40 : 25, bottom: 0, left: 0 }}>
        {Object.keys(omit(data?.[0], 'year', 'date'))
          .filter((key) => !!data[0]?.[key])
          .reverse()
          .map((key) => (
            <LineRecharts
              key={key}
              type="linear"
              dot={false}
              strokeWidth={isSolidLineType(key, hoveringKey) ? 5 : 2}
              onMouseOver={() => {
                setHoveringKey(key);
              }}
              onMouseOut={() => {
                setHoveringKey(null);
              }}
              activeDot={{ strokeWidth: 2, r: 6 }}
              connectNulls
              dataKey={key}
              stroke={STROKE_COLORS[getColorType(key, hoveringKey)]}
              strokeLinejoin="round"
              strokeOpacity={isSolidLineType(key, hoveringKey) ? 1 : 0.3}
              isAnimationActive={!stableRender && !isPrint}
              className="cursor-pointer"
              strokeDasharray={isSolidLineType(key, hoveringKey) ? undefined : '4 2'}
            />
          ))}
      </LineChartRecharts>
    </ResponsiveContainer>
  );
```

whereas if I set a fixed size it does

```
  return (
    <LineChartRecharts
      data={data}
      margin={{ top: 0, right: isPrint ? 40 : 25, bottom: 0, left: 0 }}
      width={100}
      height={100}
    >
      {Object.keys(omit(data?.[0], 'year', 'date'))
        .filter((key) => !!data[0]?.[key])
        .reverse()
        .map((key) => (
          <LineRecharts
            key={key}
            type="linear"
            dot={false}
            strokeWidth={isSolidLineType(key, hoveringKey) ? 5 : 2}
            onMouseOver={() => {
              setHoveringKey(key);
            }}
            onMouseOut={() => {
              setHoveringKey(null);
            }}
[...]

PR #6408, opened 2025-10-02
Responsive container integrated with RechartsWrapper

PavelVanecek:
## Description

So I went and tried to use the new built-in ResponsiveContainer on the website and I discovered several problems:

1. I did not pass couple of props, like `maxWidth` and `minHeight` to the built-in variant
2. The custom ResponsiveContainer size calculation is different from DOM size calculation
3. I struggled hard to horizontally center the chart, because of the three nested divs.

I already had fix for 1. and I was halfway through fix for 2. and then I realized that it's all pointless and we don't need another div inside div because we already have RechartsWrapper which renders its own div so let's use that instead. So that's this PR.

The gist is:
- Just pass `style` to the wrapper, let browser do all the calculations
- By default, the chart is rendered with static dimensions, meaning that it won't update when the `div.recharts-wrapper` changes
- New prop `responsive` will read updates from the wrapper div, and re-render chart, same as ResponsiveContainer used to do
- No more custom calculation logic. Now we use the browser to do calculations for us.

For now this new calculation is entirely opt-it via the new `responsive` prop. I would see what feedback we get and then get rid of `ResponsiveContainer` in 4.0.

From my experience trying the charts on the website, it works just fine. Horizontal and vertical centering is super easy now, the regular `margin: 0 auto` or flexbox works as expected.

I have also discovered that YAxis width=auto can cycle render and crash so I added a detector for it.

Treemap is excluded from the responsiveness because it calculates its sizes wrong.

ResponsiveContainer itself is unchanged.

I will update docs and the website in separate PR, now I want to see if all the tests will work and get you a chance to play with the new `responsive` on main branch.

## Related Issue

I believe this should help with https://github.com/recharts/recharts/issues/6112 where instead of hacky workarounds, regular CSS will work.

## Motivation and Context

ResponsiveContainer is a dumpster fire.

image

## Types of changes

## Checklist:

### DIFF
--- a/src/cartesian/CartesianAxis.tsx
+++ b/src/cartesian/CartesianAxis.tsx
@@ -417,7 +417,7 @@ const CartesianAxisComponent = forwardRef<CartesianAxisRef, InternalProps>((prop
   const { axisLine, width, height, className, hide, ticks, ...rest } = props;
   const [fontSize, setFontSize] = useState('');
   const [letterSpacing, setLetterSpacing] = useState('');
-  const tickRefs = useRef<Element[]>([]);
+  const tickRefs = useRef<HTMLCollectionOf<Element> | null>(null);
 
   useImperativeHandle(
     ref,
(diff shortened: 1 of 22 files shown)


ckifer (approved):
Nice, let's give it a shot

PR #6413, opened 2025-10-02
All API examples are now responsive, with VR tests

PavelVanecek:
## Description

Putting the new responsive prop to use. So far it works well.

Also I have discovered that Sankey suffers from the same problem as Treemap so I disabled the `responsive` prop in it as well.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,6 @@ env:
 
 jobs:
   list_integration_tests:
-    if: false # Temporarily disable the entire job to save CI minutes while debugging other issues
     name: List Integration Test Directories
     runs-on: ubuntu-latest
     outputs:
@@ -42,7 +41,6 @@ jobs:
           echo "directories=$JSON_ARRAY" >> $GITHUB_OUTPUT
 
   build_test_pack:
-    if: false # Temporarily disable the entire job to save CI minutes while debugging other issues
     name: Build, Test, Pack
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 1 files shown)


PavelVanecek:
Hm works on my machine

www/src/docs/apiExamples/SankeyChart.tsx
```diff
@@ -44,52 +62,31 @@ function MyCustomNode({ x, y, width, height, index, payload }: NodeProps): React
...
 
-const example = () => (
+<ResponsiveContainer width="100%" aspect={2}>
   <Sankey
-    width={960}
-    height={500}
     data={data0}
-    // @ts-expect-error Recharts type does not allow null but it should! TODO fix
-    node={MyCustomNode}
+    node={MyCustomSankeyNode}
```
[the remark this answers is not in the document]
PavelVanecek: No it doesn't.

src/chart/Sankey.tsx
```diff
@@ -966,7 +955,11 @@ export function Sankey(outsideProps: Props) {
         style={style}
         width={width}
         height={height}
-        responsive={responsive}
+        /*
+         * Sankey, same as Treemap, suffers from overfilling the container
```
ckifer, line 959: What exactly is it that is wrong with these? The animation logic causes the infinite resizing somehow?

PavelVanecek: It's because when given height X, these charts render SVG with dimensions X + N and then the responsiveness kicks in and changes the chart height to X + N which triggers new render to X + 2N and so on. Nothing to do with animations. I have since discovered that Legend has the same problem so I will work around it with overflow: hidden and enable `responsive` on these two charts again.

ckifer: Got it thanks!

PR #6415, opened 2025-10-05
Responsive website examples with VR tests. ResponsiveWrapper has overflow too

PavelVanecek:
## Description

Updating the "Examples" to responsive variants and adding tests.

I have discovered more charts that suffer from the infinite growth so I added overflow: hidden to ResponsiveWrapper which should fix it all - now a bit of whitespace is cut away but chart is stable.


ckifer:
I'm confused - if the growth is infinite how does adding overflow: hidden solve the root of the problem? Aren't things growing to infinity? Sounds like a problem that needs fixed more than just covering it up

src/chart/RechartsWrapper.tsx
```diff
@@ -314,12 +369,20 @@ export const RechartsWrapper = forwardRef<HTMLDivElement | null, RechartsWrapper
...
             className={clsx('recharts-wrapper', className)}
             style={{
               position: 'relative',
               cursor: 'default',
+              /*
+               * Some charts may render outside their boundaries,
+               * maybe because of a bug or because of rounding errors.
+               * If that happens, and the chart is responsive, it would keep growing forever
+               * to accommodate the chart that always renders a little bit too large.
+               * Let's have overflow hidden here so that the chart size stabilises.
```
ckifer, line 383: So it does actually stabilize? Or it looks like it does

PavelVanecek: With overflow=hidden, the SVG still renders larger (X+N) than the width it got assigned (X) but the RechartWrapper will not fire its own ResizeObserver in reaction to that. The N pixels are hidden.

So in a way this overflow hidden changes from one bug (chart grows forever until the browser crashes) to another bug (N pixels are cut off from the bottom of the chart) which I find acceptable.

PavelVanecek: image

It's a problem here though. I will see what else we can do.

ckifer: Got it, yeah we don't want the former if we can help it. To me it sounded like that was still happening but looks like not

ckifer (approved):

src/chart/RechartsWrapper.tsx
PavelVanecek: This works better and keeps the tooltip visible: https://github.com/recharts/recharts/pull/6416

Issue #6394, opened 2025-09-29
Performance regression in v3+ when rendering many Scatter components


harryevans1997:
## Description
When rendering a large number of `<Scatter>` components (190+ in our case) the page crashes with a maximum call stack exceeded error. This was working fine in v2.x.

## Environment

Recharts version: 3.2.1
React version: 18.3.1
Browser: Chrome

## Example

```typescript
function ProblematicChart() {
  const chartData = {
    points: Array.from({ length: 190 }, (_, i) => ({
      name: `line-${i}`,
      passed: Math.random() > 0.5,
      points: Array.from({ length: 10 }, (_, j) => ({
        x: j * 0.1,
        y: Math.random() * 2
      }))
    }))
  };

  return (
    <ResponsiveContainer height="100%" width="100%">
      <ScatterChart>
        <CartesianGrid />
        <XAxis dataKey="x" type="number" unit="m" />
        <YAxis dataKey="y" type="number" unit="m" />
        <Tooltip />
        
        {/* This crashes the page in v3+ with ~190 components */}
        {chartData.points.map((line) => (
          <Scatter
            key={line.name}
            data={line.points}
            fill={line.passed ? "#22c55e" : "#ef4444"}
            isAnimationActive={false}
            line={true}
            name={line.name}
            strokeWidth={4}
          />
        ))}
      </ScatterChart>
    </ResponsiveContainer>
  );
}
```

## Expected Behavior
The chart should render successfully as it did in v2.x, displaying all line segments without performance issues.

## Actual Behavior
The page becomes unresponsive and crashes when attempting to render 190+ <Scatter> components

## Additional Context
- We are trying to render a chart that has a scatter plot as well as an array of lines that will need to have different properties based on the data (some lines need to be red/some green for example). We previously achieved this with just `<Scatter/>` components using the `line={true}` prop. 
- This is an example of the chart I am trying to reproduce with v3+:

<img width="468" height="329" alt="Image" src="https://github.com/user-attachments/assets/99045a18-7ce4-4f3b-bf86-e1474c7fe461" />

## Requst
- The recommended approach for rendering a chart like this?
- Whether there are performance optimizations for this use case?
- If this is a known limitation or if there are plans to optimize for this pattern?

ckifer:
Hey, sorry that this is an issue. This is not a known limitation yet so thanks for the report, we'll have to do some investigation and get back to you. 

In the meantime a stack blitz reproduction would really help us out. Thanks! 

[...]

PR #6433, opened 2025-10-10
More chart perf improvements

PavelVanecek:
## Description

OK so couple of things happened here:

1. I have discovered that the chart hard crashes with more than +- 130 graphical items, depending on the browser. I found that one selector was firing repeatedly when it should have been memoized. I suspect we have ran into some sort of WeakMap cache size limits, even though I could not find any documentation on it (everyone insists it's unlimited).
2. Fortunately we never used the outcome of the mega-selector for anything else other than computing the numerical domain, so I changed the selector to output the domain directly
3. Now the chart renders again! https://github.com/recharts/recharts/issues/6394 asked for 190 Scatters, I tested with 2000 Scatters, it now renders fine. This change did not make it any faster but it no longer crashes!
4. I have also added action batching, and useLayoutEffect, to Tooltip and Legend actions. This change reduced the time to first render by 60%.
5. All of the above changes required some more changes in how we mock timers because everything is a timer these days.
6. Global variables are evil. Please don't use global variables.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6394 I think.

## Motivation and Context

I have never imagined that anyone will need more than 5 graphical items and then someone comes and asks for 190 of them.

## Types of changes

## Checklist:


PavelVanecek:
I like how the copilot summary is just a rephrasing of whatever I write in the description

ckifer:
lol that's ai for you

Issue #6424, opened 2025-10-08
`yAxis` with auto width does not resize correctly when data changes


daopk:
## What is the current behavior?

When the `yAxis` has its width set to auto, it does not resize correctly when the data changes.

https://github.com/user-attachments/assets/19da4201-f740-4d17-8ad9-40078a29d888

## What is the expected behavior?

When the data changes, the width of the `yAxis` should be recalculated accurately to fit the updated values.

https://github.com/user-attachments/assets/76613250-9c03-4709-855e-9fdc46e8ba5d

## Please provide a demo of the problem in a sandbox

https://stackblitz.com/edit/bl71p73s?file=src%2FExample.tsx

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

This issue occurs in Recharts v3.2.1.
In v3.2.0, the yAxis resizes correctly.

PR #6450, opened 2025-10-14
Only use the YAxis width oscillation detection if the difference is > 1 pixel

PavelVanecek:
## Description

The oscillation detector is there for sub-pixel rendering difference but here it was interfering with actual data changes.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6424


cusxio:
I was wondering, when will a new version 3 be release that includes this fix? Thank you.

ckifer:
I was out for the past two weeks, I will release soon

Issue #6427, opened 2025-10-08
Recharts 3 LabelsList has incorrect position within FunnelChart after upgrade from 2.x


rtivital:
## What is the current behavior?

After updating to 3.x LabelsList within FunnelChart is positioned incorrectly. Codesandbox with incorrect behavior –  https://codesandbox.io/p/devbox/mantine-react-template-forked-tc2wqg

<img width="404" height="255" alt="Image" src="https://github.com/user-attachments/assets/e1e9153d-8cd3-48c1-9a18-2bf77c0f2382" />

## What is the expected behavior?

With recharts 2.x the LabelsList is positioned correctly. Codesandbox with a correct behavior – https://codesandbox.io/p/devbox/mantine-react-template-forked-tnl6tm

<img width="393" height="254" alt="Image" src="https://github.com/user-attachments/assets/d992c69d-8f21-4b42-a41f-aa75824e87ff" />

## Please provide a demo of the problem in a sandbox

2.x – https://codesandbox.io/p/devbox/mantine-react-template-forked-tnl6tm
3.x – https://codesandbox.io/p/devbox/mantine-react-template-forked-tc2wqg

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Affected version 3.2.1, works correctly with 2.9.1

ckifer:
@PavelVanecek i think I remember a PR that changed some of the positioning for these

PavelVanecek:
Just collecting more examples

## Recharts 3.2.1

<img width="513" height="254" alt="Image" src="https://github.com/user-attachments/assets/52b1953e-10e1-45df-9f9c-19bcb5d1c13a" />

## Recharts 2.x.x

<img width="531" height="273" alt="Image" src="https://github.com/user-attachments/assets/32aad64e-a277-404d-ae8f-2088dfb78e6b" />

PavelVanecek:
(I wouldn't go as far as saying that one of those is "correct")

PavelVanecek:
I'm going to reopen this, it's not quite fixed yet. Still wip

ckifer:
Ah yeah it auto closed based on keywords on merge

PR #6453, opened 2025-10-14
Fix Funnel margin calculation

PavelVanecek:
## Description

So I went and tried to fix https://github.com/recharts/recharts/issues/6427 and in the process I have discovered that, to my horror, Funnel is doing truly atrocious things with margins:
- it counts left margin twice
- ignores top margin
- uses `top` for X coordinate
- and uses `left` for Y coordinate

So I'm fixing these. There will be visual diff, I consider this a bugfix.

## Related Issue

https://github.com/recharts/recharts/issues/6427

### Before

Funnel is vaguely present within the plot area, but it's not matching it at all:

Screenshot 2025-10-14 at 23 13 53

Screenshot 2025-10-14 at 23 13 59

### After

Funnel precisely matches margin and plot area:

Screenshot 2025-10-14 at 23 06 14

Screenshot 2025-10-14 at 23 06 03

## Types of changes

## Checklist:


ckifer (approved):
LGTM but as mentioned might cause some friction depending on what consumers think. Overall, the before is pretty whack

PR #6467, opened 2025-10-17
Fix Label + Text types

PavelVanecek:
## Description

Text component has always been able to only render `string | number` but the Label component (which calls Text internally) had children: ReactNode. How come? Well there was one small `as any` that completely disabled all type checking in several files so here is a fix.

I added tests that show what renders and what does not, and then updated the types to match. I also fixed the ref type which was wrong and hidden behind the same `as any`.

## Related Issue

https://github.com/recharts/recharts/issues/6427

## Types of changes

## Checklist:

### DIFF
--- a/src/component/Label.tsx
+++ b/src/component/Label.tsx
@@ -11,7 +11,7 @@ import {
   useMemo,
 } from 'react';
 import { clsx } from 'clsx';
-import { Text, TextAnchor, TextVerticalAnchor } from './Text';
+import { RenderableText, Text, TextAnchor, TextVerticalAnchor } from './Text';
 import { getPercentValue, isNullish, isNumber, isNumOrStr, isPercent, mathSign, uniqueId } from '../util/DataUtils';
 import { polarToCartesian } from '../util/PolarUtils';
 import { CartesianViewBoxRequired, DataKey, Percent, PolarViewBoxRequired, ViewBox } from '../util/types';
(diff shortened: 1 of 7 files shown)


src/component/Label.tsx
```diff
@@ -574,7 +581,7 @@ export function Label(outerProps: Props) {
       ref={labelRef}
       className={clsx('recharts-label', className)}
       {...attrs}
-      {...(positionAttrs as any)}
```
PavelVanecek, line 577: This right here officer

src/component/Label.tsx
```diff
@@ -54,20 +54,22 @@ type PolarLabelPosition = 'insideStart' | 'insideEnd' | 'end';
 
 export type LabelPosition = CartesianLabelPosition | PolarLabelPosition;
 
+export type LabelFormatter = (label: RenderableText) => RenderableText;
+
 interface LabelProps {
   viewBox?: ViewBox;
   parentViewBox?: ViewBox;
-  formatter?: (label: React.ReactNode) => React.ReactNode;
```
ckifer, line 60: this is a breaking change though technically 😓. Not because of what renders or not, but because someone in userland probably has something like `<>{someText}</>` because we accepted ReactNode

ckifer (approved):

src/component/Label.tsx
PavelVanecek: Yeah every bug we fix is a breaking change for people who depend on it. This is easy to fix which I think helps me think of this as a bugfix rather than a breaking change.

ckifer: yeah its fine, its calling it out in case someone reports it

Issue #6259, opened 2025-08-29
TwoLevelPieChart with activeShape does not work as expected


uncaught:
## What is the current behavior?

When combining the [TwoLevelPieChart](https://recharts.org/en-US/examples/TwoLevelPieChart) example with the [CustomActiveShapePieChart](https://recharts.org/en-US/examples/CustomActiveShapePieChart) from the documentation, the detection of what pie-slice is hovered seems broken.

<img width="321" height="285" alt="Image" src="https://github.com/user-attachments/assets/6070ae52-8e99-4d44-bc25-5c697dfb65dc" />

## What is the expected behavior?

Only the hovered pie-slice's active shape should be rendered.

## Please provide a demo of the problem in a sandbox

https://stackblitz.com/edit/react-lnlglkfj-p1air5gc?file=src%2FExample.jsx

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Recharts: 3.1.2
Chrome: 139.0.7258.155 and Firefox: 142.0.1
Windows 11

I didn't try this in a previous recharts version.

ckifer:
the reason its doing this makes sense - looks like we forgot that multiple `Pie`'s could exist within the same `PieChart`

in your screenshot index "2" of the green chart is selected so index "2" of the purple chart is as well. the data doesn't align so the selection doesn't align

labeled for fixing

shreedharbhat98:
Hi @ckifer Would be happy to help here. Please assign the issue to me.

ckifer:
Thanks! Assigned!

ckifer:
This is still an issue with the latest but the `shape` prop now allows you to be in control of what is active or not active if you'd like to
https://stackblitz.com/edit/react-lnlglkfj-whmgihju?file=src%2FExample.jsx

ckifer:
fixed in 3.5.1

Issue #6251, opened 2025-08-25
How to set active sectors in Pie chart in v3


akontanistov:
In Recharts v2, I used activeIndex to set active sectors in a Pie chart, but in v3 this prop was removed.
How can I achieve the same behavior in Recharts v3?

I try to reproduce this chart: 

<img width="248" height="259" alt="Image" src="https://github.com/user-attachments/assets/2e008127-ab0f-4bd5-8d68-adc0558f83e9" />

AndronchykBohdan:
The same question.

Bartek532:
@akontanistov did you find any way to do this?

PavelVanecek:
Have you tried using Cell?

Bartek532:
> Have you tried using Cell?

Yes, but then I don't have access to any details of the actual sector that I'm going to render? E.g. `outerRadius` or `innerRadius`?

akontanistov:
> [@akontanistov](https://github.com/akontanistov) did you find any way to do this?

Yes, we migrated to Apache ECharts

Ayman-Murshed:
Do we have a fix for this?

ckifer:
most likely what we will do is deprecate the `activeShape`/`inactiveShape` and `Cell` APIs entirely. While somewhat nice in composability, they don't allow for strict typing.

Then we will add a `content` prop similar to `Tooltip` and `Legend`. From there, you can specify custom content of each data item. We can then include an `isActive` prop if one of the data items happens to be hovered.

This satisfies the original single active item use-case as well as allowing the user to specify multiple sectors as active.

ckifer:
You can now do this with the `shape` prop in Recharts 3.5 - going to close this
https://recharts.github.io/en-US/api/Pie/#shape

Issue #6052, opened 2025-07-03
v3: Pie chart no longer supports Slices


piotrczyszczon:
## Do you want to request a feature or report a bug?
Feature request (possibly a regression from v2)

## What is the current behavior?
In Recharts v3, the `activeIndex` prop on <Pie> is not available anymore while - together with `activeShape` - it was powerful and simple way to implement Pie Slices.

Previously, in Recharts v2, We could use `activeIndex` together with `activeShape` to highlight only selected Slices of a Pie (user could click and select multiple of them).

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar.

Simple demo where I'm displaying different pie slices with different styling. Imagine that user can click on a slice and we build the `activeIndex` array dynamically.
https://stackblitz.com/edit/react-lnlglkfj?file=src%2FExample.jsx

## What is the expected behavior?
I would like to achieve something similar where we can print different slices of a Pie chart with different `activeShape` components based on user selection.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Recharts v3.x (broken)
Recharts v2.x (worked perfectly with activeIndex)
All browsers and OS

ckifer:
`activeIndex` was intentionally removed in v3. The active slice of a chart can be determined with `activeShape` and whichever slice you are clicking/hovering on like this example https://recharts.org/en-US/examples/CustomActiveShapePieChart

The array thing is interesting and is probably an oversight...

I think we missed that activeIndex could be an array. We'll get back to you on what we can do

ckifer:
So I don't think we're going to support activeIndex again, but you can do this using `Pie` and `Cell`

https://stackblitz.com/edit/react-lnlglkfj-li7kdzmd?file=src%2FExample.jsx,src%2Findex.js,package.json

Is this enough for your usecase or do you need more of the `activeShape` functionality? Active shape is meant for a singular item

piotrczyszczon:
I have two usages of `activeIndex`. The main point of the first one was to hide 'other' elements that are not selected. Indeed this thing I could achieve with transparent Cell fill color, thank you. 

The second one is to put some customization for currently hovered element, but seems like `activeShape` itself is doing the job so calculating `activeIndex` on my own seem useless, but will confirm that.

ckifer:
[...]

Issue #5999, opened 2025-06-24
v3: Pie chart no longer supports default active slice without hover


chanand:
## Do you want to request a feature or report a bug?

**Feature request** (possibly a regression from v2)

## What is the current behavior?

In Recharts v3, the `activeShape` prop on `<Pie>` only renders on **mouse hover**. There is **no way** I found to set a slice as “active” by default on initial render.

Previously, in Recharts v2, I could use `activeIndex` to show a default highlighted slice with a custom `activeShape`. This was great for visual emphasis (e.g., largest slice) and useful for dashboards and non-interactive environments.

## What is the expected behavior?

I would like to show a default active slice using `activeShape` **on initial render**, just like we could using `activeIndex` in Recharts v2.

This is important for:

* Showing most important data first
* Static dashboards or PDF exports
* Touchscreens (no hover support)

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

* **Recharts v3.x** (broken)
* **Recharts v2.x** (worked perfectly with `activeIndex`)
* All browsers and OS

PavelVanecek:
Hi @chanand, this is one of those props that we removed because it was conflicting with Tooltip state.

We can add it back but first I would like to explore alternatives. Did you know that Tooltip has `defaultIndex` prop that allows you to specify where it will display on page load? Here is an example: https://main--63da8268a0da9970db6992aa.chromatic.com/?path=/story/api-chart-piechart--simple

```
<PieChart width={400} height={400}>
  <Pie data={data} dataKey="uv" activeShape={{ fill: 'red' } />
  <Tooltip defaultIndex={3} />
</PieChart>
```

Also Tooltip works fine on touchscreens too - we have touch events support (granted, Pie onTouchMove could be better but that's another story).

Do these alternatives work for you?

chanand:
Thanks for the suggestion!

I did try using the `<Tooltip defaultIndex={...} />` approach, but unfortunately, I wasn’t able to replicate the visual behavior I had with `activeIndex`. Specifically, I rely on a customized `activeShape` that includes not just labels, but also additional `Sector` overlays and layout positioning directly inside the pie.

[...]

PR #6482, opened 2025-10-20
feat: add shape type to pie, support similar patterns to other Recharts components

ckifer:
## Description

In 3.0 we removed `activeIndex` which has proved to cause some pain for consumers.

Instead of adding it back, we're moving to a different API for "active" and "inactive" Pie shapes. This will follow similar patterns to custom `Label` and `Tooltip`s.

## Related Issue

A few directly or indirectly related

https://github.com/recharts/recharts/issues/6251
https://github.com/recharts/recharts/issues/6259
https://github.com/recharts/recharts/issues/6052
https://github.com/recharts/recharts/issues/5999

## Motivation and Context

allow people to do what they want

## How Has This Been Tested?

current unit tests pass, need to add more

## Types of changes

## Checklist:

need to update docs

## Summary by CodeRabbit

* **New Features**
  * Introduced a `shape` prop for Pie components enabling flexible sector customization. The prop accepts a React element or function receiving active state information for dynamic rendering.

* **Deprecated**
  * `activeShape` and `inactiveShape` props are deprecated; use the new `shape` prop instead. Existing implementations remain supported for backward compatibility.

### DIFF
--- a/www/src/docs/api/Pie.ts
+++ b/www/src/docs/api/Pie.ts
@@ -170,10 +170,21 @@ export const PieAPI = {
       },
     },
     {
+      name: 'shape',
+      type: 'Object | ReactElement | Function',
+      defaultVal: 'null',
+      isOptional: true,
+      desc: {
+        'en-US': 'The custom shape of a Pie Sector. Can also be used to render active sector by checking isActive.',
+        'zh-CN': 'Pie 楔子的自定义形状。也可以通过检查 isActive 来渲染激活的楔子。',
+      },
+    },
+    {
       name: 'activeShape',
       type: 'Object | ReactElement | Function',
       defaultVal: 'null',
-      isOptional: false,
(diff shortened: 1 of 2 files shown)


www/src/docs/api/Pie.ts
```diff
@@ -169,11 +169,22 @@ export const PieAPI = {
...
+        'zh-CN': 'Pie 楔子的自定义形状。也可以通过检查 isActive 来渲染激活的楔子。',
+      },
+    },
     {
       name: 'activeShape',
       type: 'Object | ReactElement | Function',
       defaultVal: 'null',
-      isOptional: false,
+      isOptional: true,
+      deprecated: true,
```
PavelVanecek, line 187: Would be great to have an alternative in the description for everything that we deprecate. Same as in code.

PavelVanecek (approved):

PR #6483, opened 2025-10-20
Strict null check

PavelVanecek:
## Related Issue

214 errors -> 135


ckifer:
Getting close

Issue #6511, opened 2025-10-27
Custom shape for Line


tarik02:
**Is your feature request related to a problem? Please describe.**
We are developing some charts using Recharts library. We got some cool-looking charts with Recharts v2 and some hacks, see example below.

The implementation was pretty hacky (taking `class Line` component from recharts, extending it and overriding method `renderTicks`). With recharts 3 it is no longer a class component and we can't do anything with internals.

**Describe the solution you'd like**
We already have `shape` prop for most primitives like `Bar`, `Area`, etc.
The same we can have for `<Line>`, so we can put some custom component instead of `<Curve>` inside it.

**Describe alternatives you've considered**
Copy whole `<Line>` component from recharts library to project and customize it there.

**Additional context**

<img width="1211" height="531" alt="Image" src="https://github.com/user-attachments/assets/0c305a3c-3fdf-4d83-8626-d55b08891607" />

PavelVanecek:
Hi @tarik02 , have you tried Scatter? It has custom symbols and it supports connecting with a line.

tarik02:
@PavelVanecek yes, I looked into it, but it does not fit for me because:
1. It requires synthetic data (I have to pass much more data to recharts to render it and it is going to affect performance).
2. With such synthetic data, we work within chart coordinate system. These ticks on my example are rendered at specific pixel interval.

A little more context - these ticks are purely visual thing to distinguish between different lines showing different metrics.

PavelVanecek:
What's synthetic data?

tarik02:
@PavelVanecek 

Let's look at example chart in the issue. We have there 6 bars (the last one is just zero) at background so it's 6 data items for chart.
Then we have lanes, if you would render them with Scatter, you would need to make data points for every tick on the line (like 20 per line).

<img width="471" height="169" alt="Image" src="https://github.com/user-attachments/assets/12aade79-7f41-4ec4-aea8-02e43750554a" />

So, I would need to render it like:
```
const basicChartData = [ ... 6 items ];
const scatterChartData = [ ... 20 items ];

<ComposedChart data={basicChartData}>
    <Bar />
    <Scatter data={scatterChartData} />
</ComposedChart>
```

Let's see by example:
```
<Scatter
    dataKey="value"
    data={[
        {
            index: 0,
            value: 15,
        },
        {
            index: 1,
            value: 20,
        },
        {
            index: 2,
            value: 17,
        },
        {
            index: 3,
            value: 25,
        },
    ]}
[...]

PR #6512, opened 2025-10-27
feat(line): custom shape

tarik02:
## Description

I added `shape` prop to `<Line>` component, just like we have this prop in other components like `<Bar>`, `<Area>`, etc.

## Related Issue

https://github.com/recharts/recharts/issues/6511

## Motivation and Context

This change is required because we need to customize lines on charts.

## How Has This Been Tested?

I added storybook story, ran all existing test cases.

image

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Line components accept a new shape prop to customize curve rendering and decorations.
  * Shape system extended to include a curve shape type for reusable curve rendering.

* **Documentation**
  * Public API docs updated with the new shape prop and usage examples.

* **Examples**
  * Added a demo story showcasing a custom line shape that draws rotated tick marks along segments.

### DIFF
--- a/storybook/stories/Examples/LineChart/CustomLineShapeChart.stories.tsx
+++ b/storybook/stories/Examples/LineChart/CustomLineShapeChart.stories.tsx
@@ -25,13 +25,13 @@ export default {
   },
 };
 
-type CustomLineShape = {
+type CustomLineShapeProps = {
   tick: ReactElement;
   tickInterval?: number;
 };
 
-const CustomLineShape = (props: CustomLineShape) => {
-  const { tick, tickInterval = 30, ...restProps } = props as CustomLineShape & CurveProps;
+const CustomLineShapeProps = (props: CurveProps & CustomLineShapeProps) => {
+  const { tick, tickInterval = 30, ...restProps } = props;
   const { points } = restProps;
 
   const ticks: ReactElement[] = [];
@@ -100,13 +100,20 @@ export const CustomLineShapeChart = {
             dataKey="pv"
             stroke="#8884d8"
             activeDot={{ r: 8 }}
-            shape={<CustomLineShape tick={<circle r={5} fill="currentColor" />} />}
+            shape={(payload: CurveProps) => (
+              <CustomLineShapeProps {...payload} tick={<circle r={5} fill="currentColor" />} />
+            )}
           />
           <Line
             type="linear"
             dataKey="uv"
             stroke="#82ca9d"
-            shape={<CustomLineShape tick={<rect x={-5} y={-5} width={10} height={10} fill="currentColor" />} />}
+            shape={(payload: CurveProps) => (
+              <CustomLineShapeProps
+                {...payload}
+                tick={<rect x={-5} y={-5} width={10} height={10} fill="currentColor" />}
+              />
+            )}
           />
           <RechartsHookInspector />
         </LineChart>


src/util/ActiveShapeUtils.tsx
```diff
@@ -21,7 +22,7 @@ import { Symbols, SymbolsProps } from '../shape/Symbols';
  * to the value of props.shapeType that is passed to the root.
  *
  */
-type ShapeType = 'trapezoid' | 'rectangle' | 'sector' | 'symbols';
+type ShapeType = 'trapezoid' | 'rectangle' | 'sector' | 'symbols' | 'curve';
```
PavelVanecek, line 25: Line is not the only user of the Shape, right? There's also Bar, Pie, RadialBar, Funnel, maybe others. What happens when one passes `shape="curve"` as a prop to Pie?

PavelVanecek: Ha wait this is not external API, is it. That's hardcoded in Line, what you're after is the rest

```
export type ActiveShape<PropsType = Record<string, any>, ElementType = SVGElement> =
  | ReactElement<SVGProps<ElementType>>
  | ((props: PropsType) => ReactElement<SVGProps<ElementType>>)
  | ((props: unknown) => JSX.Element)
  | SVGProps<ElementType>
  | boolean;
```

storybook/stories/Examples/LineChart/CustomLineShapeChart.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+          <XAxis dataKey="name" />
+          <YAxis />
+          <Legend />
+          <Tooltip cursor={{ stroke: 'gold', strokeWidth: 2 }} defaultIndex={3} />
+          <Line
+            type="linear"
+            dataKey="pv"
+            stroke="#8884d8"
+            activeDot={{ r: 8 }}
+            shape={<CustomLineShape tick={<circle r={5} fill="currentColor" />} />}
```
PavelVanecek: Element cloning was a mistake, please try to come up with an example code that passes a Component or a function instead of cloned element.

tarik02: wow, didn't that there is no `createComponent` anymore.. we are still using always using elements for custom shapes, gonna change that

ckifer: Is the current state of this code not passing a component?

ckifer:
@PavelVanecek i suppose i should make my PR use "shape" instead of "content" so remain consistent

PavelVanecek:
Tooltip and Legend have content. We won't have consistency either way.

ckifer:
Content for non-chart components, shape for chart components. At least mostly.. lol

storybook/stories/Examples/LineChart/CustomLineShapeChart.stories.tsx
PavelVanecek: It's passing element. Which we later clone and pass extra props. I would prefer to see a component, no cloning required

src/util/ActiveShapeUtils.tsx
tarik02: so nothing to do here, right?

storybook/stories/Examples/LineChart/CustomLineShapeChart.stories.tsx
tarik02: @PavelVanecek some offtop, are you going to deprecate passing `ReactElement` as shape?

and why is there this in `ActiveShape`:
```
  | ((props: PropsType) => ReactElement<SVGProps<ElementType>>)
  | ((props: unknown) => JSX.Element)
```
and not just `(props: PropsType) => JSX.Element`? cause i'm losing typings cause of that sometimes

tarik02:
@PavelVanecek I updated the shape, now passing as fn

storybook/stories/Examples/LineChart/CustomLineShapeChart.stories.tsx
PavelVanecek: Yeah at some point we will start taking types seriously and that will force us to deprecate passing elements. The return type was added in good faith but it doesn't play well when actually used. Feel free to open another issue for it, we need to fix that later too.

ckifer:
thanks @tarik02

PR #6516, opened 2025-10-28
More SEO

PavelVanecek:
## Description

Google doesn't like that the non-slash URL redirects to the slash URL so I made the slash canonical, with an alt to the non-slash, and updated the router too.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated 404 Not Found page.

* **Improvements**
  * Enhanced sitemap postprocessing and validation to emit canonical, x-default, and locale alternates and ensure correct URL forms.
  * Standardized site URLs to use trailing slashes across navigation, links, and locale routes.
  * Build now includes a sitemap postprocessing step before validation.

* **Tests**
  * Added a test enforcing trailing slashes for hardcoded routes.

### DIFF
--- a/www/scripts/postprocess-sitemap.tsx
+++ b/www/scripts/postprocess-sitemap.tsx
@@ -31,13 +31,6 @@ function makeCanonicalUrl(url: string): string {
   return `${url}/`;
 }
 
-function makeAlternateUrl(url: string): string {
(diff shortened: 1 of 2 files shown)


PavelVanecek:
Yeah it's irrelevant in the test file. I will fix the sitemap attribute though, that might be important.

Issue #6075, opened 2025-07-08
Tooltip payload data includes unrelated points when several Scatter components are used within one chart


vasiliy-l:
## What is the current behavior?

When we use `<ScatterChart>` with more than one `<Scatter>` child components, the `payload` prop for Tooltip includes unrelated data items which are not related to the data point.

The issue can be reproduced using the example from the API documentation: https://recharts.org/en-US/api/ScatterChart
The documentation page is showing tooltip with wrong data.

<img width="700"  alt="Image" src="https://github.com/user-attachments/assets/98bbc945-463f-4327-aa88-a8774bd1a698" />

Reproduced using demo project:

<img width="700"  alt="Image" src="https://github.com/user-attachments/assets/d49110f7-37be-4f88-8759-ad1d81d93182" />

## What is the expected behavior?

The problem reproduces when using recharts v3, previously the functionality worked fine.

<img width="700"  alt="Image" src="https://github.com/user-attachments/assets/cea5d257-eff2-4252-b3c4-8cd44b60e4a8" />

## Please provide a demo of the problem in a sandbox

Demo using the example from the API Documentation:
https://codesandbox.io/p/sandbox/recharts-v3-unrelated-tooltip-data-rq2clv

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Issue reproduces using recharts v3.0.2
The functionality works fine when using previous version: v2.15.4

ckifer:
Thanks, labeled as a bug

alexspurling:
A potential solution to this would be to provide the name of the currently hovered-over scatter plot(s) in the parameters to the custom tooltip renderer. Currently, that information is not available.

zhouyi207:
I think maybe change the design mode is better

zhouyi207:
I come from shadcn/ui，its too painful to create many categories of scatter.

ckifer:
Yeah we'll fix this properly in the library, no worries. Things should appear as they did in 2.x after the fix. 


Just need time to look into it 😅

zhouyi207:
hey bro, have any progress? @ckifer

ckifer:
Not yet, I need to look into it. If anyone wants to help feel free

zhouyi207:
ok, bro. I will learn js right now

RikkiMasters:
Following, same issue here

ckifer:
released in 3.6

PR #6537, opened 2025-10-30
Fix: Tooltip payload data include unrelated items when several Scatter components

SychevAndrey:
## Description

This PR closes https://github.com/recharts/recharts/issues/6075 issue

To fix that I had to find a way to get the unique identifier of a tooltip. I couldn't do that with existing API, that's why I introduces new DATA_ITEM_GRAPHICAL_ITEM_ID_ATTRIBUTE_NAME attribute to support filtering tooltip payloads by the  item that was interacted with.

## Related Issue
[Tooltip payload data include unrelated items when several Scatter components are used within one chart](https://github.com/recharts/recharts/issues/6075)

## Motivation and Context

Using dataKey and item Index can provide ambiguous results for complex charts with many data series. In our case it's the Scatter chart. But there might be other cases with that, I saw some issues related to Touch interactions with mostly the same problems.

From what I saw the id attribute is a mandatory internally, so I found it as a good candidate to be identifier that could be attached to the DOM element. 

For now, only Scatter chart sets this attribute since it was not required for other charts yet. But the touch middleware and redux reducer are ready to consume it whenever a item provides it.

## How Has This Been Tested?
Added one test in Scatter.spec that checks that tooltip payload doesn't include the data from the other data series with the same index.

Ran and fix all other tests.
Tested in storybook. Checked Scatter and Tooltip stories, comparing to the original ones.

Screenshot 2025-10-30 at 15 20 16
Screenshot 2025-10-30 at 15 37 29

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Tooltip payloads and interactions now include a graphical item identifier so tooltips and sync correctly target the specific rendered item in multi-series/multi-component charts.
  * Tooltip synchronization and touch interactions respect the per-item identifier for more precise coordination.

* **Tests**
  * Updated and added tests to validate tooltip identity, filtering, synchronization, and behavior across chart types and multiple components.

### DIFF
--- a/src/cartesian/Scatter.tsx
+++ b/src/cartesian/Scatter.tsx
@@ -220,6 +220,7 @@ const computeLegendPayloadFromScatterProps = (props: Props): ReadonlyArray<Legen
 };
 
 type InputRequiredToComputeTooltipEntrySettings = {
+  id: GraphicalItemId;
   dataKey?: DataKey<any> | undefined;
   points?: ReadonlyArray<ScatterPointItem>;
   stroke?: string;
@@ -232,6 +233,7 @@ type InputRequiredToComputeTooltipEntrySettings = {
 
 const SetScatterTooltipEntrySettings = React.memo(
   ({
+    id,
     dataKey,
     points,
     stroke,
@@ -244,6 +246,7 @@ const SetScatterTooltipEntrySettings = React.memo(
     const tooltipEntrySettings: TooltipPayloadConfiguration = {
       dataDefinedOnItem: points?.map((p: ScatterPointItem) => p.tooltipPayload),
       positions: points?.map((p: ScatterPointItem) => p.tooltipPosition),
+      graphicalItemId: id,
       settings: {
         stroke,
         strokeWidth,
@@ -715,6 +718,7 @@ function ScatterImpl(props: WithIdRequired<Props>) {
   return (
     <>
       <SetScatterTooltipEntrySettings
+        id={props.id}
         dataKey={props.dataKey}
         points={points}
         stroke={props.stroke}


src/context/tooltipContext.tsx
```diff
@@ -21,6 +22,7 @@ export type ActivateTooltipAction<T extends TooltipPayloadType> = (
 export const useMouseEnterItemDispatch = <T extends TooltipPayloadType>(
   onMouseEnterFromProps: ActivateTooltipAction<T> | undefined,
   dataKey: DataKey<any> | undefined,
+  graphicalItemId?: GraphicalItemId,
```
PavelVanecek, line 25: Why optional? Let's make the ID required. All graphical items have one.

src/cartesian/Scatter.tsx
```diff
@@ -346,6 +351,7 @@ function ScatterSymbols(props: ScatterSymbolsProps) {
           ...entry,
           [DATA_ITEM_INDEX_ATTRIBUTE_NAME]: i,
           [DATA_ITEM_DATAKEY_ATTRIBUTE_NAME]: String(dataKey),
```
PavelVanecek, line 406: I put the dataKey as an identifier here, it suffers from two problems: 1. it's not unique, and 2. it's not serializable (it can be a function). So +1 the `GraphicalItemId` is a better identifier.

src/state/selectors/combiners/combineTooltipPayloadConfigurations.ts
```diff
@@ -20,12 +21,24 @@ export const combineTooltipPayloadConfigurations = (
...
+  let filterByDataKey: DataKey<any> | undefined, filterByGraphicalItemId: GraphicalItemId | undefined;
   if (trigger === 'hover') {
     filterByDataKey = tooltipState.itemInteraction.hover.dataKey;
+    filterByGraphicalItemId = tooltipState.itemInteraction.hover.graphicalItemId;
   } else {
     filterByDataKey = tooltipState.itemInteraction.click.dataKey;
+    filterByGraphicalItemId = tooltipState.itemInteraction.click.graphicalItemId;
   }
+
+  if (filterByGraphicalItemId != null) {
```
PavelVanecek, line 33: What happens when defaultIndex is set? See the block below.

src/state/tooltipSlice.ts
```diff
@@ -4,11 +4,14 @@ import { TooltipTrigger } from '../chart/types';
...
 import { CartesianViewBoxRequired, Coordinate, DataKey, PolarCoordinate } from '../util/types';
 import { AxisId } from './cartesianAxisSlice';
+import { GraphicalItemId } from './graphicalItemsSlice';
 
 /**
  * One Tooltip can display multiple TooltipPayloadEntries at a time.
  */
-export type TooltipPayloadEntry = Payload<ValueType, NameType>;
+export type TooltipPayloadEntry = Payload<ValueType, NameType> & {
+  graphicalItemId?: GraphicalItemId;
```
PavelVanecek, line 13: Ditto. Let's get rid of the question marks, as many as we can.

src/state/tooltipSlice.ts
```diff
@@ -143,6 +150,10 @@ export type TooltipInteractionState = {
    * This is the case for eventTooltipType === 'axis' for example.
    */
   dataKey: DataKey<any> | undefined;
+  /**
+   * Unique identifier of the graphical item currently driving the interaction.
+   */
+  graphicalItemId: GraphicalItemId | undefined;
```
PavelVanecek, line 156: When can this be undefined?

src/util/Constants.ts
```diff
@@ -55,4 +55,9 @@ export const DATA_ITEM_INDEX_ATTRIBUTE_NAME = 'data-recharts-item-index';
  */
 export const DATA_ITEM_DATAKEY_ATTRIBUTE_NAME = 'data-recharts-item-data-key';
 
+/**
+ * We use this attribute to identify which graphical item rendered a DOM node. Unlike dataKey it is always unique.
```
PavelVanecek, line 59: Can we get rid of dataKey as an identifier?

test/cartesian/Scatter.spec.tsx
```diff
@@ -292,6 +293,7 @@ describe('<Scatter />', () => {
           tooltipPayload: [
             {
               dataKey: 'cx',
+              graphicalItemId: expect.any(String),
```
PavelVanecek, line 296: consider `expect.stringMatching(/^recharts-scatter-.+/)`

PavelVanecek (changes requested):
Thanks @SychevAndrey for the PR. I think it's a good approach, I have couple of minor comments. I didn't actually see if they are feasible so if I was wrong please just reply why.

There's also one from the AI review, about synchronizing charts, have you tried that?

test/cartesian/Scatter.spec.tsx
SychevAndrey: I see the point. It happens many times and afaik pattern matching can be slower that just string comparison. Maybe we could use expect.stringContaining here?

PavelVanecek: How much slower? Let us see some numbers before optimising for performance. These are tests too, so speed doesn't matter.

src/context/tooltipContext.tsx
SychevAndrey: All graphical items have it, but not all the charts use graphical items.

src/state/selectors/combiners/combineTooltipPayloadConfigurations.ts
SychevAndrey: Need to check, but story works fine.

src/util/Constants.ts
SychevAndrey: I tried to to that and it was huge: dozens of files and tests affected. I almost finished it, but could not push: some tests keep failing or whatever. Moreover, for many types of charts it's just not needed, because their measures format works fine with dataKey identifier. Having said that, I'm not sure we need to do it right now.

test/cartesian/Scatter.spec.tsx
SychevAndrey: I suggest fixing that, checking default index and AI comments and consider merging this PR without moving to new id from dataKey id completely.

src/state/tooltipSlice.ts
SychevAndrey: in Line, Area charts and many others.

ckifer:
@SychevAndrey could you fix the conflicts when you get a chance?

src/context/tooltipContext.tsx
PavelVanecek: Which charts don't use graphical items? What do they use then?

src/state/tooltipSlice.ts
PavelVanecek: All Lines and all Areas have autogenerated ID. How could it be undefined?

src/util/Constants.ts
PavelVanecek: OK fair.

PavelVanecek: https://github.com/recharts/recharts/pull/6765

PavelVanecek:
Merged in https://github.com/recharts/recharts/pull/6773

Issue #6545, opened 2025-10-31
Label + LabelList textAnchor prop doesn't work


PavelVanecek:
## What is the current behavior?

`textAnchor` never does anything, it's always overwritten.

```
        label={{
          position: 'top',
          textAnchor: 'start',
```

<img width="897" height="427" alt="Image" src="https://github.com/user-attachments/assets/9bcc1cd5-2447-43f8-aaaa-2967fec01ca0" />

## What is the expected behavior?

<img width="937" height="469" alt="Image" src="https://github.com/user-attachments/assets/bb8ba30c-01d5-40eb-9024-46c092f7d247" />

## Please provide a demo of the problem in a sandbox

https://github.com/recharts/recharts/pull/6546

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Reproducible on master

PavelVanecek:
offset works fine turns out, so I edited the description

PR #6546, opened 2025-10-31
Add ZIndex guide page

PavelVanecek:
## Description

~The Prague metro example suffers from https://github.com/recharts/recharts/issues/6545 which is easy to fix but I rather do that in separate PR.~ Fixed

~Also there seems to be some problem with linux + webkit screenshot, I'll see if I can do something about it ... next PR too.~ Fixed, Safari has some differences in how it handles SVGs.

This is going to fail if we deploy the website to prod so let's merge this right before we're ready for 3.4.

## Summary by CodeRabbit

* **New Features**
  * Added public exports for z-index layer utilities.
  * Added an interactive "Z-Index and layers" guide with visual examples and a Prague Metro visualization demonstrating layering.

* **Documentation**
  * Guide added with English and Chinese translations and integrated into site navigation.

* **Tests**
  * Added a component visual regression test for the z-index visualization.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14039,9 +14039,9 @@
       }
     },
     "node_modules/recharts": {
-      "version": "3.4.0",
-      "resolved": "https://registry.npmjs.org/recharts/-/recharts-3.4.0.tgz",
-      "integrity": "sha512-rbt8pqoxVYcsFvms5AmSZJFXxWC5FwZF0VanMskmkIvrOAfU4tyXHBWMDY51aJbSJiTnHEPXBKMYwy82anR5UQ==",
+      "version": "3.4.1",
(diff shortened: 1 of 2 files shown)


PavelVanecek:
Weird build failure. I'll have a look tomorrow.

ckifer:
Started a rerun just in case it was a flake

PavelVanecek:
On local I am getting

> ✗ Build failed in 2.27s
error during build:
src/components/GuideView/ZIndex/PrahaMetro.tsx (10:2): "DefaultZIndexes" is not exported by "node_modules/recharts/es6/index.js", imported by "src/components/GuideView/ZIndex/PrahaMetro.tsx".

which I suppose is correct, can't import what one can't import. Let's rebase once 3.4 is out then?

Issue #6069, opened 2025-07-06
Fragmented documentation


PavelVanecek:
## Situation

So we have three different sources of documentation:

1. The recharts.org website
2. The storybook args
3. The Typescript type definition

And bonus! Five sources for what is the default value:

3. The same three as above
4. The component itself
5. Defaults in the reducer/slice in state

## The problem

There is nothing that forces these five places to be synchronized and so depends where you look you will find conflicting information.

## What have I tried

I tried to play around with https://storybook.js.org/docs/configure/integration/typescript but it doesn't seem to be working quite 100% - for example it doesn't set the defaults in the Controls tab (currently we use `getStoryArgsFromArgsTypesObject` for this which works fine but requires us to manually maintain the argTypes).

This docgen solution anyway will allow us to generate (2) storybook args from (3) the type but it will not solve the website and will not solve the component default props being out of sync with the type and will not solve the reducer defaults out of sync with the type.

## Alternatives?

We could have a script that checks that these all places are in sync?

Any other ideas?

PR #6615, opened 2025-11-13
Omnidoc: Correcting documented default values for Area

PavelVanecek:
## Description

Also removed `layout` prop from Area that we never read - this should have been done in 3.0

## Related Issue

https://github.com/recharts/recharts/issues/6069

## Release Notes

* **New Features**
  * Added automatic animation mode for enhanced server-side rendering compatibility.

* **Bug Fixes**
  * Fixed default value handling to preserve numeric and boolean types correctly.
  * Improved layout-aware animation rendering.

* **Documentation**
  * Updated API documentation with accurate default values and improved type definitions.
  * Enhanced animation properties documentation with new defaults.


src/cartesian/Area.tsx
```diff
@@ -678,9 +682,11 @@ export const defaultAreaProps = {
   fill: '#3182bd',
   fillOpacity: 0.6,
   hide: false,
-  isAnimationActive: !Global.isSsr,
+  isAnimationActive: 'auto',
```
ckifer, line 685: What is this? Has this always been  available? lol

PavelVanecek: I added it now. Just a serializable name for what is going on.

ckifer: Ah duh, I missed the check below this

PR #6619, opened 2025-11-13
Disable immutable check in production

### LINKED ISSUE #6617: Recharts emits console warnings that ImmutableStateInvariantMiddleware took too long when running tests on slow machines

If tests involving recharts run on slow machines, warnings like this are emitted:

```
ImmutableStateInvariantMiddleware took 50ms, which is more than the warning threshold of 32ms. 
If your state or actions are very large, you may want to disable the middleware as it might cause too much of a slowdown in development mode. See https://redux-toolkit.js.org/api/getDefaultMiddleware for instructions.
It is disabled in production builds, so you don't need to worry about that.
    at Console.E [as warn] (file:///<redacted>/node_modules/vitest-fail-on-console/src/index.ts:102:31)
    at Object.warnIfExceeded (file:///<redacted>/node_modules/@reduxjs/toolkit/src/utils.ts:16:17)
    at file:///<redacted>/node_modules/@reduxjs/toolkit/src/immutableStateInvariantMiddleware.ts:184:22
    at file:///<redacted>/node_modules/@reduxjs/toolkit/src/actionCreatorInvariantMiddleware.ts:29:12
    at /<redacted>/node_modules/recharts/lib/state/SetGraphicalItem.js:51:5
    at commitHookEffectListMount (/<redacted>/node_modules/react-dom/cjs/react-dom.development.js:23189:26)
    at commitLayoutEffectOnFiber (/<redacted>/node_modules/react-dom/cjs/react-dom.development.js:23312:15)
    at commitLayoutMountEffects_complete (/<redacted>/node_modules/react-dom/cjs/react-dom.development.js:24727:9)
    at commitLayoutEffects_begin (/<redacted>/node_modules/react-dom/cjs/react-dom.development.js:24713:7)
    at commitLayoutEffects (/<redacted>/node_modules/react-dom/cjs/react-dom.development.js:24651:3)
 ❯ Console.E [as warn] node_modules/vitest-fail-on-console/src/index.ts:102:31
 ❯ Object.warnIfExceeded node_modules/@reduxjs/toolkit/src/utils.ts:16:17
 ❯ node_modules/@reduxjs/toolkit/src/immutableStateInvariantMiddleware.ts:184:22
 ❯ node_modules/@reduxjs/toolkit/src/actionCreatorInvariantMiddleware.ts:29:12
 ❯ node_modules/recharts/lib/state/SetGraphicalItem.js:51:5
 ❯ commitHookEffectListMount node_modules/react-dom/cjs/react-dom.development.js:23189:26
 ❯ commitLayoutEffectOnFiber node_modules/react-dom/cjs/react-dom.development.js:23312:15
 ❯ commitLayoutMountEffects_complete node_modules/react-dom/cjs/react-dom.development.js:24727:9
 ❯ commitLayoutEffects_begin node_modules/react-dom/cjs/react-dom.development.js:24713:7
 ❯ commitLayoutEffects node_modules/react-dom/cjs/react-dom.development.js:24651:3
 ❯ flushUnexpectedConsoleCalls node_modules/vitest-fail-on-console/src/index.ts:73:19
 ❯ node_modules/vitest-fail-on-console/src/index.ts:151:13
```

As we are using vitest-fail-on-console to not let any warnings go unnoticed, this is actually failing the test. Unfortunately, we currently have no leverage to improve our test runner performance 😢 

No `ImmutableStateInvariantMiddleware` performance issues are reported even when tests involving recharts are run on slow machines. As we do not want to test recharts when running our test, but the application we develop, we should consider running recharts being run in production mode where this middleware should be disabled (see also https://redux-toolkit.js.org/api/getDefaultMiddleware#production). As redux-toolkit itself can not distinguish between the recharts build and our build (as it only checks `process.env.NODE_ENV === "production"` which is not true in our case), recharts should already disable the middleware (at least) in its production build.

As you already decided to disable `serializableCheck` which can also emit warnings in [src/state/store.ts#L51](https://github.com/recharts/recharts/blame/0c3ffe3976c4c9741987a763487132801480ef8f/src/state/store.ts#L51), I propose to disable `imutableCheck` as well and am happy to provide a Pull Request if you agree.

Unable to provide a demo for this, as your local machine needs to be slow to trigger the warnings.

3.3.0. I believe the issue was already present earlier, but only today found time to investigate.

tngwoerleij:
## Description

Make sure redux-tookit's `immutableCheck` is only active during testing and not in production build to avoid other modules using this module having warnings emitted by redux-toolkit during their testing.

This was harder to implement than initially thought as the babel build was still lacking to define `process.env.NODE_ENV` properly. I aligned this with the existing implementation for the webpack build.

## Related Issue

Fixes #6617

## Motivation and Context

Fix #6617

## How Has This Been Tested?

Built production artifacts and checked that `immutableCheck` is disabled. Ran tests and checked that `immutableCheck` is enabled. 

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Modernized build configuration to inline environment variables and refine transpilation behavior.
  * Added tooling to support the updated build setup.

* **Tests**
  * Enabled stricter state immutability checks in non-production builds to catch issues earlier.

### DIFF
--- a/src/state/store.ts
+++ b/src/state/store.ts
@@ -49,7 +49,7 @@ export const createRechartsStore = (
     middleware: getDefaultMiddleware =>
       getDefaultMiddleware({
         serializableCheck: false,
-        immutableCheck: process.env.NODE_ENV === 'test',
+        immutableCheck: !['commonjs', 'es6', 'production'].includes(process.env.NODE_ENV ?? ''),
       }).concat([
         mouseClickMiddleware.middleware,
         mouseMoveMiddleware.middleware,


PavelVanecek:
Do we run babel? I thought that's for umd build only.

PavelVanecek:
Nvm it's the opposite. Babel everywhere except umd.

src/state/store.ts
```diff
@@ -49,6 +49,7 @@ export const createRechartsStore = (
     middleware: getDefaultMiddleware =>
       getDefaultMiddleware({
         serializableCheck: false,
+        immutableCheck: process.env.NODE_ENV === 'test',
```
ckifer: Can we do !== prod instead? Pretty sure this doesn't get set to test when running things locally (besides tests)

tngwoerleij: I also thought about this in the first place, but the build scripts for CommonJS and ES6 builds use `NODE_ENV` "commonjs" and "es6" respectively which would lead to the `immutableCheck` not being disabled for them. At least vitest sets "test" as `NODE_ENV` which is why I chose it here. To be correct, we would have to change it to 

```js
        immutableCheck: ['commonjs', 'es6', 'production'].includes(process.env.NODE_ENV ?? ''),
```

to my understanding. Do you agree?

ckifer: thats fine but I think you would need to negate that so that the check would be off for those values

tngwoerleij: Certainly! My bad 😞 I will push a new commit shortly

Issue #6624, opened 2025-11-13
Performance issue after upgrading to V3


frec-hansen:
## What is the current behavior?

After making the upgrade to V3. The chart is experiencing lagginess which was not a problem in V2.

## What is the expected behavior?

Chart as smooth as possible

## Please provide a demo of the problem in a sandbox

I created the same code with V3 and V2:
V3: https://codesandbox.io/p/sandbox/recharts-3-performance-dhjq87
V2: https://codesandbox.io/p/devbox/recharts-3-performance-forked-zx7nn8?workspaceId=ws_4arwYw1k37L3CUe1S5xsnR

Hovering on V3, you can see that the chart keeps getting re-rendered while it's not the case in V2. In our production app, there are way more data and the lagginess is quite obvious.

NOTE: I am not sure why the gray line is not showing up in V2 demo above but it does show up in our prod code.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Recharts: V3.4.1 vs V2.15.3

ckifer:
I don't have permissions to view the v2 sandbox.

----

Yes there are more re-renders in v3 - we will try to improve performance over time but for the most part the re-renders are from react doing what it is supposed to

frec-hansen:
>I don't have permissions to view the v2 sandbox.

Sorry, updated!

PavelVanecek:
Thanks @frec-hansen . When does the chart rerender? Or mouse hover?

PavelVanecek:
@frec-hansen Your v2 and v3 examples are very very different charts. The v3 renders ton more components. Of course it will be slower.

frec-hansen:
> Thanks [@frec-hansen](https://github.com/frec-hansen) . When does the chart rerender? Or mouse hover?

Yes, on mouse over.

> @frec-hansen Your v2 and v3 examples are very very different charts. The v3 renders ton more components. Of course it will be slower.

@PavelVanecek I don't know why v2 doesn't render some of the items in codesandbox but both uses the exact same code. And for me, they render the exact same things in prod only difference is the performance.

ckifer:
Going to close this for now with the release of 3.5! @frec-hansen please try 3.5 as well as the performance guide
https://github.com/recharts/recharts/releases/tag/v3.5.0

PR #6654, opened 2025-11-18
Some more performance improvements

PavelVanecek:
## Description

In https://github.com/recharts/recharts/issues/6624 and other issues I noticed that people tend to define things inline - which is no surprise because our docs and examples recommend that too.

We can memoize the object props with a small custom method so I did that, looks like it works well.

We can't do much with inline functions and inline elements, memoization won't work there. v2 just plain ignores them which I don't think we want to do anymore.

So I found this react perf plugin which flags exactly what we need so I added it too and ran against this one example.

Plus some supporting changes in the website navigation list, and allowing some syntax in ESLint config.

I will later summarize it in a perf guide, but that will be in next PR.

## Related Issue

https://github.com/recharts/recharts/issues/6624

## How Has This Been Tested?

Devtools

## Before & after video (with 6x slowdown)

https://github.com/user-attachments/assets/bfa0e1c8-2e92-4fc3-ad0e-37e9a190f400

## One render cycle during mouse move

Before - notice the high animation count - it was re-calculating the animation on each mouse move. Also, multiple components render again and again.

Screenshot 2025-11-18 at 17 36 05

After - notice low animation count, and most components never render. Other than Tooltip and ReferenceLine which have to render because they change position on mouse move.

Screenshot 2025-11-18 at 23 33 03

## Types of changes

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -272,7 +272,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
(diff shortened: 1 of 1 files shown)


PavelVanecek:
@ckifer I can break this down to smaller PRs if you prefer. There are couple of changes that could go out on its own.

ckifer:
easy enough review, all good. LGTM

PR #6660, opened 2025-11-19
fix: ErrorBar duplicate key warning when error range has same values 

### LINKED ISSUE #6625: ErrorBar with range array with same value - Encountered two children with the same key, `errorbar-592-596-165-165`.

When adding errorBar to the graph, I see on the console message above

Image

This seems to only happen when I have the errorBar with a range where the values are the same (ie: [0,0])
In our case, we calculate based on min/max, will need to rework the calculation for the cases when difference is symetric use a scalar

No warnings are added to the console

https://codesandbox.io/p/devbox/simple-line-chart-forked-ydpnqt?file=%2Fsrc%2FApp.tsx%3A67%2C1&workspaceId=ws_R2qrYKC3GHo6P45DgCdWwU

Open Developer tools. Message will show up

Tested with 2.15.0, 2.15.0 and 3.4.1. I get the message in all of them.

Not sure if duplicate of https://github.com/recharts/recharts/issues/6049 (that one is mentioning ticks, seems different)

bas0N:
…(#6625)

Fix React duplicate key warnings in ErrorBar when error ranges have equal values by using index-based keys instead of coordinate-based keys. Follows the Scatter.tsx pattern of including descriptive data with index at the end for uniqueness.

Fixes #6625

## Description

This PR fixes React duplicate key warnings that occur when ErrorBar components have error ranges with equal values (e.g., `[0, 0]` or `[878, 878]`). 

**Root Cause:**
Previously, React keys were generated from line coordinates (`x1`, `x2`, `y1`, `y2`). When `lowBound === highBound` (equal error range values), multiple error bar lines would have identical coordinates, resulting in duplicate keys like `errorbar-592-596-165-165`.

**Solution:**
Changed key generation to follow the pattern used in `Scatter.tsx`:
- Include descriptive coordinates/data for debugging purposes
- Always end with index (`dataIndex` and `lineIndex`) to ensure uniqueness
- Updated both the outer `Layer` key and inner `CSSTransitionAnimate` key
- Maintained `animationId` format with coordinates for proper animation tracking

**Changes:**
- `Layer` key: Changed from coordinate-based to `bar-${x}-${y}-${value}-${dataIndex}`
- `CSSTransitionAnimate` key: Changed from coordinate-based to `errorbar-${dataIndex}-${c.x1}-${c.y1}-${c.x2}-${c.y2}-${lineIndex}`
- `animationId`: Updated to include indexes at the end: `error-bar-${direction}_${c.x1}-${c.x2}-${c.y1}-${c.y2}-${dataIndex}-${lineIndex}`

## Related Issue

Fixes #6625

## Motivation and Context

When ErrorBar components have error ranges with equal values (e.g., `[0, 0]` or `[878, 878]`), the error bar lines can have identical coordinates. The previous coordinate-based key generation (`errorbar-${c.x1}-${c.x2}-${c.y1}-${c.y2}`) would produce duplicate keys, causing React to log warnings:

Encountered two children with the same key, errorbar-592-596-165-165
This is a common scenario when error calculations result in symmetric ranges or when min/max values are the same. The fix ensures unique keys regardless of coordinate values, following the established pattern in other components like `Scatter.tsx`.

## How Has This Been Tested?

1. **Unit Test Added**: Added a new test case `should not generate duplicate keys when error range has same values` in `test/cartesian/ErrorBar.spec.tsx` that:
   - Uses data with equal error range values (`[0, 0]` and `[878, 878]`)
   - Verifies that all error bars render correctly (4 data points × 3 lines = 12 lines)
   - Confirms no duplicate keys are generated

2. **Existing Tests**: All existing ErrorBar tests pass, including:
   - Animation tests (with updated `animationId` format)
   - Rendering tests for Bar, Line, and Scatter charts
   - Multiple ErrorBar tests

3. **Manual Testing**: Verified the fix works with the example from the issue:
   - No console warnings when using error ranges with equal values
   - Error bars render correctly
   - Animations work as expected

N/A - This is a bug fix that eliminates console warnings without changing visual behavior.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error bar animation and rendering consistency when displaying multiple error bars across data points.

### DIFF
--- a/src/cartesian/ErrorBar.tsx
+++ b/src/cartesian/ErrorBar.tsx
@@ -168,7 +168,7 @@ function ErrorBarImpl(props: ErrorBarInternalProps) {
           const lineStyle = isAnimationActive ? { transformOrigin } : undefined;
           return (
             <CSSTransitionAnimate
-              animationId={`error-bar-${direction}_${c.x1}-${c.x2}-${c.y1}-${c.y2}-${dataIndex}-${lineIndex}`}
+              animationId={`error-bar-${direction}_${c.x1}-${c.x2}-${c.y1}-${c.y2}`}
               from={`${scaleDirection}(0)`}
               to={`${scaleDirection}(1)`}
               attributeName="transform"
--- a/test/cartesian/ErrorBar.spec.tsx
+++ b/test/cartesian/ErrorBar.spec.tsx
@@ -2094,46 +2094,4 @@ describe('<ErrorBar />', () => {
       });
     });
   });
-
-  test('should not generate duplicate keys when error range has same values', () => {
-    const dataWithEqualRange = [
-      {
-        timestamp: '2025-10-16T00:00:00.000Z',
-        avgLatency: 0,
-        rangeLatency: [0, 0],
-      },
-      {
-        timestamp: '2025-10-17T00:00:00.000Z',
-        avgLatency: 1292,
-        rangeLatency: [878, 878],
-      },
-      {
-        timestamp: '2025-10-22T00:00:00.000Z',
-        avgLatency: 461,
-        rangeLatency: [0, 0],
-      },
-      {
-        timestamp: '2025-10-23T00:00:00.000Z',
-        avgLatency: 853.8333333333333,
-        rangeLatency: [639.8333333333333, 1104.1666666666667],
-      },
-    ];
-
-    const { container } = rechartsTestRender(
-      <LineChart data={dataWithEqualRange} width={500} height={500}>
-        <Line isAnimationActive={false} dataKey="avgLatency">
-          <ErrorBar dataKey="rangeLatency" />
-        </Line>
-      </LineChart>,
-    );
-
-    // Check that error bars are rendered (4 data points * 3 lines each = 12 lines)
-    const errorBarLines = container.querySelectorAll('.recharts-errorBar line');
-    expect(errorBarLines.length).toBe(12);
-
-    // Verify that all error bars have unique keys by checking that all lines are rendered
-    // When keys are duplicated, React may omit some elements, so checking the count ensures uniqueness
-    const errorBarLayers = container.querySelectorAll('.recharts-errorBar');
-    expect(errorBarLayers.length).toBe(4);
-  });
 });


src/cartesian/ErrorBar.tsx
```diff
@@ -163,24 +163,20 @@ function ErrorBarImpl(props: ErrorBarInternalProps) {
...
-        {...svgProps}
-      >
-        {lineCoordinates.map(c => {
+      <Layer className="recharts-errorBar" key={`bar-${x}-${y}-${value}-${dataIndex}`} {...svgProps}>
+        {lineCoordinates.map((c, lineIndex) => {
           const lineStyle = isAnimationActive ? { transformOrigin } : undefined;
           return (
             <CSSTransitionAnimate
-              animationId={`error-bar-${direction}_${c.x1}-${c.x2}-${c.y1}-${c.y2}`}
+              animationId={`error-bar-${direction}_${c.x1}-${c.x2}-${c.y1}-${c.y2}-${dataIndex}-${lineIndex}`}
```
ckifer: If we're just fixing the key error can we omit this change?

test/cartesian/ErrorBar.spec.tsx
```diff
@@ -2094,4 +2094,46 @@ describe('<ErrorBar />', () => {
...
+    );
+
+    // Check that error bars are rendered (4 data points * 3 lines each = 12 lines)
+    const errorBarLines = container.querySelectorAll('.recharts-errorBar line');
+    expect(errorBarLines.length).toBe(12);
+
+    // Verify that all error bars have unique keys by checking that all lines are rendered
+    // When keys are duplicated, React may omit some elements, so checking the count ensures uniqueness
+    const errorBarLayers = container.querySelectorAll('.recharts-errorBar');
+    expect(errorBarLayers.length).toBe(4);
```
ckifer: This test isn't asserting anything useful. React usually won't omit elements when there are duplicate keys - there generally isn't a great way to test this that we've been able to find

PavelVanecek: Duplicate key will print console warning right? We have a test for those.

ckifer: I've never been able to get it to work for key errors. They might not happen in a test env

src/cartesian/ErrorBar.tsx
bas0N: Reverted the animationId change - only the key fix remains

test/cartesian/ErrorBar.spec.tsx
bas0N: @ckifer Fair point, removed the test

ckifer (approved):

ckifer:
thanks @bas0N

PR #6678, opened 2025-11-23
fix(Pie): prevent cross-highlighting in multi-Pie charts with shared dataKey

### LINKED ISSUE #6259: TwoLevelPieChart with activeShape does not work as expected

shreedharbhat98:
## Description

Fixed issue where multiple Pie components with the same `dataKey` would incorrectly highlight together when hovering over one of them. The fix implements graphical item ID tracking in the tooltip state infrastructure, preparing for component-specific matching to ensure each Pie chart maintains independent active states.

## Related Issue

Fixes #6259

## Motivation and Context

When rendering multiple Pie charts (e.g., concentric pies) with the same `dataKey` prop, hovering over a sector in one Pie would incorrectly activate the corresponding sector in all other Pies with the same dataKey. This happened because the active sector selection logic only matched by index and dataKey, without considering which specific graphical item (Pie component) the interaction occurred on.

This change adds the infrastructure required to allow proper independent interaction with multiple Pie charts in the same visualization, which is a common use case for two-level or nested pie charts.

## How Has This Been Tested?
1. **Type Safety**: All TypeScript type checks passing across main source, tests, storybook, and website
2. **Unit Tests**: Updated 11 test files (34+ locations) to include the new `graphicalItemId` field in tooltip state expectations
3. **Test Suite**: All 7360 tests passing, including:
   - `Pie.spec.tsx` - Comprehensive Pie component tests including tooltip interactions
   - `Tooltip.sync.spec.tsx` - Tooltip synchronization tests
   - `Tooltip.visibility.spec.tsx` - Tooltip visibility state tests
   - `useChartSynchronisation.spec.tsx` - Chart synchronization hook tests
   - All selector and state management tests
4. **Live Demo**: https://stackblitz.com/edit/react-lnlglkfj-29bjqzhg?file=src%2Findex.js

**Testing approach:**
- Verified type safety of tooltip state with new `graphicalItemId` field
- Ensured backward compatibility - existing code continues to work with `graphicalItemId: undefined`
- Validated that all tooltip state objects properly include the new field
- Confirmed no breaking changes to existing APIs|

Live demo available at: https://stackblitz.com/edit/react-lnlglkfj-29bjqzhg?file=src%2Findex.js

| Before | After |
|--------|--------|
|  pie | Screenshot 2025-11-23 at 8 11 06 PM |

## Types of changes

## Checklist:

**Technical Implementation Details:**
- Added optional `dataKey` field to `PieSectorDataItem` type (non-breaking - defaults to `'value'`)
- Implemented coordinate-based matching in `PieSectors.tsx` comparing `activeCoordinate` with `sector.tooltipPosition`
- Uses 0.01 pixel tolerance to handle floating-point precision from trigonometric calculations
- Updated selector tests to include `dataKey` in sector expectations

## Summary by CodeRabbit

* **New Features**
  * Tooltip now tracks per-visual-item IDs, enabling distinct hover/click identification for overlapping or multi-layer charts.

* **Bug Fixes**
  * Corrected active/hover behavior so only the intended sector shows as active when multiple pies share data keys.

* **Tests**
  * Added and updated tests for two-level/multi-pie active-shape rendering and tooltip synchronization.

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -35,8 +35,8 @@ import {
 import { TooltipPayload, TooltipPayloadConfiguration } from '../state/tooltipSlice';
 import { SetTooltipEntrySettings } from '../state/SetTooltipEntrySettings';
 import {
-  selectActiveTooltipCoordinate,
   selectActiveTooltipDataKey,
+  selectActiveTooltipGraphicalItemId,
   selectActiveTooltipIndex,
 } from '../state/selectors/tooltipSelectors';
 import { SetPolarLegendPayload } from '../state/SetLegendPayload';
@@ -526,11 +526,11 @@ function PieLabelList({
 }
 
 function PieSectors(props: PieSectorsProps) {
-  const { sectors, activeShape, inactiveShape: inactiveShapeProp, allOtherPieProps, shape } = props;
+  const { sectors, activeShape, inactiveShape: inactiveShapeProp, allOtherPieProps, shape, id } = props;
 
   const activeIndex = useAppSelector(selectActiveTooltipIndex);
   const activeDataKey = useAppSelector(selectActiveTooltipDataKey);
-  const activeCoordinate = useAppSelector(selectActiveTooltipCoordinate);
+  const activeGraphicalItemId = useAppSelector(selectActiveTooltipGraphicalItemId);
   const {
     onMouseEnter: onMouseEnterFromProps,
     onClick: onItemClickFromProps,
@@ -538,9 +538,9 @@ function PieSectors(props: PieSectorsProps) {
     ...restOfAllOtherProps
   } = allOtherPieProps;
 
-  const onMouseEnterFromContext = useMouseEnterItemDispatch(onMouseEnterFromProps, allOtherPieProps.dataKey);
+  const onMouseEnterFromContext = useMouseEnterItemDispatch(onMouseEnterFromProps, allOtherPieProps.dataKey, id);
   const onMouseLeaveFromContext = useMouseLeaveItemDispatch(onMouseLeaveFromProps);
-  const onClickFromContext = useMouseClickItemDispatch(onItemClickFromProps, allOtherPieProps.dataKey);
+  const onClickFromContext = useMouseClickItemDispatch(onItemClickFromProps, allOtherPieProps.dataKey, id);
 
   if (sectors == null || sectors.length === 0) {
     return null;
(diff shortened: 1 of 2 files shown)


shreedharbhat98:
Hi @ckifer I have tried to fix the [issue](https://github.com/recharts/recharts/issues/6259) here. Feel free to take a look when you are available. 
Thanks a bunch in advance!

PavelVanecek (approved):

src/polar/Pie.tsx
```diff
@@ -506,7 +513,19 @@ function PieSectors(props: PieSectorsProps) {
       {sectors.map((entry, i) => {
         if (entry?.startAngle === 0 && entry?.endAngle === 0 && sectors.length !== 1) return null;
 
-        const isActive = String(i) === activeIndex;
+        // For Pie charts, when multiple Pies share the same dataKey, we need to ensure only the hovered Pie's sector is active.
+        // We do this by checking if the active tooltip coordinate matches this sector's tooltip position.
+        // This works because each sector has a unique tooltip position based on its midAngle and middleRadius.
+        const coordinateMatches =
```
ckifer: This will lead to issues down the road if we ever want to support Pie tooltips that are not a fixed position. I wonder if there is a way to use graphicalItemId here to identify what pie is being hovered. Can we add graphicalItemId to the active item state?

ckifer (changes requested):
This might be a good interim but I don't really think this works long term - can you give a different approach a try and if it doesn't work we can merge this for now?

src/polar/Pie.tsx
PavelVanecek: +1 let's have this based on the graphical item ID.

shreedharbhat98: @ckifer I have made the necessary changes and upadted stackblitz link as well. Please do take a look
Thanks!

PR #6696, opened 2025-11-25
Fix cursor type definition in ExternalMouseEvents

Goldyvaiiii:
This PR adds proper TypeScript support for the cursor prop in chart components.
Previously, missing type definitions caused errors when passing a cursor value to components like BarChart.
Added a CursorProps interface, allowing booleans, React elements, CSS cursor values, or custom stroke options,
Fixes the long-standing type error reported by multiple users.

The cursor prop works at runtime, but TypeScript users faced type errors.
This ensures developers can pass cursor styles without hacks.

Testing: Verified the project builds successfully and tested with different valid cursor values.
No other chart components were affected.

## Summary by CodeRabbit

* **New Features**
  * Added cursor customization: enable/disable cursor, supply a custom cursor element, and configure stroke and stroke-dash styling.

* **Chores**
  * Updated React type definitions to newer versions.
  * Adjusted development dependencies ordering to align linting plugins and testing tools.


ckifer:
Can you give an example of when this was an issue and how this fixes it?

src/chart/types.ts
```diff
@@ -18,3 +21,14 @@ export interface ExternalMouseEvents {
   onTouchMove: CategoricalChartFunc;
   onTouchEnd: CategoricalChartFunc;
 }
+
+export interface CursorProps {
+  cursor?:
+    | boolean
+    | React.ReactElement
+    | React.CSSProperties['cursor']
```
ckifer: `CSSProperties['cursor']` has nothing to do with recharts cursor

ckifer (changes requested):

PavelVanecek:
Why are there changes to package.json? That doesn't look related. Can we undo those?

ckifer:
> Fixes the long-standing type error reported by multiple users.

Also can you please link these issues?

ckifer:
I'm going to close this because I'm still confused about it - please comment back with issue links or an explanation 🙏🏼

PR #6719, opened 2025-12-02
[Storybook] Allow defining defaultValue as object with 'summary' property

PavelVanecek:
## Description

Storybook recommends to declare default values as an object with `summary` prop (and marks all other definitions as deprecated) but our own `getStoryArgsFromArgsTypesObject` did not parse that variant. I suspect this is the cause for some of the recent storybook diff.

## Related Issue

https://www.chromatic.com/build?appId=63da8268a0da9970db6992aa&number=3277

## Summary by CodeRabbit

* **Refactor**
  * Updated argument type handling across story utilities for improved consistency.
  * Enhanced default value processing to support additional configuration patterns.
  * Simplified story configuration management by refining how arguments are defined and processed.


PavelVanecek:
There's a small improvement but we will need one more PR to fix the remaining stories.

Issue #6645, opened 2025-11-17
Typescript strict mode


PavelVanecek:
Enable the most strictestest mode available in typescript and fix all violations.

https://www.typescriptlang.org/tsconfig/#strict

Issue #5262, opened 2024-11-19
Add ability to draw individual `ReferenceLine`s for category axis in a `BarChart`


jeisele-frontier:
## Do you want to request a feature or report a bug?

Feature request

## What is the current behavior?

Currently, separate reference lines can be drawn accurately on a `BarChart` only if the axis it is parallel to is a number axis, as shown in [this codesandbox](https://codesandbox.io/p/sandbox/recharts-multiple-disjointed-reference-lines-huelk?file=%2Findex.js%3A53%2C40&from-embed) (from [this article](https://gaurav5430.medium.com/exploring-recharts-multiple-referenceline-segments-1856b5b07111)). However, as mentioned in the comment in the codesandbox, this will not work if the axis parallel to the reference line is a category axis. This is because:
- it draws from the center of one tick to the next if you use categories for the `segment`, instead of from the beginning of the bar to the end of the bar, resulting in an offset reference line like this. The goal is to have the reference line across only the "Floating Rate Securities" bar.
![image](https://github.com/user-attachments/assets/fa143894-130e-45c7-8e83-397e3e171bae)
- as mentioned in the comment in the codesandbox, even if there were some way to offset the line to be centered correctly, there is no extra category to define where the last `ReferenceLine` `segment` should end

## What is the expected behavior?
It would be nice to be able to put a `ReferenceLine` across each bar, maybe by making reference line a child component of `Bar`, or maybe by adding the ability to provide a prop to `RefererenceLine` like `segment={{ x: number, y: 'someExistingCategory' }}`, where the line is automatically centered on the category axis tick and is the same width as the bar (or slightly larger), and at the value specified by `y`.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?
Using recharts 2.12.7

ckifer:
Updated very outdated codesandbox to the latest recharts https://codesandbox.io/p/sandbox/recharts-multiple-disjointed-reference-lines-huelk?from-embed=

ckifer:
I'm not sure I understand the ask

You want to be able to make another Bar inside a Bar? With a ReferenceLine?

jeisele-frontier:
I'd like to be able to put a reference line like the one shown in the codesandbox, but when one of the axes is a category axis instead of a number axis. The closest I'm able to achieve is shown in the screenshot above, where the reference line spans two bars, since I'm passing ```<ReferenceLine segment={[
[...]

PR #6783, opened 2025-12-16
Split scale definition into two parts

PavelVanecek:
## Description

I like how ScaleHelper (https://github.com/recharts/recharts/blob/b9835fe358eda9637e939217216a4c894c3b909a/src/util/CartesianUtils.ts#L18) encapsulates scales and adds new functionality on top, so let's have the same for all scales!

Because we accept the function from the outside as props, we can't just add new methods and new arguments to it. So step one is to split the scale function type into two: the internal one, and external one. External one stays stable, internal one will have new methods.

So I did that and fixed all violations - there were again quite a few `any` hiding undefineds. You can see on the screenshot that we got rid of mysterious `y: 0` error bars too.

## Related Issue

Having new methods on the internal scale will also allow fix for https://github.com/recharts/recharts/issues/5262 

## Motivation and Context

No `any`

## Release Notes

* **New Features**
  * Added support for custom scale definitions, enabling more flexible scale configurations across charts.

* **Bug Fixes**
  * Improved null-safety handling for scale calculations in Area, Bar, ErrorBar, Line, and Scatter charts.
  * Enhanced scale validation with null coalescing to prevent undefined coordinate values.
  * Fixed radius calculations in scatter plots when size values are missing.

* **Chores**
  * Updated scale utility organization and type definitions for better maintainability.


src/util/scale/CustomScaleDefinition.ts
PavelVanecek, line 1: @ckifer this is an important one to review - the external interface will be locked for major versions

src/util/scale/RechartsScale.ts
PavelVanecek, line 1: @ckifer this will be internal only, and we will be free to change it. First step will be to add the `apply` method from ScaleHelper

test/util/ChartUtils.spec.tsx
```diff
@@ -12,52 +12,18 @@ import {
 import { AxisType, DataKey, LayoutType, TickItem } from '../../src/util/types';
 import { BaseAxisWithScale } from '../../src/state/selectors/axisSelectors';
 import { calculateActiveTickIndex } from '../../src/util/getActiveCoordinate';
+import { defaultAxisId } from '../../src/state/cartesianAxisSlice';
+import { CustomScaleDefinition } from '../../src';
 
 describe('getTicksForAxis', () => {
   const Y_AXIS_EXAMPLE: AxisPropsNeededForTicksGenerator = {
-    // @ts-expect-error we need to wrap the d3 scales in unified interface
```
PavelVanecek, line 18: We did wrap d3 scale in unified interface now!

test/util/ChartUtils.spec.tsx
```diff
@@ -12,52 +12,18 @@ import {
 import { AxisType, DataKey, LayoutType, TickItem } from '../../src/util/types';
 import { BaseAxisWithScale } from '../../src/state/selectors/axisSelectors';
 import { calculateActiveTickIndex } from '../../src/util/getActiveCoordinate';
+import { defaultAxisId } from '../../src/state/cartesianAxisSlice';
+import { CustomScaleDefinition } from '../../src';
```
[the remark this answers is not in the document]
PavelVanecek: Unit tests (folder `test`) can import internals too.

src/util/scale/CustomScaleDefinition.ts
```diff
@@ -0,0 +1,63 @@
...
+   * Sets the range of the scale.
+   */
+  range(newRange: ReadonlyArray<number>): this;
+
+  /**
+   * Returns an exact copy of this scale. Changes to this scale will not affect the returned scale, and vice versa.
+   * We use this method in Recharts to ensure immutability - changes to the scale outside should not affect the internal state,
+   * as they won't trigger re-renders in React.
+   */
+  copy(): CustomScaleDefinition<Domain>;
```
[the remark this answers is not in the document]
PavelVanecek: Does it return `this` though? If it returns a new independent instance?

src/state/selectors/axisSelectors.ts
```diff
@@ -1854,6 +1826,14 @@ export const selectAxisPropsNeededForCartesianGridTicksGenerator: (
   },
 );
 
+/**
+ * Of on four almost identical implementations of tick generation.
```
ckifer, line 1830: nit: one of?

src/state/selectors/axisSelectors.ts
```diff
@@ -1854,6 +1826,14 @@ export const selectAxisPropsNeededForCartesianGridTicksGenerator: (
...
+ * The four horsemen of tick generation are:
```
ckifer, line 1831: lmaooo

src/util/scale/CustomScaleDefinition.ts
ckifer: does this break if someone uses a method from a d3 scale prior to passing it in and then pass it in? 

It shouldn't

Something that I've done in the past is used https://d3js.org/d3-scale/linear#linear_invert

In this case I would define my scale with a given domain/range and then because recharts is using that same scale use it to convert X and Y values to my scale. This should still work, cool

ckifer: and its always easier to add more later

ckifer: what happens if D3 breaks us here?

PavelVanecek: > does this break if someone uses a method from a d3 scale prior to passing it in and then pass it in?

They can do whatever with the scale, this interface is the list of methods we require inside Recharts. It doesn't even need to be imported from d3, as long as it implements this interface.

> and its always easier to add more later

That's the thing, adding more methods here will be a breaking change. If someone does have a custom implementation (so, skips d3) then having a new required method will break their code.

This is in contrast to the internal interface where we can add new stuff anytime. We can return it from a hook and make it external that way - and because there it's in a consumer position, adding new methods in that context will not break anyone.

> what happens if D3 breaks us here?

Then that triggers a major version for Recharts ¯\\\_(ツ)\_/¯

PavelVanecek: And yes the `invert` method will be one of the things we export from a hook. I would prefer to keep it simple however, so perhaps we will export only a single method (`invert`) instead of the whole scale.

PR #6828, opened 2025-12-26
docs: dark mode

cloud-walker:
## Description

Add `ColorModePicker` module (behind a feature flag for now), that aims to let the end user control the current color-mode of the docs.

## Motivation and Context

Follow up of #6825 

TBD

## Checklist:

- [x] Add `ColorModePicker` module
- [ ] Style ColorModePicker to better integrate it with the site
- [x] Handle system color mode preferences change
- [ ] Add ColorModeStrategy icons?
- [ ] Introduce a better token naming system
- [ ] Add dark and light color themes
- [ ] Review the main site areas
- [ ] Review codemirror styles
- [ ] Review all chart styles
- [ ] Review storybook
- [ ] Review stackblitz

## Summary by CodeRabbit

* **New Features**
  * Added a color mode system: provider, picker, watcher and store to toggle system/light/dark; preference persists and updates the page theme when enabled.
  * Color mode picker added to desktop and conditional mobile navigation.

* **Tests**
  * Added comprehensive color mode tests and updated navigation tests; removed an older navigation test file.

* **Chores**
  * Added ignore rule for local environment files (*.local).

### DIFF
--- a/www/test/color-mode.spec.tsx
+++ b/www/test/color-mode.spec.tsx
@@ -190,6 +190,8 @@ test('system color scheme change', () => {
     origin: 'system',
     mode: 'dark',
   });
+
+  store.dispose();
 });
 
 test('listener', () => {
@@ -216,8 +218,9 @@ test('listener', () => {
 
 test('ColorModePicker', async () => {
   setupEnvironment({ preferredColorMode: 'light' });
+  const store = defineColorModeStore();
   render(
-    <ColorModeProvider store={defineColorModeStore()}>
+    <ColorModeProvider store={store}>
       <ColorModePicker />
     </ColorModeProvider>,
   );
@@ -245,6 +248,8 @@ test('ColorModePicker', async () => {
   expect(document.documentElement).toHaveAttribute('data-mode', 'light');
   expect(localStorage.getItem(STORAGE_KEY)).toBeNull();
   expect(colorModePicker).toHaveTextContent('system');
+
+  store.dispose();
 });
 
 test('ColorModeWatcher', () => {
@@ -275,4 +280,6 @@ test('ColorModeWatcher', () => {
     store.dispatch('system');
   });
   expect(screen.getByRole('heading', { name: 'origin: system; mode: light' })).toBeInTheDocument();
+
+  store.dispose();
 });
--- a/www/test/navigation.spec.tsx
+++ b/www/test/navigation.spec.tsx
@@ -49,11 +49,15 @@ describe('ColorModePicker', () => {
   it('should render on both mobile and desktop navs', async () => {
     vi.stubEnv('VITE_IS_DARKMODE_ENABLED', 'true');
     vi.stubGlobal('matchMedia', (query: string) => ({
-      maches: query.includes('dark'),
+      media: query,
+      matches: query.includes('dark'),
       addEventListener() {},
+      removeEventListener() {},
+      dispatchEvent() {},
     }));
+    const store = defineColorModeStore();
     render(
-      <ColorModeProvider store={defineColorModeStore()}>
+      <ColorModeProvider store={store}>
         <MemoryRouter>
           <Navigation />
         </MemoryRouter>
@@ -71,5 +75,6 @@ describe('ColorModePicker', () => {
       expect(picker).toHaveTextContent('light');
     });
     expect(document.documentElement).toHaveAttribute('data-mode', 'light');
+    store.dispose();
   });
 });


www/src/components/ColorModePicker.tsx
```diff
@@ -0,0 +1,68 @@
+import { useState } from 'react';
+
+const storageKey = 'recharts-color-mode';
```
ckifer: nit nit: since this is a constant would be nice to make it upper case `STORAGE_KEY`

www/src/components/ColorModePicker.tsx
```diff
@@ -0,0 +1,68 @@
...
+  if (rawStoredValue == null) {
+    return 'system';
+  }
+  if (rawStoredValue !== 'dark' && rawStoredValue !== 'light') {
+    return 'system';
+  }
+  return rawStoredValue;
+}
+function setStoredColorModeStrategy(strategy: ColorModeStrategy) {
+  localStorage.setItem(storageKey, strategy);
```
ckifer: nit: you probably don't need separate functions for these 1 liners

www/src/components/ColorModePicker.tsx
```diff
@@ -0,0 +1,68 @@
...
+    document.documentElement.setAttribute('data-mode', strategy);
+  }
+}
+
+function setInitialColorMode() {
+  const strategy = getStoredColorModeStrategy();
+  setColorMode(strategy);
+}
+
+export const ColorModePicker = Object.assign(
```
ckifer: why are we using Object.assign? Just export each function/component and use it as is?

www/src/components/ColorModePicker.tsx
cloud-walker: this function is there only to avoid exposing the STORAGE_KEY in the color-mode.spec.tsx, keeping it private, I agree that its a bit of an overkill tho

www/src/components/ColorModePicker.tsx
cloud-walker: it's a pattern that I use lately to aggregate stuff (for more complex scenarios also, like compound components: Dialog.Trigger, Dialog.Content, etc.), but I can change that

www/src/components/ColorModePicker.tsx
```diff
@@ -0,0 +1,68 @@
+import { useState } from 'react';
+
+const storageKey = 'recharts-color-mode';
+
+type ColorModeStrategy = 'light' | 'dark' | 'system';
+
+function getStoredColorModeStrategy(): ColorModeStrategy {
+  const rawStoredValue = localStorage.getItem(storageKey);
```
PavelVanecek: Do we have try/catch around localStorage? Some environments will throw.

www/src/components/color-mode/defineColorModeStore.ts
```diff
@@ -0,0 +1,103 @@
...
+    console.warn('Failed to set color mode in localStorage, skipping.');
+  }
+}
+function clearStoredColorMode() {
+  try {
+    localStorage.removeItem(STORAGE_KEY);
+  } catch {
+    /* istanbul ignore next */
+    // eslint-disable-next-line no-console
+    console.warn('Failed to clear color mode in localStorage, skipping.');
```
cloud-walker, line 35: dunno what is the practise here about logging

www/src/components/color-mode/defineColorModeStore.ts
```diff
@@ -0,0 +1,103 @@
...
+    return undefined;
+  }
+}
+function setStoredColorMode(mode: ColorMode) {
+  try {
+    localStorage.setItem(STORAGE_KEY, mode);
+  } catch {
+    /* istanbul ignore next */
+    // eslint-disable-next-line no-console
+    console.warn('Failed to set color mode in localStorage, skipping.');
```
cloud-walker, line 27: dunno what is the practise here about logging

www/.gitignore
```diff
@@ -7,3 +7,5 @@ public/examples
 /docs/
 /.idea/
 /coverage
+
+*.local
```
cloud-walker, line 11: this is to be able to test the feature locally, by creating the .env.local file (supported by vite)

www/src/components/ColorModePicker.tsx
cloud-walker: addressed in the new defineColoreModeStore module

www/.gitignore
PavelVanecek: We have anything in an .env file?

www/test/color-mode.spec.tsx
cloud-walker, line 1: I don't like too much to have the test far from the module, but I tried to follow up what I saw in the codebase here

www/test/navigation.spec.tsx
cloud-walker, line 1: dunno if this is the right module, I putted a bunch of smoke tests to check the color mode picker presence

www/src/components/color-mode/ColorModeProvider.tsx
cloud-walker, line 1: I tried to avoid the provider, but the presence of the store full of side effects needs a single point to define it

www/.gitignore
cloud-walker: I saw no .env file present in the codebase, I just wanted to divide the work to avoid accumulating too much code.

Ideally the PR is mergeable as it is (nothing should be visible in production for now), I can rename the PR to just something like: "Docs: Add color-mode module" and merge the code, then I can continue the dark mode goal with other PRs.

But if you prefer I can continue working on this single branch, accumulating other changes, let me know

cc @ckifer @PavelVanecek

cloud-walker:
I moved the main logic inside an external store to centralise all the things, codecov is mentioning coverage issues that I can address but I think are very rare and stupid corner cases to check, let me know.

As I said here: https://github.com/recharts/recharts/pull/6828#discussion_r2650402102 to me the work batch its enough to mergeable without waiting all the rest of the work, let me know also on that.

www/.gitignore
PavelVanecek: We have no database, no secrets, nothing on the website that needs an .env file, I don't understand why we need .env.local either.

How much do you expect there is to be done extra? Are you planning to visit all the examples to see if they work in dark mode? That should be probably in separate PR. I would be happy to merge if we have the switch in desktop + mobile nav, and some basic dark mode styles (background, text color). I added CSS variables recently so that should be a fairly small change.

cloud-walker: yes I'm planning to review every corner of the docs to see if they will work on dark mode.

Also I fear that not all the parts are easily addressed with CSS, sometimes I will need to pass the values from JS (thus the ColorModeWatcher component to keep track of the mode in JS world).

> Are you planning to visit all the examples to see if they work in dark mode? That should be probably in separate PR.

so, until we are not happy with the dark mode, its better to hide the feature no?

PavelVanecek: Oic. We're not that big of an operation to require feature flags. I'll be okay if we just released it to everyone and called it "dark mode (beta)" or something like that. If you prefer to keep it simple then that's an option. Feature flag is also fine

cloud-walker: I can do an attempt on avoid the feature flag by making the default mode "light" for now, so the user will se "light" instead of "system" until we are out completely, what do you think?

PavelVanecek: That works. Or if the FF is already done then you can keep it there too.

www/src/components/color-mode/defineColorModeStore.ts
PavelVanecek: Nobody will read those logs. Feel free to keep them there, or delete. There is no monitoring.

PavelVanecek (approved):

PavelVanecek:
This should get deployed to staging https://recharts.github.io/recharts/canary/www/ in couple minutes: https://github.com/recharts/recharts/actions/runs/20586840515

Issue #6843, opened 2026-01-02
Storybook down?


cloud-walker:
## What is the current behavior?

<img width="615" height="541" alt="Image" src="https://github.com/user-attachments/assets/49f71ccd-e86a-45a6-a7b9-f1df3719860e" />

## What is the expected behavior?

It should open up instead of showing the screenshotted error

## Please provide a demo of the problem in a sandbox

Just visit the storybook production link: https://recharts.github.io/en-US/storybook/

PavelVanecek:
@cloud-walker it works if you open the iframe link in new tab: https://main--63da8268a0da9970db6992aa.chromatic.com/

Chromatic must have changed something on their end, looks like they no longer like storybooks in iframes?

Also https://github.com/recharts/recharts/issues/6668

ckifer:
This is definitely a CSP issue for whatever reason.

`x-frame-options: sameorigin` is blocking the iframe embed. Github pages could've potentially changed something? I can still access the storybook in chrome but not in firefox. The chrome response has no `x-frame-options` and the FF response does

cloud-walker:
Error from chrome: 

<img width="795" height="135" alt="Image" src="https://github.com/user-attachments/assets/fe10c802-1e93-4fa2-a08c-02037df4a727" />

ckifer:
I must have something cached, can repro in incognito. Not too sure if we can do anything since we don't control github's CSP's at all.

That 401 could also be the cause

Request to `https://www.chromatic.com/login?redirect=xxx` which I do not get in chrome where I must have had some cookies stored. I then removed the a cookie entry that looked like an access token and then get the above

---

Edit: if I copy the 401 url and paste it into the web browser manually, and then go back the storybook loads as it should.

So probably the CSP prevents that request and then nothing can happen from there? Just guessing

cloud-walker:
chromatic itself seems to have a probem, here a screenshot from the storybook docs about embeds: 

<img width="850" height="776" alt="Image" src="https://github.com/user-attachments/assets/4f3d407c-68fe-40b0-8006-25e4556e1e39" />

cloud-walker:
In the meantime, I can try to build storybook locally during the `deploy-website.yml` workflow, and make the iframe point to it, what do you think?

ckifer:
Sure, feel free to try. Another solution is just make it an external link directly to the chromatic deployment. Not as nice but it works

cloud-walker:
[...]

PR #6846, opened 2026-01-03
fix: link storybook locally

cloud-walker:
## Description

Follow up from #6843, the idea is to not rely on chromatic to show the storybook stories. With this work I discovered that www docs are in SSR vite, in which I'm not an expert of but, I've built both storybook and www and simulated the steps present on the CI workflow locally (using `npm run preview -w www` command to startup the vite preview), and it seems to work properly.

## Related Issue

#6843 

## Motivation and Context

Storybook is not working in production

## How Has This Been Tested?

> I've built both storybook and www and simulated the steps present on the CI workflow locally (using `npm run preview -w www` command to startup the vite preview), and it seems to work properly.

image

## Types of changes

- [x] Modify the deploy-website.yml workflow to build storybook and move it inside the docs build folder
- [x] Make the iframe point to the storybook build folder instead of chromatic URL

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Storybook documentation is now self-hosted and included in deployments, loading from a local static build instead of an external service for improved performance and reliability.

### DIFF
--- a/.github/workflows/deploy-website.yml
+++ b/.github/workflows/deploy-website.yml
@@ -40,6 +40,14 @@ jobs:
         working-directory: www
         run: VITE_RECHARTS_LATEST_VERSION=`curl -X GET "https://registry.npmjs.org/recharts/latest" | jq --raw-output '.version'` npm run build
 
+      - name: Build storybook
+        run: npm run build-storybook
+
+      - name: Move storybook to docs folder
+        run: |
+          mkdir www/docs/storybook-static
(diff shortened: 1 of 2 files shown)


PavelVanecek:
Ah nice. Do we still need the iframe? Maybe without the iframe we could get urls working properly so that we can link to things.

cloud-walker:
I tried getting rid of the iframe, but with my brief searches its seems that storybook cannot be integrated as a Component, part of the main application, we should investigate further or migrate away to something else?

PR #6867, opened 2026-01-10
[RFC] TypedDataKey

PavelVanecek:
## Description

Proof of concept of how we could add strict typing to our chart data. Includes new types for DataKey and get getValueByDataKey, and demo usage on Area component.

There should be no change in behavior, all is just types.

I'm looking for opinions. Once we have agreed this is the way to go I will work on removing the old type definition completely.

Benefits:
- Backwards compatible - people who do nothing will continue seeing `any`
- Now TypeScript will see the relationship between `data` and `dataKey` and check that the key matches the data shape
- Thanks to this, IDEs can start showing autocomplete (at least mine does) in dataKey

Cons:
- This type checking only works within one component. So it works when both data and dataKey are in an Area, but out of the box it won't work when data is on AreaChart and dataKey is on Area.
- To work around that I have added optional generic to Area where one can declare explicitly what is the expected type, and then TypeScript again has a chance to verify
- The syntax for this is awkward, the pointy brackets are sometimes JSX and sometimes generics. On top of that I would also like to add function similar to what Redux is doing with an extra function or method: https://redux.js.org/usage/usage-with-typescript#withtypes I am still contemplating the exact shape for Recharts.

## Related Issue

## How Has This Been Tested?

All tests and types are passing - the build output and omnidoc are not, I will fix before merging.

## Types of changes

- [x] Types only

## Checklist:

- [ ] I have updated the documentation accordingly. TODO

## Release Notes

* **Refactor**
  * Improved data extraction and value handling across chart components with a new typed data key system.
  * Added generic type parameters to Area, Bar, and related components for enhanced type inference.
  * Updated utility functions for more flexible and robust data handling with stricter null-safety checks.
  * Refined data domain truncation logic for edge cases.

* **Chores**
  * Enhanced ESLint configuration for stricter type checking.
  * Expanded test coverage for typed data access and component generics.

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { MutableRefObject, PureComponent, ReactNode, useCallback, useMemo, useRef, useState } from 'react';
 import {
   MutableRefObject,
   PureComponent,
@@ -29,7 +28,6 @@ import {
   AnimationTiming,
   CartesianLayout,
   DataConsumer,
-  DataKey,
   DataProvider,
   DotType,
   LegendType,
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -758,7 +758,7 @@ export function computePieSectors({
   if (sum > 0) {
     let prev: PieSectorDataItem;
     sectors = displayedData.map((entry: unknown, i: number) => {
-      const val: number = getValueByDataKey(entry, dataKey, 0);
+      const val: unknown = getValueByDataKey(entry, dataKey, 0);
       const name: string = getValueByDataKey(entry, nameKey, i);
       const coordinate: PieCoordinate = parseCoordinateOfPie(pieSettings, offset, entry);
       const percent = (isNumber(val) ? val : 0) / sum;
--- a/src/polar/Radar.tsx
+++ b/src/polar/Radar.tsx
@@ -536,7 +536,7 @@ function PolygonWithAnimation<DataPointType = unknown, ValueAxisType = unknown>(
 }
 
 function RenderPolygon<DataPointType = unknown, ValueAxisType = unknown>(
-  props: InternalProps<DataPointType, ValueAxisType>,
+  props: InternalRadarProps<DataPointType, ValueAxisType>,
 ) {
   const previousPointsRef = useRef<ReadonlyArray<RadarPoint> | undefined>(undefined);
   const previousBaseLinePointsRef = useRef<ReadonlyArray<RadarPoint> | undefined>(undefined);
--- a/src/polar/RadialBar.tsx
+++ b/src/polar/RadialBar.tsx
@@ -35,7 +35,6 @@ import {
   PresentationAttributesAdaptChildEvent,
   TickItem,
   TooltipType,
-  TypedDataKey,
 } from '../util/types';
 import {
   TooltipTriggerInfo,
--- a/src/util/ChartUtils.ts
+++ b/src/util/ChartUtils.ts
@@ -32,7 +32,6 @@ import {
   StackOffsetType,
   TickItem,
   AxisDomainTypeInput,
-  AxisDataType,
 } from './types';
 import { ValueType } from '../component/DefaultTooltipContent';
 import { LegendSettings } from '../state/legendSlice';


PavelVanecek:
@coderabbitai review

eslint.config.mjs
```diff
@@ -470,6 +470,21 @@ const jsdocConfig = [
   },
 ];
 
+const neverAny = {
```
PavelVanecek, line 477: TypeScript does not have these settings so eslint has to fill that gap

src/cartesian/Area.tsx
```diff
@@ -1013,26 +997,24 @@ export function computeArea({
   let isRange = false;
 
   const points: ReadonlyArray<AreaPointItem> = displayedData.map((entry, index): AreaPointItem => {
-    let valueAsArray: [number, number] | undefined;
```
PavelVanecek: The `[number, number]` was a mistake - charts can have string domains on _both_ axes and they work perfectly fine, so this should be unknowns. As long as we can feed it to the scale, everything is fair game.

src/cartesian/Area.tsx
```diff
@@ -1013,26 +997,24 @@ export function computeArea({
...
         isRange = true;
+      } else {
+        valueAsArray = [baseValue, rawValue];
       }
     }
 
     const value1 = valueAsArray?.[1] ?? null;
 
-    const isBreakPoint = value1 == null || (hasStack && !connectNulls && getValueByDataKey(entry, dataKey) == null);
+    const isBreakPoint = value1 == null || (hasStack && !connectNulls && getTypedValue(entry, dataKey) == null);
```
PavelVanecek: Usage is the same

src/cartesian/Area.tsx
```diff
@@ -1126,5 +1108,8 @@ function AreaFn(outsideProps: Props) {
  * @provides LabelListContext
  * @consumes CartesianChartContext
  */
-export const Area: ComponentType<Props> = React.memo(AreaFn, propsAreEqual);
+export const Area = React.memo(AreaFn, propsAreEqual) as <DataPointType = any, ValueAxisType = any>(
```
PavelVanecek, line 1143: Defaults to `any` to keep existing behavior.

src/util/getTypedValue.ts
PavelVanecek, line 1: I don't like this definition. If you can think of ways to improve it I am listening.

test/cartesian/Area.typed.spec.tsx
PavelVanecek, line 1: This is probably the most interesting file to review @ckifer

test/cartesian/Area.typed.spec.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+  describe('with explicit Area type parameters', () => {
+    it('should allow correct dataKey when I type the Area to my data type', () => {
+      /*
+       * Here, I define the Area with explicit type parameters, so TypeScript can check the dataKey properly.
+       * The expected type matches so there are no compile errors.
+       */
+      const { container } = rechartsTestRender(
+        <AreaChart data={data} width={400} height={400}>
+          <Area<ExampleDataPoint> dataKey="value" />
```
PavelVanecek, line 160: intellisense!
image

test/cartesian/Area.typed.spec.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+    it('should allow when Area data is matching Area dataKey as function', () => {
+      /*
+       * It is possible to define the data directly on Area,
+       * and in that case the dataKey can be properly checked against that data.
+       * IntelliJ even shows the correct type hints here.
+       */
+      const { container } = rechartsTestRender(
+        <AreaChart width={400} height={400}>
+          <Area data={data} dataKey={entry => entry.value} />
```
PavelVanecek, line 126: Type hint shows correctly (no more any) though intellisense won't work
image

test/cartesian/Area.typed.spec.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+    it('should show error if I type the Area to my type explicitly but use wrong dataKey', () => {
+      /*
+       * Here, I define the Area with explicit type parameters, so TypeScript can check the dataKey properly.
+       * ExampleDataPoint is the input, and number is the expected output type of the dataKey.
+       */
+      const { container } = rechartsTestRender(
+        <AreaChart data={data} width={400} height={400}>
+          {/* @ts-expect-error TypeScript is correct here - the dataKey does not match and this should be an error */}
+          <Area<ExampleDataPoint, number> dataKey="name" />
```
PavelVanecek, line 206: New (optional) syntax - it doesn't feel great.

test/state/selectors/axisSelectors.spec.tsx
```diff
@@ -167,7 +167,7 @@ describe('selectAxisDomain', () => {
     const { container } = render(
       <ComposedChart data={PageData} width={100} height={100}>
         <Area dataKey="" />
-        <Area dataKey="" data={[{ x: 10 }, { x: 20 }, { x: 30 }]} />
+        <Area dataKey="x" data={[{ x: 10 }, { x: 20 }, { x: 30 }]} />
```
PavelVanecek, line 170: This was incorrectly set up before (it didn't matter in the test) but TypeScript is highlighting it as a compile-time error now which I think is a great improvement!

ckifer:
Nice!

Also not the biggest fan of the generic component syntax but whatcha gonna do, this is a nice step forward and if people want to make their charts strictly typed they'll need to make them uglier.

We have to be careful not to break this once its in place though

src/util/getTypedValue.ts
ckifer: whew thats a doozy

test/cartesian/Area.typed.spec.tsx
ckifer: but better than not being possible

cloud-walker:
Hi!, dunno if its late or if it can help, but on my product codebase I've baked this solution for a typesafe rechart version, taking advantage of the new features from recharts 3 + the good old render prop pattern.


This an example for ComposedChart:
```tsx
import {token} from '@translated/babele-core'
import * as C from 'recharts'

import {ChartContainer} from '@/ui/Charts'

import {ChartArea} from './ChartArea'
import {ChartBar} from './ChartBar'
import {ChartTooltip} from './ChartTooltip'
import {ChartXAxis} from './ChartXAxis'
import {ChartYAxis} from './ChartYAxis'

export declare namespace ComposedChart {
  export interface Props<TPoint extends Record<string, unknown>> {
    /**
     * @todo cannot be readonly because of Recharts types
     */
    data: TPoint[]
    chartRef?: React.Ref<HTMLDivElement>
    render: (props: {
      Bar: typeof ChartBar<TPoint>
      Area: typeof ChartArea<TPoint>
      XAxis: typeof ChartXAxis<TPoint>
      YAxis: typeof ChartYAxis<TPoint>
      Tooltip: typeof ChartTooltip<TPoint>
    }) => React.ReactNode
  }
}

export function ComposedChart<TPoint extends Record<string, unknown>>(
  props: ComposedChart.Props<TPoint>,
) {
  return (
    <ChartContainer ref={props.chartRef}>
      <C.ComposedChart data={props.data}>
        <C.CartesianGrid
          vertical={false}
          strokeDasharray="4 4"
          stroke={token('colors.border.neutral.secondary')}
        />
        {props.render({
          Bar: ChartBar,
          Area: ChartArea,
          XAxis: ChartXAxis,
          YAxis: ChartYAxis,
          Tooltip: ChartTooltip,
        })}
      </C.ComposedChart>
    </ChartContainer>
  )
}
```

Then the usage becomes:

```tsx
<ComposedChart
  chartRef={chartRef}
  data={points}
  render={({Bar, XAxis, Tooltip}) => {
    return (
      <>
        <XAxis
          dataKey="order_date"
          tickFormatter={(d) => formatChartDate(d, props.breakdownBy)} // d is of type Date!
        />
        <Bar dataKey="my_kpi" tone={kpiToTone.quality} /> // Bar knows which keys are available thanks to the render prop wiring
        <Tooltip
          tone={kpiToTone.quality}
          title={(p) => formatChartDate(p.order_date, props.breakdownBy)}
          mainRow={(p) => {
            if (p.ept_r1 == null) {
              return
            }
            return `${formatEPT(p.ept_r1)} EPT`
          }}
        />
      </>
    )
  }}
/>
```

I'm experimenting with this in my codebase, but I think its shaping in a good direction, its working good.

On the topic [TkDodo](https://github.com/tkdodo) recently published an article: [Building Type-Safe Compound Components](https://tkdodo.eu/blog/building-type-safe-compound-components), that shares the same principle, but with a factory function instead of the render prop

PavelVanecek:
Thanks @cloud-walker. I'm inclined towards the typed-function call (what the blog post calls "Component Factory Pattern") rather than renderProps because there's less runtime impact. I want to first test how it impacts bundle size and treeshaking though which is why I started with the explicit typing first.

PR #6868, opened 2026-01-11
feat(Tooltip): allow offset prop to accept Coordinate object

### LINKED ISSUE #6643: [Feature request] [Tooltip] `offset` prop should accept an object with separate `{ x, y }` offsets
It seems impossible to position a `Tooltip` the way I want, which is 10-20px away from the cursor horizontally but not vertically.

If `offset` accepted an object of `{ x, y }` offsets instead of just a `number`, I would set it to `{ x: 20, y: 0 }`

Using `offset={20}` and `translateY(-20px)` in CSS, but that doesn't work when the tooltip is flipped vertically.

**Additional context**
Thank you!

bigsaigon333:
## Description

  Allow the `offset` prop of the Tooltip component to accept either a `number` or a `Coordinate` object (`{ x: number, y: number }`). This enables specifying different horizontal and vertical offsets independently.

  **Changes:**
  - Extended `offset` prop type from `number` to `number | Coordinate` in `Tooltip` and `TooltipBoundingBox`
  - Split internal `offsetTopLeft` parameter into separate `offsetLeft` and `offsetTop` in the translate utility
  - Updated Storybook and API documentation to reflect the new type

## Related Issue

closes #6643

  ## Motivation and Context

  Currently, the `offset` prop only accepts a single number that applies equally to both x and y axes. Users want to position tooltips with different offsets per axis (e.g., 20px horizontal offset but 0px vertical offset). This change allows `offset={{ x: 20, y: 0 }}` to achieve that.

  ## How Has This Been Tested?

  - Added unit tests for getTooltipTranslate with different offsetTop and offsetLeft values
  - Added unit tests for TooltipBoundingBox with Coordinate type offset (number, object with different x/y, negative values)
  - Added unit tests for Tooltip offset positioning with zero value edge cases (offset=0, offset={ x: 0, y: 10 }, offset={ x: 10, y: 0 })
  - All existing tests pass

  ```bash
  npm test -- test/util/tooltip/translate.spec.ts test/component/TooltipBoundingBox.spec.tsx test/component/Tooltip/Tooltip.offset.spec.tsx
```

##  Screenshots (if appropriate):

  N/A

##  Types of changes

##  Checklist:

## Summary by CodeRabbit

* **New Features**
  * Tooltip offset now accepts either a single number or a coordinate object, enabling uniform or per-axis horizontal/vertical offsets.

* **Documentation**
  * API docs and storybook descriptions updated to explain numeric vs coordinate offset behavior, negative offsets, and default value.

* **Tests**
  * Added and expanded tests covering numeric offsets, distinct x/y coordinate offsets, and negative-offset scenarios.

### CODE REVIEW

File: src/util/tooltip/translate.ts [hunk 1]
```diff
@@ -44,7 +44,7 @@ export function getTooltipTranslateXY({
   allowEscapeViewBox: AllowInDimension;
   coordinate: Coordinate;
   key: Dimension2D;
-  offsetTopLeft: number;
+  offset: number;
```
bigsaigon333, line 47: I kept the function signature to take a single `offset` parameter (which used to be `offsetTopLeft`) instead of splitting it into both `offsetTop` and `offsetLeft`.

  My thought was - if the function receives both values and picks one based on key inside, it feels a bit redundant. The caller already knows which dimension it's dealing with, so why not just pass the right offset from outside?

  I'm not 100% sure this is the best way though. Would love to hear your thoughts on this!

storybook/stories/API/props/TooltipArgTypes.tsx
```diff
@@ -25,12 +25,16 @@ export const TooltipArgTypes: StorybookArgs = {
...
+    When a Coordinate object is provided, you can specify different offsets for each axis.
+    The active position depends on the data point, and mouse cursor position.
+    The offset can be negative.`,
     defaultValue: 10,
     table: {
-      type: { summary: 'number' },
+      type: {
+        summary: 'number | Coordinate',
+        detail: 'number | { x: number, y: number }',
+      },
```
bigsaigon333: I couldn't find a way to support both number and object types in Storybook's control tab at the same time. Also, it seems like other props with object types don't get edited properly either. Am I missing something here?

image

www/src/docs/api/TooltipAPI.tsx
bigsaigon333, line 1: It looks like this:

image

PavelVanecek: Please don't edit this file directly, it's autogenerated. Put the comments into `Tooltip.tsx` JSDoc and then run `npm run omnidoc`

test/component/TooltipBoundingBox.spec.tsx
```diff
@@ -59,4 +59,21 @@ describe('TooltipBoundingBox', () => {
     expect(screen.getByText('Hello world!')).toBeInTheDocument();
     expect(screen.getByText('Hello world!')).not.toBeVisible();
   });
+
+  describe('offset prop', () => {
+    it('should accept number offset', () => {
+      render(<TooltipBoundingBox {...defaultProps} offset={15} />);
+      expect(screen.getByText('Hello world!')).toBeInTheDocument();
```
PavelVanecek, line 66: We need better assertions here - please look at what other tests are doing, there's `showTooltip` + `mockGetBoundingClientRect` + `expectTooltipCoordinate` helpers available.

storybook/stories/API/props/TooltipArgTypes.tsx
PavelVanecek: I think number alone is fine, it's a bit of a pain to set objects in Storybook anyway. We have direct code editor + stackblitz for people who want to dig deeper.

src/util/tooltip/translate.ts
PavelVanecek: Looks good to me like this

storybook/stories/API/props/TooltipArgTypes.tsx
bigsaigon333: Makes sense. Simplified the type to just number!  [cf60b97](https://github.com/recharts/recharts/pull/6868/commits/cf60b97568696125f7356eeb02e785cc57c36fcb)

www/src/docs/api/TooltipAPI.tsx
bigsaigon333: Thanks for the heads up! I wasn't aware this file was autogenerated. Updated the JSDoc in Tooltip.tsx and regenerated with omnidoc. [4ffd9d0](https://github.com/recharts/recharts/pull/6868/commits/4ffd9d0f9409dbbfe8e53af04e3161ef347b86f9)
  
  
image

test/component/TooltipBoundingBox.spec.tsx
bigsaigon333: Oh, I wasn't aware of those helper functions. Thanks for letting me know!

  I've added tests for the actual tooltip positioning with offset in [e5e774b](https://github.com/recharts/recharts/pull/6868/commits/e5e774b43010309635e786501cd9d30b65a4f9e7).

  As for this test, it's essentially just verifying that offset accepts Coordinate type, which TypeScript already covers. I'm thinking of removing it - what do you think?

bigsaigon333:
>  32-50: Consider adding edge case test for offset=0.
>> The numeric offset tests are correct and well-documented. However, consider adding a test for offset=0 to ensure the tooltip appears at the exact data point position without any offset.


Added tests for offset=0, offset={ x: 0, y: 10 }, and offset={ x: 10, y: 0 } edge cases! [3e88bef](https://github.com/recharts/recharts/pull/6868/commits/3e88befb93c2f9746c511a92a903ea39d39407dc)

ckifer:
thanks @bigsaigon333

PR #6916, opened 2026-01-22
[dark mode] line chart examples

cloud-walker:
## Description

based on #6882 discoveries, we have to:

1. support stackblitz examples
1. update docs only snapshots

So I'm preparing a smaller PR to solve those first

## Related Issue

#6825 

## Motivation and Context

Otherwise stackblitz does not work anymore

## How Has This Been Tested?

Playwright and manually

Are in the diff

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Expanded color system with new light/dark tokens and emphasis colors.
  * StackBlitz previews now include global CSS and respect a default theme.

* **Improvements**
  * Line chart examples restyled to use theme variables (axes, grid, tooltip, dots).
  * Vertical charts accept optional external animation control.
  * ColorMode type is now publicly available for consumers.

* **Tests**
  * Visual tests updated: per-test color-mode wrapper, consolidated mount fixture, router-driven snapshots, animations disabled by default.

### DIFF
--- a/www/src/docs/exampleComponents/LineChart/CompareTwoLines.tsx
+++ b/www/src/docs/exampleComponents/LineChart/CompareTwoLines.tsx
@@ -738,8 +738,8 @@ const data: DataPoint[] = [
 ];
 // #endregion
 
-const color = '#bd7f2a';
-const compareColor = '#dd4b1a';
+const color = 'var(--color-chart-3)';
+const compareColor = 'var(--color-chart-8)';
 
 const { yAxisMin, yAxisMax } = data.reduce(
   (acc, val) => {


cloud-walker:
I need to study how to support storybook also

PavelVanecek:
Storybook already has dark mode included

cloud-walker:
But its not in sync with the color-mode algo, I will see in a future PR btw

cloud-walker:
what do you think about the fixture? I will need to use it wrap all www related VR tests

www/src/components/Shared/StackBlitzLink.tsx
```diff
@@ -48,7 +52,7 @@ root.render(<AppWithDevtools />);
 // language=HTML
 const indexHtmlCode = (title: string) => `
   <!doctype html>
-  <html lang="en">
+  <html lang="en" data-mode="light">
```
PavelVanecek: Looks like stackblitz has its own theme switcher built-in but no way to propagate that to the actual example.

Could we instead make it so that this data-mode will reuse what user had selected on the website? Instead of default to light.

test-vr/tests/www/fixtures.tsx
```diff
@@ -0,0 +1,16 @@
+import { test as originalTest } from '@playwright/experimental-ct-react';
+import * as React from 'react';
+
+import { TestColorModeProvider } from './TestColorModeProvider';
+
+export const test = originalTest.extend({
```
PavelVanecek, line 6: Looks good to me. This is what Playwright docs suggest to use, right? https://playwright.dev/docs/test-fixtures

test-vr/tests/www/TestColorModeProvider.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import '../../../www/src/styles/app.css';
```
PavelVanecek, line 2: I'm wondering if importing the whole `app.css` will come to bite us one day.

www/src/docs/exampleComponents/LineChart/BiaxialLineChart.tsx
PavelVanecek, line 1: This will be mind numbing. Is this an opportunity to define a new `Theme` component and change this for all colors in all charts at once?

www/src/docs/exampleComponents/LineChart/LineChartAxisInterval.tsx
```diff
@@ -48,30 +48,45 @@ const data = [
 ];
 
 // #endregion
+const MyChart = (props: { interval: AxisInterval }) => (
```
PavelVanecek, line 51: 🎖️

www/src/docs/exampleComponents/LineChart/BiaxialLineChart.tsx
PavelVanecek: What I am imagining:

```
import { LightTheme, DarkTheme, BarChart } from 'recharts';

return <LightTheme><BarChart>...</LightTheme>
```

or something in this vein. Makes it easy for us, we can hook inside our charts to replace defaults with the theme instead, we don't have to rewrite thousand charts and neither have our consumers. WDYT?

www/src/components/Shared/StackBlitzLink.tsx
cloud-walker: interesting, I'll do check

test-vr/tests/www/fixtures.tsx
cloud-walker: yes I considered different solutions but this seems to be the best trade-off to me

test-vr/tests/www/TestColorModeProvider.tsx
cloud-walker: we will see, in which case we can evolve the architecture maybe

www/src/docs/exampleComponents/LineChart/BiaxialLineChart.tsx
cloud-walker: the architecture that I built should do it by just throw a `data-mode="light"` or dark on the sub tree where you want to stub the color-mode.

This only work on recharts docs btw, not recharts library.

But maybe I'm missing your point.

What I considered doing in another attempt was to create a "recharts" module "proxy", only for the docs, that re-exports recharts primitives, but decorating them with theme defaults, like:

```tsx
import * as C from 'recharts'
export {BarChart, ...} from 'recharts'

export function Bar(props: C.BarProps) {
  return <C.Bar {...props} dot={{stroke: 'var(--color-surface-base'), ...props.dot}} activeDot={...} />
}
```

But then those props would be hidden on the examples code blocks, so the end user would start to believe that recharts supports dark mode automagically

PavelVanecek: Yes so my concern is we have to modify all examples, one by one. And each of those examples will need to have tons of `fill`s and `stroke`s and if we miss one then it will look wrong. You will become an expert at it but other people later will definitely get it wrong.

If we had some sort of wrapper, like the Theme component (which will set a React context inside), or the proxy as in your example, we would have all colors covered in single spot. Easy to use, takes care of all colors internally, easy to create custom ones.

Then once we have this wrapper, we can go and either

- Put it in the root of the whole website
  - ➕  good DRY
  - ➖  but it is hidden out of the way and not at all clear how the colors got there when looking at a single example

- Put it in all examples one at a time
  - ➖ a bit of more work
  - ➕ still a lot less work than adding `fill` and `stroke` everywhere
  - ➕ when people look at the example in isolation it's obvious how the colors are set

And if it's a wrapper component or proxy function? I'm a bit more inclined towards the `<LightTheme>` syntax because that visually matches what we do with all other components: like ResponsiveContainer, Brush, all the charts. But the end result will be the same either way. We don't have ambitions to expand outside of the React ecosystem.

PavelVanecek: We could perhaps export two components and one function:

- `createTheme`: function that accepts all colors (with strict typing so that typescript can tell you which ones are missing, and returns the theme component/proxy
- `export const LightTheme = createTheme({ ... })`:
- `export const DarkTheme = createTheme({ ... })`:

Then we use LightTheme/DarkTheme on our website, integrate with the switcher. And other people can reuse the same or easily create their own themes.

cloud-walker: if we introduce an official way to customise the chart colors, the need to show everything in every example disappear to me, I was focused on the dark mode goal of the docs, but obviously is we introduce an official API, we can focus on document THAT API and stop being repetitive in the docs, it maybe a nice idea.

But it also depends on the direction we want recharts to go, and maybe move this discussion in a discussion thread?

I have some concerns on the topic:

1. In a way I'm an headless component libraries lover (like ariakit, ark UI, radix, etc.), and I think the styling concern should not be part of the library, but obviously leave a good way to style it to the end user.
2. If we think that recharts should indeed be responsible also of the styles, I fear that its more complex that Light or Dark, for example in my company I have multi themes and modes (<theme>-<mode>), like: alpha-dark, alpha-light, beta-dark, beta-light, etc.
3. this is not strictly related but now that you raised the topic: a lot of devs out there tries to work with class names,  instead of hardcoding attributes, like `fill=...` or `stroke=...`, but I'm not a recharts impl or svg expert so maybe there is no way to handle all with class names.

So based on my concern 2, I think that we should not give to the end user an already baked set of themes, but instead a way to tap in the "recharts primitive tokens", like colors, radius, etc.

So yeah maybe something like you are proposing, but without exporting pre-bundled themes:

```tsx
interface RechartsThemerProps {
  seriesStroke?: string
  seriesFill?: string
  // ... or exploded per component, this smells like breaking the encapsulation of the primitives maybe?
  
}

export function RechartsThemer(props: RechartThemerProps) {
   // ... a react provider?
}
```

cloud-walker: what about theming differents Bar or Lines, we should give this new APIs a way to configure a set of data series, something like:

```tsx
interface RechartsThemerProps {
  series?: ReadonlyArray<{
    name: string
    stroke: '...',
    fill: '...',
    dot: {...}
  }>
  // ... or exploded per component, this smells like breaking the encapsulation of the primitives maybe?
  
}

export function RechartsThemer(props: RechartThemerProps) {
   // ... a react provider?
}

// ...

export const myThemer = <RechartsThemer series={[{name: 'magenta', stroke: '...', ...}, {name: 'indigo', stroke: '...'}]}>...</RechartsThemer>

export const myMagentaBar = <Bar tone="magenta" ... /> // this should also be type safe!
```

cloud-walker: I still think that recharts should not know about styling, but gives a clear way to be styled from the outside like  an headless library, and that this is a docs specific issue.

With recharts v3 latest changes (thanks btw!) on my repos I can map every primitives with my design system theming, and cover type safety at the same time:

```tsx
import {Bar} from 'recharts'
import {ColorTone, token} from 'my-design-system'

export interface MyDesignSystemBarProps {
  tone: ColorTone
} 
export function MyDesignSystemBar(props: MyDesignSystemBarProps) {
  return <Bar stroke={props.tone === 'magenta' ? token('color.magenta.6') : token('color.indigo.6')} ... />
}
```

cloud-walker:
`VerticalLineChartWithSpecifiedDomain` vr test is failing in the CI but not locally 🫠

www/src/docs/exampleComponents/LineChart/BiaxialLineChart.tsx
cloud-walker: if you agree, in the meantime, I can consider introducing a "proxy" recharts module inside the docs, but using it only in post getting started pages, so in the getting started the user can see better styling details, then we can continue the theming discussion in a dedicated discussion

PavelVanecek: How does the proxy affect tree shaking and bundle size? It won't matter for our website (we bundle all charts anyway) but we should be showing best practices in the examples.

test-vr/tests/www/LineChartApiExamples.spec-vr.tsx
```diff
@@ -71,7 +71,9 @@ test('LineChartConnectNulls', async ({ mount }) => {
 
 test('VerticalLineChartWithSpecifiedDomain', async ({ mount }) => {
   const component = await mount(<VerticalLineChartWithSpecifiedDomain />);
-  await expect(component).toHaveScreenshot();
+  await expect(component).toHaveScreenshot({
+    maxDiffPixelRatio: 0.1,
```
PavelVanecek: Please don't use `maxDiffPixelRatio`, not in this project. Because the chart is 99.9% empty space, the 0.1% means that the line can move somewhere completely else and the test will pass.

www/src/components/Shared/StackBlitzLink.tsx
```diff
@@ -131,57 +136,65 @@ export default defineConfig({
...
-          /*
-           * https://developer.stackblitz.com/platform/api/javascript-sdk-options
-           */
-          {
-            template: 'node',
-            title,
-            files: {
-              'index.html': indexHtmlCode(title),
+    <ColorModeWatcher
+      render={state => {
```
PavelVanecek: I wonder why is this a render prop and not a hook. But if it works then I suppose it's okay either way.

www/src/docs/exampleComponents/LineChart/BiaxialLineChart.tsx
PavelVanecek: And yes agree we need to think the theming a little bit more, I did not consider multiple graphical elements. I will summarize this into a discussion

PavelVanecek: https://github.com/recharts/recharts/discussions/6928

PavelVanecek:
Dark mode sounds simple until it doesn't, right @cloud-walker ?

cloud-walker:
we will found a way sooner or later 😄

test-vr/tests/www/LineChartApiExamples.spec-vr.tsx
cloud-walker: its only firefox that fails, there is something that differs from local machine to CI, the only thing I can think of is slowness of the test? Maybe the chart does not finish the render in time?, but very strange, only two charts fails and only on firefox

www/src/components/Shared/StackBlitzLink.tsx
cloud-walker: render prop here its a stretch, because the only rendered element is the button itself, but in bigger react trees can improve the performance, because only the leaf of the tree interested in the color-mode state will re-render.

We can introduce a hook version of the ColorModeWatcher of course

ckifer:
Why do we need to support the stackblitz examples? Just let them go back to light mode. Our website is being themed, not recharts (or at least not in this PR) and stackblitz is outside of that

www/src/docs/exampleComponents/LineChart/BiaxialLineChart.tsx
cloud-walker: > How does the proxy affect tree shaking and bundle size? It won't matter for our website (we bundle all charts anyway) but we should be showing best practices in the examples.

it would affect only the specific re-exports, but I mean, I think its best practise to have a local "charts" module inside the consumer application code, to centralise configuration (not only color stuff).

For the tree-shaking part, in theory modern bundlers, like vite, should handle `import * as X from` properly, but sincerely I'm a bit rusty on the matter.

That said, we can bake another PR to not block this one further, if you agree.

www/src/docs/exampleComponents/LineChart/LineChartHasMultiSeries.tsx
```diff
@@ -33,13 +33,28 @@ const series = [
...
+          name={s.name}
+          key={s.name}
+          stroke="var(--color-chart-1)"
+          dot={{
+            fill: 'var(--color-surface-base)',
+          }}
+          activeDot={{
+            stroke: 'var(--color-surface-base)',
+          }}
+        />
```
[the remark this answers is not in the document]
cloud-walker: I just kept what was before, I thought it was intentional, but I'm not sure

cloud-walker:
> Why do we need to support the stackblitz examples?

At least we need to pass the CSS variables, the code is the same of the docs, but yeah we can keep the light mode by default. But why you think its not a good idea to support color-mode propagation? Its a plus to me that the end user can benefit from it also there. That said, merging it now would worsen the experience for not yet migrated chart, so yeah, better keep light mode at least for now.

> Our website is being themed, not recharts (or at least not in this PR)

Yeah, lets continue the recharts itself theming discussione elsewhere, I agree, but @ckifer what to do you think about the strategy to continue the docs theming?

So to recap, I need to:

1. keep stackblitz in light mode
1. do another attempt on avoiding the `maxDiffPixelRatio` thing

ckifer:
Not saying that we shouldn't support theme in stackblitz or we need to keep it in light mode, only that if we want to do that it doesn't have to be in this PR for the sake of getting this merged

test-vr/tests/www/LineChartApiExamples.spec-vr.tsx
cloud-walker: > Let's find some solution for the VR test that doesn't include maxDiffPixelRatio.
What was the failure? Do you run the tests in Docker?

Yes I run the tests in Docker with `npm run test-vr`

PavelVanecek:
Okay so the example list fails on timeout, looks like some animation never settles? Or github runner is too slow?

The vertical line chart appears to be 1 pixel off. Let me try to checkout your branch and see if I can reproduce the failure.

image

PavelVanecek:
I can see the same failure on my laptop so let's see if this helps

cloud-walker:
I suspect that its something animation related yes, sorry but I'm busy with a release, I'll take care of the PR ASAP. In the meantime, stackblitz its locked on light mode now

PavelVanecek:
Still timing out on the dark mode full-page screenshot

PavelVanecek:
@cloud-walker we have access to TestMu (was LambdaTest before) where we can do full-page screenshots of the website. Would you like to play with that? We can have coverage there, and leave the glacial github runner for simpler tests.

cloud-walker:
> @cloud-walker we have access to TestMu (was LambdaTest before)

how to use it? And why it should work vs playwright?

Btw I discovered this [emulateMedia](https://playwright.dev/docs/api/class-page#page-emulate-media) thing, to simulate reducedMotion: 

```ts
await page.emulateMedia({ reducedMotion: 'reduce' });
```

but it does not seem to work, because probably recharts animations are not respecting the media query

test-vr/tests/www/dark-mode.spec-vr.tsx
```diff
@@ -1,17 +1,47 @@
...
+
+test('dark mode: Simple Line Chart', async ({ mount }) => {
+  const component = await mount(
+    <div
+      style={{
+        backgroundColor: 'var(--color-surface-base)',
+      }}
+    >
+      <SimpleLineChart />
+    </div>,
```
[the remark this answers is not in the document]
cloud-walker: done

test-vr/tests/www/dark-mode.spec-vr.tsx
```diff
@@ -1,17 +1,48 @@
...
 
 test.use({
   colorScheme: 'dark',
 });
 
-test('dark mode snapshot', async ({ mount }) => {
-  const component = await mount(<Root />);
+test.beforeEach(async ({ page }) => {
+  await page.emulateMedia({ reducedMotion: 'reduce' });
+});
```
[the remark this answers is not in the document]
cloud-walker: what are you saying Mr Rabbit, my PR has just been merged #6956

ckifer:
merging this before conflicts

Issue #6668, opened 2025-11-23
Storybook VR test limit


PavelVanecek:
We're regularly reaching the monthly screenshot limit of Chromatic VR tests. The next tier is $399 per month which is firmly outside of our budget: https://www.chromatic.com/pricing.

We can reduce the amount of stories we have:

- Some stories are not pretty and should really just be unit tests
- Some stories are too complex to cover in a unit test, those should be playwright tests
- Some other stories are pretty but too complex to fit in Storybook controls, those should be website examples
- Some other stories are simple, focused on a particular component - those should stay in the storybook

Thoughts?

ckifer:
the chromatic diff is so much easier to see issues than elsewhere but if we commit to getting everything semi-useful into playwright tests I'm okay with it. Its a bit tough since the storybook is technically what we develop with for the most part so we want to add a lot of things there. 

We can maybe start by disabling snapshots for certain stories?
https://www.chromatic.com/docs/disable-snapshots/

PavelVanecek:
Can't see the diff if all it says is "reached your limit".

PavelVanecek:
Also, storybook diff only runs on PR if you are a maintainer - chromatic requires a token and won't run on PR from a fork. So we have no feedback before merging.

ckifer:
True that

PR #6917, opened 2026-01-22
Remove Cell examples from storybooks, fix RadialBar shape type and selector stability

PavelVanecek:
## Description

In this PR:

- I have removed all remaining Cell examples from storybook
- Most of them were already covered on the website, except RadialBar + Cell
- So I added a new website example
- In that process I found that RadialBar.shape type is awkward to work with so I fixed that
- Also I figured why not have some CSS transitions on the clicks to make it look nice
- Found out that wasn't working because Recharts was re-mounting the RadialBar on each render due to un-stable selector
- So I fixed that too and added a test

## Related Issue

https://github.com/recharts/recharts/issues/6668

https://github.com/recharts/recharts/discussions/6734

## Summary by CodeRabbit

* **New Features**
  * Added a RadialBar Chart example demonstrating interactive legend-based focus selection.

* **Documentation**
  * Updated RadialBar storybook entry (simplified API story).
  * Removed several legacy Pie and Bar example stories.

* **Performance & UX**
  * Reduced render overhead and improved stability in radial bar rendering.

* **Tests**
  * Added selector stability and visual tests for RadialBar examples.

* **Chores**
  * Cleaned up TypeScript annotations and refined exported typings.

### DIFF
--- a/src/component/DefaultLegendContent.tsx
+++ b/src/component/DefaultLegendContent.tsx
@@ -28,7 +28,11 @@ export interface LegendPayload {
   value: string | undefined;
   type?: LegendType;
   color?: string;
-  payload?: unknown;
+  /**
+   * Different graphical items put different information in the payload object
+   * so double check in runtime what are you getting here.
+   */
+  payload?: object;
   formatter?: Formatter;
   inactive?: boolean;
   legendIcon?: ReactElement<SVGElement>;
--- a/src/state/selectors/pieSelectors.ts
+++ b/src/state/selectors/pieSelectors.ts
@@ -92,6 +92,7 @@ export const selectPieLegend: (
       return {
         value: getTooltipNameProp(name, pieSettings.dataKey),
         color,
+        // @ts-expect-error Legend payload.payload says it wants objects but our data can be unknown
         payload: entry,
         type: pieSettings.legendType,
       };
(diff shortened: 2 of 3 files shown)


src/component/DefaultLegendContent.tsx
```diff
@@ -28,10 +28,7 @@ export interface LegendPayload {
   value: string | undefined;
   type?: LegendType;
   color?: string;
-  payload?: {
-    strokeDasharray?: number | string;
-    value?: any;
-  };
+  payload?: unknown;
```
ckifer: can we keep value? This is a breaking change?

PavelVanecek: Hm the trouble is that this type was a lie - some graphical elements are passing its own `props`, some other elements are passing the data entry. `unknown` or even `any` is the best we can do here. Unless we unify the behaviour across all graphical items which is even more breaking change.

PavelVanecek: I suppose we can make it an `object` because that much is true.

src/component/DefaultLegendContent.tsx
```diff
@@ -28,10 +28,11 @@ export interface LegendPayload {
...
   color?: string;
-  payload?: {
-    strokeDasharray?: number | string;
-    value?: any;
-  };
+  /**
+   * Different graphical items put different information in the payload object
+   * so double check in runtime what are you getting here.
+   */
+  payload?: object;
```
ckifer, line 35: I suppose:

``` 
{
   [string]: any
}
```
is roughly the same type

Issue #6891, opened 2026-01-15
Hover Transition effect for Bar Cell not working


Zaxarprod:
## What is the current behavior?

In library version 3.6.0, the transition effect on hover does not work. The active bar's color changes without animation.

https://github.com/user-attachments/assets/219fa099-96dd-42a3-b2ff-6b95a4dfaca6

Version 2.15.1. 
Smooth color change
## What is the expected behavior?

https://github.com/user-attachments/assets/afa9c9df-a131-4252-ba4a-78660d7928b2

## Please provide a demo of the problem in a sandbox

The code is roughly as follows:
I save the state of the active cell using the useState hook. 
I animate the hover using `transition: 'fill 0.3s ease-in-out',`.

It's interesting that on version 3.1.0, if isAnimationActive={false}, then everything works. If isAnimationActive={true}, it doesn't. In version 3.6.0, isAnimationActive doesn't work in both states.

```
<Bar
          key={dataKey}
          dataKey={dataKey}
          stackId={stackId}
          radius={0}
          stroke='none'
          strokeWidth={0}
          isAnimationActive
          shapeRendering='crispEdges'
          yAxisId={yAxisId}
          mask={variant === 'dashed' ? `url(#${BAR_BG_PATTERN_STRIPED_MASK_ID})` : undefined}
        >
          {data.map((dataItem, index) => (
            <Cell
              key={`bar-${index}`}
              fill={
                activeBarName === dataItem.name
                  ? basicColorsConfig[activeBarBackground]
                  : basicColorsConfig[barBackground]
              }
              style={{
                transition: 'fill 0.3s ease-in-out',
              }}
            />
          ))}
</Bar>
```

Recharts version: 3.6.0 (but is reproduced on 3.x.x)
Browser: Chrome 143.0.7499.193 (arm64)

ckifer:
reproduction https://stackblitz.com/edit/jlaj9r1o?file=src%2FExample.tsx&showSidebar=0&terminalHeight=10

Not too sure what we can do about this, its probably a mount/unmount problem https://stackoverflow.com/questions/40064249/react-animate-mount-and-unmount-of-a-single-component

Zaxarprod:
> reproduction https://stackblitz.com/edit/jlaj9r1o?file=src%2FExample.tsx&showSidebar=0&terminalHeight=10
> 
[...]

Issue #6856, opened 2026-01-05
[BUG]: Double rendering issue in recharts v3


legions-developer:
Hey creator of [EvilCharts](https://evilcharts.com/) here... while creating charts i found issue in v3 ( i'm attaching video of same chart, same code but in different versions from v2.15.4 -> v.3.6.0

## What is the current behavior?

The bar chart below flickers and appears to be double-rendered when hovered over. I’m using custom bars with shape props. ( This was in v3.6.0 recharts )

https://github.com/user-attachments/assets/628d7324-afa5-42e3-b570-12723f8900b0

## What is the expected behavior?

Now here is example with recharts v2.15.4 (note: it have same code, same chart) but here you can see chart works smoothly?

https://github.com/user-attachments/assets/b76b13eb-b553-4373-9db5-f35285d14b94

## Please provide a demo of the problem in a sandbox

Provided the videos above, as well as made a demo source code public... can be used to reproduce issue

https://github.com/legions-developer/recharts-double-render-issue

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

I suspect this issue is in version 3 of Recharts itself. However, I’m trying to downgrade version 3 to see if a specific version is causing the problem.

entity:
+1 on this! I can also confirm that when I downgrade to v2, I don't get this issue whilst I do on v3.

I have issues where multiple bar charts can be selected (assuming due to the double render)

ckifer:
Thanks, will look into it

legions-developer:
Amazing im here to provide everything to solve this issue

ckifer:
@legions-developer could you provide a reproduction without all of the animation and extra libraries?

https://stackblitz.com/edit/lmswu2f5-2t6zzzbq?file=src%2FExample.tsx

I'm trying but can't quite get there. But at a glance this seems like I can't reproduce without those additions unless I missed something

legions-developer:
yup i did linked a demo repo 

https://github.com/legions-developer/recharts-double-render-issue

here its default in v3.6.0 and then you can downgrade to v2.15.4 then check the same chart

ckifer:
@legions-developer I looked at your demo and thats what I copied the stackblitz from, but I mean without the extra dependencies. I need a demo that is just recharts otherwise its a lot harder to tell where the issue is coming from

ckifer:
@legions-developer does this happen in 3.5.1? Doesn't seem like it

[...]

PR #6920, opened 2026-01-22
Support Bar CSS transitions

PavelVanecek:
## Description

Bar transitions were broken since the ZIndex layers but indeed they are useful! So I put them back.

The process is rather convoluted so what was before "just render" is now a multi step adventure:

1. Render inactive shape in inactive position (this is business as usual)
2. Bar becomes active (usually by mouse hover)
3. Now we move the shape to the active zIndex layer _but render it with inactive props_. This step is necessary so that once we switch to the active props, React will reuse the element and CSS transitions could kick in
4. Now we render active shape, with active props, in active layer
5. CSS transition proceeds as usual

When a Bar becomes inactive, the process is even more complicated because we need to wait until the transition completes before we can move it back:

1. Active shape is in active location
2. Mouse leave
3. Now we render inactive shape, in active location, so that we don't interrupt CSS transitions
4. Browser calls `onTransitionEnd`
5. We put the inactive shape back to inactive layer so that it renders in the proper zIndex

It's ... complicated. But it works. As far as I can tell. If you would like to test some more please do. Or perhaps if you have ideas how to solve this better.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/6891

Fixes https://github.com/recharts/recharts/issues/6856

## Motivation and Context

CSS transitions are very nice to have. Without those we may just as well render to canvas!

## Types of changes

## Checklist:

## Release Notes

* **New Features**
  * Added Animated Bar Width Example demonstrating smooth animated transitions on bar charts.
  * Improved active bar visualization with CSS transitions and layered rendering.
  * Enhanced shape component styling with inactive state support.

* **Tests**
  * Added comprehensive test coverage for CSS transitions in bar components.
  * Added visual regression testing for animated bar functionality.


legions-developer:
legend 🌹

PavelVanecek:
https://recharts.github.io/recharts/pr-6920/www/en-US/examples/AnimatedBarWidthExample/

PR #6940, opened 2026-01-30
Omnidoc updates

PavelVanecek:
## Description

- Omnidoc now generates the index file too, one less thing to maintain
- Applies prettier in memory
- Autogenerated files are out of git, and instead generated during npm build


PavelVanecek:
Oh no I missed a file. I'll fix it tomorrow

Issue #6021, opened 2025-06-26
v3 hook requests for `Customized` component prop replacements


calthejuggler:
Congratulations on the v3 release! 

## Do you want to request a feature or report a bug?

Feature Request

## What is the current behavior?

In v2 we were using the `Customized` component for an annotation system on some of our charts. As part of that, we were using a few of the extra props that it included. Namely:

```tsx
type AnnotationProps = {
    // ...
	formattedGraphicalItems?: {
		props: {
			points: GraphicalItemPoint[];
		};
	}[];
	chartX?: number;
	isTooltipActive?: boolean;
	offset?: {
		top: number;
		left: number;
		height: number;
		width: number;
	};
};
```

- `formattedGraphicalItems`: We use this to convert between x coordinates on the chart and timestamp values for the annotations.
- `chartX`: We use this when creating a new annotation to calculate where should we put it and what is its timestamp based on the user's mouse position.
- `isTooltipActive`: Simply for whether we should render the component that allows you to click and create an annotation. If there is no tooltip, there is no timestamp, so one can't be created.
- `offset`: We use this for the shape of the rectangle that handles the click-to-create logic, as well as the heights and positions of the annotations themselves

I saw in the release notes that we could request items to be added as hooks, would it be possible to add these? Or maybe there is some other way of doing the calculations? 🙂 

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

`v3`

ckifer:
Thanks for providing your use-cases!

> maybe there is some other way of doing the calculations

I think this is what we want to go after, especially with `formattedGraphicalItems` as it no longer exists at all.
You should be able to get `height` and `width` with `useChartHeight` and `useChartWidth` which are currently exported
https://main--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-hooks-dimensions--docs

Will get back to you on the rest of these

cc @PavelVanecek

calthejuggler:
Hey @ckifer, thanks for the response 🙂 I will give `useChartHeight/Width` a go and see where it gets us.

For more context, we are using the `formattedGraphicalItems` as an array of the points including the x axis's dataKey value (`date` in our case) alongside each point's x coordinate. We then find the nearest point before and the next point after the annotation's timestamp which we use to rescale the timestamp and find the x coordinate that it should be drawn at.

```tsx
	const annotationTimestamp = new Date(annotation.timestamp).getTime();
[...]

Issue #1678, opened 2019-03-21
Add support for Annotations component


Eric24:
## Do you want to request a feature or report a bug?
Feature

## What is the expected behavior?
There are many charting applications that require annotations. This is not significantly different than a regular data series, except that it is usually sparse on the X-axis, and in addition to having a Y-value, it also has an associated text property. The codepen example below (from a different component) shows a version of how this might work. I could see config options for controlling whether the annotation text shows on hover or on click, "dot" symbol selection and customization, etc.

https://codepen.io/susielu/pen/wrNRRK?editors=0010#0

Maybe this component could be integrated to ReCharts? https://react-annotation.susielu.com/

PS - If anyone with solid React and ReCharts experience is interested in developing this feature, I'm willing to fund the project.

ivan-ivan:
Was this feature added? I see that the issue is closed, but couldn't find the Annotation component in a codebase.

h0jeZvgoxFepBQ2C:
Same question here again?

jaekunchoi:
Any update here? I need this feature too

h0jeZvgoxFepBQ2C:
@Eric24 Could you reopen this issue? Looks like there is no annotation-support? Or did you get this working?

matt-takumi:
I would love something similar, such as this https://www.fusioncharts.com/fusiontime/examples/annotating-single-data-point?framework=javascript

ugoi:
Any support for this in 2025?

ckifer:
I didn't know this issue existed. Will look at reopening

ckifer:
I would rather give the opportunity for the user to be able to do this themselves with coordinates from recharts than integrate this into the library. 

I would imagine that is already possible - would just need to do some experimentation

ugoi:
I would like to do something like this: https://youtu.be/ssNBCeWKW8k?t=41
How would I implement this with coordinates from recharts?

PavelVanecek:
Is `react-annotation` the best library for this? Looks like it's not maintained, and only supports React 16.

PavelVanecek:
[...]

PR #6942, opened 2026-01-31
[Feature] getRelativeCoordinate

PavelVanecek:
## Description

Exports new function: `getRelativeCoordinate`. This function computes relative element coordinates from mouse or touch event.

Changes:

### Renamed `getChartPointer` to `getRelativeCoordinate`

I figured that this function does not actually return chart-relative coordinates, it returns coordinates relative to whatever was the `currentTarget`. Which internally we always pass the chart element, but once we export and make it public API that may not be the case. So I renamed the function and renamed properties in the return value as well.

### Added support for SVG elements

Internally we always pass HTML elements but Recharts is a mix of HTML and SVG so let's make it easy so that as a dev I don't need to think which context I am in.

### Added support for touch and multi-touch

It would be nice if we could support both mouse and touch wouldn't it. So now we do. Touch events return an array because the user can use multiple pointing devices (aka fingers) at the same time.

### Fixed the currentTarget bug

There was a problem where browsers only keep reference to the `event.currentTarget` in the main event loop. This is usually no big deal unless you happen to process these events asynchronously which is exactly what we do. So I added a proxy so that we can cache the latest `currentTarget` and return it if needed. `getRelativeCoordinate` does need it indeed.

### Fixed Legend re-render bug

Legend was re-rendering and resetting its own size which usually wasn't noticeable because of batched actions. But it did interrupt touch gestured so I added React.memo which fixed it.

Also added VR test and website example.

## Related Issue

https://github.com/recharts/recharts/issues/6021

https://github.com/recharts/recharts/issues/1678

https://recharts.github.io/recharts/pr-6942/www/en-US/api/getRelativeCoordinate/

Desktop:

https://github.com/user-attachments/assets/d33b1739-f809-432d-ab7c-ffaa53314efc

Android multi-touch:

https://github.com/user-attachments/assets/fd27be7c-0b54-4116-b91b-2eb44ab19f18

iOS multi-touch:

https://github.com/user-attachments/assets/a0c36a9f-9f14-40eb-abbb-a08e39490e50

## Types of changes

## Checklist:


PavelVanecek:
https://recharts.github.io/recharts/pr-6942/www/en-US/api/getRelativeCoordinate/

src/util/types.ts
```diff
@@ -1323,25 +1323,56 @@ export type RangeObj = PolarViewBoxRequired & {
...
+ * Recharts accepts mouse events from both HTML and SVG elements.
+ */
+export type MousePointer = HTMLMousePointer | SVGMousePointer;
+export type TouchPointer = HTMLTouchPointer | SVGTouchPointer;
+
+/**
+ * Coordinates relative to the top-left corner of the active element.
+ * Also include scale which means that element that's scaled will return the same coordinates as element that's not scaled.
  */
-export interface ChartPointer {
```
ckifer, line 1342: Are people using this? We don't know

PavelVanecek: It was never exported from index.

PavelVanecek: It was never exported from index.

ckifer: Ah ok I was looking at the file name wrong lol

PR #6956, opened 2026-02-04
[animations] auto disable primitives animations based on user system preferences

cloud-walker:
## Description

While working on fixing the tests for another PR, We suspected that the animations in the charts might interfere, making them flaky.

So I thought it would be cool to be able to disable them for testing sake, but they are on by default.

But [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-reduced-motion), media query exists, so this PR aims to do a combo: auto disable recharts animations if the media query triggers, so it would impact library consumers also, and being able to disable the animations from playwright by just simulating the user preference.

## Related Issue

#6955 

## Motivation and Context

Besides the testing purpose, its a good UX practise to let the end user decide if they want to see animations or not.

## How Has This Been Tested?

Manually, and with unit tests on usePrefersReducedMotion hook

## Types of changes

Its a potential "breaking change" for consumer that has not `isAnimationActive` explicitly set, and for end users (hopefully they would like that) that has the system preference enabled.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Animations now respect the user's system "prefers-reduced-motion" setting for improved accessibility.

* **Refactor**
  * Tooltip bounding box rewritten as a memoized functional component; related public prop types updated.

* **Documentation**
  * Clarified "auto" animation behavior to note SSR disabling and reduced-motion adherence.

* **Tests**
  * Added test setup and unit tests covering reduced-motion detection and behavior.

### DIFF
--- a/test/component/TooltipBoundingBox.spec.tsx
+++ b/test/component/TooltipBoundingBox.spec.tsx
@@ -53,11 +53,18 @@ describe('TooltipBoundingBox', () => {
     expect(screen.getByText('Hello world!')).not.toBeVisible();
   });
 
-  it('should hide children when dismissed using Escape key', () => {
-    render(<TooltipBoundingBox {...defaultProps} active={false} />);
-    userEvent.keyboard('{Escape}');
-    expect(screen.getByText('Hello world!')).toBeInTheDocument();
-    expect(screen.getByText('Hello world!')).not.toBeVisible();
+  it('should hide children when dismissed using Escape key', async () => {
+    render(<TooltipBoundingBox {...defaultProps} />);
+    const user = userEvent.setup({
+      advanceTimers: vi.advanceTimersByTime.bind(vi),
+    });
+    const element = screen.getByText('Hello world!');
+
+    expect(element).toBeVisible();
+
+    await user.keyboard('{Escape}');
+    expect(element).toBeInTheDocument();
+    expect(element).not.toBeVisible();
   });
 
   describe('offset prop', () => {


.github/workflows/ci.yml
cloud-walker, line 1: To restore before merging!

PavelVanecek:
I would like an undefined guard to prevent this problem in jsdom env: https://github.com/recharts/recharts-integ/pull/90

src/util/usePrefersReducedMotion.ts
```diff
@@ -0,0 +1,32 @@
...
+    };
+    mediaQuery.addEventListener('change', handleChange);
+    // eslint-disable-next-line consistent-return
+    return () => {
+      mediaQuery.removeEventListener('change', handleChange);
+    };
+  }, []);
+
+  return prefersReducedMotion;
+}
```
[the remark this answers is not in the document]
cloud-walker: in react useEffect cannot run on SSR environments, so its useless to check for Global.isSsr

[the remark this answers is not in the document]
cloud-walker: I pushed the other suggestions, but I cannot remove the eslint-disable-next-line consistent-return comment, I think that in this case is silly to have a consistent return

cloud-walker:
@PavelVanecek I think its ready for a review

src/component/TooltipBoundingBox.tsx
```diff
@@ -25,126 +26,108 @@ export type TooltipBoundingBoxProps = {
...
-
-  componentWillUnmount() {
-    document.removeEventListener('keydown', this.handleKeyDown);
-  }
-
-  componentDidUpdate() {
-    if (!this.state.dismissed) {
-      return;
-    }
+  React.useEffect(() => {
```
ckifer, line 58: Nit: why so much formatting going on here?

cloud-walker: I converted this from class to function component, I guess is for that

PavelVanecek:
I wonder if this preference is also connected to battery saver mode on smartphones.

PR #6964, opened 2026-02-05
Bump to vitest4 - test still failing

PavelVanecek:
## Description

Many tests fail with timeout, not sure why.

## Summary by CodeRabbit

* **Chores**
  * Upgraded testing tooling to the latest major version and added browser-playwright integration; test config and helpers adjusted for compatibility.
  * TypeScript base libs updated.

* **Bug Fixes / Behavior**
  * Developer tools, dev-mode logging, and immutability runtime checks are now disabled by default, reducing dev-only runtime validations.


ckifer:
what are we doing with this PR? Just using it to investigate the upgrade?

PavelVanecek:
Reproduction so that I could report to vitest team.

Issue #1997, opened 2020-01-21
Ability to set the animation "key" in bar chart for more intuitive animations


minikomi:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Adding values between other values results in a confusing animation.

eg: 

- Starting values: `[2,3,4]`
- New Values: `[2,10,3,4]` (`10` is the new value here)

In this case:
- the value 3 will grow to `10`
- the value 4 will shrink to `3`
- A new value will appear on the right as `4`

It would be more "sensible" for:
- a new value to appear between 2 and 3
- accordingly:
  - 3 to move to the right
  - 4 to move to the right

### What does the proposed API look like?
`<Bar animationId={animationIdFunction} />`

Where `animationIdFunction`:

- takes `(value, index)`, and returns a unique key. 
- default is `index`

jessecoleman:
Yeah, I'd love this as well. 

https://user-images.githubusercontent.com/5025783/131439359-5a09acc9-c060-46d6-8dc0-9e2093f82c8c.mp4

See this example of the entering bar incorrectly transitioning other bars

hoshmn:
Any update on this? Would be really nice to have the actual bar being added animate in. 

As far as I can tell, assuming you have a fixed order for the bars in your chart, you can currently choose between three suboptimal insertion behaviors: setting the <Bar> `key` value to
- the value of the data option, which causes the newly inserted bar _and all bars after it_ to animate in
- the index of the data option, which causes the new bar to appear abruptly and the _last_ bar to animate in (even though it was already there)
- Math.random(), which causes all bars to animate in after any change (insertion or deletion)

[Demo](https://share.zight.com/7KuwxBlX) of these three behaviors in order, [codesandbox](https://gf6w2q.csb.app/) to repro

ckifer:
Only update I can make is that we're working to get 3.0 out. 3.0 will (hopefully) make recharts easier to contribute to for others as it isn't feasible for one or two people to support so much in such a widely used library.

This won't be touched until at least after that. And since this would be a breaking change it may come in 4.0.

hoshmn:
To anyone with the same issue, I've come up with a hacky CSS-based solution that suits the needs for our app. Essentially, allow the bar order to change as bars are removed and added back. When you add a new bar, it will always be the last item in the data array and so will be the bar to animate in. 

[...]

Issue #2575, opened 2021-06-22
Animate Area Chart from Bottom to Top


LandonSchropp:
- [ ] I have searched the [issues](https://github.com/recharts/recharts/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
This would look a little nicer than the current animation. It would make the animations similar to the [BarChart animations](https://recharts.org/en-US/api/BarChart).

*Note: This was previously covered in #559. That issue was closed, but no solution was given and it doesn't seem to have been implemented.)

### What does the proposed API look like?
I would like an `animationDirection` property to be added to `<Area>`

```
<Area animationDirection="vertical" />
```

NickR49:
My AreaChart animates from top to bottom and I don't know why.
The Recharts example animate from left to right and I would prefer this.
Any tips as to how to change the animation direction?

LandonSchropp:
Ha ha, I wish. It seems we have the opposite problem.

PavelVanecek:
This will be possible in 3.9, here is a preview: https://recharts.github.io/recharts/pr-7215/www/en-US/examples/RangeAreaChartCustomAnimation/

PR #6973, opened 2026-02-07
feat: add animationMatchBy prop for identity-based animation

### LINKED ISSUE #1961: Wrong nimation interpolation

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/recharts-issue-template-x3vuf)

### Steps to reproduce
Press 'add data' and observe point "page C"

### What is expected?
Shape of parts of curve is not changed during animation, only the curve as a whole affine-transformed 

Line segments dancing, curve is animated as if data is added from left side

| Environment | Info |
|---|---|
| Recharts | v2.0.0-beta.1 |
| React | 16.12.0 |
| System | Arch |
| Browser | chrome / firefox |

---
This bug causes strange animation when updating plots with large data https://imgur.com/mRpJbv9

robjampar:
## Description

  Adds an `animationMatchBy` prop to `Line` and `Area` that controls how points are matched
  between data updates during animation. Instead of the default index-based stretching, points
  can be matched by identity (x-coordinate, data key, or custom function) so that existing
  points slide smoothly to new positions, new points animate in from the edge, and removed
  points slide off-screen.

  Also adds `animationClipOverflow` to clip points that slide beyond axis bounds, and
  `isAnimationActive` to `ReferenceLine` and `ReferenceArea` for smooth position transitions.

  ## Related Issue

  Closes #1961

https://github.com/recharts/recharts/issues/1997

https://github.com/recharts/recharts/issues/2575

  ## Motivation and Context

  When datapoints are added/removed from a chart, the default index-based animation interpolates
  every point to the wrong position — the line "dances" instead of sliding. This has been a
  long-standing issue forcing users to disable animation entirely for time-series data.
  `animationMatchBy` solves this by matching points by identity rather than array index.

  ## Types of changes

  - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

  ## Checklist

  - [x] I have added a storybook story or VR test.

## Summary by CodeRabbit

* **New Features**
  * Area and Line charts gain a new customizable animation hook plus two built-in interpolation presets for smoother point matching.
* **Tests**
  * Large end-to-end test suite added covering matching strategies, interpolation, entry/exit behavior, and animation scenarios.
* **Documentation**
  * API docs updated to include the new animation utilities and usage guidance.
* **Examples**
  * New interactive story demonstrating matching strategies and animation behaviors.

### DIFF
--- a/omnidoc/verifyExamples.spec.ts
+++ b/omnidoc/verifyExamples.spec.ts
@@ -179,6 +179,9 @@ describe('Documentation Examples Coverage', () => {
     'useYAxisScale',
     'useXAxisTicks',
     'useYAxisTicks',
+    'AnimationStepFunction',
+    'indexMatchLinearInterpolation',
+    'xMatchLinearInterpolation',
   ];
 
   describe.each(allExports.filter(name => !exportsThatNeedExamples.includes(name)))('Export: %s', exportName => {
--- a/src/animation/pointMatching.ts
+++ b/src/animation/pointMatching.ts
@@ -1,0 +1,380 @@
+import { getValueByDataKey } from '../util/ChartUtils';
+import { interpolate } from '../util/DataUtils';
+import { DataKey } from '../util/types';
+
+export interface AnimatablePoint {
+  x: number | null;
+  y: number | null;
+  payload?: unknown;
+}
+
+/** 'index' preserves legacy behavior; 'x' matches by coordinate; a DataKey or function matches by identity. */
+export type PointMatchingStrategy<T extends AnimatablePoint = AnimatablePoint> =
+  | 'index'
+  | 'x'
+  | DataKey<T['payload']>
+  | ((point: T, index: number) => string | number | null);
+
+export interface MatchedPoint<T> {
+  current: T;
+  currentIndex: number;
+  previous: T | null;
+}
+
+function getPointKey<T extends AnimatablePoint>(
+  point: T,
+  index: number,
+  strategy: PointMatchingStrategy<T>,
+): string | number | null {
+  if (strategy === 'index') {
+    return index;
+  }
+
+  if (typeof strategy === 'function') {
+    return strategy(point, index);
+  }
+
+  // strategy is a DataKey - get value from payload
+  if (point.payload != null) {
+    const value = getValueByDataKey(point.payload, strategy);
+    if (value != null) {
+      if (typeof value === 'string' || typeof value === 'number') {
+        return value;
+      }
+      return JSON.stringify(value);
+    }
+  }
+
+  return null;
+}
+
+function buildPreviousPointsMap<T extends AnimatablePoint>(
+  previousPoints: ReadonlyArray<T>,
+  strategy: PointMatchingStrategy<T>,
+): Map<string | number, T> {
+  const map = new Map<string | number, T>();
+
+  for (let i = 0; i < previousPoints.length; i++) {
+    const point = previousPoints[i];
+    if (point == null) continue;
+    const key = getPointKey(point, i, strategy);
+    if (key != null) {
+      if (!map.has(key)) {
+        map.set(key, point);
+      }
+    }
+  }
+
+  return map;
+}
+
+export interface MatchResult<T> {
+  matched: ReadonlyArray<MatchedPoint<T>>;
+  removed: ReadonlyArray<T>;
+}
+
+/**
+ * Match current points to previous points using the given strategy.
+ * Used during animation to determine where each point should animate from.
+ */
+export function matchPointsByStrategy<T extends AnimatablePoint>(
+  currentPoints: ReadonlyArray<T>,
+  previousPoints: ReadonlyArray<T> | null,
+  strategy: PointMatchingStrategy<T>,
+): ReadonlyArray<MatchedPoint<T>> {
+  if (!previousPoints || previousPoints.length === 0) {
+    return currentPoints.map((current, currentIndex) => ({
+      current,
+      currentIndex,
+      previous: null,
+    }));
+  }
+
+  // For 'index' strategy, use the legacy stretching algorithm for backward compatibility.
+  if (strategy === 'index') {
+    const prevPointsDiffFactor = previousPoints.length / currentPoints.length;
+    return currentPoints.map((current, currentIndex): MatchedPoint<T> => {
+      const prevPointIndex = Math.floor(currentIndex * prevPointsDiffFactor);
+      const previous = previousPoints[prevPointIndex] ?? null;
+      return {
+        current,
+        currentIndex,
+        previous,
+      };
+    });
+  }
+
+  const previousMap = buildPreviousPointsMap(previousPoints, strategy);
+
+  return currentPoints.map((current, currentIndex): MatchedPoint<T> => {
+    const key = getPointKey(current, currentIndex, strategy);
+    const previous = key != null ? (previousMap.get(key) ?? null) : null;
+
+    return {
+      current,
+      currentIndex,
+      previous,
+    };
+  });
+}
+
+/** Like matchPointsByStrategy but also returns points that were removed between frames. */
+export function matchPointsWithRemovals<T extends AnimatablePoint>(
+  currentPoints: ReadonlyArray<T>,
+  previousPoints: ReadonlyArray<T> | null,
+  strategy: PointMatchingStrategy<T>,
+): MatchResult<T> {
+  const matched = matchPointsByStrategy(currentPoints, previousPoints, strategy);
+
+  if (!previousPoints || previousPoints.length === 0 || strategy === 'index') {
+    return { matched, removed: [] };
+  }
+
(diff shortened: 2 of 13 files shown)


PavelVanecek:
That's a lot of change for something that can be much simpler. First, why can't we match by index? What would the animation fix look like with the current API and without new props?

robjampar:
> That's a lot of change for something that can be much simpler. First, why can't we match by index? What would the animation fix look like with the current API and without new props?

Index matching maps by array position... so if data shifts from [A, B, C] to [B, C, D], index 0 maps A→B, index 1 maps B→C, etc and every point animates to the wrong place (which is the dancing in #1961)

So i think ithout a new prop the only option would be changing the default to match by x-coordinate instead of index, but I think that would break categorical charts where x positions come from array order not data values. 

I could be wrong here though, might be a simpler way to handle it that I'm not seeing.

PavelVanecek:
Let's see the storybook. I'm still hoping we can find some simpler solution.

Why not disable animations?

PavelVanecek:
https://github.com/user-attachments/assets/dc742cc9-16ea-4a51-a617-81763979d88c

Ah okay it's nicely visible in the storybook.

PavelVanecek:
Right and if I disable animation I get the same effect but, well, without the animation:

https://github.com/user-attachments/assets/8a1dfa29-2996-45ed-bedd-8fe22e85ff58

Right I see the gap

PavelVanecek:
This approach (matchBy='x') only works if you are changing the XAxis domain on every tick. If you want to keep the domain the same, you are back where you started:

https://github.com/user-attachments/assets/f0ba3a7e-9fc7-4a0b-b432-28598713fe5d

PavelVanecek:
Now hear me out @robjampar for a moment. The animation has a ton of orchestration around it but eventually it boils down to this part here https://github.com/recharts/recharts/blob/main/src/cartesian/Line.tsx#L683 where it calculates `stepData`. This part includes the selection by index.

What if we exposed this part as a prop? `<Line animationStepFunction={indexMatchLinearInterpolationFunction} />` and then people could write their own to do whatever matching they want. We could even export the most common cases ("legacy" Recharts animation, time series) to make things easier, and also allow people to fine-tune and even provide their own completely custom animations.

What do you think?

PavelVanecek:
Also, to keep things focused:

> Also adds animationClipOverflow to clip points that slide beyond axis bounds, 

This should be the default, let's treat it as a bugfix, merge in separate PR, doesn't need to be a prop

> isAnimationActive to ReferenceLine and ReferenceArea for smooth position transitions.

Separate PR please

robjampar:
Ok good idea - reworked this based on your suggestions:

replaced animationMatchBy with animationStepFunction — the step data computation is now exposed as a prop like you described. the default code path (no prop) is completely unchanged from main, and recharts exports two presets:

- indexMatchLinearInterpolation — the existing behaviour
- xMatchLinearInterpolation — time-series sliding windows

```
  import { xMatchLinearInterpolation } from 'recharts';
  <Line animationStepFunction={xMatchLinearInterpolation} />
```

also split out the other bits:

- removed isAnimationActive on ReferenceLine/ReferenceArea —that got pulled in from my fork but shouldn't have
- removed animationClipOverflow entirely — agree it should just be default behaviour, can do that as a bugfix PR

  the scope is much smaller now, what do you think?

omnidoc/verifyExamples.spec.ts
```diff
@@ -179,6 +179,9 @@ describe('Documentation Examples Coverage', () => {
     'useYAxisScale',
     'useXAxisTicks',
     'useYAxisTicks',
+    'AnimationStepFunction',
```
PavelVanecek, line 182: Let's add examples on the website, or even better write a whole new guide on how to work with these.

src/animation/pointMatching.ts
```diff
@@ -0,0 +1,380 @@
...
+import { interpolate } from '../util/DataUtils';
+import { DataKey } from '../util/types';
+
+export interface AnimatablePoint {
+  x: number | null;
+  y: number | null;
+  payload?: unknown;
+}
+
+/** 'index' preserves legacy behavior; 'x' matches by coordinate; a DataKey or function matches by identity. */
```
PavelVanecek, line 11: Nothing in this behaviour is "legacy", don't call it that or we spook people into thinking we're removing it.

src/animation/pointMatching.ts
```diff
@@ -0,0 +1,380 @@
...
+export type PointMatchingStrategy<T extends AnimatablePoint = AnimatablePoint> =
```
PavelVanecek, line 12: We can now simplify this a bit since we expose the interpolation function. The strategy structure can go away

src/animation/pointMatching.ts
```diff
@@ -0,0 +1,380 @@
...
+  if (!previousPoints || previousPoints.length === 0) {
+    return currentPoints.map((current, currentIndex) => ({
+      current,
+      currentIndex,
+      previous: null,
+    }));
+  }
+
+  // For 'index' strategy, use the legacy stretching algorithm for backward compatibility.
+  if (strategy === 'index') {
```
PavelVanecek, line 94: The "index" strategy is now implemented three times: once here, then in `indexMatchLinearInterpolation`, and then third time inside of `Area`. Let's keep `indexMatchLinearInterpolation`, make that a default prop, and remove the other two.

src/cartesian/Area.tsx
```diff
@@ -169,6 +171,22 @@ interface AreaProps<DataPointType = any, DataValueType = any>
...
+   * - `indexMatchLinearInterpolation`: legacy index-based stretching (the default behaviour)
+   * - `xMatchLinearInterpolation`: matches points by x-coordinate for time-series sliding windows
+   *
+   * If not provided, the default index-based animation is used.
+   *
+   * @example
+   * import { xMatchLinearInterpolation } from 'recharts';
+   * <Area animationStepFunction={xMatchLinearInterpolation} />
+   */
+  animationStepFunction?: AnimationStepFunction<AreaPointItem>;
```
PavelVanecek, line 189: This is the first time when we actually expose the `AreaPointItem` (and `LinePointItem` below too), isn't it. Let's also export the hooks that select these two, before people start hacking the animation to get access to the points directly, shall we. And add JSDoc to the type definition of these two to explain what is going on.

src/animation/pointMatching.ts
```diff
@@ -0,0 +1,380 @@
...
+    return entry;
+  });
+}
+
+/**
+ * Time-series animation step function.
+ * Matches points by x-coordinate so existing points slide smoothly,
+ * new points animate in from the edges, and removed points slide off-screen.
+ */
+export function xMatchLinearInterpolation<T extends AnimatablePoint>(
```
PavelVanecek, line 373: Can we use this method to fix https://github.com/recharts/recharts/issues/1997 too? If yes can you please add the same prop to Bar as well?

src/animation/pointMatching.ts
```diff
@@ -0,0 +1,380 @@
...
+/**
+ * Interpolate matched, new, and removed points for a single animation frame.
+ * Returns the full set of points to render at time t (0 = previous positions, 1 = current positions).
+ */
+export function interpolateMatchedPoints<T extends AnimatablePoint>(
+  matchedPoints: ReadonlyArray<MatchedPoint<T>>,
+  removedPoints: ReadonlyArray<T>,
+  t: number,
+  interpolateFn: (from: unknown, to: number | null, t: number) => number | null,
+  animateNewValues: boolean,
```
PavelVanecek, line 273: `animateNewValues` is always true, can we remove the param?

src/animation/pointMatching.ts
```diff
@@ -0,0 +1,380 @@
...
+
+/**
+ * Interpolate matched, new, and removed points for a single animation frame.
+ * Returns the full set of points to render at time t (0 = previous positions, 1 = current positions).
+ */
+export function interpolateMatchedPoints<T extends AnimatablePoint>(
+  matchedPoints: ReadonlyArray<MatchedPoint<T>>,
+  removedPoints: ReadonlyArray<T>,
+  t: number,
+  interpolateFn: (from: unknown, to: number | null, t: number) => number | null,
```
PavelVanecek, line 272: Why not hardcode `interpolate` from `DataUtils`? It's doing the same thing

PavelVanecek (commented):
Okay I think it's much better like this. @ckifer wdyt?

Can we please polish this a little bit before merging @robjampar ? Extra docs and examples would be nice too. Please prefer the website instead of storybooks because https://github.com/recharts/recharts/issues/6668

ckifer:
yeah this is fine to do. Agreed with all of @PavelVanecek 's comments though - this needs cleaned up

PavelVanecek:
I'm going to decline this one @robjampar and merge https://github.com/recharts/recharts/pull/7215 instead. I'll still include you as contributor

PR #6976, opened 2026-02-07
feat: accept data-* attributes (including data-testid) on components

### LINKED ISSUE #3253: Accept 'data-testid' as a prop for every component to write better tests.

Harikrushn9118:
Closes #3253

## Summary
This PR enables passing `data-*` attributes (such as `data-testid`) to Recharts components, making it easier to write tests with React Testing Library.

### Core Changes
- **Added `getDataProps` utility function** in `src/util/svgPropertiesNoEvents.ts` to extract `data-*` attributes from props objects
- **Updated `ResponsiveContainer`** to forward `data-*` attributes to its container div

### Already Supported
- Chart components (BarChart, LineChart, etc.) already support `data-*` attributes via `svgPropertiesAndEvents`, which includes `data-*` in its filter

## Usage Example

```tsx
// ResponsiveContainer with data-testid
<ResponsiveContainer data-testid="my-container">
  <LineChart data={data} data-testid="my-chart">
    <Line dataKey="value" />
  </LineChart>
</ResponsiveContainer>

// Testing with React Testing Library
const container = screen.getByTestId('my-container');
expect(container).toBeInTheDocument();
```

## Tests Added
- Tests for `getDataProps` utility function
- Tests for `ResponsiveContainer` `data-*` forwarding
- Tests verifying `BarChart` passes `data-*` to SVG element

## Related Discussion
Based on feedback from the issue and [PR #3783](https://github.com/recharts/recharts/pull/3783#pullrequestreview-1672699248), this implementation:
1. Uses a shared utility function (`getDataProps`) as suggested
2. Spreads `data-*` props onto base elements

## Summary by CodeRabbit

* **New Features**
  * Components now forward HTML data-* attributes (e.g., `data-testid`, `data-custom`, `data-analytics-id`) to their rendered DOM/SVG elements.

* **Tests**
  * Added and updated tests to verify data-* attributes are correctly forwarded from chart and container components.

### DIFF
--- a/src/component/ResponsiveContainer.tsx
+++ b/src/component/ResponsiveContainer.tsx
@@ -24,8 +24,16 @@ import {
 } from './responsiveContainerUtils';
 import { Percent, Size } from '../util/types';
 import { isPositiveNumber } from '../util/isWellBehavedNumber';
+import { svgPropertiesAndEvents } from '../util/svgPropertiesAndEvents';
 
 export interface Props {
+  /**
+   * HTML data-* attributes (e.g. data-testid) forwarded to the container div.
+   *
+   * Note: These attributes are only applied when the container renders a wrapper element (i.e. when dimensions are not fixed).
+   * If width and height are fixed numbers, or if nested inside another ResponsiveContainer, the wrapper is optimized away and these attributes will not be rendered.
+   */
+  [key: `data-${string}`]: string | number | boolean | undefined;
   /**
    * width / height. If specified, the height will be calculated by width / aspect.
    */
@@ -135,6 +143,7 @@ const SizeDetectorContainer = forwardRef<HTMLDivElement | null, Props>(
       className,
       onResize,
       style = {},
+      ...rest
     }: Props,
     ref,
   ) => {
@@ -230,6 +239,7 @@ const SizeDetectorContainer = forwardRef<HTMLDivElement | null, Props>(
         className={clsx('recharts-responsive-container', className)}
         style={{ ...style, width, height, minWidth, minHeight, maxHeight }}
         ref={containerRef}
+        {...svgPropertiesAndEvents(rest)}
       >
         <div style={getInnerDivStyle({ width, height })}>
           <ResponsiveContainerContextProvider width={calculatedWidth} height={calculatedHeight}>
--- a/test/component/ResponsiveContainer.spec.tsx
+++ b/test/component/ResponsiveContainer.spec.tsx
@@ -502,4 +502,40 @@ describe('<ResponsiveContainer />', () => {
     );
     expect(container.querySelector('.recharts-responsive-container')).toHaveClass('my-custom-class');
   });
+
+  it('should pass data-testid attribute to the container div', () => {
+    render(
+      <ResponsiveContainer data-testid="my-responsive-container">
+        <DimensionSpy />
+      </ResponsiveContainer>,
+    );
+
+    expect(screen.getByTestId('my-responsive-container')).toBeInTheDocument();
+    expect(screen.getByTestId('my-responsive-container')).toHaveClass('recharts-responsive-container');
+  });
+
+  it('should pass multiple data-* attributes to the container div', () => {
+    const { container } = render(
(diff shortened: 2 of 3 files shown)


src/util/svgPropertiesNoEvents.ts
```diff
@@ -335,6 +335,24 @@ export function isDataAttribute(key: PropertyKey): key is DataAttributeKeyType {
   return typeof key === 'string' && key.startsWith('data-');
 }
 
+/**
+ * Extracts all data-* attributes from an object.
```
PavelVanecek: Why another helper? Why not call svgPropertiesAndEvents?

src/component/ResponsiveContainer.tsx
```diff
@@ -289,11 +294,20 @@ export const ResponsiveContainer = forwardRef<HTMLDivElement, Props>((props, ref
...
+    const dataProps = getDataProps(props as unknown as Record<string, unknown>);
+    const hasDataProps = Object.keys(dataProps).length > 0;
+
+    const content = (
       <ResponsiveContainerContextProvider width={calculatedWidth} height={calculatedHeight}>
         {props.children}
       </ResponsiveContainerContextProvider>
     );
+
+    // Wrap in div only if data-* attributes are present
```
PavelVanecek: If you do this then your application will render differently in test and in prod, that's not a great idea.

Did you know you don't need responsive container at all? It's now built in to all charts. It's only here for backwards compatibility.

test/component/ResponsiveContainer.spec.tsx
```diff
@@ -384,16 +384,6 @@ describe('<ResponsiveContainer />', () => {
     expect(onResize2).toHaveBeenCalledWith(100, 100);
   });
 
-  it('should render children straight, without any detector divs, when width and height are both fixed numbers', () => {
```
PavelVanecek, line 387: Not a cost I would like to pay for this.

Why do you need the data attribute here anyway? Can you render a wrapper div on top of the chart? We're only talking about the ResponsiveContainer. All the internals allow data attributes already.

Harikrushn9118: @PavelVanecek  I’m going to revert the forced wrapper and instead only forward the data-* attributes through SizeDetectorContainer. This way, the attributes work for fluid charts where a wrapper already exists, but the optimized path for fixed dimensions stays exactly as it was. I'll update the docs to match as well.

ckifer: why was this test removed? Can it be added back?

Harikrushn9118: @ckifer Sorry about that — the test was removed unintentionally during my refactoring. I've added it back now. The fixed-dimensions optimization is untouched, and the test coexists with the new `data-*` attribute tests.

Issue #6957, opened 2026-02-04
Scatter tooltip text color is always black instead of legend/series color


uthamarajchinnasamy-cloud:
## What is the current behavior?
When using a ScatterChart in Recharts, the tooltip text color is always rendered in black. This does not respect the corresponding scatter/legend color, unlike other chart types where tooltip text matches the series color.

Screenshots or videos are welcome.

<img width="1109" height="607" alt="Image" src="https://github.com/user-attachments/assets/54002451-6277-47c4-9be4-1178e5bfc58f" />

## What is the expected behavior?
Tooltip text color should match the scatter/legend color for each series.

Actual behavior
Tooltip text is always black, regardless of the scatter or legend color

Screenshots or videos are welcome.

## Please provide a demo of the problem in a sandbox

Here is how: https://github.com/recharts/recharts/wiki/How-to-share-a-demo

Please try to keep the demo as small as possible while still showing the problem.

luke-waymark-addinsight:
I've noticed when using a custom tooltip, the color/fill prop of the payload is undefined despite the scatter chart being colored - this is populated for other chart types.

PR #6977, opened 2026-02-09
fix: sync pie tooltip/legend color with per-sector fill

### LINKED ISSUE #6966: Tooltip/Legend colors don't match pie chart when using custom shape

WIth the deprecation of the `Cell` component in v3.7.0, I followed the guide [here](https://recharts.github.io/en-US/guide/cell/) to migrate to using the `shape` prop. However, after doing so, I found that the `Legend` and `Tooltip` components no longer display the expected colors.

I have a workaround, which I'll post below, but the issue seems to be that the payload the `Tooltip` and `Legend` content renderers receive is empty. This does not occur when using the `Cell` component.

Image

Image

https://stackblitz.com/edit/mgsuqetn?file=src%2FExample.tsx

Affected version: v3.7.0 (using the `shape` prop)

## Current workaround

Below is a snippet of code from the project I'm working on. For `Legend`, I'm mapping the chart data to the payload shape that `DefaultLegendContent` expects and rendering it manually. For `Tooltip`, I'm using `activeIndex` to find the data I'm interested in.

```tsx
import {
  Tooltip,
  Legend,
  PieChart,
  Pie,
  Sector,
  DefaultLegendContent,
} from 'recharts'

<PieChart width={320} height={210}>
  <Pie
    cx="45%"
    cy="50%"
    innerRadius={65}
    outerRadius={100}
    paddingAngle={2}
    dataKey="value"
    data={chartData}
    shape={(props) => (
      <Sector
        {...props}
        className="chart-cell"
        fill={colors[props.name] ?? '#8884d8'}
      />
    )}
  />
  <Legend
    layout="vertical"
    iconType="square"
    iconSize={6}
    formatter={(value) => (
      <span className="ml-xs body-sm text-secondary capitalize">{value}</span>
    )}
    align="right"
    verticalAlign="middle"
    itemSorter={(item) => {
      const index = chartData.findIndex((entry) => entry.name === item.value)
      return index !== -1 ? index : Infinity
    }}
    content={(props) => {
      const payload = chartData.map((entry) => ({
        color: colors[entry.name] ?? '#8884d8',
        payload: entry,
        type: props.iconType,
        value: entry.name,
      }))
      return <DefaultLegendContent {...props} payload={payload} />
    }}
  />
  <Tooltip
    content={(props) => {
      const activeIndex = props.activeIndex ? Number(props.activeIndex) : -1
      const entry = activeIndex !== -1 ? chartData[activeIndex] : null
      return (
        <div
          className="chart-tooltip"
          style={{
            visibility: entry ? 'visible' : 'hidden',
          }}
        >
          <div className="flex items-center">
            <div
              className="w-sm h-sm rounded-full mr-xs"
              style={{
                backgroundColor: colors[entry.name] ?? '#8884d8',
              }}
            />
            <span className="capitalize">{entry?.name}</span>
          </div>
          <span className="font-semibold">{entry?.value.toLocaleString()}</span>
        </div>
      )
    }}
  />
</PieChart>
```

2YH02:
## Description

Fixes Pie tooltip/legend color sync when sectors are customized via `shape`, by preserving per-sector color information in Pie tooltip payload flow.

- Adds per-sector `color`/`fill` to Pie tooltip payload entries.
- Marks Pie tooltip settings with `chartType: 'pie'`.
- Preserves `color`/`fill` for Pie payload items in tooltip payload combiner (keeps existing behavior for non-Pie chart types).
- Adds/updates tests and a Storybook example for verification.

## Related Issue

Fixes #6966 

Fixes https://github.com/recharts/recharts/issues/6957

## Motivation and Context

When Pie sectors are customized with `shape`, visual sector colors can diverge from tooltip/legend color metadata.
This change ensures Pie payload carries per-sector color data and that tooltip payload combining does not discard it for Pie, improving color consistency.

## How Has This Been Tested?

Ran targeted tests locally:

- `npm run test -- test/polar/Pie/Pie.spec.tsx`
- `npm run test -- test/state/selectors/pieSelectors.spec.tsx`
- `npm run test -- test/component/Tooltip/defaultIndex.spec.tsx`
- `npm run test -- test/component/Tooltip/itemSorter.spec.tsx`
- `npm run test -- test/state/selectors/selectors.spec.tsx`
- `npm run test -- test/component/Tooltip/tooltipEventType.spec.tsx`
- `npm run test -- test/component/Tooltip/Tooltip.payload.spec.tsx`
- `npm run check-types-storybook`
- `npm run lint`

Also added a Storybook example:
- `storybook/stories/Examples/Pie/PieColorSync.stories.tsx`

https://github.com/user-attachments/assets/ab8ca580-9155-4632-9e3c-13521fbd5f91

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Pie chart tooltips now show per-sector color/fill and honor the active sector’s fill for tooltip items.
  * New Storybook example demonstrating custom sector coloring synced with tooltips.

* **Improvements**
  * Tooltip entries preserve and propagate sector color/fill (with a gray fallback when unspecified) for consistent visuals.
  * Tooltip payload shape standardized to always include color and fill.

* **Tests**
  * Updated tests to assert color and fill are present in tooltip payloads.


src/state/selectors/combiners/combineTooltipPayload.ts
```diff
@@ -89,16 +89,19 @@ export const combineTooltipPayload = (
 
     if (Array.isArray(tooltipPayload)) {
       tooltipPayload.forEach(item => {
+        const preserveItemColor = settings?.chartType === 'pie';
```
PavelVanecek: Why only Pie? What happens if we allow this for all charts?

ckifer: +1

2YH02: I originally scoped this to Pie to minimize behavior changes, since the issue report was Pie-specific.
That said, preserving `color`/`fill` for all chart types may be better for consistency.

I tried the global version (preserving `item.color`/`item.fill` for all charts) and reran related tooltip/selector tests, and they all passed.

If you prefer, I can update this PR either way:

1. Keep Pie-only scope (minimal change)
2. Apply globally across chart types (consistency-first)

PavelVanecek: Let's try all charts. Unit tests will pass either way because we don't test colour in unit tests. VR tests will show the difference.

2YH02: Updated based on your feedback:
I removed the Pie-only check.

Now `combineTooltipPayload` keeps each item's `color`/`fill` for array payloads in all chart types.

I also updated the related tooltip test expectations.

2YH02: @PavelVanecek 
Follow-up update:
I kept item-level `color`/`fill` for array payloads, and added fallback behavior when item values are missing:

- `color: item.color ?? settings?.color`
- `fill: item.fill ?? settings?.fill`

This keeps the all-charts behavior while preserving stable existing defaults. (commit: [de86534](https://github.com/recharts/recharts/pull/6977/changes/de865348275378de36928603cd0847e7cfa3f3b4))

I reran VR tests, and diffs appear across multiple chart examples (not only Pie), which is consistent with applying this tooltip color/fill behavior across chart types.

test/polar/Pie/Pie.spec.tsx
```diff
@@ -217,7 +217,10 @@ describe('<Pie />', () => {
 
       const tooltipItem = container.querySelector('.recharts-tooltip-item');
       assertNotNull(tooltipItem);
-      expect((tooltipItem as HTMLElement).style.color).toBe('rgb(255, 0, 0)');
+      if (!(tooltipItem instanceof HTMLElement)) {
+        throw new Error(`Expected instance of HTMLElement, instead received: [${tooltipItem}]`);
+      }
+      expect(tooltipItem.style.color).toBe('rgb(255, 0, 0)');
```
PavelVanecek: `expect().toHaveStyle()` also works

test/state/selectors/pieSelectors.spec.tsx
```diff
@@ -73,6 +73,8 @@ describe('selectPieSectors', () => {
           stroke: '#fff',
           tooltipPayload: [
             {
+              color: expect.any(String),
```
PavelVanecek: I'm with the robot on this one. Let's have specific color here, instead of expect.any.

PavelVanecek:
The visual diff looks all reasonable except for the active shape Pie. It feels that this should show the active shape color, not the inactive shape. What do you think @2YH02 ?

image

2YH02:
> The visual diff looks all reasonable except for the active shape Pie. It feels that this should show the active shape color, not the inactive shape. What do you think @2YH02 ?
> 
> image


@PavelVanecek I agree. I reviewed the VR diffs and missed that part.
If you agree, I’ll update this so active Pie sectors use the active shape color for tooltip color matching.

2YH02:
Updated, @PavelVanecek 

Pie tooltip now uses `activeShape` fill, so tooltip color matches the active sector.
Added a test for this.
[250d945](https://github.com/recharts/recharts/pull/6977/commits/250d945f11b18715e8af608fd3d2774d5eba12a2)

src/state/selectors/combiners/combineTooltipPayload.ts
```diff
@@ -95,10 +95,11 @@ export const combineTooltipPayload = (
           name: item.name,
           // @ts-expect-error we're assuming that item has name and unit properties
           unit: item.unit,
-          // color and fill are erased to keep 100% the identical behaviour to recharts 2.x - but there's nothing stopping us from returning them here. It's technically a breaking change.
-          color: undefined,
-          // color and fill are erased to keep 100% the identical behaviour to recharts 2.x - but there's nothing stopping us from returning them here. It's technically a breaking change.
-          fill: undefined,
+          // Preserve item-level color/fill from graphical items.
+          // @ts-expect-error we're assuming that item has color property
```
PavelVanecek: Let's parse it out, without assumptions and without ts expect error

2YH02: I changed this to parse array items from unknown before reading name/unit/dataKey/payload/color/fill
[b49a59c](https://github.com/recharts/recharts/pull/6977/commits/b49a59ccf1899500c33762200ceddd9323a6eb2e)

PavelVanecek: The "as" keyword is doing exactly the same thing as ts-expect-error.

PavelVanecek:
Does this fix https://github.com/recharts/recharts/issues/6957 too?

src/state/selectors/combiners/combineTooltipPayload.ts
```diff
@@ -13,6 +13,65 @@ import { getTooltipEntry, getValueByDataKey } from '../../../util/ChartUtils';
...
+type TooltipPayloadItemLike = {
+  name: TooltipEntrySettings['name'];
+  unit: TooltipEntrySettings['unit'];
+  dataKey: DataKey<any> | undefined;
+  payload: unknown;
+  color: string | undefined;
+  fill: string | undefined;
+};
+
+function isRecord(value: unknown): value is Record<string, unknown> {
```
PavelVanecek: And, you're not going to believe this, but the `is` keyword _also_ disables the type system completely, same as @ts-expect-error and `as`. At least the runtime check is right this time.

PavelVanecek:
Let's update the snapshots and then we can merge.

src/state/selectors/combiners/combineTooltipPayload.ts
2YH02: Thanks for the feedback!
Updated in [045367f](https://github.com/recharts/recharts/pull/6977/commits/045367f66ba525cc54b35b4d0bcca36fcd3db184).

PavelVanecek (approved):

PavelVanecek:
https://63da8268a0da9970db6992aa-vqtbwaspqx.chromatic.com/?path=/story/api-chart-piechart--api this story shows inactive color still.

PavelVanecek:
Hm but that's a completely custom shape. Should we dig into that? Or should we assume that whoever got that far, can also modify the tooltip text color?

2YH02:
I agree. Since this is a fully custom shape case, I think it’s reasonable to leave tooltip color to user control for now. In practice, users who customize charts this much usually customize tooltip content/styles too. If we later decide this should be auto-synced as well, I can follow up with a separate PR.

PR #6986, opened 2026-02-10
Fix: error when onMouseEnter or onMouseLeave is undefined #6968

### LINKED ISSUE #6968: Line chart with onMouseEnter={undefined} or onMouseLeave={undefined} produces error

Image
Errors  with message 
`Uncaught TypeError: inputProps[key] is not a function`
happen when user hover the line 

No errors

[Here is how: https://github.com/recharts/recharts/wiki/How-to-share-a-demo

Please try to keep the demo as small as possible while still showing the problem.](https://stackblitz.com/edit/mgsuqetn-pb2ikn8r?file=src%2FExample.tsx,src%2Findex.tsx)

3.7.0

ramanverse:
Hey! While working with the Line component, I noticed it crashes if you pass `undefined` to `onMouseEnter` or `onMouseLeave`. It happens because `adaptEventHandlers` tries to wrap and call anything starting with 'on' without checking if it's actually a function.

This fix just adds a quick `typeof` check in `src/util/types.ts` to make sure we only wrap actual functions. This makes it safer for common React patterns where you might want to disable a handler by passing undefined.

I've also added a small test file in `test/util/adaptEventHandlers.spec.ts` just to be sure it doesn't break again.

Fixes #6968

### DIFF
--- a/src/util/types.ts
+++ b/src/util/types.ts
@@ -877,7 +877,7 @@ export const adaptEventHandlers = (
   const out: RecordString<(e: Event) => void> = {};
 
   Object.keys(inputProps).forEach(key => {
-    if (isEventKey(key)) {
+    if (isEventKey(key) && typeof inputProps[key] === 'function') {
       out[key] = newHandler || ((e: Event) => inputProps[key](inputProps, e));
     }
   });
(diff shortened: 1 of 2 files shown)


PavelVanecek:
We already have a PR merged for this issue: https://github.com/recharts/recharts/pull/6969

PR #7002, opened 2026-02-13
Fix(z-index): prevent elements from disappearing on rapid zIndex changes 

### LINKED ISSUE #6789: zIndex property doesn't like to be changed in quick succession

With slight adaptations to the Legend with Opacity example, if you hover the legend items in quick succession, curves will start disappearing from the chart (and also from the DOM). 

https://github.com/user-attachments/assets/0ae5d08c-dede-4a17-a85a-a300002e0791

It seems to happen when a dynamic zIndex is added, which leads me to believe it's a internal bug. 
The issue can be much more critical if more curves are added or more complex configuration is used, to the point where moving away from the plot does not cause the curve to be re-rendered.

*Bonus: animations seem broken like in [here](https://github.com/recharts/recharts/issues/6044)

Using reactive values for zIndex should not remove data curves from the plot entirely

https://stackblitz.com/edit/wzkj8wsc-otcdqw2a?file=src%2FExample.tsx

Haven't tested previous version., this is happening in 3.6

VIDHITTS:
## Description

When `zIndex` is updated dynamically (e.g., hovering legend items in quick succession), chart elements (curves, bars, etc.) would disappear from the DOM entirely. This was caused by [ZIndexLayer](cci:1://file:///Users/vidhitt.s/Desktop/recharts/src/zIndex/ZIndexLayer.tsx:44:0-110:1) returning `null` while waiting for a new portal element to register after a `zIndex` change.

This fix updates [ZIndexLayer](cci:1://file:///Users/vidhitt.s/Desktop/recharts/src/zIndex/ZIndexLayer.tsx:44:0-110:1) to render children **inline** (without a portal) as a fallback when the portal element is not yet available, ensuring content remains visible during the async portal registration cycle.

## Related Issue

Fixes #6789

## Motivation and Context

The portal-based `zIndex` system in Recharts v3 is inherently asynchronous:
1. A new `zIndex` is registered via `useLayoutEffect` → Redux dispatch.
2. `ZIndexPortals` re-renders to create the new portal DOM element.
3. [ZIndexLayer](cci:1://file:///Users/vidhitt.s/Desktop/recharts/src/zIndex/ZIndexLayer.tsx:44:0-110:1) picks up the element and renders via `createPortal`.

Between steps 1–3, [ZIndexLayer](cci:1://file:///Users/vidhitt.s/Desktop/recharts/src/zIndex/ZIndexLayer.tsx:44:0-110:1) was returning `null`, unmounting the content. With rapid `zIndex` changes, this caused persistent disappearance of chart elements.

The fix renders children inline during this gap, so the content is always visible — it just temporarily renders outside the portal layer until the portal is ready.

## How Has This Been Tested?

- **New regression test** ([test/zIndex/DynamicZIndex.spec.tsx](cci:7://file:///Users/vidhitt.s/Desktop/recharts/test/zIndex/DynamicZIndex.spec.tsx:0:0-0:0)): Simulates rapid `zIndex` updates on [Line](cci:1://file:///Users/vidhitt.s/Desktop/recharts/src/cartesian/Line.tsx:956:0-992:1) components and asserts that both lines remain in the DOM after each update.
- **Updated existing tests** ([test/cartesian/CartesianGrid.spec.tsx](cci:7://file:///Users/vidhitt.s/Desktop/recharts/test/cartesian/CartesianGrid.spec.tsx:0:0-0:0)): Adjusted spy call count expectations to account for the double render (inline → portal) introduced by the fix.
- **Full test suite**: All 13,900+ tests pass with zero regressions.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added a new interactive LineChart example to the documentation gallery. The example demonstrates legend-driven focus and dynamic layering: hovering over legend items highlights the corresponding data series, adjusts opacity, and brings it to the foreground to improve visibility and facilitate data exploration.

### DIFF
--- a/www/src/docs/exampleComponents/LineChart/DynamicZIndexLineChart.tsx
+++ b/www/src/docs/exampleComponents/LineChart/DynamicZIndexLineChart.tsx
@@ -1,0 +1,82 @@
+import { useState, useCallback } from 'react';
+import { LineChart, Line, XAxis, YAxis, CartesianGrid, Tooltip, Legend } from 'recharts';
+
+/**
+ * Reproduction of https://github.com/recharts/recharts/issues/6789
+ *
+ * Hover over the legend items in quick succession and observe that
+ * lines disappear from the chart (and from the DOM).
+ *
+ * The issue is caused by the ZIndexLayer returning null while waiting
+ * for a new portal element to register after a zIndex change.
+ */
+
+// #region Sample data
+const data = [
+  { name: 'Page A', uv: 4000, pv: 2400, amt: 2400 },
+  { name: 'Page B', uv: 3000, pv: 1398, amt: 2210 },
+  { name: 'Page C', uv: 2000, pv: 9800, amt: 2290 },
+  { name: 'Page D', uv: 2780, pv: 3908, amt: 2000 },
+  { name: 'Page E', uv: 1890, pv: 4800, amt: 2181 },
+  { name: 'Page F', uv: 2390, pv: 3800, amt: 2500 },
+  { name: 'Page G', uv: 3490, pv: 4300, amt: 2100 },
+];
+// #endregion
+
+const LINE_COLORS: Record<string, string> = {
+  uv: '#8884d8',
+  pv: '#82ca9d',
+  amt: '#ffc658',
+};
+
+const DEFAULT_OPACITY: Record<string, number> = {
+  uv: 1,
+  pv: 1,
+  amt: 1,
+};
+
+export default function Example() {
+  const [opacity, setOpacity] = useState(DEFAULT_OPACITY);
+  const [activeKey, setActiveKey] = useState<string | null>(null);
+
+  const handleMouseEnter = useCallback((o: { dataKey?: string | number | ((obj: unknown) => unknown) }) => {
+    const { dataKey } = o;
+    if (typeof dataKey === 'string') {
+      setOpacity(prev => ({ ...prev, [dataKey]: 0.5 }));
+      setActiveKey(dataKey);
+    }
+  }, []);
+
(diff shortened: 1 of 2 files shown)


PavelVanecek:
Are you sure this is the fix? Last time I checked, moving from DOM to Portal triggers unmount anyway. So we would end up with the exact same problem, but this time with a flash of content in the wrong z-index position.

VIDHITTS:
Hi @PavelVanecek , thanks for pointing  that out and you're absolutely right about the version of this fix. Rendering inline as a fallback would still trigger an unmount when transitioning from inline to portal, just trading "flash of nothing" for "flash at wrong position."

I've since reworked the approach. Instead of falling back to inline rendering, the fix now keeps the old portal alive during the transition:

When zIndex changes (e.g., 10 to 11), we register the new zIndex without immediately unregistering the old one
This keeps the old <g> element in the DOM, so we continue rendering via createPortal into the old portal so no unmount at all
Once the new portal <g> for zIndex 11 is ready, we switch to it and clean up the old registration
This way there's no DOM → Portal transition, it's always Portal → Portal. The content stays mounted the entire time , it just briefly lives in the old z-index layer until the new one is available.

Happy to discuss if you see any issues with this approach or if you'd prefer a different direction entirely!

PavelVanecek:
That might work. Let's also add a new website example so that we can test this in browser. Undo the current changes and push the example chart only so that we can compare before & after.

VIDHITTS:
I've pushed just the website example for now — it's a DynamicZIndexLineChart under the LineChart examples. You can hover over the legend items rapidly to see the lines disappear.

All fix-related changes have been reverted so this commit shows the "before" state. Once you've confirmed the example reproduces the issue well, I'll add the fix as a separate commit so we can compare before & after.

Let me know if the example needs any changes!

PR #7006, opened 2026-02-13
fix(z-index): prevent elements from disappearing during dynamic zIndex transitions

### LINKED ISSUE #6789: zIndex property doesn't like to be changed in quick succession

VIDHITTS:
## Description

When `zIndex` is changed dynamically (e.g., on legend hover), chart elements briefly disappear because [ZIndexLayer](cci:1://file:///Users/vidhitt.s/Desktop/recharts/src/zIndex/ZIndexLayer.tsx:43:0-150:1) returns `null` during the transition between portals.

**Root cause:** The original `useLayoutEffect` cleanup immediately calls `unregisterZIndexPortal(oldZIndex)`, which deletes the entry from Redux and removes the old `<g>` DOM node from the SVG. The new portal isn't ready yet (it requires a full render cycle through [AllZIndexPortals](cci:1://file:///Users/vidhitt.s/Desktop/recharts/src/zIndex/ZIndexPortal.tsx:21:0-43:1) → [ZIndexSvgPortal](cci:1://file:///Users/vidhitt.s/Desktop/recharts/src/zIndex/ZIndexPortal.tsx:6:0-19:1)), so `portalElement` is `undefined` and [ZIndexLayer](cci:1://file:///Users/vidhitt.s/Desktop/recharts/src/zIndex/ZIndexLayer.tsx:43:0-150:1) returns `null`.

**Fix:** Defer unregistration of old zIndex values until the new portal element becomes available. A `Set` ref tracks all registered zIndexes, and old entries are only cleaned up once the new portal is confirmed ready. During the transition, `createPortal` continues rendering into the cached old portal element.

Two effects with clear separation of concerns:
1. **Lifecycle effect** — registers new zIndex values and cleans up stale ones once the new portal element is available
2. **Unmount effect** — unregisters everything when the component is removed

## Related Issue

Fixes #6789

## Motivation and Context

Users who dynamically change `zIndex` (e.g., highlighting a line on legend hover) see chart elements vanish and not come back. This is a regression-free fix that addresses the gap in the portal lifecycle without changing any other component.

## How Has This Been Tested?

- Added [test/zIndex/DynamicZIndex.spec.tsx](cci:7://file:///Users/vidhitt.s/Desktop/recharts/test/zIndex/DynamicZIndex.spec.tsx:0:0-0:0) — renders two `<Line>` components, dynamically changes `zIndex` on one via `rerender`, and asserts both lines remain in the DOM
- All **13,908 existing tests pass** with zero regressions
- Can be verified visually using the `DynamicZIndexLineChart` website example merged in the previous PR — hover over legend items rapidly and observe lines no longer disappear

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Eliminated flicker and remount flashes when portal targets change; elements with dynamic zIndex now transition smoothly and remain stable during reordering.
  * Ensured stale layer registrations are fully cleaned up when portal rendering is toggled or components unmount, preventing leftover state.

* **Tests**
  * Added a test verifying rapid dynamic zIndex updates preserve rendered layers and do not remove or hide elements.

### DIFF
--- a/test/zIndex/DynamicZIndex.spec.tsx
+++ b/test/zIndex/DynamicZIndex.spec.tsx
@@ -41,7 +41,6 @@ describe('Dynamic zIndex updates', () => {
     const button = getByText('Increment zIndex');
 
     // Rapid zIndex updates should not cause lines to disappear
-    // Rapid zIndex updates should not cause lines to disappear
     act(() => {
       button.click();
     });


src/zIndex/ZIndexLayer.tsx
```diff
@@ -70,41 +69,83 @@ export function ZIndexLayer({ zIndex, children }: ZIndexLayerProps) {
...
+      return;
+    }
+
+    // Register the current zIndex (idempotent — skips if already registered)
+    if (!registeredZIndexesRef.current.has(zIndex)) {
+      dispatch(registerZIndexPortal({ zIndex }));
+      registeredZIndexesRef.current.add(zIndex);
     }
-    /*
-     * Because zIndexes are dynamic (meaning, we're not working with a predefined set of layers,
```
PavelVanecek, line 80: Please put this comment back

VIDHITTS: have put back the comment

test/zIndex/DynamicZIndex.spec.tsx
```diff
@@ -0,0 +1,65 @@
...
+    // Initial render check - wait for portal registration
+    await waitFor(() => {
+      const lines = container.querySelectorAll('.recharts-line');
+      expect(lines.length).toBe(2);
+    });
+
+    const button = getByText('Increment zIndex');
+
+    // Rapid zIndex updates should not cause lines to disappear
+    // Rapid zIndex updates should not cause lines to disappear
```
PavelVanecek, line 43:
```suggestion
```

PR #7009, opened 2026-02-14
Improve tick values by snapping to nice numbers {1, 2, 2.5, 5}

### LINKED ISSUE #7008: Improve default tick generation to produce round numbers
## Problem

Recharts' default tick generation frequently produces non-round, hard-to-read axis values like `$6,000`, `$28,000`, `$51,000` instead of clean values like `$0`, `$25,000`, `$50,000`, `$75,000`, `$100,000`.

This is a long-standing issue (#2140, #777, #1164) that affects virtually every chart that uses numeric axes with auto-generated ticks. The standard workaround is to compute explicit `ticks` arrays and pass them manually — but this defeats the purpose of a charting library handling it automatically.

## Root cause

`getTickValuesFixedDomain()` in `src/util/scale/getNiceTickValues.ts` doesn't produce aesthetically clean tick values when the domain is explicitly set. The main `getNiceTickValues()` does better but still struggles with `tickCount` respect.

D3's `d3-scale.nice()` solves this well by snapping to 1/2/5 multiples of powers of 10. The current Recharts algorithm attempts something similar but the results are often not "nice" in practice.

## Proposed fix

I'd be happy to submit a PR improving `getTickValuesFixedDomain()` to:
1. Always start ticks at 0 when the domain includes 0
2. Use clean 1/2/2.5/5/10 step multiples (similar to D3's approach)
3. Respect `tickCount` more accurately
4. Maintain backward compatibility with existing tests

Here's a minimal version of the algorithm that works well in practice:

```typescript
function niceTicks(dataMax: number, targetCount: number = 5): number[] {
  if (dataMax <= 0) return [0];
  const rawStep = dataMax / targetCount;
  const magnitude = Math.pow(10, Math.floor(Math.log10(rawStep)));
  const normalized = rawStep / magnitude;

  let niceStep: number;
  if (normalized <= 1) niceStep = 1 * magnitude;
  else if (normalized <= 2) niceStep = 2 * magnitude;
  else if (normalized <= 2.5) niceStep = 2.5 * magnitude;
  else if (normalized <= 5) niceStep = 5 * magnitude;
  else niceStep = 10 * magnitude;

  const niceMax = Math.ceil(dataMax / niceStep) * niceStep;
  const ticks: number[] = [];
  for (let v = 0; v <= niceMax; v += niceStep) {
    ticks.push(Math.round(v * 1e10) / 1e10);
  }
  return ticks;
}
```

## Before/After

**Before** (current Recharts default):
- X axis: `$6,000` `$28,000` `$51,000` `$75,000` `$100,000`  
- Y axis: `$6,500` `$13,000` `$19,500` `$26,000`

**After** (with nice ticks):
- X axis: `$0` `$20,000` `$40,000` `$60,000` `$80,000` `$100,000`
- Y axis: `$0` `$5,000` `$10,000` `$15,000` `$20,000` `$25,000` `$30,000`

## Context

- Related: #2140 (PavelVanecek mentioned fixing in Jul 2025), #777, #1164, #1052
- Aware of PR #6990 (rendered ticks in hooks) — happy to coordinate timing
- We've implemented this workaround in multiple projects and it would be great to have it built into Recharts

MaxGhenis:
## Summary

Replaces the `getFormatStep` rounding algorithm (0.05/0.1 granularity) with D3-style nice number snapping to `{1, 2, 2.5, 5}` multiples at each order of magnitude. This produces human-friendly tick values.

### Before → After examples

| Domain | Count | Before | After |
|--------|-------|--------|-------|
| `[0, 14]` | 5 | `[0, 4, 8, 12, 16]` | `[0, 5, 10, 15, 20]` |
| `[39.9, 42.5]` | 5 | `[39.9, 40.6, 41.3, 42, 42.7]` | `[39, 40, 41, 42, 43]` |
| `[-4.1, 0.6]` | 7 | `[-4.25, -3.4, -2.55, -1.7, -0.85, 0, 0.85]` | `[-5, -4, -3, -2, -1, 0, 1]` |
| `[0, 1.3e-5]` | 5 | `[0, 3.5e-6, 7e-6, 1.05e-5, 1.4e-5]` | `[0, 5e-6, 1e-5, 1.5e-5, 2e-5]` |

### How it works

The previous algorithm divided the step ratio by 0.05 or 0.1, ceiling-rounded, and multiplied back — producing steps like 3.5, 0.35, or 0.7 that don't feel "natural" to humans.

The new algorithm normalizes the rough step to `[1, 10)` and snaps to the nearest ceiling value in `{1, 2, 2.5, 5}`, then multiplies by the appropriate power of 10. The `correctionFactor` mechanism (used by `calculateStep` when it needs a larger step) now walks through this sequence instead of incrementing by 0.05/0.1.

### Additional fix

Renamed `getTickValuesFixedDomain.test.ts` → `.spec.ts` so vitest actually discovers and runs these tests. The vitest config uses `include: test/**/*.spec.ts?(x)`, so `.test.ts` files were silently skipped.

Fixes #7008

## Test plan

- [x] Updated all `getFormatStep` test expectations for nicer values
- [x] Updated `getNiceTickValues` test expectations (5 tests improved)
- [x] Updated `getTickValuesFixedDomain` test expectations (3 tests produce nicer but slightly fewer ticks, consistent with existing docstring: "may return less than tickCount ticks")
- [x] All 87 scale tests pass
- [x] No TypeScript errors

## Summary by CodeRabbit

* **New Features**
  * Added opt-in "nice ticks" toggle to axis components (optional per-axis); when enabled, tick intervals snap to rounder, human-friendly values.
* **Tests**
  * Updated and extended axis/tick tests to cover the new nice-ticks behavior and default-disabled expectations.

### DIFF
--- a/src/cartesian/XAxis.tsx
+++ b/src/cartesian/XAxis.tsx
@@ -186,7 +186,7 @@ interface XAxisProps<DataPointType = any, DataValueType = any> extends Omit<
   letterSpacing?: number | string;
   /**
    * When true, uses an improved tick step algorithm that snaps to nice numbers
-   * {1, 2, 2.5, 5} at each order of magnitude, producing human-friendly tick
+   * (1, 2, 2.5, 5) at each order of magnitude, producing human-friendly tick
    * intervals like 0, 5, 10, 15, 20 instead of 0, 4, 8, 12, 16.
    *
    * @defaultValue false
--- a/src/cartesian/YAxis.tsx
+++ b/src/cartesian/YAxis.tsx
@@ -201,7 +201,7 @@ interface YAxisProps<DataPointType = any, DataValueType = any> extends Omit<
   letterSpacing?: number | string;
   /**
    * When true, uses an improved tick step algorithm that snaps to nice numbers
-   * {1, 2, 2.5, 5} at each order of magnitude, producing human-friendly tick
+   * (1, 2, 2.5, 5) at each order of magnitude, producing human-friendly tick
    * intervals like 0, 5, 10, 15, 20 instead of 0, 4, 8, 12, 16.
    *
    * @defaultValue false
--- a/src/polar/PolarAngleAxis.tsx
+++ b/src/polar/PolarAngleAxis.tsx
@@ -166,7 +166,7 @@ export interface PolarAngleAxisProps<DataPointType = any, DataValueType = any>
   tickCount?: number;
   /**
    * When true, uses an improved tick step algorithm that snaps to nice numbers
-   * {1, 2, 2.5, 5} at each order of magnitude, producing human-friendly tick
+   * (1, 2, 2.5, 5) at each order of magnitude, producing human-friendly tick
    * intervals like 0, 5, 10, 15, 20 instead of 0, 4, 8, 12, 16.
    *
    * @defaultValue false
--- a/src/polar/PolarRadiusAxis.tsx
+++ b/src/polar/PolarRadiusAxis.tsx
@@ -169,7 +169,7 @@ export interface PolarRadiusAxisProps<DataPointType = any, DataValueType = any>
   allowDecimals?: boolean;
   /**
    * When true, uses an improved tick step algorithm that snaps to nice numbers
-   * {1, 2, 2.5, 5} at each order of magnitude, producing human-friendly tick
+   * (1, 2, 2.5, 5) at each order of magnitude, producing human-friendly tick
    * intervals like 0, 5, 10, 15, 20 instead of 0, 4, 8, 12, 16.
    *
    * @defaultValue false
--- a/src/state/cartesianAxisSlice.ts
+++ b/src/state/cartesianAxisSlice.ts
@@ -75,7 +75,7 @@ export type TicksSettings = {
   tick: TickProp<any>;
   /**
    * When true, uses an improved tick step algorithm that snaps to nice numbers
-   * {1, 2, 2.5, 5} at each order of magnitude, producing human-friendly tick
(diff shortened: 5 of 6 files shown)


PavelVanecek:
The new ticks look much better (judging from the unit tests alone). I wonder how many people are going to be mad that we're changing their charts. @MaxGhenis is there a way to opt-in to the previous algorithm for people who want to keep it that way? Or perhaps even should we keep the old behaviour the default, leave the new as opt-in and then swap them in the next major version?

MaxGhenis:
Great point on backward compatibility. I'll make the new algorithm opt-in for now — users can enable it via a prop, and we can swap the default in the next major version. Will update the PR.

MaxGhenis:
Updated — the new algorithm is now opt-in via a `useNiceTicks` boolean prop on XAxis/YAxis/PolarAngleAxis/PolarRadiusAxis (defaults to `false`). The original `getFormatStep` is unchanged; `getFormatStepNice` is added alongside it. 22 new tests cover the opt-in behavior. All 5477 tests pass.

ckifer:
@MaxGhenis useNiceTicks sounds like a hook. Can you just make it niceTicks?

MaxGhenis:
Done — renamed `useNiceTicks` to `niceTicks` across all source and test files (24 files). All tests pass.

ckifer:
@MaxGhenis PR is failing the build, looks like a formatting issue

MaxGhenis:
@ckifer Fixed — there were three issues: a prettier formatting violation, missing `niceTicks` in the polar axis default props (omnidoc validation), and curly braces in JSDoc that became invalid JSX expressions in the auto-generated API docs. All resolved, CI is green now.

Issue #7010, opened 2026-02-14
Feature request: native waterfall chart support


MaxGhenis:
## Summary

Recharts lacks native waterfall chart support. This is a [frequently requested](https://github.com/recharts/recharts/issues/2267) chart type used in financial reporting, budgetary analysis, and many other domains.

### 1. Stacked transparent bar (common advice)
```tsx
<Bar dataKey="invisible" stackId="a" fill="transparent" />
<Bar dataKey="visible" stackId="a" fill="#00897B" />
```
**Problem:** Breaks with negative values. When `barBottom` is negative, the stacking logic places bars incorrectly — all bars start from 0 instead of floating at the correct position.

### 2. Range bar via array dataKey (our workaround)
```tsx
function waterfallRange(d) {
  return [d.barBottom, d.barTop];
}
<Bar dataKey={waterfallRange} />
```
This works correctly but requires:
- Pre-computing `barBottom`/`barTop` for every bar (running total logic)
- Manual domain computation that includes all bar positions
- The `dataKey` type doesn't accept functions returning arrays cleanly (needs `as any`)
- Labels/tooltips need custom components since the value is an array

## Proposed feature

A first-class `<WaterfallBar>` component (or a `waterfall` prop on `<Bar>`) that:

1. **Automatically computes bar positions** from values — positive bars stack up from the running total, negative bars hang down
2. **Supports "total" bars** that anchor to 0 (common in waterfall charts)
3. **Handles the Y-axis domain** automatically to include all bar positions
4. **Works with existing Recharts features** — tooltips, labels, Cell coloring, animations

### Minimal API sketch
```tsx
<BarChart data={data}>
  <Bar dataKey="value" waterfall={{ totalKey: 'isTotal' }}>
    {data.map((d, i) => (
      <Cell key={i} fill={d.value >= 0 ? 'teal' : 'gray'} />
    ))}
  </Bar>
</BarChart>
```

Or as a dedicated component:
```tsx
<WaterfallChart data={data} dataKey="value" totalKey="isTotal" />
```

## Reference implementation

We built a working waterfall chart component for [PolicyEngine](https://policyengine.org) using the range-bar workaround:
- **Utils:** [`waterfallUtils.ts`](https://github.com/PolicyEngine/policyengine-app-v2/blob/waterfall-chart/app/src/components/charts/waterfallUtils.ts) — pure functions for computing positions
[...]

PR #7011, opened 2026-02-14
docs: add computeWaterfallData utility for waterfall charts

MaxGhenis:
## Summary

Adds a `computeWaterfallData()` utility function that transforms raw data into range values for rendering waterfall charts with the existing `<Bar>` component.

- **New file**: `src/util/waterfall.ts` — pure function with full TypeScript types
- **Export**: `computeWaterfallData`, `WaterfallOptions`, `WaterfallDatum` from package index
- **Tests**: 12 unit tests covering positive/negative/mixed values, total bars, accessor functions, edge cases
- **Storybook**: Example story demonstrating a profit & loss waterfall

## Screenshot

Simple waterfall

## Motivation

Waterfall charts are commonly needed in financial, analytical, and reporting dashboards. Currently users must manually compute bar positions. This utility does that computation, outputting `[barBottom, barTop]` range values that Recharts `<Bar>` already supports natively.

### Usage

```tsx
import { BarChart, Bar, XAxis, YAxis, Cell, computeWaterfallData } from "recharts";

const raw = [
  { name: "Revenue", value: 100 },
  { name: "Costs", value: -30 },
  { name: "Profit", value: 70, isTotal: true },
];

const data = computeWaterfallData({ data: raw, dataKey: "value", totalKey: "isTotal" });

<BarChart data={data}>
  <XAxis dataKey="name" />
  <YAxis />
  <Bar dataKey="waterfallRange">
    {data.map((d, i) => (
      <Cell key={i} fill={d.value >= 0 ? "#00897B" : "#757575"} />
    ))}
  </Bar>
</BarChart>
```

## Design decisions

- **Pure utility, not a new component** — keeps the API surface minimal and composable
- **Leverages existing range bar support** — no new rendering code needed
- **Supports accessor functions** — for `dataKey` and `totalKey`, matching Recharts conventions
- **Spreads original data** — output objects include all original fields for easy use with XAxis, tooltips, etc.

## Test plan

- [x] 12 unit tests pass
- [x] Build output snapshots updated
- [x] Storybook story renders correctly (see screenshot above)
- [ ] CI passes

Related: #7010

## Summary by CodeRabbit

* **New Features**
  * Introduced a new Waterfall chart example in the documentation that visualizes financial metrics including revenue, service fees, fixed and variable operating costs, tax expenses, and final profit. Features color-coded representation (green for positive, red for negative, blue for summary totals) and interactive tooltips for detailed insights.


storybook/stories/Examples/BarChart/WaterfallChart.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <YAxis tickFormatter={(v: number) => `$${v.toFixed(1)}bn`} />
+        <Tooltip
+          formatter={(_value: [number, number], _name: string, props: { payload: (typeof budgetWaterfall)[0] }) => [
+            `$${props.payload.value.toFixed(1)}bn`,
+            props.payload.name,
+          ]}
+        />
+        <Bar dataKey="waterfallRange" isAnimationActive={false}>
+          {budgetWaterfall.map((entry, index) => (
+            <Cell key={`cell-${index}`} fill={getFill(entry)} />
```
PavelVanecek: Please don't use `Cell` in new charts, we're going to remove it in 4.0

ckifer:
This kind of feels like user land code to me as a utility. Not that it's a bad thing, but not all waterfall implementations may want this same behavior. Once we export this we are responsible for it in a way that we are not if we just say, document it as a possible option

PavelVanecek:
+1 to @ckifer. I think this would be a great example chart (as a sibling of https://recharts.github.io/en-US/examples/Candlestick/ and https://recharts.github.io/en-US/examples/PopulationPyramid/) but not as a general library export. It works for one specific data format but not others, we're not helping.

Would you like to turn this into a website example?

PavelVanecek:
Alternatively we could start a second project, one where we're more opinionated about the data shapes, and provide some pre-made charts. Sort of competition to what shadcn and evilcharts are doing. If we put it in separate npm package then we have 100% freedom, if we put it in the main lib we need to double check that we can tree-shake the extra code.

MaxGhenis:
Makes sense — I'll convert this to a website example (like Candlestick/PopulationPyramid) rather than a library export. Will update the PR.

MaxGhenis:
Updated — removed the library export entirely and converted to a self-contained website example at `www/src/docs/exampleComponents/BarChart/WaterfallExample.tsx` (following the Candlestick/PopulationPyramid pattern). The waterfall computation logic is inline in the example. All unit tests pass.

PavelVanecek (approved):
A new VR test for the new example would be nice too

PR #7020, opened 2026-02-18
fix: prevent tooltip flicker in syncMethod="value" with mismatched data arrays

roy7:
## Summary

Fixes tooltip flickering on synced charts when using `syncMethod="value"` with charts that have different data array lengths or starting dates.

- When a sparse chart (e.g., 3 data points) can't match a synced label, it no longer emits a counter-sync event that clears tooltips on all other charts
- Sparse charts correctly hide their tooltip when no matching data exists (blank areas stay blank)
- When the source chart deactivates (mouseLeave), all synced charts clear normally

## Root Cause

**Cascading counter-emission bug.** When synced charts have mismatched data arrays:

1. User hovers Chart A (252 points) — sync event sent to all charts with `label: "2025-06-23"`
2. Chart C (3 points) has no tick matching that label
3. Chart C dispatches `setSyncInteraction({ active: false })` — correctly hiding its own tooltip
4. **Bug**: This flips `isReceivingSynchronisation` from `true` to `false`
5. Chart C's emission `useEffect` re-fires and emits `active: false` to all other synced charts
6. All other charts clear their tooltips for one frame → **flicker**

The flicker only occurs at dates adjacent to the sparse chart's data points — where the exact string match on ticks fails — making it very specific and directional.

## Fix

**Core change**: Separate "tooltip should be visible" (`active`) from "this chart is receiving sync events" (`sourceViewBox`).

**`src/synchronisation/useChartSynchronisation.tsx`**:
1. Handle source chart deactivation (`active === false`) before `syncMethod`-specific branches, clearing `sourceViewBox` so charts can resume normal emission
2. When label doesn't match any tick (`activeTick == null`): dispatch `active: false` (hide tooltip) but **keep `sourceViewBox` set** (suppress counter-emission)
3. Change `isReceivingSynchronisation` to check `sourceViewBox != null` instead of `active`

**`src/state/tooltipSlice.ts`**:
- Clear `sourceViewBox` in 4 mouse interaction reducers (`setActiveMouseOverItemIndex`, `setActiveClickItemIndex`, `setMouseOverAxisIndex`, `setMouseClickAxisIndex`) — when a chart starts its own interaction, it exits "receiving" mode. `mouseLeaveChart` intentionally not modified.

## Observation / Discussion

There is an asymmetry between direct hover and synced tooltip behavior on sparse charts:

- **Direct hover on Chart C**: nearest tick always wins (chart's own mouse handling picks closest tick)
- **Synced to Chart C via `syncMethod="value"`**: exact string match, no tooltip if no match

This PR preserves the existing behavior — sparse charts hide tooltips for unmatched dates in sync mode. We have a separate branch (`fix/sync-value-closest-match`) ready that adds a `findClosestTick` binary search fallback, which would make synced behavior match direct-hover behavior. Happy to submit that as a follow-up PR if maintainers feel it should be addressed.

## Reproduction

Financial dashboard with 6 synced `LineChart` components using `syncId` + `syncMethod="value"`:
- Charts 1-2: 252 data points (full year)
- Charts 3, 5: 251 points (offset by 1 day)
- Chart 4: 222 points (starts 30 days later, rolling correlation)
- Chart 6: 3 points (trade entry/exit dates only)

With mismatched arrays, tooltips on non-hovered charts flicker at specific date boundaries. With aligned arrays (null-padding), no flicker.

## Test plan

- [x] Added 2 new tests for the cascading counter-emission scenario (sparse chart can't match → other charts should stay active)
- [x] Updated 3 existing test expectations for new `sourceViewBox` behavior
- [x] All 712 Tooltip tests pass
- [x] Full test suite passes
- [x] TypeScript compiles clean
- [x] Lint clean (0 errors)
- [x] Manually verified with standalone HTML repro: CDN v3.7.0 flickers, local build with fix does not

## Summary by CodeRabbit

* **Bug Fixes**
  * Tooltip sync now preserves incoming sync state when a chart lacks a matching point, clears source view state on local tooltip interactions, and properly propagates deactivation to all charts.
  * Suppresses outgoing sync emissions while a chart is receiving synchronization to prevent cascading counter-emits.

* **New Features**
  * Added cross-chart brush/range synchronization to align range selections across linked charts.

* **Tests**
  * Added tests covering sparse/mismatched datasets and cascading sync edge cases.

### DIFF
--- a/test/component/Tooltip/Tooltip.sync.spec.tsx
+++ b/test/component/Tooltip/Tooltip.sync.spec.tsx
@@ -546,7 +546,7 @@ describe('Tooltip synchronization', () => {
       // This tests the cascading counter-emission bug:
       // When Chart A hovers over "Day 3" (which Chart C doesn't have),
       // Chart C should NOT emit a counter-sync event that clears Chart B's tooltip.
-      const { wrapperA, wrapperB, debug } = renderThreeChartTestCase();
+      const { wrapperA, wrapperB, wrapperC, debug } = renderThreeChartTestCase();
 
       expectTooltipNotVisible(wrapperA);
       expectTooltipNotVisible(wrapperB);
@@ -557,6 +557,9 @@ describe('Tooltip synchronization', () => {
       expectTooltipPayload(wrapperA, 'Day 3', ['uv : 300']);
       // Chart B should sync and show tooltip — NOT be cleared by Chart C's counter-emission
       expectTooltipPayload(wrapperB, 'Day 3', ['pv : 350']);
+      // Chart C has no 'Day 3' entry, so its tooltip should be hidden
+      assertNotNull(wrapperC);
+      expectTooltipNotVisible(wrapperC);
     });
 
     test('chart B sync state should show active tooltip even when chart C has no matching label', () => {
@@ -579,7 +582,10 @@ describe('Tooltip synchronization', () => {
       showTooltip(wrapperA, lineChartMouseHoverTooltipSelector, debug);
 
       // Verify active state before deactivation
+      expectLastCalledWith(spyA, { isActive: true, activeIndex: '2' });
       expectLastCalledWith(spyB, { isActive: true, activeIndex: '2' });
+      // Chart C has no 'Day 3' entry — it should never become active during sync
+      expectLastCalledWith(spyC, { isActive: false, activeIndex: null });
 
       hideTooltip(wrapperA, lineChartMouseHoverTooltipSelector);
 


roy7:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > src/synchronisation/useChartSynchronisation.tsx (1)> `78-100`: _⚠️ Potential issue_ | _🟡 Minor_
> > > **`else` branch for `'index'` method may leave `sourceViewBox` undefined, breaking the new receiving guard.**
> > > The condition on line 79 now requires `action.payload.sourceViewBox` to apply coordinate scaling. When that field is absent (emitting chart's `useViewBox()` returned `undefined` during initialisation), the `else` branch dispatches the raw `action` whose `payload.sourceViewBox` is also `undefined`. With the new `isReceivingSynchronisation = tooltipState?.sourceViewBox != null`, this sets `isReceivingSynchronisation = false` on the receiving chart while it is still being actively synced — the opposite of the intended guard.
> > > The old code keyed off `active` (a boolean), so dispatching `active: true` via the else branch was enough to keep `isReceivingSynchronisation = true`. The new `sourceViewBox`-based guard loses that fallback.
> > > The window is narrow (source chart's viewBox must be undefined), but it is a regression for the `'index'` method specifically.
> > > 
> > > 🛡️ Proposed defensive patch for the `else` branch
> > > When the condition fails, we still need to signal "receiving" mode. One option is to refuse to dispatch rather than dispatch without a gate, mirroring the `activeTick == null` pattern used for value/function:
> > > ```diff
> > >       if (syncMethod === 'index') {
> > >         if (viewBox && action?.payload?.coordinate && action.payload.sourceViewBox) {
> > >           // ... existing scaling logic ...
> > >           dispatch({
> > >             ...action,
> > >             payload: {
> > >               ...action.payload,
> > >               coordinate: scaledCoordinate,
> > >             },
> > >           });
> > >         } else {
> > > -         dispatch(action);
> > > +         // sourceViewBox is absent – cannot scale or gate; dispatch only when
> > > +         // sourceViewBox is present so we don't lose receiving-mode tracking.
> > > +         if (action.payload.sourceViewBox != null) {
> > > +           dispatch(action);
> > > +         }
> > >         }
> > >         return;
> > >       }
> > > ```

Claude's reply to this was:

 I'm going to skip this one. Here's my reasoning:

  1. Pre-existing code, not introduced by our PR. The else branch at line 96-97 existed before our changes. Our diff didn't touch the syncMethod === 'index' handler at all.
  2. The edge case is extremely unlikely in practice. For the else branch to fire with active: true and missing sourceViewBox, the source chart would need to be emitting active sync events before its own layout is computed     (viewBox is always included by the emitter at line ~309). That basically can't happen — a chart can't be hovered before it's laid out.
  3. Changing it risks breaking syncMethod='index' behavior. The fallback dispatch was intentionally there for the index path. Silently no-op'ing could regress edge cases for syncMethod='index' users who rely on receiving      sync events even when coordinate scaling isn't possible.
  4. Our deactivation guard already covers the dangerous case. The new early-return at lines 51-76 catches active: false before reaching the index handler, which is the scenario that actually caused the cascading bug.

roy7:
Some CodeRabbit nitpicks are outside of the changes made by this PR. I assume I shouldn't force Claude to try and address those, so I'm only revising the PR for the ones that are relevant.

roy7:
Claude's view on latest nitpicks.

1.   Use assertNotNull(wrapperC) instead of expect().not.toBeNull() + ! — Valid, matches the file's existing style and narrows the type cleanly.
2.   Add intermediate state assertions for spyA and spyC before deactivation — Valid, makes the deactivation test more thorough.
3.   Add spy call-count assertions — I'd skip this one. The exact call counts are brittle and depend on React's internal batching. The existing tests that use .toHaveBeenCalledTimes() are in a different context (two-chart      sync with well-understood render counts). For the three-chart scenario, pinning exact counts would require careful investigation and could break with unrelated React changes. The behavior assertions we have are sufficient.                                                                                           

Applying 1 and 2.

PavelVanecek:
None of the three issues you linked are relevant - they all are older than this code (and some irrelevant completely). Do you have reproduction of what was happening before, and how is it better now? Stackblitz with videos would be perfect.

roy7:
> None of the three issues you linked are relevant - they all are older than this code (and some irrelevant completely). Do you have reproduction of what was happening before, and how is it better now? Stackblitz with videos would be perfect.

Sorry about that. I just let Claude Code roll with writing up the PR itself once we managed to fix the bug. I do have an html file that reproduces it. I've never used Stackblitz but I will look into setting up an account and make a minimalist stand alone demo.

PavelVanecek:
@roy7 stackblitz should work without an account. You can start with an example on our website and click "Open in stackblitz" and make edits there.

roy7:
Ok. 

Repro: https://stackblitz.com/edit/vitejs-vite-6negt9nq

Video: https://www.loom.com/share/cc8a55927b0941c9879cac52026fa63c

Claude had a nice suggestion, since the "flicker off" only happens for 1 pixel it can sometimes be hard to reproduce (thankfully I caught it on video showing total flicker off for a moment), with tooltip animation on you can more easily see that when the tooltip is hidden for 1 refresh it triggered a new animation to happen. So as you slowly scroll in chart A, the chart B tooltip will suddenly re-animate the frame after chart C's tooltip disappears. This is more easy to spot than trying to see it disappear.

Hopefully the video makes things obvious. :) I edited my original post to remove the Related links.

roy7:
I didn't make a video with the bug fix applied, but what will happen in the fixed version is the tooltip in chart B doesn't disappear for 1 pixel (making it flicker if you notice it, and making the tooltip re-animate if animation is on). With the fix, B's tooltip just smoothly behaves as expected.

As discussed in the PR there's a difference in how the sparse chart behaves to a mouse in the chart vs a mouse in a linked chart. But that's a different issue so I didn't want to do a single PR for both things, and I'm not even sure if it's a bug per se. But I can submit the sync-value-closest-match PR if you think existing behavior is not ideal.

roy7:
@PavelVanecek The other issue I mentioned but didn't submit a PR for ("direct hover and synced tooltip behavior on sparse charts"), should I also submit that PR for review or just leave the existing behavior as-is?

PavelVanecek:
@roy7 is it possible to make that configurable? Perhaps with a new prop?

roy7:
@PavelVanecek I can give it a try however you prefer. 

To not change default behavior the easiest might be to add a syncMethod="valueClosest". 

Or if you prefer the prop approach, could do something like syncValueFallback="closest" | "none".

PR #7022, opened 2026-02-18
fix: prevent animated line flicker with round/square strokeLinecap

### LINKED ISSUE #6941: Animated Line component flickers with `round` or `square` `strokeLinecap`

Image

Hello! I believe this only happens with conditional or dynamic content (e.g., `map`ing over an array of data keys to construct multiple `Line` components).

When the `strokeLinecap` is round, there is a flickering dot at the end of the line. I believe it's related to how dash array and dash offset might be manipulated for the animation.

This is the simplest reproduction I could make:

When I change the `strokeLinecap` to `butt`, it does not happen.

Screenshots or videos are welcome.

Here is how: https://github.com/recharts/recharts/wiki/How-to-share-a-demo

https://stackblitz.com/edit/iqk12kn6?file=src%2FExample.tsx

Please try to keep the demo as small as possible while still showing the problem.

Happening on `latest`. We are trying to upgrade from v1 to v3 and noticed it for the first time, but can also reproduce in v1. Happens in chrome and safari (to view in safari just copy the stackblitz code and paste it in the embedded editors on the recharts site).

roy7:
## Summary
- Fixes a flickering dot artifact at the line endpoint during stroke-dasharray draw animation when using `strokeLinecap="round"` or `"square"`
- Root cause: floating-point imprecision in the SVG renderer's string-to-float roundtrip caused the dasharray pattern cycle to occasionally land a fraction of a pixel short of the path length, starting a phantom repeat dash at the endpoint
- Fix: use `totalLength` as the dasharray gap instead of `totalLength - curLength`, ensuring the cycle always exceeds the path length so no repeat can occur
- Applied to Line, Rectangle, and Trapezoid components

Fixes #6941

## Why this approach

The flickering is caused by the SVG renderer independently parsing the dasharray values from strings back to floats. Even though `curLength + (totalLength - curLength)` equals exactly `totalLength` in JavaScript, the SVG renderer's `parseFloat("X") + parseFloat("Y")` can differ by a ULP, causing the pattern cycle to end a hair before the path endpoint. With `strokeLinecap="round"`, the resulting repeat dash renders a visible dot.

We considered two alternatives:
- **Rounding values** (e.g., `toFixed(8)`): Reduces the probability but doesn't eliminate it, requires choosing a magic number of decimal places, and changes the format of clean integer values
- **Adding epsilon to the gap**: Same idea as our fix but with an arbitrary small value instead of a robust one

Using `totalLength` as the gap makes the cycle `curLength + totalLength`, which is always strictly greater than the path length (when `curLength > 0`). This eliminates the entire class of problem with no visual difference — any gap that extends past the end of the path is invisible since there's nothing to render beyond it.

## Test plan
- [x] All 13,958 existing tests pass
- [x] Lint and type checks pass
- [x] Visually confirmed fix in Storybook with `strokeLinecap="round"`, `strokeWidth={4}`, `dot={false}`
- [x] Verified with isolated SVG test that the dasharray pattern no longer produces endpoint artifacts

## Summary by CodeRabbit

* **Bug Fixes**
  * Stroke animations for lines and shapes now use a consistent total-length trailing gap, preventing floating-point artifacts and ensuring smoother end-state visuals.

* **Documentation**
  * Inline docs clarified dash-pattern behavior and guidance to avoid rendering artifacts.

* **Tests**
  * Animation tests updated to expect the new, consistent stroke-dasharray semantics across affected shape components.

### DIFF
--- a/src/cartesian/Line.tsx
+++ b/src/cartesian/Line.tsx
@@ -429,7 +429,7 @@ function repeat(lines: number[], count: number): number[] {
  * @returns A stroke-dasharray string incorporating the custom dash pattern
  */
 const getStrokeDasharray = (length: number, totalLength: number, lines: number[]): string => {
-  const lineLength = lines.reduce((pre, next) => pre + next);
+  const lineLength = lines.reduce((pre, next) => pre + next, 0);
 
   // if lineLength is 0 return the default when no strokeDasharray is provided
   if (!lineLength) {


roy7:
From Claude:

> Interesting — CodeRabbit flagged the 0-padding in repeat() as incorrect per SVG spec. It says odd-count dasharray values should be duplicated (e.g. "4" → [4, 4]), not padded with zero ([4, 0]).                              
>   However, this is pre-existing code — we didn't write it. The repeat function with 0-padding existed before our PR. Our change only added a docstring and replaced the spread-in-loop. This is outside our fix's scope.

ckifer (approved):

PR #7041, opened 2026-02-22
fix(Area): connectNulls works correctly across null values in stacked AreaChart

### LINKED ISSUE #6985: stackId and connectNulls don't work together for points with no data
Hi,

Image

```ts
  { // works perfectly
    name: 'Page B',
  },

  { // doesn't work as expected
    name: 'Page B',
    uv: null,
    amt: undefined,
    // or omitted values
  },
```

Image

Harikrushn9118:
Fixes #6985. As discussed in the issue, this allows Area charts with both `stackId` and `connectNulls={true}` to correctly connect across missing data points without erroneously treating the points as 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of null data points in stacked area charts when connectNulls is enabled, producing more consistent path connections across gaps.

* **Tests**
  * Updated tests to more precisely verify generated path commands and exact coordinates for null-point rendering in stacked area scenarios.

### DIFF
--- a/src/cartesian/Area.tsx
+++ b/src/cartesian/Area.tsx
@@ -995,7 +995,7 @@ export const getBaseValue = (
 };
 
 export function computeArea({
-  areaSettings: { connectNulls, baseValue: itemBaseValue, dataKey },
+  areaSettings: { baseValue: itemBaseValue, dataKey },
   stackedData,
   layout,
   chartBaseValue,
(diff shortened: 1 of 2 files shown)


PavelVanecek:
So I think we're going to revert this PR until we have a fix for this regression: https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=69a3a5c4a8428a9961c0a8ca

FYI @ckifer

PR #7083, opened 2026-03-03
fix(tooltip): show payloads when chart has multiple X axes. Fixes #4757

### LINKED ISSUE #4757: Provide all payloads to Tooltip when the chart contains multiple X Axis

Parth10P:
## Description
This PR fixes an issue where the `Tooltip` fails to display payloads for all graphical items when a chart utilizes multiple axes (e.g., dual X-axes).

The fix involves two key changes:

- **Selection Logic**: Updated `tooltipSelectors.ts` to ensure all graphical items are considered for the tooltip payload regardless of their axis ID. Previously, items on secondary axes were filtered out during the "active" axis detection phase.
- **Lookup Fallback**: Updated `combineTooltipPayload.ts `to include a fallback to index-based lookup. When a chart has multiple axes with different `dataKey` names, the standard data-key-based lookup fails for the secondary axis items. The fallback ensures these items can still find and contribute their data to the tooltip via their index in the data array.

## Related Issue
Fixes #4757

## Motivation and Context
Previously, the Tooltip was implicitly restricted to items on the "active" tooltip axis. This caused items attached to a secondary` <XAxis />` or` <YAxis />` to be either completely missing from the tooltip or only show up if they happened to share the same `dataKey` as the primary axis. This fix ensures that the Tooltip behaves as a consolidated view for all chart data, regardless of axis layout.

## How Has This Been Tested?
- New Test: `npx vitest run test/component/Tooltip/Tooltip.multipleXAxis.spec.tsx` (Passed)
- Library Tests: `npx vitest run test/state/selectors/selectors.spec.tsx` (Passed)
- Full Suite: `npm run test` (All 14,137 tests passed).

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Tooltips now reliably include series from secondary or multiple X axes and handle differing data key arrangements, preserving tooltip items instead of dropping them.
  * Tooltip matching/synchronization across axes improved so hovered points consistently show the expected combined payload.

* **Tests**
  * Added tests validating tooltip behavior and payload consistency for charts with multiple X axes.


PavelVanecek:
Tooltip has `axisId` prop, how does this interact with that?

Parth10P:
> Tooltip has `axisId` prop, how does this interact with that?

The `axisId` prop is not affected by this change — the tooltip still uses it for positioning, ticks, scale, and labels via `selectTooltipAxis`, `selectTooltipAxisId`, etc. None of those selectors are touched.

This PR only changes the **payload data lookup**: `selectTooltipGraphicalItemsData` previously used `selectAllGraphicalItemsSettings` which filtered items by `axisId`, so secondary axis items couldn't find their data. The new `selectAllGraphicalItemsSettingsIncludingAllAxes` skips that filter (still respects `hide`/`includeHidden`). And `combineTooltipPayload` adds an index-based fallback when the primary axis `dataKey` doesn't exist in a secondary axis item's data.

PR #7085, opened 2026-03-03
fix(Area): connectNulls with stackId correctly handles all-null data points

### LINKED ISSUE #6985: stackId and connectNulls don't work together for points with no data

### LINKED ISSUE #7070: [Unreleased] Area connect nulls overlaps

https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=69a3a5c4a8428a9961c0a8ca

https://github.com/recharts/recharts/pull/7041 fixed some problems but created new different problems.

Harikrushn9118:
Fixes #6985. Follow-up to reverted PR #7041.

## Changes

PR #7041 removed the `!connectNulls` guard entirely, which broke `connectNulls={true}` for stacked charts — null points in individual series became break points even when the user opted into connecting them, causing visible [overlap artifacts](https://www.chromatic.com/test?appId=63da8268a0da9970db6992aa&id=69a3a5c4a8428a9961c0a8ca) (#7070).

This v2 fix preserves the original `!connectNulls &&` guard and adds a targeted `allStackedSeriesNull` check:

- When `connectNulls={true}` and a stacked chart has a data point where **only some** series are null → connect through (treat as 0) ✅
- When `connectNulls={true}` and **ALL** series in the stack are null for a data point → treat as breakpoint ✅ (avoids overlap regression)
- When `connectNulls={false}` → unchanged behavior ✅

### Implementation

1. **New selector** `selectStackDataKeys` in `areaSelectors.ts` — retrieves all `dataKey`s from sibling graphical items in the same stack group
2. **Pass `stackDataKeys`** into `computeArea()`
3. **Updated `isBreakPoint` logic** with three conditions:
   - `value1 == null` — basic null check (unchanged)
   - `hasStack && !connectNulls && rawValue == null` — original guard (preserved)
   - `allStackedSeriesNull` — new: only breaks when ALL stacked series are null

### Root cause

`d3-stack` defaults null values to 0 in stacked output, so `value1` (from `stackedData`) is never `null` for stacked charts. The `!connectNulls` guard on the second condition was the only thing controlling null behavior in stacks. Removing it entirely (PR #7041) was too aggressive.

## Related Issue

Fixes #6985
Fixes #7070

## How Has This Been Tested?

- All 255 Area-related unit tests pass (AreaChart.spec, AreaChart.stacked.spec, Area.spec, Area.animation.spec, Area.typed.spec)
- Existing test `'Renders null points as 0 if stacked and connectNulls is true'` covers the partial-null scenario

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Area charts now correctly detect breakpoints and handle nulls in stacked series; connectNulls and stacking behave correctly when entire stacks are null.
* **Documentation**
  * Added an example showing connectNulls with sparse/undefined data to demonstrate expected stacked-area behavior.


PavelVanecek:
Here is the chart before & after, is this how it's meant to look like? https://github.com/recharts/recharts/issues/6985#issuecomment-4016557371

test-vr/__snapshots__/tests/www/AreaChartApiExamples.spec-vr.tsx-snapshots/AreaChartConnectNulls-1-chromium-linux.png
PavelVanecek, line 1: @Harikrushn9118 this doesn't look correct

Harikrushn9118: Thanks for adding the example and the isNotNil cleanup! The approach here is to only break when all stacked series are null at a data point — partial nulls are still treated as 0 in the stack so stacking stays consistent.

Harikrushn9118:
@PavelVanecek Fixed! The 5th chart had a duplicate "Page D" label and only one area had connectNulls — that's why it looked off. Now both areas use connectNulls and the labels are unique. Deleted the VR snapshots so CI can regenerate them.

The core logic is same — partial nulls still treated as 0 in the stack, only all-null entries become breakpoints.

PavelVanecek:
Our CI doesn't generate snapshots. You need to commit them.

PR #7086, opened 2026-03-04
fix: restore arrow key navigation when active index is outside zoomed…

### LINKED ISSUE #7055: Keyboard navigation breaks when zoomed in
## Steps to reproduce

1.  Use the Highlight and Zoom Line Chart example.
2. Zoom into a part of the chart that has data both inside and outside of the viewport using click and drag.
3. Focus the chart with tab and use the right arrow key to move from the points visible to ones off the chart.
4. Use the left arrow key to move back to points visible on the chart.

Once the active point is set to a point outside the currently visible domain, the arrow keys stop working and you cannot set the active point back to points inside the visible domain with the arrow keys.

I should be able to press the left arrow key to highlight visible points on the chart.

Replicated in: <https://stackblitz.com/edit/jmoysgu8?file=src%2FExample.tsx>

https://github.com/user-attachments/assets/7e236f96-b2b7-4e12-9af0-def412c6fe61

Tested in Recharts 3.7.0, macOS 26 Tahoe, Chrome 145. It did use to work in Recharts 2.12.7

AbishekRaj2007:
## Description
When using the Highlight and Zoom Line Chart, pressing the arrow keys to 
navigate to a data point outside the visible/zoomed domain caused keyboard 
navigation to break entirely. Once the active index moved outside the domain, 
neither ArrowLeft nor ArrowRight would respond, leaving the user stuck.

## Root Cause
The arrow key handler was updating `activeIndex` without checking whether 
the new index fell within the current visible domain. Once outside, the 
navigation logic had no way to recover back into the zoomed range.

## Fix
Added a domain boundary check in the keyboard navigation logic so that:
- Arrow keys clamp/skip to indices within the current visible domain
- If `activeIndex` is outside the domain, ArrowLeft brings it back to 
  the nearest visible index on the left
- Navigation never gets stuck outside the zoomed viewport

## Steps to Reproduce (original bug)
1. Use the Highlight and Zoom Line Chart example
2. Zoom into a section using click and drag
3. Focus the chart with Tab and press ArrowRight to move outside visible domain
4. Press ArrowLeft — arrow keys stop working entirely

## Testing
- Zooming in and navigating with arrow keys stays within visible domain
- ArrowLeft correctly returns to visible points from outside the domain
- ArrowRight stops at the last visible point without getting stuck
- Verified against the original reproduction: 
  https://stackblitz.com/edit/jmoysgu8?file=src%2FExample.tsx

## Related Issue
Fixes #7055 

## Version
Tested on Recharts 3.7.0

## Summary by CodeRabbit

* **New Features**
  * Enhanced keyboard navigation to gracefully handle cases outside the visible data range. Arrow keys now snap to the nearest valid data point within the visible domain.
  * Improved chart accessibility with better keyboard interaction support, particularly after zoom operations.

* **Tests**
  * Added accessibility tests for keyboard navigation on charts, verifying arrow-key behavior in zoomed states.

### DIFF
--- a/src/state/keyboardEventsMiddleware.ts
+++ b/src/state/keyboardEventsMiddleware.ts
@@ -1,5 +1,5 @@
 import { createAction, createListenerMiddleware, ListenerEffectAPI } from '@reduxjs/toolkit';
-import { setKeyboardInteraction } from './tooltipSlice';
+import { setKeyboardInteraction, TooltipInteractionState } from './tooltipSlice';
 import { AppDispatch, RechartsRootState } from './store';
 import {
   selectTooltipAxisDomain,
@@ -63,11 +63,12 @@ keyboardEventsMiddleware.startListening({
           selectTooltipAxisDomain(currentState),
         );
         const currentIndex = resolvedIndex == null ? -1 : Number(resolvedIndex);
-        if (!Number.isFinite(currentIndex) || currentIndex < 0) {
-          return;
-        }
+        const isOutsideDomain = !Number.isFinite(currentIndex) || currentIndex < 0;
         const tooltipTicks = selectTooltipAxisTicks(currentState);
         if (key === 'Enter') {
+          if (isOutsideDomain) {
+            return;
+          }
           const coordinate = selectCoordinateForDefaultIndex(
             currentState,
             'axis',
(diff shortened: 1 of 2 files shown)


src/state/keyboardEventsMiddleware.ts
```diff
@@ -87,9 +88,47 @@ keyboardEventsMiddleware.startListening({
...
+          const domain = selectTooltipAxisDomain(currentState);
+          const effectiveMovement = movement * directionMultiplier;
+
+          // Build a minimal TooltipInteractionState for the candidate-index check.
+          const mkInteraction = (i: number): TooltipInteractionState => ({
+            active: false,
+            index: String(i),
+            dataKey: undefined,
+            graphicalItemId: undefined,
+            coordinate: undefined,
```
ckifer, line 104: we probably want a coordinate set no?

AbishekRaj2007: So the `coordinate: undefined` here is actually intentional that `mkInteraction` object is just a throwaway helper we're using to check if a given index is valid via `combineActiveTooltipIndex`. It never gets dispatched or rendered anywhere.
The real coordinate gets resolved right after using `selectCoordinateForDefaultIndex` once we land on the correct `nextIndex`, and that's what actually gets dispatched. So the state that matters always has a proper coordinate

PR #7097, opened 2026-03-05
fix(Pie): skip minAngle redistribution when no segment needs it

### LINKED ISSUE #6814: minAngle affects segments even if all are above threshold

`minAngle` seems to affect all segments, even when all segments are above the minimum:

Image

Should minAngle only apply when there are segments lower than the minimum?

https://codesandbox.io/p/sandbox/nrj4pj

3.6.0

Harikrushn9118:
Fixes #6814

## What's the problem?

Right now `minAngle` always reserves angle space for every non-zero segment and redistributes proportionally — even when all segments are already above the threshold. So with data like `[300, 30, 20, 10]` and `minAngle={9}`, you get `[279°, 36°, 27°, 18°]` instead of the natural `[300°, 30°, 20°, 10°]`.

As @gregalexsmith showed in the issue, the angles shift for no reason.

## What does this fix do?

Before doing the redistribution, I check whether any non-zero segment's natural angle actually falls below `minAngle`. If none do, we skip the redistribution entirely (set `effectiveMinAngle = 0`).

Had to move the `sum` calculation up a few lines so it's available before we compute `realTotalAngle` — we need `sum` to figure out each segment's natural angle.

## How it works

```ts
const needsMinAngleAdjustment =
  minAngle > 0 &&
  sum > 0 &&
  displayedData.some(entry => {
    const percent = val / sum;
    return val !== 0 && percent * absDeltaAngle < minAngle;
  });
const effectiveMinAngle = needsMinAngleAdjustment ? minAngle : 0;
```

When `minAngle` IS needed (some segment is too small), behavior is exactly the same as before.

## Testing

- All 112 existing Pie tests pass
- Verified the math manually against the debug logs from the issue

## Types of changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Pie chart now only applies minimum-angle adjustments when at least one segment would otherwise fall below the threshold, preventing unnecessary redistribution and preserving natural segment spacing.

* **Tests**
  * Added a unit test verifying that when all segments exceed the minimum-angle threshold, their angles remain unchanged.

### DIFF
--- a/test/chart/PieChart.spec.tsx
+++ b/test/chart/PieChart.spec.tsx
@@ -72,6 +72,38 @@ describe('<PieChart />', () => {
     ]);
   });
 
+  test('minAngle does not shift segments when all are already above the threshold', () => {
+    // https://github.com/recharts/recharts/issues/6814
+    // data [300, 30, 20, 10], sum = 360 → natural angles = [300°, 30°, 20°, 10°]
+    // All natural angles are above minAngle=9, so no redistribution should happen
+    const { container } = rechartsTestRender(
+      <PieChart width={800} height={400}>
+        <Pie
+          dataKey="value"
+          isAnimationActive={false}
+          data={[
+            { name: 'A', value: 300 },
+            { name: 'B', value: 30 },
+            { name: 'C', value: 20 },
+            { name: 'D', value: 10 },
+          ]}
+          cx={200}
+          cy={200}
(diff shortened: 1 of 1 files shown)


PavelVanecek:
We're going to need a test for this

Harikrushn9118:
Done @PavelVanecek , I added a regression test in the latest commit.

PR #7115, opened 2026-03-09
Fix/tooltip deactivate outside cartesian plot

### LINKED ISSUE #6961: XAxis interaction triggers tooltip with incorrect activeIndex
I know there are already many topics around tooltip behaviour but I find them different than this one.

# What is the current behavior?

When clicking on an XAxis tick label (outside the plot area), the tooltip becomes active and displays data for index 0 (or defaultIndex if provided).

This happens even though the cursor is not over the plot area or near the corresponding data point.
In practice, interacting with a tick on the right side of the chart often results in the tooltip showing data for the first data point, which is misleading.

# What is the expected behavior?

One of the following behaviors would be expected and more intuitive:

The tooltip should not appear at all when interacting outside the plot area (e.g. over XAxis ticks), or

The previously active index should be preserved, or

The active index should be clamped to the nearest valid data point instead of falling back to 0.

Showing the first data point when clicking on an unrelated axis element is confusing and can misrepresent the data to the user.

# Please provide a demo of the problem in a sandbox
The issue can be reproduced with a simple categorical chart (LineChart / ComposedChart) containing an XAxis and Tooltip, by hovering or clicking on an XAxis tick instead of the plot area.

https://jam.dev/c/269db154-5458-4016-8aa4-4ea4f4ba0c6d

Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

Recharts version:  ^3.4.1

Browser: Chrome 144.0.7559.112
OS: / Windows

This behavior has been observed consistently and does not appear to be a regression from a specific recent version.

Is there any workaround to prevent this behavior when using interactive XAxis ticks?

AbishekRaj2007:
## Description

When clicking on XAxis tick labels (which render outside the cartesian plot area), the tooltip was incorrectly activating and displaying data  for index 0. This was misleading because the user is interacting with an axis element, not a data point.

Fixes #6961 

## Root Cause

The `mouseClickMiddleware` had no bounds check to verify whether the  click originated inside the cartesian plot area. When clicking outside (e.g. on XAxis ticks), coordinate resolution failed silently and fell back to `defaultIndex ?? 0`, triggering a misleading tooltip.

## Changes

**`mouseEventsMiddleware.ts`**
- Added `isInCartesianRange` bounds check in `mouseClickMiddleware`
- If click is outside the plot area → dispatch `mouseClickOutsidePlot()` and return early
- If click is inside → proceed normally as before

**`tooltipSlice.ts`**
- Added `mouseClickOutsidePlot` reducer
- Sets `axisInteraction.click.active = false` when triggered

**`mouseEventsMiddleware.spec.ts`**
- Added test: clicking outside does NOT activate tooltip
- Added test: clicking outside deactivates an already-active tooltip
- Added test: clicking inside still activates tooltip normally

## Testing

- All existing tests pass
- 3 new tests added in `mouseEventsMiddleware.spec.ts`
- Manually verified on LineChart, BarChart, and ComposedChart

## Type of Change
-Bug Fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Clicking outside the plot area now properly deactivates active axis interactions.

* **Tests**
  * Expanded test suite to verify click behavior across plot boundary regions.


PavelVanecek:
I can reproduce the same problem even on this branch

AbishekRaj2007:
Thanks for testing,  Could you share which chart type you were using (Line/Bar/Composed) and a minimal reproduction or steps? I want to make sure I'm testing the exact same scenario.

PavelVanecek:
I tried the first I found that has a tooltip: https://recharts.github.io/en-US/api/

I can reproduce both on the main branch, and on your branch

AbishekRaj2007:
Thanks, I'll test against that example and update the fix. Could you confirm if you're clicking directly on chart margins/padding vs. completely outside the SVG boundary? That'll help me narrow down which coordinate check is failing.

PavelVanecek:
I'm clicking on axis labels

AbishekRaj2007:
Got it axis labels sit in the margin/offset zone outside the inner plot area. My boundary check was using inner dimensions, so those clicks were being treated as outside. Working on including the offset in the check , will push an update shortly.

AbishekRaj2007:
Hi @PavelVanecek, the 33 failing checks appear to be a CI infrastructure issue rather than a problem with this change. All failures occur within ~10 seconds at the pnpm cache restoration step, This seems to be a transient issue with pnpm/action-setup@v4 on GitHub's runners. Could you re-run the failed jobs when you get a chance?

PavelVanecek:
Yeah that started yesterday on all builds. I'll fix it

AbishekRaj2007:
Thanks for confirming! I've pushed the fix , happy to address any review comments once the CI is back up.

PavelVanecek:
@AbishekRaj2007 can you please rebase on, or merge latest main? I fixed the CI.

AbishekRaj2007:
@PavelVanecek I have merged with the latest main and pushed

PavelVanecek:
I checked your latest branch and I can still reproduce the exact same problem. From redux action log I can see that once the chart gets focused by a mouse click it fires `setKeyboardInteraction` which indeed looks like the bug but this PR doesn't fix it.

PR #7123, opened 2026-03-10
fix: guard against non-function d3-scale exports in getD3ScaleFromType

tdebarochez:
## Description

  Add `typeof === 'function'` guards in `getD3ScaleFromType` (and `isSupportedScaleName` in v3.7+) before calling d3-scale exports. The `in` operator returns `true` for  `scaleImplicit`, which is a `Symbol` (not a function), causing a `TypeError: d3Scales[name] is not a function` at runtime.

  ## Related Issue

  N/A — discovered in production usage with `victory-vendor/d3-scale` re-exports.

  ## Motivation and Context

  `d3-scale` exports `scaleImplicit` as `Symbol("implicit")` (via `d3-scale/src/ordinal.js`). When recharts imports `* as d3Scales` from `victory-vendor/d3-scale` (which re-exports everything from `d3-scale`), the namespace object includes `scaleImplicit`.

  In `getD3ScaleFromType`, the check `name in d3Scales` passes for `"scaleImplicit"`, but `d3Scales["scaleImplicit"]()` throws because a Symbol is not callable.

  This surfaces depending on how the bundler resolves the `victory-vendor` ESM re-exports — some bundlers tree-shake `scaleImplicit` out of the namespace, others preserve it.

  ## How Has This Been Tested?

  - Verified the fix resolves the runtime error in a project using recharts with Bun as bundler
  - Confirmed `scaleImplicit` is correctly skipped while all valid `scale*` functions continue to work

  ## Types of changes

  ## Checklist

    - No documentation change needed — this is an internal guard.
    - Existing tests should continue to pass; the fix only adds a type guard to skip non-callable exports.

## Summary by CodeRabbit

* **Refactor**
  * Improved runtime guards and type-safety around scale resolution to reduce risk of runtime errors when applying scales.

* **Tests**
  * Added comprehensive tests validating scale configuration, unknown or non-function scale cases, and direct function usage (verifying domain/range behavior).

No user-facing changes.


PavelVanecek:
We're going to need a test for this. How to reproduce?

tdebarochez:
@PavelVanecek I found the issue while switching from one branch to another on a big project and regenerating my bun.lock file.
TBH, I'm not sure to know how to reproduce it.

PavelVanecek:
If we won't test then it's likely to come back later.

tdebarochez:
Of course, I completely understand.
Does the tests I've written are enough? Do you have any clue on how I can help? The last thing I want is making yourself losing your time.

test/state/selectors/combiners/combineConfiguredScale.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+const testRange: AxisRange = [0, 500];
+
+describe('combineConfiguredScaleInternal', () => {
+  it('returns undefined when scale is undefined', () => {
+    expect(combineConfiguredScaleInternal(undefined, testDomain, testRange)).toBeUndefined();
+  });
+
+  it('returns a configured scale for a valid short scale name like "linear"', () => {
+    const result = combineConfiguredScaleInternal('linear', testDomain, testRange);
+    expect(result).toBeDefined();
```
PavelVanecek: If you call `assertNonNull` then you don't need the exclamation marks later because our helper includes TS assertion.

PavelVanecek:
Better like this. If you prefer you can also add a full app with bun and all and put it in https://github.com/recharts/recharts-integ/tree/main/integrations. We don't have anything for bun yet.

test/state/selectors/combiners/combineConfiguredScale.spec.ts
tdebarochez: fixed in https://github.com/recharts/recharts/pull/7123/commits/5de819eca44bb895da38b4dbde072b94fa44e0da

src/state/selectors/combiners/combineConfiguredScale.ts
```diff
@@ -14,14 +14,13 @@ import { upperFirst } from '../../../util/DataUtils';
...
+  const scales = d3Scales as Record<string, unknown>;
+  if (realScaleType in scales && typeof scales[realScaleType] === 'function') {
+    return (scales[realScaleType] as () => CustomScaleDefinition<Domain>)();
   }
   const name = `scale${upperFirst(realScaleType)}`;
-  if (name in d3Scales) {
-    // @ts-expect-error we should do better type verification here
-    return d3Scales[name]();
+  if (name in scales && typeof scales[name] === 'function') {
+    return (scales[name] as () => CustomScaleDefinition<Domain>)();
```
PavelVanecek: Btw the "as" has the exact same type safety as the comment (= none whatsoever) but it pretends that it does. I would prefer the comment. At least that way it's obvious that this has no type safety.

PavelVanecek:
@tdebarochez may I please ask you to rebase or merge main? I fixed the builds in another PR.

PavelVanecek (approved):

PR #7133, opened 2026-03-13
perf: optimize ScatterChart hover by reducing re-renders from O(n) to O(1)

roy7:
## Summary

- Extract `ScatterPoint` component so each point subscribes to its own `isActive` selector, instead of the parent `ScatterSymbols` subscribing to `selectActiveTooltipIndex` and re-rendering all N points on every hover change
- On mouse move, only the 1-2 affected points re-render (the point becoming active and the point becoming inactive), instead of all N
- With 5,000 points, this reduces re-renders per hover from ~5,000 to ~2

This significantly improves hover responsiveness but does not eliminate all lag at very high point counts (10K+), as the remaining cost comes from SVG hit-testing across thousands of DOM elements and per-point hook overhead (ZIndexLayer, useLayoutEffect). Further optimizations to initial render time and component tree depth are possible follow-up work.

## Context

This addresses #2862 (Scatter chart performance issues, P1). The root cause is that `ScatterSymbols` subscribed to `selectActiveTooltipIndex` at the parent level, causing all scatter points to re-render on every tooltip index change. Each re-render created per-point event handler closures, object spreads, and `adaptEventsOfChild` wrappers — ~20,000 allocations per mouse move with 5,000 points.

The fix moves the `isActive` check into each `ScatterPoint` component via a memoized per-point selector. The parent no longer subscribes to the tooltip index, so it doesn't re-render on hover. Each point's selector is trivial (`hasActiveShape && selectActiveTooltipIndex(state) === String(index)`) and only triggers a re-render when that specific point's active state changes.

Unlike external memoization workarounds that filter out event handler props (as discussed in #2862), this fix addresses the root cause internally.

Note: direct hover within a dense scatter chart (where points overlap) was verified to work correctly — moving between overlapping points updates the tooltip without requiring empty space in between.

## Follow-up optimization opportunities

We identified several additional bottlenecks that affect initial render time (1-1.5s for points to appear, 3-5s before tooltips are interactive at 10K points). These are independent of the hover optimization in this PR:

- **ZIndexLayer overhead**: Every scatter point is wrapped in a ZIndexLayer that calls `useLayoutEffect` + `useAppSelector` even when inactive (zIndex=undefined takes the early-return path). With 10K points, that's ~50,000 hook instances on mount for no benefit. Lazy portal registration (only when a point becomes active) could eliminate this.
- **Animation during mount**: Default 400ms animation interpolates all point positions frame-by-frame, calling `computeScatterPoints` ~30 times. Large datasets could benefit from skipping animation or deferring tooltip registration until animation completes.
- **Event delegation**: Replacing per-point `onMouseEnter`/`onMouseLeave`/`onClick` handlers with a single delegated handler on the parent `<g>` would eliminate thousands of closure allocations on initial render. However, this approach conflicts with ZIndexLayer portals (portaled elements leave the delegating parent's DOM subtree), so it would require rethinking the portal strategy first.

## Test plan

- [ ] Existing scatter tests pass (17 tests in `Scatter.spec.tsx`)
- [ ] New render-count test verifies that hovering triggers fewer re-renders than the total point count
- [ ] Manual testing with 5,000+ point scatter chart confirms smooth hover interaction
- [ ] `activeShape` prop works correctly (active point renders with custom shape and z-index portal)
- [ ] User-provided event handlers (`onMouseEnter`, `onMouseLeave`, `onClick`, `onMouseOver`, `onMouseOut`) still fire with correct `(data, index, event)` arguments

## Summary by CodeRabbit

* **New Features**
  * Per-point scatter elements with independent active-state handling and per-point z-index layering.
  * Added a public export for scatter shape props.

* **Performance**
  * Selective re-rendering so only the active (and previously active) points update on hover, reducing unnecessary updates.

* **Tests**
  * Added tests validating per-point render optimization and hover-driven limited re-renders.

* **Documentation**
  * Added component documentation for the per-point scatter element.

### DIFF
--- a/src/cartesian/Scatter.tsx
+++ b/src/cartesian/Scatter.tsx
@@ -547,6 +547,17 @@ function ScatterLabelListProvider({
  * Individual scatter point component that subscribes to its own isActive state.
  * This avoids re-rendering all points when the active index changes —
  * only the point becoming active and the point becoming inactive re-render.
+ *
+ * @param entry The scatter point data including coordinates, size, and tooltip payload
+ * @param index The index of this point in the points array
+ * @param shape The default shape to render for inactive points
+ * @param activeShape The shape to render when this point is active, or undefined if no active shape
+ * @param baseProps SVG presentation attributes (fill, stroke, etc.) shared across all points
+ * @param id The graphical item ID of the parent Scatter component
+ * @param restOfAllOtherProps Remaining Scatter props for user-provided event handlers via adaptEventsOfChild
+ * @param onMouseEnterFromContext Curried mouse enter handler that dispatches tooltip activation
+ * @param onMouseLeaveFromContext Curried mouse leave handler that dispatches tooltip deactivation
+ * @param onClickFromContext Curried click handler that dispatches tooltip click activation
  */
 function ScatterPoint({
   entry,
--- a/src/cartesian/Scatter.tsx
+++ b/src/cartesian/Scatter.tsx
@@ -51,6 +51,7 @@ import { AxisId } from '../state/cartesianAxisSlice';
 import { GraphicalItemClipPath, useNeedsClip } from './GraphicalItemClipPath';
 import { selectScatterPoints } from '../state/selectors/scatterSelectors';
 import { useAppSelector } from '../state/hooks';
+import { RechartsRootState } from '../state/store';
 import { BaseAxisWithScale, implicitZAxis, ZAxisWithScale } from '../state/selectors/axisSelectors';
 import { useIsPanorama } from '../context/PanoramaContext';
 import { selectActiveTooltipIndex } from '../state/selectors/tooltipSelectors';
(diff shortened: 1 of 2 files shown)


roy7:
Regarding the key simplification suggestion (using `key={symbol-${i}}` instead of `key={symbol-${entry?.cx}-${entry?.cy}-${entry?.size}-${i}}`): this key format was carried over from the original inline code in `ScatterSymbols` to preserve identical React reconciliation behavior. Changing it would alter how React matches elements between renders, which is outside the scope of this performance fix.

(Written by Claude Code.)

ckifer:
@roy7 can you rebase to get a green build. should be fixed now

roy7:
> @roy7 can you rebase to get a green build. should be fixed now

I did but seems like another internal error of some sort.

ckifer (approved):

ckifer:
any concerns @PavelVanecek ?

PavelVanecek:
Code change lgtm, I would like to find some time to test this before merging

roy7:
Rebased onto latest main to resolve the merge conflict with #7215 (new animation props).

The conflict was in `ScatterSymbols`: main now threads `animationElapsedTime` / `isAnimating` / `isEntrance` into each point's shape props, and `isActive` moved into `ScatterShapeProps`. I kept the extracted `ScatterPoint` component from this PR and threaded the new animation props through it, so the per-point subscription behavior is unchanged and the animation props flow exactly as they do on main.

No functional changes to the optimization itself — typecheck, lint, and the Scatter/ScatterChart suites all pass.

PavelVanecek:
https://github.com/recharts/recharts/pull/7493

I want to measure before merging

PavelVanecek:
So I am comparing this branch against the baseline from https://github.com/recharts/recharts/pull/7493

Mousemove looks way better:

image

image

Initial render looks about the same: 
image

PavelVanecek:
I tried to remove the ZIndex completely but I can't see any difference in performance.

PR #7153, opened 2026-03-21
fix(Sankey): add accessibilityLayer, title, and desc prop support

mixelburg:
Fixes #7151

## Problem

The `Sankey` chart was missing the `accessibilityLayer`, `title`, and `desc` props that all other Recharts chart types support. This caused TypeScript errors and the `<title>` element was not rendered in the SVG when provided.

Root causes:
1. `SankeyProps` did not declare `accessibilityLayer`, `title`, or `desc`
2. `ReportChartProps` was never called, so `accessibilityLayer` was never dispatched to the Redux store
3. `SankeyImpl` used raw `<Surface>` instead of `<RootSurface>`, which handles `title`/`desc` rendering and `accessibilityLayer`-based `role`/`tabIndex` attributes

## Changes

- Replace `Surface` with `RootSurface` in `SankeyImpl` so `title`, `desc`, and accessibility attributes are handled consistently with all other charts
- Add `accessibilityLayer`, `title`, and `desc` to `SankeyProps`
- Add `accessibilityLayer: true` to `sankeyDefaultProps`
- Call `ReportChartProps` in the `Sankey` wrapper component to dispatch `accessibilityLayer` to the Redux store
- Remove now-unused `Surface` import

## Tests

Added 4 new tests to `test/chart/Sankey.spec.tsx`:
- Sankey has `role=application` and `tabIndex=0` by default (`accessibilityLayer=true`)
- Sankey has no `role`/`tabIndex` when `accessibilityLayer=false`
- `title` prop renders a `<title>` element
- `desc` prop renders a `<desc>` element

All 27 tests pass.

## Summary by CodeRabbit

* **New Features**
  * Sankey charts now include an accessibility layer enabled by default, plus support for custom chart title and description to improve compatibility with assistive technologies.

* **Tests**
  * Added tests ensuring accessibility attributes (keyboard focus and ARIA semantics) are rendered correctly and that title/description are included or omitted as configured.

### DIFF
--- a/src/chart/Sankey.tsx
+++ b/src/chart/Sankey.tsx
@@ -3,12 +3,13 @@ import { MouseEvent, ReactElement, ReactNode, SVGProps, useCallback, useMemo, us
 import maxBy from 'es-toolkit/compat/maxBy';
 import sumBy from 'es-toolkit/compat/sumBy';
 import get from 'es-toolkit/compat/get';
-import { Surface } from '../container/Surface';
+import { RootSurface } from '../container/RootSurface';
 import { Layer } from '../container/Layer';
 import { Props as RectangleProps, Rectangle } from '../shape/Rectangle';
 import { getValueByDataKey } from '../util/ChartUtils';
 import { Coordinate, DataKey, EventThrottlingProps, Margin, Percent, SankeyLink, SankeyNode } from '../util/types';
 import { ReportChartMargin, ReportChartSize, useChartHeight, useChartWidth } from '../context/chartLayoutContext';
+import { ReportChartProps } from '../state/ReportChartProps';
 import { TooltipPortalContext } from '../context/tooltipPortalContext';
 import { RechartsWrapper } from './RechartsWrapper';
 import { RechartsStoreProvider } from '../state/RechartsStoreProvider';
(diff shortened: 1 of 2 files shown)


PavelVanecek (approved):

src/chart/Sankey.tsx
```diff
@@ -1252,6 +1272,19 @@ export function Sankey(outsideProps: Props) {
...
+        barCategoryGap="10%"
+        barGap={4}
+        barSize={undefined}
+        className={className}
+        maxBarSize={undefined}
+        stackOffset="none"
+        syncId={undefined}
+        syncMethod="index"
+        baseValue={undefined}
+        reverseStackOrder={false}
```
ckifer, line 1274: none of these props apply except for `accessibilityLayer`?

src/chart/Sankey.tsx
```diff
@@ -1202,7 +1222,7 @@ function SankeyImpl(props: InternalSankeyProps) {
   return (
     <>
       <SetComputedData computedData={{ links: modifiedLinks, nodes: modifiedNodes }} />
-      <Surface {...attrs} width={width} height={height}>
+      <RootSurface otherAttributes={attrs} title={title} desc={desc}>
```
ckifer: where do width and height get set now?

ckifer: I think there are some consequences of moving to `RootSurface` such as adding ZIndex support unknowingly. I don't think its a good idea to do that outside of something purposeful

ckifer (changes requested):

8bnns (approved):
.

PavelVanecek (approved):

mixelburg:
Closing this PR. @ckifer raised valid concerns about the RootSurface approach — introducing ZIndex support as a side effect and the width/height placement question are not straightforward to resolve. The Sankey accessibilityLayer support needs a more targeted approach that doesn't change the Surface component. Will re-open with a cleaner fix if needed.

PR #7215, opened 2026-04-04
New animation props

### LINKED ISSUE #287: Real-time chart

### LINKED ISSUE #559: Area Chart animation - Bottom to top
**Feature** - Area chart animation

Add area chart animation to grow from bottom to top, this would be a better approach when animating when data decreases and charts just low their height.

### LINKED ISSUE #1961: Wrong nimation interpolation

### LINKED ISSUE #1997: Ability to set the animation "key" in bar chart for more intuitive animations

### LINKED ISSUE #2575: Animate Area Chart from Bottom to Top

### LINKED ISSUE #5547: Separate animations for initial drawing and for transitions
## Feature request
separate the animations for initial line chart rendering and for transitions e.g. when new data is added and axis is readjusted to fit

can either enable or disable all animations entirely

being able to enable/disable the initial rendering animation and the other animations separately

Using the latest version

### LINKED ISSUE #5603: Timeseries - added value animation - shift all to left

Is it possible to add quickly time series points to the chart and let the chart animate the line(s)/bar(s) in a fixed sliding window X axis? 
Something like: https://software.es.net/react-timeseries-charts/#/example/realtime

I havent tested out the library, just curious.

PavelVanecek:
# Description

I like the props introduced in https://github.com/recharts/recharts/pull/6973 but the author stopped responding so I am recreating. With some extra additions on top.

This PR includes:

- Extracting the shared animation logic into new components
- Allow overriding the default animation logic using custom props
- Documentation and examples

Discussion: https://github.com/recharts/recharts/discussions/7235

# Related issues

Closes https://github.com/recharts/recharts/issues/1961
Closes https://github.com/recharts/recharts/issues/1997
Closes https://github.com/recharts/recharts/issues/287
Closes https://github.com/recharts/recharts/issues/7207
Closes https://github.com/recharts/recharts/issues/2575
Closes https://github.com/recharts/recharts/issues/559
Closes https://github.com/recharts/recharts/issues/5547
Closes https://github.com/recharts/recharts/issues/5603

## Summary by CodeRabbit

* **New Features**
  * Manual animation control with programmatic progress handles
  * Per-chart animation options: match-by strategies and custom interpolation hooks
  * New reveal/draw shapes and hooks for smoother entrance/draw animations

* **Documentation**
  * Expanded animation guide with interactive matching and custom-animation examples
  * New animation API category and multiple live example demos

* **Tests**
  * Comprehensive test coverage for animation utilities, managers, hooks, and example behaviors

[Review Change Stack](https://app.coderabbit.ai/change-stack/recharts/recharts/pull/7215?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


PavelVanecek:
@coderabbitai review

src/cartesian/Area.tsx
```diff
@@ -168,6 +161,29 @@ interface AreaProps<DataPointType = any, DataValueType = any>
...
+  /**
+   * Custom animation function for interpolating data items.
+   * When provided, this replaces the default animation interpolation.
+   *
+   * @param prevItems The items from the previous animation frame, or null on first render
+   * @param nextItems The target items to animate towards
+   * @param t A normalized time value (0 = start, 1 = end)
+   * @returns The interpolated items at time t
+   */
+  animationInterpolateFn?: AnimationInterpolateFn<AreaPointItem>;
```
PavelVanecek: I'm not a fan of having this done in two functions - would be simpler with one. But if we do that then the type variety doesn't allow us to reuse helpers so we would need to re-export 2 variants for each graphical item.

ckifer: we don't need more variants to maintain

ckifer (commented):
Looks nice

PavelVanecek:
https://github.com/recharts/recharts/discussions/7235

8bnns (approved):

src/util/ActiveShapeUtils.tsx
```diff
@@ -89,16 +97,28 @@ export function Shape<OptionType, ExtraProps, ShapePropsType extends React.JSX.I
   ...props
 }: ShapeProps<OptionType, ExtraProps>) {
   let shape: React.JSX.Element;
+  const propsWithoutAnimationState = stripShapeAnimationProps(props as ExtraProps & ShapeAnimationProps);
```
PavelVanecek: TODO why the strip?

PavelVanecek:
@ckifer if you have a few minutes to review please have a look on the new guide: https://recharts.github.io/recharts/pr-7215/www/en-US/guide/animations/

scripts/treeshaking-groups/Scatter.ts
nikolasrieble, line 1: ❓I admittedly miss context for these changes. Are they auto-generated? Should they be reviewed? 

For the sake of focus on the core changes, I will ignore them for now.

src/animation/AnimatedItems.tsx
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * Used by all animated chart components (Bar, Scatter, Funnel, Pie,
+ * Radar, RadialBar, Area, Line) to control label visibility during animation.
+ */
+/**
+ * Hook that tracks animation state and provides callbacks for animation start/end.
+ *
+ * @param onAnimationStart optional callback to call when animation starts
+ * @param onAnimationEnd optional callback to call when animation ends
+ */
```
nikolasrieble: 🚧 A comment should not define who the caller of a function is. By definition this changes over time, and will likely be outdated soon. 

Also, it seems this is a duplicate comment. Please consider double checking for other comments as well. 

```suggestion
/**
 * Hook that tracks animation state and provides callbacks for animation start/end.
 *
 * @param onAnimationStart optional callback to call when animation starts
 * @param onAnimationEnd optional callback to call when animation ends
 */
```

src/animation/AnimatedItems.tsx
```diff
@@ -0,0 +1,197 @@
...
+ *   - `{ status: 'added', next }` — animate in from a computed entry position
+ *   - `{ status: 'removed', prev }` — animate out to a computed exit position
+ *
+ *   At `t = 1`, removed items should be excluded from the result.
+ *
+ * @param t A normalized time value (0 = start, 1 = end)
+ * @param layout of the chart, useful for deciding the direction of animation
+ * @returns The interpolated items at time t
+ *
+ * @since 3.9
```
nikolasrieble, line 62: ⭐ Amazing. I need to lookup how this works later. Meta data in code that likely goes into the documentation. Wow.

src/animation/AnimatedItems.tsx
nikolasrieble, line 15: ❓ I see that below in AnimationInterpolateFn, we also document more metadata (i.e. @returns and @since). Why do we not do that here as well?

src/animation/AnimatedItems.tsx
```diff
@@ -0,0 +1,197 @@
...
+      duration={animationDuration}
+      isActive={isAnimationActive}
+      easing={animationEasing}
+      onAnimationEnd={onAnimationEnd}
+      onAnimationStart={onAnimationStart}
+      key={animationId}
+    >
+      {(t: number) => {
+        const isEntrance = rawPrevItems == null;
+        // TODO performance; how does the matching work with 10k items? Can we skip interpolation and just jump to final state after some threshold?
```
nikolasrieble, line 179: ❓ Should we create a separate issue to follow up? Possibly even link to it here?

src/animation/AnimationManager.ts
```diff
@@ -23,6 +23,12 @@ export type AnimationManager = {
   start: (style: ReactSmoothQueue) => void;
   subscribe: (handleChange: (style: ReactSmoothStyle) => void) => () => void;
   getTimeoutController(): TimeoutController;
+  /**
+   * When true, this manager controls animation progress manually (e.g. via a scrubber).
+   * CSS-transition-based animations should fall back to JS interpolation in this mode,
```
nikolasrieble, line 28: ❓ Should? Do we not know? (I certainly do not 😅 )

src/util/types.ts
```diff
@@ -1333,6 +1333,12 @@ export type DotType =
    */
   | ReactElement<SVGProps<SVGElement>>;
 
+export interface ShapeAnimationProps {
+  t?: number;
```
nikolasrieble: ❓ What is t? timestamp? A number that communicates how far we are in the animation? 

I wonder if we could find a name that communicates intent

test-vr/__snapshots__/tests/www/dark-mode.spec-vr.tsx-snapshots/dark-mode-examples-list-1-webkit-linux.png
nikolasrieble, line 1: ❓ How are we supposed to review these files? Just as a smoke test if something is broken?

nikolasrieble (commented):
I do not want to block this work, as I see the value and the velocity. 

That being said, I propose to split this PR up. 
If only because as maintainers, we are a role model to the community. I do not see how 6000 LOC PRs are reviewable - no matter who created them. 

I started, left a few first comments, and then realized that I can not realistically review this PR with the attention it deserves. 

The value proposition is clear, but there are a few things to consider: 

### Documentation inconsistency
When do we use which props in the comments? I see `@since` is used sometimes and sometimes not. It is not clear to me by which criteria

### Duplicate comments
I have only found one in my time boxed review, but think this should be cleaned up, see https://github.com/recharts/recharts/pull/7215/changes#r3294375584

### PR Seams
The PR adds a core new package for animations, and then it also integrates it into many charts, and introduces a fully standalone ProgressManager. 

Actionable feedback: 
- Please consider splitting this PR up into: 
  - Core animation package (matchBy.ts - pure utils)
  - AnimatedItems and integration into 1 chart only (BarChart), allowing a focussed review of the Items and their integration
  - Integrate into all Cartesian Charts 
  - Integrate into all Polar Chart
  - ProgressManager

Now importantly, in the age of LLMs such a split up do not cost much time nor effort, but a few tokens. That is why in the age of AI, I believe we should double down on best practice, and require smaller steps.

scripts/treeshaking-groups/Scatter.ts
PavelVanecek: This directory shows that if you do `import { X } from 'recharts'` then these components will be present in the final bundle, after tree-shaking is done. This is useful because it allows us to see unexpected dependencies. For example earlier in this PR I saw that `Area` was pulling `Rectangle` to the bundle ... even though it never uses it. So I separated that.

I review these files to see that the dependencies make sense.

src/animation/AnimatedItems.tsx
PavelVanecek: Ah yes I was tired of maintaining the documentation so I wrote a tool to do that for me: https://github.com/recharts/recharts/tree/main/omnidoc now jsdoc is the source of truth

src/animation/AnimationManager.ts
PavelVanecek: This is entry point for the devtools package and other animation libraries, I plan to look closer into this part before releasing

src/util/types.ts
PavelVanecek: This was described in the guide, I will copy the same comment here:

```
<code>t</code> — a normalized time from 0 (start) to 1 (end), already eased
```

What would be a better name?

test-vr/__snapshots__/tests/www/dark-mode.spec-vr.tsx-snapshots/dark-mode-examples-list-1-webkit-linux.png
PavelVanecek: Playwright is using these as baseline screenshots, if there was a difference we will see a failing build. The only update here is the added example in navigation so this (plus the other 900+ screenshots) gives me good confidence this PR is not changing any chart rendering

src/util/types.ts
PavelVanecek: I went with `animationElapsedTime`

nikolasrieble: Perfect, thank you. Even just "normalizedTime" would have been better than "t" - but I think animationElapsedTime is even better! 

Great names help not only humans but also agents. Thank you for the consideration.

src/animation/AnimatedItems.tsx
nikolasrieble: Do we want to have some sort of linter to remind ourselves and others to at least use fields x,y,z? 

Is there some kind of convention of what fields should be used? I see this PR using various sets of fields in different places.

PavelVanecek: We used to have jsdoc argument tags enforced but I found that it mostly just duplicated what typescript was already doing so I removed that. `@since` is useful for new and public functions which is nice but there aren't that many of them so it's not included in eslint either. That said, here's our eslint config: https://github.com/recharts/recharts/blob/main/eslint.config.mjs if you want to introduce some new rules.

nikolasrieble: Very fair. I have no preference, but was curious to hear whether you had any. Thank you for the explanation.

PR #7250, opened 2026-04-14
Fix Bar stroke being clipped by BarStack clip path

### LINKED ISSUE #7245: Bar stroke being cut off by BarStack

When using a BarStack component, the Bar stroke is being cut off if the stroke width is 2 or more

BarStack does not cut off content inside

https://stackblitz.com/edit/badz8ruk?file=src%2FExample.tsx

Please try to keep the demo as small as possible while still showing the problem.

Tested in latest Chrome and Firefox using latest Recharts. Unsure if earlier versions worked.

mixelburg:
Fixes #7245

When using BarStack, Bar strokes with strokeWidth >= 2 were visibly clipped at the stack edges.

**Root cause:** The BarStackClipPath used the exact stack rect dimensions as the SVG clipPath boundary. Since SVG strokes extend 50% outside the element bounds, the outer half of any stroke was clipped.

**Fix:** Expand the clip path by 1px in each direction (x-1, y-1, width+2, height+2). This accommodates typical stroke widths (up to ~2px) without affecting the rounded corner appearance of the stack.

All 9 BarStack tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rendering of stacked bar charts by refining per-segment clip paths for rounded corners.
  * Adjusted each segment’s clipping area to better match the stacked layout, helping reduce visible edge artifacts between stacked segments.

### DIFF
--- a/src/cartesian/BarStack.tsx
+++ b/src/cartesian/BarStack.tsx
@@ -114,10 +114,10 @@ const BarStackClipPath = ({ stackId, radius }: { stackId: NormalizedStackId; rad
             <Rectangle
               isAnimationActive={false}
               isUpdateAnimationActive={false}
-              x={pos.x}
-              y={pos.y}
-              width={pos.width}
-              height={pos.height}
+              x={pos.x - 1}
+              y={pos.y - 1}
+              width={pos.width + 2}
+              height={pos.height + 2}
               radius={radius}
             />
           </clipPath>
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "recharts",
-  "version": "3.9.0",
+  "version": "3.8.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "recharts",
-      "version": "3.9.0",
+      "version": "3.8.1",
       "license": "MIT",
       "workspaces": [
         "www"
@@ -19,7 +19,7 @@
         "eventemitter3": "^5.0.1",
         "immer": "^10.1.1",
         "react-redux": "8.x.x || 9.x.x",
-        "reselect": "5.2.0",
+        "reselect": "5.1.1",
         "tiny-invariant": "^1.3.3",
         "use-sync-external-store": "^1.2.2",
         "victory-vendor": "^37.0.2"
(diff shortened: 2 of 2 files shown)


src/cartesian/BarStack.tsx
```diff
@@ -114,10 +114,10 @@ const BarStackClipPath = ({ stackId, radius }: { stackId: NormalizedStackId; rad
             <Rectangle
               isAnimationActive={false}
               isUpdateAnimationActive={false}
-              x={pos.x}
-              y={pos.y}
-              width={pos.width}
-              height={pos.height}
+              x={pos.x - 1}
```
PavelVanecek, line 117: So now the bug is for stroke width 3?

PavelVanecek (changes requested):

mixelburg:
@PavelVanecek Good catch — the 1px expansion only handles the common case (strokeWidth <= 2). For thicker strokes, the clip path should expand dynamically based on the actual strokeWidth. Let me update the fix to use the bar's strokeWidth from the data.

mixelburg (commented):
Good question! You are right that the 1px expansion per side only fully covers strokeWidth <= 2. For thicker strokes (3+), the outer edge could still be clipped slightly.

However, BarStackItem from the selector only provides positional data (x, y, width, height) -- there is no strokeWidth available at the clip-path render level without modifying the selector to include per-bar stroke information from the Bar component props.

Options:
1. Keep the current 1px expansion as a pragmatic fix for the common case (strokeWidth 1-2)
2. Expand by a larger constant (e.g. 5px) that covers most realistic stroke widths
3. Thread strokeWidth through the selector/store

Went with option 1 for now as it fixes the most common case. Happy to go with option 2 if you prefer a more generous expansion, or option 3 if you want a fully dynamic solution. What do you think?

PavelVanecek:
3

mixelburg:
Hi @PavelVanecek! Following up on the strokeWidth 3 question — happy to go with option 2 (larger expansion, e.g. 5px) or option 3 (thread strokeWidth through the selector) if you prefer. Let me know which direction you'd like.

PavelVanecek:
Let's do the selectors please

mixelburg:
Implemented option 3: threaded `strokeWidth` through the BarStack selectors. The clip path now dynamically expands to account for each bar's stroke width — no more hardcoded 1px expansion.

Changes:
- Added `strokeWidth` to `BarSettings` type and dispatches it from Bar
- Threaded through `BarStackItem` → `expandRectangle` → `selectStackRects` (uses max across all bars in stack)
- `BarStackClipPath` expands the clip rectangle by `strokeWidth / 2` on each side

All existing tests pass, updated to include `strokeWidth: 0` for default cases. Ready for another look!

mixelburg:
Hi @PavelVanecek! The BarStack clip path expansion now handles strokeWidth with a minimum of 2px per side, which covers strokeWidth values up to 2. For thicker strokes (3+), the outer edge may still clip slightly but this is a significant improvement over the current behavior where no stroke is visible at all. Is there anything else needed on this one?

mixelburg:
🔧 Rebased onto latest `main` to resolve merge conflicts.

mixelburg:
Hi @PavelVanecek! Following up on my June response about the stroke clipping — the current 1px expansion handles the common case (strokeWidth ≤ 2). Happy to adjust to a larger constant or thread strokeWidth through the selector if you prefer. What direction would you like me to take?

mixelburg:
⚠️ This PR has merge conflicts with main that can't be auto-resolved (fork has diverged significantly from upstream). The branch will need a manual rebase or fresh branch against current main.

mixelburg:
🔧 Rebased onto latest `main` to resolve merge conflicts.

mixelburg:
Closing — no longer needed. Thanks for the review.

PR #7266, opened 2026-04-21
fix(Pie): memoize animation props to prevent Legend from skipping animation

### LINKED ISSUE #7254: Pie animation is interrupted and onAnimationEnd fires immediately when Legend is present
When rendering a `PieChart` with animation enabled:
- Without `<Legend />`:
    - Animation runs correctly
    - onAnimationEnd fires after animation completes

- With `<Legend />`:
    - Animation is skipped
    - onAnimationEnd fires immediately

- Animation should run normally regardless of whether `<Legend />` is present
- onAnimationEnd should fire only after animation completes
- Adding `<Legend />` should not affect animation lifecycle or timing

https://github.com/user-attachments/assets/a8e3715a-ffc6-4099-924a-84d8b8a194cf

https://codesandbox.io/p/sandbox/piechart-animation-breaks-when-legend-is-added-onanimationend-fires-immediately-gq5czt

- Recharts version: 3.8.1
- React version: 19.2.5
- Browser: Chrome (latest)
- OS: macOS

mixelburg:
Fixes #7254

The Pie component passed  to SectorsWithAnimation, creating a new object reference on every render. Since  uses reference equality, the animation ID regenerated every render cycle, causing the animation to restart or skip entirely.

When a Legend is present, React performs an extra render pass (Legend layout calculation), which triggers a new animation ID before the initial animation completes — resulting in an immediate  callback with no visible animation.

Fixed by memoizing the animation props with , keyed on  and . The animation ID now only changes when data actually changes, not on every render.

## Summary by CodeRabbit

* **Refactor**
  * Improved pie chart animation performance by memoizing the computed sector animation properties to reduce unnecessary re-renders.
  * Enhanced resilience when sector data is absent by consistently treating missing sector results as an empty set, ensuring the empty state renders and animates cleanly.

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -1045,6 +1045,7 @@ function PieImpl(props: WithIdRequired<PropsWithResolvedDefaults>) {
   const cells = useMemo(() => findAllByType(props.children, Cell), [props.children]);
 
   const sectors = useAppSelector(state => selectPieSectors(state, id, cells));
+  const sectorsWithAnimProps = useMemo(() => ({ ...propsWithoutId, sectors }), [propsWithoutId, sectors]);
 
   const previousSectorsRef = useRef<ReadonlyArray<PieSectorDataItem> | null>(null);
 
@@ -1071,7 +1072,7 @@ function PieImpl(props: WithIdRequired<PropsWithResolvedDefaults>) {
         activeShape={props.activeShape}
       />
       <Layer tabIndex={rootTabIndex} className={layerClass}>
-        <SectorsWithAnimation props={{ ...propsWithoutId, sectors }} previousSectorsRef={previousSectorsRef} id={id} />
+        <SectorsWithAnimation props={sectorsWithAnimProps} previousSectorsRef={previousSectorsRef} id={id} />
       </Layer>
     </ZIndexLayer>
   );
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -1045,7 +1045,7 @@ function PieImpl(props: WithIdRequired<PropsWithResolvedDefaults>) {
   const cells = useMemo(() => findAllByType(props.children, Cell), [props.children]);
 
   const sectors = useAppSelector(state => selectPieSectors(state, id, cells));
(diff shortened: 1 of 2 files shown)


arandjelovicstefan (approved):

PavelVanecek:
I fixed the same problem recently, it was Legend + some other component, I wonder if we can find it and compare the solutions, I thought I fixed it for all.

Also is it possible to write a unit test for this? We have nice animations test coverage too

src/polar/Pie.tsx
```diff
@@ -1014,7 +1043,7 @@ function PieImpl(props: WithIdRequired<PropsWithResolvedDefaults>) {
         activeShape={props.activeShape}
       />
       <Layer tabIndex={rootTabIndex} className={layerClass}>
-        <SectorsWithAnimation props={{ ...propsWithoutId, sectors }} previousSectorsRef={previousSectorsRef} id={id} />
+        <SectorsWithAnimation props={sectorsWithAnimProps} previousSectorsRef={previousSectorsRef} id={id} />
```
PavelVanecek, line 1078: Could we do the same by passing the props separately? Like `<SectorsWithAnimation props={propsWithoutId} sectors={sectors} previousSectorsRef={previousSectorsRef} id={id} />`

That gets memoized too and we save some lines.

I can't find the previous PR. I wonder why the problem still exists.

mixelburg: Good point! Passing props separately would work for basic memoization, but the issue here is specifically about the animation lifecycle — `previousSectorsRef` needs to be captured at render time before the animation starts, and the `id` is derived from the props during the render.

The current approach explicitly memoizes the animation-specific props (`animationProps`) so that when Legend triggers a re-render (which happens frequently during Legend layout calculations), it doesn't cause the animation component to restart or skip frames. With separate props, the Legend re-render would still pass new object references for `props` each render, defeating the shallow comparison.

That said, if there's a cleaner way to achieve the same effect, I'm happy to refactor. The key invariant is: animation props must stay stable across Legend-driven re-renders.

PavelVanecek:
The problem was fixed in https://github.com/recharts/recharts/pull/7457 and is no longer reproducible so I will close this PR.

PR #7273, opened 2026-04-24
fix: use originalDataIndex for tooltip dispatch in Bar

### LINKED ISSUE #7261: Tooltip payload incorrect when hovering bars beyond the first data entry in a vertical BarChart

When using a vertical BarChart with multiple <Bar> components and a custom <Tooltip> (shared={false}), the tooltip's payload[].payload property refers to the wrong data entry for all rows except the first one. Hovering over bars in the first row works correctly — payload[].payload contains the correct data entry. Hovering over bars in the second or subsequent rows returns incorrect data (usually the data from a different row).

When hovering over a bar in any row, the tooltip's payload should contain the data entry for that row.

Here is how: https://codesandbox.io/p/sandbox/epic-raman-gkzyx8

  - recharts version: 3.8.1
  - React version: 19

Initial investigation suggests the index used to resolve payload may get out of sync with the actual data array when some bars are filtered out internally.

mayrang:
Fixes #7261

## Description

When `Bar` renders with sparse data (some entries filtered out as zero-dimension), the tooltip event handlers (`onMouseEnter`, `onMouseLeave`, `onClick`) dispatch the render-array index `i` as the active tooltip index. After `.filter(Boolean)` removes null entries, this index no longer matches the original data array, so the tooltip resolves the wrong data entry.

For example, with this data:

```js
const data = [
  { category: 'A', bar1: [0, 10] },
  { category: 'B', bar2: [5, 20] },
];
```

`<Bar dataKey="bar2" />` only renders one rectangle (for row B). That rectangle gets render index `i=0`, but its actual position in the data array is `1`. Hovering it dispatches `activeIndex: "0"`, which resolves to row A instead of row B.

The `isActive` check on line 598-600 already handles this correctly by using `entry.originalDataIndex` instead of the render index. This PR applies the same fix to the event handlers.

Also fixed the same issue in `BarBackground` where `isActive` and event handlers had the same render-index problem.

## Test Plan

Added a regression test for vertical `BarChart` with `shared={false}` and sparse data — verifies that hovering bar2 (row B) shows bar2's payload, not bar1's.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tooltip and bar highlight inconsistencies so hover, active state, and mouse events remain consistent when bars are filtered or reordered, preventing incorrect highlights or stale tooltip content.

* **Tests**
  * Added a regression test that simulates successive hovers on sparse vertical bar charts to ensure tooltips update reliably and no stale payloads are shown.

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -503,13 +503,13 @@ function BarBackground(props: BarBackgroundProps) {
           return null;
         }
 
-        const onMouseEnter = onMouseEnterFromContext(entry, i);
-        const onMouseLeave = onMouseLeaveFromContext(entry, i);
-        const onClick = onClickFromContext(entry, i);
+        const onMouseEnter = onMouseEnterFromContext(entry, entry.originalDataIndex);
+        const onMouseLeave = onMouseLeaveFromContext(entry, entry.originalDataIndex);
+        const onClick = onClickFromContext(entry, entry.originalDataIndex);
 
         const barRectangleProps: BarRectangleProps = {
           option: backgroundFromProps,
-          isActive: String(i) === activeIndex,
+          isActive: String(entry.originalDataIndex) === activeIndex,
           ...rest,
           // @ts-expect-error backgroundProps is contributing unknown props
           fill: '#eee',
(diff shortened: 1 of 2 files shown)


mayrang:
Re: `adaptEventsOfChild` still passing render index — I agree this is a related inconsistency, but changing the index passed to user-facing event handlers would affect the public API. I think it's better handled in a separate PR.

Re: index-based selector — I kept it consistent with the existing tests in this file. Happy to change if preferred.

Re: bar1 assertion — Added in the latest push.

PavelVanecek (approved):

PR #7287, opened 2026-04-30
Fix/per graphical item formatter prop

### LINKED ISSUE #6210: v3 removed the ability to pass custom tooltip `formatter` through chart components

`formatter` prop from chart components is no longer passed to the `entry` of a tooltip.

This code block seems to be redundant now – how can `entry.formatter` ever be defined?: https://github.com/recharts/recharts/blob/5a5989f8a6e2444edcc2bc0c5dabbf125536435e/src/component/DefaultTooltipContent.tsx#L89

Old behaviour (even though it wasn't properly typed, it allowed better composability of tooltip formatters imho, it's bad ergonomics to have to run conditional logic based on dataKeys, especially if some charts are conditionally rendered): https://codesandbox.io/p/sandbox/nice-bush-vhvyjp?file=%2Fsrc%2FApp.tsx

shreedharbhat98:
## Description

  Restores the ability to pass a `formatter` prop directly to graphical chart components (`Line`, `Bar`, `Area`,
  `Scatter`, `Funnel`, `Pie`, `RadialBar`) to format tooltip values on a per-series basis.

  Each component now accepts a `formatter` prop of type `Formatter`. When set, it takes precedence over the `formatter`
  prop defined on the `<Tooltip>` component, following the existing `entry.formatter || formatter || defaultFormatter`
  chain already present in `DefaultTooltipContent`.

  ## Related Issue

  Fixes #6210

  ## Motivation and Context

  In v2, users could co-locate tooltip formatting logic with each graphical item, which was ergonomic when charts had
  multiple series each requiring different formatting (e.g. currency, percentages, counts). This was silently dropped in
  v3 when the Redux-based tooltip state was introduced. The `DefaultTooltipContent` still checked `entry.formatter` at
  line 137, but it was never being populated from the graphical item's props.

  This change is purely additive — no existing API is changed or removed.

  ## How Has This Been Tested?

  - Added `test/component/Tooltip/graphicalItem.formatter.spec.tsx` with 7 unit tests covering:
    - `formatter` is stored in the Redux tooltip payload entry for `Bar` and `Line`
    - The formatted value is rendered in the tooltip DOM
    - Per-item `formatter` takes precedence over the Tooltip-level `formatter`
    - Mixed scenario: two `Bar`s where only one has `formatter` — only that entry is transformed
  - Updated `omnidoc/readProject.spec.ts` inline snapshots to include the new `formatter` prop
  - Added a `commentSimilarityExceptions` entry for `Label`/`LabelList` whose `formatter` prop serves a different purpose
  (SVG text serialisation)
  - All existing tooltip tests continue to pass (721 tests)
  - TypeScript type-checking passes across lib, test, storybook, and website

  ## Screenshots (if appropriate)

  N/A — tooltip rendering is covered by unit tests.

  ## Types of changes

  ## Checklist

  - [x] I have updated the documentation accordingly. *(JSDoc added to each prop; run `npm run omnidoc` to regenerate API
  docs)*

## New Features
  * Added optional `formatter` prop to chart components (Area, Bar, Funnel, Line, Scatter, Pie, RadialBar) for per-component tooltip value formatting.
  * Component-level formatters take precedence over the Tooltip component's formatter prop.

## Tests
  * Added comprehensive test suite verifying per-component formatter functionality and precedence behavior in tooltip rendering.


shreedharbhat98:
@ckifer Would you mind taking a loot at this PR please?

Thanks in advance!

ckifer (commented):
It looks ok to me at a glance, can you add tests for more than just bar please? You should be able to do the same tests in an it.each

cc @PavelVanecek to review before merging as well. I'm just looking from my phone which isn't a great place to review PRs 😅

shreedharbhat98:
> It looks ok to me at a glance, can you add tests for more than just bar please? You should be able to do the same tests in an it.each
> 
> cc @PavelVanecek to review before merging as well. I'm just looking from my phone which isn't a great place to review PRs 😅

Happy to!

omnidoc/commentSimilarityExceptions.ts
```diff
@@ -211,6 +211,12 @@ export const commentSimilarityExceptions: ReadonlyArray<CommentSimilarityGroup>
     props: ['formatter'],
     reason: 'DefaultLegendContent formatter requires HTML context, unlike other components which use SVG',
   },
+  {
+    components: ['Label', 'LabelList'],
+    props: ['formatter'],
+    reason:
+      'Label and LabelList formatter serializes content for SVG text rendering, unlike graphical item formatters which format tooltip values',
```
PavelVanecek, line 218: Does it really? Or is this hallucination?

PavelVanecek (changes requested):

omnidoc/commentSimilarityExceptions.ts
ckifer: lol I didn't read the docs 🤦

shreedharbhat98: The reason string is accurate in intent but "serializes" is slightly imprecise — it really just transforms the value.
The Label/LabelList formatter has the signature:
```sh
  type LabelFormatter = (label: RenderableText) => RenderableText;
```
It takes a raw value and returns something renderable inside an SVG <text> element — typically a formatted string like 4000 → "4.0k". It never touches the tooltip.

The Bar/Line formatter on the other hand has a much richer signature:
```sh
  (value, name, entry, index, payload) => ReactNode
```
 and formats values shown in the HTML tooltip overlay.
Same prop name, completely different context and signature — hence the exception. Happy to update the wording to may be "transforms the label value for SVG text display" if that reads clearer.

shreedharbhat98: @PavelVanecek Just bumping this up incase if you have missed it!

PavelVanecek (approved):

PavelVanecek (approved):

PR #7297, opened 2026-05-04
fix(bar): use Math.round instead of bitwise truncation for bar positioning

### LINKED ISSUE #6115: Scatter and Bar components have misaligned X coordinates in ComposedChart

I'm trying to align custom `<Scatter>` shapes to their respective `<Bar>` shapes using `<ComposedChart>`, however I'm seeing slight differences between their X coordinates.

I had assumed that using `props.cx` within the `<Scatter>` shape alongside a predictable bar size would allow me to calculate the X coordinate of the `<Bar>` so that they align, however the values come out slightly different (see the demo's console output). I can't figure out a way around this.

Is this expected? Or am I trying to do something silly here?

The X coordinate of a `<Scatter>` component can be calculated to match the associated `<Bar>` component.

https://stackblitz.com/edit/react-pna3dnxs?file=src%2FExample.tsx

All versions/browsers.

EduardF1:
## Problem

In ComposedChart, Bar and Scatter components can have misaligned X coordinates due to inconsistent use of the bitwise OR (>> 0) operator for truncating fractional pixel values.

### Root cause

In combineAllBarPositions.ts, two places use >> 0 (bitwise right-shift) to truncate a float to an integer:

`	s
// Case 1 – user-set barSize path
const offset = ((bandSize - sum) / 2) >> 0;  // truncates toward zero

// Case 2 – auto-size path
if (originalSize > 1) {
  originalSize >>= 0;  // truncates toward zero
}
`

>> 0 always **floors toward zero** (i.e. truncates the decimal), while Scatter uses raw floating-point values for its cx. This mismatch causes visible sub-pixel (and sometimes full-pixel) misalignment when bar widths or gaps result in fractional offsets.

## Fix

Replace both truncations with Math.round(), which rounds to the nearest integer consistently with how other positional values are computed:

`	s
// Case 1
const offset = Math.round((bandSize - sum) / 2);

// Case 2
if (originalSize > 1) {
  originalSize = Math.round(originalSize);
}
`

Fixes #6115

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined bar sizing and centering with rounded layout calculations, improving alignment and producing slightly adjusted bar widths across chart types.
  * Fixed stacking and background layering so bars and backgrounds render in the correct order.

* **Tests**
  * Expanded and added regression tests for bar sizing, sub-pixel/zero-value edge cases, radial geometry, animations, tooltip payloads, and selector behavior.


PavelVanecek:
Do we have reproduction, screenshots perhaps, with before & after? Also could we cover this with an eslint rule to prevent regressions? Or unit test?

Edit: judging by the failures I guess the test part is covered.

EduardF1:
@PavelVanecek thanks for the review. Addressing your questions:

### Reproduction + before/after screenshots

Side-by-side reproduction of issue #6115 (`bandSize=91`, `barCategoryGap='10%'`, single Bar+Scatter where the Scatter dot's `cx` should sit at the bar's centre). The vertical red dashed line is the Scatter `cx`; the bar should be centred on it.

side-by-side

Before/after as separate images: [before.png](https://raw.githubusercontent.com/EduardF1/recharts/pr-7297-screenshots/before.png) · [after.png](https://raw.githubusercontent.com/EduardF1/recharts/pr-7297-screenshots/after.png)

Numeric proof from the same render (delta = `scatter cx − bar centre`):

| cat | before delta | after delta | bar width before/after |
|-----|-------------:|------------:|------------------------|
| A | +0.4 px | −0.1 px | 72 / 73 |
| B | +0.4 px | −0.1 px | 72 / 73 |
| C | +0.4 px | −0.1 px | 72 / 73 |
| D | +0.4 px | −0.1 px | 72 / 73 |
| E | +0.4 px | −0.1 px | 72 / 73 |

That matches @VIDHITTS's analysis on #6115 exactly: with `>> 0`, `originalSize` truncates `72.8 → 72` and the bar centre lands at `9.1 + 36 = 45.1` while Scatter `cx = bandSize/2 = 45.5` (0.4 px error). With `Math.round` we land on `73` and bar centre `45.6`, matching Scatter to within 0.1 px.

### Unit test coverage

Added a focused spec at [`test/state/selectors/combiners/combineAllBarPositions.spec.ts`](https://github.com/EduardF1/recharts/blob/fix/bar-position-rounding/test/state/selectors/combiners/combineAllBarPositions.spec.ts) that exercises both branches:

- The else-branch fractional `originalSize` (issue #6115 scenario): asserts `73`, not `72`.
- The if-branch centring `offset` for user-set `barSize`: asserts `Math.round`-based centring.
- A `len > 1` overflow guard: ensures sub-pixel `originalSize ≤ 1` is not rounded away.

I sanity-checked these on `main` (pre-fix) and they fail with the expected `−1` deltas, so they lock in the rounding direction.

### ESLint rule to prevent regressions

`>> 0` truncation patterns are easy to miss in review. I'd suggest enabling [`no-bitwise`](https://eslint.org/docs/latest/rules/no-bitwise) for the `src/` tree, with an opt-out comment on any genuinely intentional bit-twiddling. Not in this PR — happy to do it as a follow-up if you want.

### Other failing checks

- The 49 unit-test failures from the previous run were all hard-coded expected values that encoded the old `>> 0` truncation residue (e.g. `offset: 5.999999999999997` from a borderline number that now lands cleanly on `6`). Those are updated in the latest commit; suite is now green locally for `unit:lib` (5800 pass, no failures attributable to this PR).
- VR Tests fail on three browsers because the rendered pixels shifted; those snapshots need regenerating (Docker-only `npm run test-vr:update` per `test-vr/README.md`). I can't run that locally on Windows; happy to hand it off, or push a separate snapshot-update commit if you point me at the runner.

### CodeRabbit's `Math.floor` suggestion

CodeRabbit flagged that `Math.round` on `originalSize` can push `len * originalSize + (len-1) * realBarGap + 2 * offset` over `bandSize` and suggested `Math.floor`. I considered it and went with `Math.round`:

- For #6115's exact case `Math.floor(72.8) = 72` puts us back at the 0.4 px error this PR is meant to fix.
- The overflow CodeRabbit is worried about is bounded by `0.5 × len` px and is absorbed into the existing `barCategoryGap` offset zone (the bars don't actually paint outside `bandSize` because `maxBarSize` already clamps the practical worst case).

If you'd rather have `Math.floor` for the overflow guarantee even at the cost of partial alignment, I'll switch — but my read is `Math.round` is the better trade for the issue at hand.

EduardF1:
Pushed [3ef336b](https://github.com/EduardF1/recharts/commit/3ef336b) which clears the Unit Tests + Eslint/Typecheck failures from the previous run.

- **Eslint and Typecheck**: ran prettier on the seven test files I touched (whitespace only, net -49 lines). Eslint is now 0 errors, 12 pre-existing `no-console` warnings.
- **Unit Tests**: `tsc --project test/tsconfig.json` was failing on two patterns. (a) `selectBarRectangles` returns `BarRectangleItem[]` but the runtime rectangles spread the original chart-data fields (`uv`, `pv`, `xKey`, `name`, `amt`) into each item, which excess-property-check rejects. Added `as unknown as BarRectangleItem[]` on the four affected assertions to match what the runtime actually emits without re-modelling the type. (b) `selectAllBarPositions` returns `BarWithPosition[]` whose `stackId: StackId | undefined` is required, so I added explicit `stackId: undefined` to the two left/right-axis literals.
- All 368 tests across the eight modified files pass locally (`npx vitest run`).

**VR Tests** are still red, and I haven't attempted to update the baselines. The diffs are exactly what you'd expect from `>> 0` → `Math.round`: most snapshots show ~160 px (0.01 ratio) sub-pixel shifts where the centring offset moves by 0.5 px, plus a handful of 0.3–0.4-ratio diffs on the charts where the rounding direction flipped a bar by a full pixel. Those are intentional consequences of the fix.

If the maintainers are happy with the change, `/update-snapshots` should regenerate the baselines cleanly. Happy to also walk through any individual diff if you want a sanity check on a specific chart before that's run.

EduardF1:
Hi @PavelVanecek, the workflow run on commit [3ef336b](https://github.com/recharts/recharts/commit/3ef336b) is sitting at `action_required` (first-time-contributor approval gate), so the CI hasn't actually run since the fix push. Could you approve the workflow run when convenient? That commit covers the prettier autofix on the new spec files plus the typecheck fixture corrections, so we should know quickly whether anything else is left.

ckifer:
running

EduardF1:
Thanks for approving the workflows, @ckifer. CI is now running on commit [3ef336b](https://github.com/recharts/recharts/commit/3ef336b), and the long-tail integration tests (React 16/17/18/19 across npm/pnpm/yarn) are already green; Eslint+Typecheck and Omnidoc passed too. Only Unit Tests, VR Tests, and Website Build are still in flight.

@PavelVanecek, the screenshots posted yesterday cover the Math.round case you asked about. I'll keep an eye on the remaining checks and push fixes promptly if any of them surface real issues. Would appreciate a review once CI settles.

EduardF1:
Pushed [5f018b0](https://github.com/EduardF1/recharts/commit/5f018b0) refreshing the VR snapshots. The Math.round change shifted bar pixels by 1 - 2 px, which affected 102 PNGs across BarChart, RadialBarChart, ComposedChart and a few API-example tests; the remaining 876 snapshots regenerated to byte-identical content (deterministic docker rendering), so the diff is scoped only to bar-related views.

@ckifer / @PavelVanecek, this push is sitting at `action_required` again (first-time-contributor gate fires per push). Could you re-approve the workflow run when convenient? After this run, the only remaining variable should be Unit Tests, which already passed on the previous run.

PavelVanecek (approved):

PR #7369, opened 2026-05-26
RFC: Theming proposal

PavelVanecek:
Not intending to merge this as-is, just showing how it might work.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive theming system with `RechartsThemeProvider` and `useRechartsTheme` hook for customizing charts
  * Introduced three built-in themes: legacy, light, and dark
  * Enabled theme-driven styling for all chart components with color, typography, and visual attribute customization

* **Documentation**
  * Added complete theming guide with examples and integration patterns for popular UI libraries

[Review Change Stack](https://app.coderabbit.ai/change-stack/recharts/recharts/pull/7369?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


PavelVanecek:
@cloud-walker what do you think?

www/src/containers/Root.tsx
```diff
@@ -19,7 +19,9 @@ export function Root() {
 
   return (
     <ColorModeProvider store={colorModeStore}>
-      <RouterProvider router={router} />
+      <ColorModeRechartsThemeProvider>
```
PavelVanecek, line 22: This is the actual user-facing implementation - one provider means all charts will inherit the same theme. We don't have to change anything in any of the examples, and all current and future charts follow the same theme. This is the main benefit imho.

cloud-walker:
I was sick, I'll check it out soon!

PavelVanecek:
https://recharts.github.io/recharts/pr-7369/www/en-US/guide/theming/ here is the guide on staging - it still has rough edges, the Tooltip doesn't follow the theme, but it shows the idea

cloud-walker:
As u already know I'm hesitant on this solution, recharts will expose more APIs and we will need to maintain them, users will start asking changes about the theming config. That said, I agree it may be a good trade-off for recharts

As long as the theming config supports (in order of priority):

1. className: so that static CSS solutions like vanilla CSS, tailwind CSS, panda CSS can take advantage of their default way of styling stuff
2. CSS vars: with panda CSS for example you have a [token()](https://panda-css.com/docs/theming/usage#style-attribute) function to retrieve design tokens (value or CSS var name which is better) at runtime 

It will interop better

PavelVanecek:
Maintenance I can handle.

classNames are not included right now, I will add that.

CSS vars are there already:

```
export const cssVariableTheme: RechartsTheme = {
  colors: {
    surface: 'var(--chart-surface)',
    text: 'var(--chart-text)',
    mutedText: 'var(--chart-muted-text)',
    grid: 'var(--chart-grid)',
    border: 'var(--chart-border)',
    axis: 'var(--chart-muted-text)',
    tooltipCursor: 'var(--chart-border)',
  },
  series: {
    palette: ['var(--chart-1)', 'var(--chart-2)', 'var(--chart-3)'],
  },
};
```

with the `token` function I think this should work just fine. It won't work with styled components - I don't think we need to pursue that.

cloud-walker:
about the rest of the code, I'm not familiar enough with recharts impl. code, but I suppose is a react context state wiring thing if I notice something I'll leave a comment, I will be happy to resume the dark mode work on the docs after that ❤️

PavelVanecek:
Yeah don't worry about the code, I will rewrite it anyway. Look at the guide

www/src/components/GuideView/Theming/index.tsx
```diff
@@ -0,0 +1,896 @@
...
+        <h2>Theme anatomy</h2>
+        <p>
+          A Recharts theme is a plain object shaped like the exported <code>RechartsTheme</code> type. Shared tokens
+          come first, component-specific overrides come second.
+        </p>
+        <table className={styles.table}>
+          <thead>
+            <tr>
+              <th>Section</th>
+              <th>Typical values</th>
```
cloud-walker, line 567: with "Typical" we mean that custom values may be passed by the end developer?

So for colors for example, "mutedText" maybe "secondaryText"?

www/src/components/GuideView/Theming/index.tsx
```diff
@@ -0,0 +1,896 @@
...
+          If you only need local overrides, prefer a nested provider over cloning the whole theme. If you need a
+          reusable branded theme package, export a plain <code>RechartsTheme</code> object from your own codebase.
+        </p>
+      </section>
+
+      <section>
+        <h2>Handling multiple series</h2>
+        <p>
+          Multi-series charts must not depend on React child order or render order for color assignment. Recharts now
+          derives a stable series identity from <code>coalesce(dataKey, name)</code>.
```
cloud-walker, line 657: this coalesce thing is something known in recharts? Or is pseudo code?

www/src/components/GuideView/Theming/index.tsx
```diff
@@ -0,0 +1,896 @@
...
+          make it easy to share branded chart defaults across projects, CDNs, CMS pipelines, or design-token builds.
+        </p>
+        <SourceCodeEditor value={cssVariablesExample} />
+        <SourceCodeEditor value={jsonThemeExample} />
+      </section>
+
+      <section>
+        <h2>Integrations with popular UI libraries</h2>
+        <p>
+          Recharts currently ships a first-party helper for Ant Design. For the other libraries, the recommended
```
cloud-walker, line 707: this part seems to early, I mean we can add it later, after a first release focused on the theme provider primitives, no?

cloud-walker (changes requested):

PavelVanecek:
https://github.com/recharts/recharts/releases/tag/v3.11.0-canary.0

Issue #7362, opened 2026-05-25
[Feature] Allow YAxis to render ticks from explicit domain+ticks when data is empty/all-null


nlenepveu:
**Is your feature request related to a problem?**

When a chart's data is entirely null/empty for a window (a real product case for analytics dashboards: a metric tile that's "quiet" for an interval), YAxis renders no ticks even when both `domain` and `ticks` are explicitly supplied. This breaks the "consistent frame" pattern dashboards rely on: each tile renders identical chrome (axes, gridlines, tick ladder) whether it has data or not, so the eye can scan a grid of tiles in one pass.

Today, when one tile in a 6-tile grid is empty, that tile collapses into a different visual (axis line only, no ladder) and the eye has to re-parse "loading vs broken vs empty" every time. You pay that cognitive cost continuously for the lifetime of the product.

Repro: pass `data` where every value for the YAxis-bound dataKey is `null`. Set explicit `domain={[0, 100]}`, `ticks={[0, 25, 50, 75, 100]}`, `allowDataOverflow={true}`. Result: the `<g class="recharts-yAxis">` is in the DOM but contains zero tick-label children. v2 used to emit the ladder from `domain` alone.

(Tested on v3.8.1. PR #4433 / issue #4426 restored the axis *line* for the empty case but explicitly left tick rendering tied to data — this request is the missing half.)

**Describe the solution you'd like**

When `YAxis` has all of:
- `type="number"`
- explicit `domain` (e.g. `[0, 100]`)
- explicit `ticks` (or `tickCount`)
- `allowDataOverflow={true}` (opt-in signal that the user really means "use my domain, ignore data")

…compute the scale from `domain` alone and render the ticks from the `ticks` prop, independent of whether any data point is numeric. That treats `domain`+`ticks`+`allowDataOverflow` as an explicit "scale spec independent of data," which is the missing primitive today.

This would let dashboards drop the workaround of injecting a hidden synthetic numeric series (`<Line dataKey="__anchor" stroke="transparent" />` with `__anchor` values ramping 0→100) just to satisfy the scale calculator.

**Describe alternatives you've considered**

1. **Hidden anchor series** (what we ship today): inject a synthetic dataKey with numeric values, render a transparent Line bound to it. Works, but leaks: the series appears in the tooltip payload, the synthetic key appears in legend lookups, and any reducer/scale-derivation code that walks the data has to know to skip `__anchor`. Each leak is its own bug.

[...]

PR #7393, opened 2026-05-31
fix(YAxis): render explicit ticks when a non-literal domain can't resolve on empty data (#7362)

nlenepveu:
> Draft — proposes the approach @PavelVanecek suggested on #7362 (derive the domain from explicit `ticks` when data is absent). Opening as a draft for a steer on the approach before polishing.

## What

Follow-up to #7384 (the failing repro). When a numeric axis supplies an explicit `ticks` array but its `domain` cannot be resolved without data — a **function** bound, or `'auto'`/`'dataMin'`/`'dataMax'` — the scale never builds on empty/all-null data, so the axis renders **zero ticks**. A literal numeric domain already works; this extends that to derived domains.

This avoids the typing problem raised on the issue (the `AxisDomain` function form is typed `(NumberDomain, boolean) => NumberDomain`, so feeding it `undefined`/`NaN` would be a breaking change): instead of calling the user's function without data, we fall back to the explicit ticks' extent.

## How

`combineNumericalDomain` falls back to `[Math.min(...ticks), Math.max(...ticks)]` of the explicit numeric ticks **only when all of**:
- the user domain could not be resolved without data (`numericalDomainSpecifiedWithoutRequiringData` → `undefined`), and
- `parseNumericalUserDomain` also returned `undefined` (no data domain — empty/all-null data and no domain-extending reference elements), and
- `allowDataOverflow` is set, and
- the axis is `type="number"` with ≥1 numeric tick.

So it **never** overrides a resolvable literal domain, **never** engages when data is present, and leaves category / polar / tooltip paths untouched.

- adds `combineNumericTicksDomain` + `selectNumericTicksDomain` (the ticks `[min,max]` extent)
- threads it as an optional input to `combineNumericalDomain` (polar/tooltip pass nothing → unchanged)
- **does not** touch `parseNumericalUserDomain` / `numericalDomainSpecifiedWithoutRequiringData` — their "undefined without data" contract is preserved
- converts the #7384 `it.fails` repro into passing tests + adds edge-case coverage (unsorted/non-numeric/single ticks, tickCount-only, literal-wins, no-overflow, data-present guards)

## Test plan

- [x] `YAxis.7362.derivedDomain.spec.tsx` — 13 cases (positive + negative guards) pass
- [x] Affected selectors/axis/scatter/polar inventory + full `unit:lib` suite green locally (vitest 4)
- [x] format / lint / typecheck clean
- [ ] Maintainer steer on the approach
- [ ] Open question: should the fallback also apply without `allowDataOverflow`? (kept it required, for parity with the literal-domain path)

Refs #7362. Follows #7384.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where axis ticks would not render properly when no data is available but explicit numeric ticks are specified. The axis domain now intelligently falls back to the explicit tick range when `allowDataOverflow` is enabled.

* **Tests**
  * Expanded test coverage to verify axis behavior with explicit ticks on empty or null data scenarios.

### DIFF
--- a/src/state/selectors/axisSelectors.ts
+++ b/src/state/selectors/axisSelectors.ts
@@ -1094,6 +1094,7 @@ export const combineNumericalDomain = (
   referenceElementsDomain: NumberDomain | undefined,
   layout: LayoutType,
   axisType: AllAxisTypes,
+  numericTicksDomain?: NumberDomain | undefined,
 ): NumberDomain | undefined => {
   if (domainFromUserPreference != null) {
     // We're done! No need to compute anything else.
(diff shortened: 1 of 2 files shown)


PavelVanecek (approved):
I am not quite certain if allowDataOverflow should make a difference here. We're talking about a chart that has all null data anyway, the overflow makes no sense. Maybe let's have the new code only when it's true so that we affect fewer people? Someone else might rely on the old behaviour.

nlenepveu:
Agreed — let's keep it gated on `allowDataOverflow` (that's what the PR does). Thanks for the review!

PR #5934, opened 2025-06-09
Feature/add zoom and pan support improved brush (vertical) and minimap

Fefedu973:
# feat: Implement Advanced Zoom, Pan, and Interaction Controls  

## Description
This pull request introduces a comprehensive and highly-configurable **zoom and pan system** for all Cartesian charts. It also includes major enhancements to the **`Brush`** component, a new **`Minimap`** component, and improved mobile/touch interactions.

### Core changes
- **`ZoomPanContainer`** – A new context provider that wraps chart children and manages all zoom/pan state and logic. It handles mouse-wheel, drag, and keyboard events for seamless navigation.  
- **`Minimap` component** – Gives a high-level overview of the entire chart, showing the current viewport and allowing quick navigation by clicking or dragging.  
- **Major `Brush` overhaul**
  - New `useZoomPan` mode lets the brush control the chart’s zoom state instead of merely slicing data.
  - Added `layout="vertical"` for Y-axis brushing.
  - Enhanced touch support (pinch-to-zoom, scroll-to-zoom).
  - Supports multiple brushes on the same chart (e.g., one per axis).
  - Introduced an isolated `BrushPreview` container to prevent rendering conflicts.  
- **State management** – Added a `zoomSlice` to the Redux store to track global zoom state (`scaleX`, `scaleY`, `offsetX`, `offsetY`).  
- **Improved axis interactions** – `CartesianAxis` now supports direct drag/scroll on axis labels to pan and zoom.  
- **Animation control** – Disables animations on graphical elements (`Area`, `Bar`, `Line`, `Scatter`) during zoom/pan for smooth performance (initial render excluded).  
- **Refined event handling** – Greatly improved touch and pointer event handling to support complex gestures (pinch, two-finger pan, double-tap-drag) and to resolve conflicts between tooltips and panning.

## Related Issue
Resolves **#710
**

## Motivation and Context
Large or dense datasets were previously hard to explore interactively. These changes provide users with built-in, powerful tools for zooming, panning, and brushing, turning static charts into dynamic, responsive interfaces suitable for advanced data-exploration use cases.

## How Has This Been Tested?
* **Unit tests** – Added for core `ZoomPanContainer` logic (`test/context/zoomPanContext.spec.tsx`) to verify zoom, pan, and reset behaviors.  
* **Storybook stories** –  
[...]

PR #7410, opened 2026-06-05
feat(zoom): zoom/pan viewport state foundation (1/n)

Fefedu973:
## Description

This is the first, deliberately small and self-contained step of re-building the cartesian
zoom/pan feature (originally #5934) as a series of independently reviewable PRs.

It adds **only the state foundation** and is **completely dormant**: nothing reads the new
state yet, so every chart renders exactly as it does today. There is zero behavior change.

What's included:

- `src/util/zoom/viewport.ts` — pure, dependency-free viewport math. Zoom/pan is represented
  as a **normalized `[0, 1]` window per axis** (`AxisViewport`) rather than the absolute pixel
  `scale`/`offset` the original prototype used. This is resize-stable and scale-agnostic (it
  applies the same way to linear, band and point scales). Helpers: `clampViewport`,
  `panViewport`, `zoomViewportAround` (zoom around a focus point).
- `src/state/zoomSlice.ts` — a Redux slice holding one `AxisViewport` per dimension, defaulting
  to the full viewport (== not zoomed). Actions: `setAxisViewport`, `setZoom`, `resetZoom`.
- `src/state/selectors/zoomSelectors.ts` — `selectZoom`, `selectAxisViewport`, `selectIsZoomed`.
- the reducer wired into the store.

## Related Issue

Re-implementation of #5934 (zoom & pan), original feature request #710.

## Motivation and Context

#5934 added the whole feature in one ~6.8k-line, tightly-coupled PR which, per maintainer
feedback, "needs to be done in smaller chunks" and "would need to start over essentially".
This series does exactly that: each PR is small, non-breaking and unit-tested, and several are
intentionally dormant until a later PR turns the feature on — as suggested in
[this comment](https://github.com/recharts/recharts/pull/5934#issuecomment-3416769135)
("we can add the changes required to disable animation upon zoom without actually having zoom
work yet ... write unit tests for each piece as we go").

**Architecture note / request for early feedback.** I represent zoom state as a normalized
per-axis viewport window and plan to apply it at the **axis range** level (feeding
`selectAxisRangeWithReverse` / `combineConfiguredScale`), rather than monkey-patching the d3
scale like the prototype did — that approach is no longer compatible with the current immutable
`RechartsScale` abstraction. Confirming this representation now, before more is built on top,
would de-risk the rest of the series.

Planned follow-up PRs, each on its own: (1) **this** state foundation → (2) suppress animation
while zoomed → (3) apply the viewport to axis scales (still identity by default, so still no
visible change) → (4) cartesian interaction layer (wheel / drag / double-click / keyboard)
behind a `zoom` prop, which is the PR that actually turns the feature on → (5) touch & pinch →
(6) minimap → (7) brush ↔ zoom integration → (8) polar & standalone charts. The vertical-brush
feature (also part of #5934, and independent) can land separately.

## How Has This Been Tested?

New unit tests, all green:

- `test/util/zoom/viewport.spec.ts` (viewport math, incl. edge/clamping cases)
- `test/state/zoomSlice.spec.ts` (reducer behavior and normalization)
- `test/state/selectors/zoomSelectors.spec.ts`

The full `test/state` suite (520 tests) still passes, and `tsc` (lib + test), `eslint` and
`prettier` are clean. There is nothing to verify visually because the state is dormant.

## Types of changes

- [x] New feature (non-breaking change which adds functionality) — dormant foundation only

## Checklist:

- [ ] I have added a storybook story or VR test — N/A, nothing is rendered yet; the story will
      come with the interaction-layer PR.

## Summary by CodeRabbit

* **New Features**
  * Added zoom & pan controls for charts with independent x/y viewport management and state integration.
  * Exposed utilities for normalized viewport math (clamping, panning, zooming) and selectors to read zoom state.

* **Tests**
  * Added comprehensive tests covering viewport utilities, state reducers, and selectors.

* **Documentation**
  * New Storybook docs describing the zoom/pan API, usage, and behavior expectations.


PavelVanecek:
Much better this way thank you. Please remind me of your plan to render this thing. I understand we're going to zoom the range instead of domain? How will the render work?

PavelVanecek:
If I may propose, would you like to start by writing a guide on how to use the new zoom features? Kinda work from the other end. First describe the final state as a documentation/ tutorial. Then work backwards to implement

Fefedu973:
Thanks! Glad the smaller shape works.

**Yes — we zoom the range, not the domain.** Recharts already maps every datum to a pixel through one per-axis scale (`domain → range`), and zoom hooks into that single spot:

1. The normalized viewport `{ x, y }` (each a `[0,1]` window, default = full) lives in the small Redux slice from this PR.
2. When a window isn't full, I stretch the axis **range** (the pixel interval the scale maps into) by `1 / windowWidth` and shift it so the visible slice fills the plot area — the **domain is left untouched**. Concretely it slots in as a `selectZoomedAxisRange` step between `selectAxisRangeWithReverse` and `combineConfiguredScale`; a full window → identity → zero change.
3. Every renderer (line / bar / area / scatter / grid / ticks / reference lines) asks that same scale for positions, so they all zoom automatically; geometry outside the plot area is hidden by the clip path Recharts already applies. **No graphical component needs to learn about zoom.**
4. Tooltip / cursor / keyboard use the scale's inverse, which is derived from the same stretched range, so hit-testing stays consistent.

Range over domain because it's uniform across linear / time / **band** scales (a domain window can't express "show 2½ categories", a stretched range can), it allows smooth sub-pixel panning even on categorical axes, it keeps the change tiny and decoupled (one selector instead of threading zoom through every renderer), and it's compatible with the immutable `RechartsScale` — we feed it a different range and never mutate or monkey-patch the scale.

**And yes — great call on docs-first.** I wrote the usage guide and pushed it to this PR: `storybook/stories/API/ZoomAndPan.mdx`. It describes the final public API — a single `zoom` prop (`boolean | { axis, minZoom, maxZoom, wheel, pan, dragToZoom, doubleClickReset, keyboard, animate, viewport, onZoomChange }`), with a controlled `viewport` expressed as those same normalized `[0,1]` windows — and the render approach above. I'll work backwards from it.

Build order from here, each as its own small PR: (2) disable animation while zoomed → (3) the `selectZoomedAxisRange` step (still identity by default, so still no visible change) → (4) the cartesian interaction layer (wheel / drag / double-click / keyboard) behind the `zoom` prop, which is the PR that actually turns the feature on.

Fefedu973:
By the way, it would make the development workflow much easier for me if you could enable the stacked PR feature: [https://github.github.com/gh-stack/](https://github.github.com/gh-stack/)

Fefedu973:
I’ve already made significant progress, but I can’t push my changes until the stacked PR feature is enabled. Otherwise, it would create the same kind of mess as the previous pull request, with too many changes bundled together at once.

I’ve been very careful to split my work into small, focused, easily reviewable stacked pull requests, each building on the previous one. This should make the review process much simpler, and the code is now much cleaner, better documented, and more modular.

PavelVanecek:
gh stacks are not available to me I'm afraid, it says closed beta.

Let's discuss the docs before merging a stack of pull requests though. First thanks for that, would you mind to port that into a guide instead of a story? Please follow https://github.com/recharts/recharts/blob/main/.github/skills/guide/SKILL.md. Also please write the guide in the final state, as it will be when all your stacked PRs are merged. Now it mentions "Roadmap" - is that the full feature? Do we release the first version to npm and then start looking at touch gestured? If you separate the guide in its new PR then we can have the other PR open, discuss the API and merge state changes in the meantime.

For the feature itself. We already have zoom, kinda sorta. It's hiding inside https://recharts.github.io/en-US/api/Brush/. How will this new zoom interact with Brush? Have you considered using the same API?

Touch gestures are not discussed. There's ton of things to consider there - how will touch zoom interact with scroll? How will it interact with Tooltip, and native browser zoom? How are these things customizable?

There are multiple gesture libraries, right? If a user wants to bring their own, or customize the gestures completely, how can they do that?

The guide mentions 0,1 as fractions of a viewport, and mentions windowWidth. Why relate this to windowWidth? Does it mean that resizing the window will zoom the chart? Or is the `window` referring to something else?

"keyboard navigation and the live-region tooltip are unaffected" when zoomed I would expect the keyboard navigation to be affected, what do you think? When zoomed in then only the currently visible (zoomed in) elements should be navigable via keyboard. Or is there a different best practice?

Do we allow grouping graphical items? The more zoomed in, the more details we show? For example scatter; once zoomed in, one large scatter dot breaks down into smaller scatter dots (because now there is more space). How do we allow that? What will API for this look like?

Did I miss anything?

Fefedu973:
Thanks for the thorough review, lots of good questions in here.

**gh stacks / closed beta**: ah right, it's a private preview so it won't work until your account is enabled. The sign-up form is here: https://gh.io/stacksbeta. Either way it's fine, I'll open the PRs individually so nothing is blocked by it.

**Docs as a real guide, in its own PR**: yep, going to do that, and I'll write it for the final state rather than the current partial one.

On the "Roadmap" section: that was just me tracking the stack, not really part of the feature, so I'll drop it from the guide.

**Brush, same API?**: yeah, that's exactly where I'd like to go. The Brush already does a kind of zoom today by slicing the data (start/end index), so rather than inventing a parallel thing, the direction I'm considering is to let the Brush act as a viewport editor over the same zoom state: a `mode: 'zoom'` where the brush window maps to the zoom x window and stays in sync with the other gestures, instead of truncating the data. The Minimap would be the same primitive in 2D. So Brush, Minimap and zoom could all share one piece of state (the viewport) and one callback (`onZoomChange` / a controlled `viewport`), with the default Brush staying exactly as it is today (slice mode).

The open question I'd love your take on: should the public surface be the `viewport` (fractions), or should we keep Brush's index-based API as the canonical one and have zoom derive from it? That feels like the main API decision, and a good thing to settle in the guide PR before I build the Brush side.

**"window" / windowWidth terminology**: sorry, that name is confusing. By "window" I mean the visible slice of an axis, two numbers in `[0, 1]` (`{ start, end }`), and `windowWidth` is just `end` minus `start` (the visible fraction). It has nothing to do with the browser window, and resizing the browser does not zoom. Storing it as fractions is actually what keeps it resize stable (the old prototype used absolute pixels and broke on resize). I'll rename it to `viewport` / "visible fraction" in the guide so it isn't ambiguous.

**Touch**: this part isn't in this PR, it's further up the stack, but here's the design I'm planning so we can poke holes in it:
- vs page scroll and native pinch: the gestures we handle call `preventDefault` (we can also set `touch-action: none`) so the page doesn't scroll and the browser doesn't pinch zoom while you're interacting with the chart.
- vs Tooltip: one finger is left entirely to the tooltip (a tap still shows it). Zoom only uses two fingers (pinch and two-finger pan), plus double-tap and double-tap-then-drag (the Google Maps gesture), so a single finger never zooms and there's no conflict.
- customisable: `touch` can be turned off, and see the next point.

There's real nuance here (should single-finger pan be opt-in, etc.), and I'm happy to set the defaults to whatever you prefer. That's part of why I'd lean towards shipping touch as a fast-follow rather than in v1.

**Bring your own / custom gestures**: the way I'd expose that is through the controlled API. Turn off the built-ins you don't want (`wheel: false`, `pan: false`, `touch: false`, ...) and drive the zoom from your own gesture library through a controlled `viewport` plus `onZoomChange`. Each gesture is its own small module internally, so if there's demand we can also expose them directly or add an imperative handle later. So full control is doable through controlled state, and deeper hooks are a small addition on top.

**Keyboard nav when zoomed**: good catch, you're right that I overstated it. The sensible behaviour is to scope arrow-key navigation to the visible window when zoomed (you shouldn't land on off-screen points, and ideally panning follows the active point to keep it in view). I'll fix that line and make the nav read the same viewport state. I already planned to add, like in the old PR a feature to automatically scale the axis and center the view to the graph.

**Level of detail / grouping (scatter splitting, etc.)**: I'd keep that out of the core, since it's really data aggregation rather than the zoom transform itself. The API does give you a way to do it in userland: `onZoomChange` hands you the visible viewport, so you can recompute or downsample your `data` for that range. A built-in aggregation helper could come later as its own discussion.

I'll start with the guide PR so we have a concrete place to settle the API.

PavelVanecek:
Sounds good to me. I'll just add couple things:

Minimap in Recharts source is called "panorama" and it's not directly publicly accessible but it's what Brush renders inside if you tell it to. Just now I am realizing we don't actually have any example on the website! But there are couple in the storybook: https://recharts.github.io/storybook-static/index.html?path=/story/api-cartesian-brush--panorama-with-bar so support for this is already baked in

I am contemplating splitting the various features as components. For example we would export `<PinchToZoom />` and `<MouseWheelZoom />` components and others. Would that be intuitive? These components would register the appropriate handlers. Or perhaps these better should be hooks? I'm thinking out loud. I expect that people will want to either integrate the zoom behavior with their existing UI, or write completely new custom UI. Some people will want implicit mouse wheel support, some people will want only clickable buttons I'm sure.

If I understand correctly the viewport that you are describing, in Recharts terminology that would be a "plot area" and we have a hook for that too: https://recharts.github.io/en-US/api/usePlotArea/

The default definitely has to be the same it is today, meaning: no zoom, no gestures, until you explicitly enable that either via a prop or component or hook. Enabling it by default would make it a major version change I think.

Can we have an overview of some popular gesture libraries and see which API they expose? Maybe if we mimic what they are doing we learn a thing or two, and make integration easier.

I had AI do a deep research on this topic a few months ago, here it is: 
[Recharts Zoom and Scroll Feature Research.pdf](https://github.com/user-attachments/files/28681246/Recharts.Zoom.and.Scroll.Feature.Research.pdf) feel free to use it but keep in mind this is Gemini 3.0 so read it with a grain of salt. It completely missed all the work on Recharts 3.x branch for example.

Fefedu973:
Sounds good, and thanks for all of this. It actually nudges my order of work, in a good way.


Rather than pushing on Brush and Minimap next, I want to land a solid, modular foundation first and get the public API right, because everything else (Brush-as-viewport-editor, Minimap, follow-line / auto-scale, and people's own custom UI) builds on it. Your components-vs-hooks question is the crux of that, so let me answer it properly.


**Components or hooks: both, layered.** The cleanest shape I see is three layers over one shared piece of state (the viewport + `onZoomChange`):

- low-level hooks for fully custom UI: `useMouseWheelZoom()`, `usePinchZoom()`, `useZoom()` to read/control the viewport,

- thin declarative components on top of those hooks for the common case: `<MouseWheelZoom />`, `<PinchToZoom />`, `<DragToZoom />`, `<ZoomPanKeyboard />`, dropped into the chart like `<Brush />` or `<Tooltip />`,

- and for the "only buttons" crowd, they add none of those and just drive a controlled `viewport` from their own buttons via `onZoomChange`.


That covers your three cases (implicit wheel, fully custom, buttons only) with one state model. Each gesture is already its own isolated module in my implementation, so wrapping each as a hook plus a component is mostly plumbing. What's left to settle is naming and granularity. (MUI X went the other way, one config object plus a controlled `zoomData`; components feel more Recharts-y to me, but I'm easy.)


**Docs.** I'll write the guide around exactly this composable API, in the final state, as its own PR following the SKILL.md format, and use it as the place to lock the API down before building the rest. That's the "work backwards from the docs" you suggested, and it doubles as the spec for the foundation.


**Panorama / Minimap.** Good to know it's already in there. Reusing the panorama for the Minimap (and for the Brush's zoom mode) is the plan, so the Minimap is "panorama plus a draggable viewport window" rather than anything new.


**viewport vs plot area.** Worth clarifying, since they're different things. The plot area (the `usePlotArea` rect) is the pixels where data is drawn, and zoom never changes that. What I've been calling the viewport is which slice of the data is currently visible inside that rect, per axis, as a fraction in `[0, 1]`. So it's closer to "visible data range" than plot area. Since the word collides, I'll rename it in the guide, probably to a per-axis `range` / "visible window".


**Default off.** Completely agree. No zoom and no gestures until you opt in (prop, component or hook), and the current Brush is untouched. On-by-default would be a breaking change, so it stays off.


**Gesture libraries overview.** Short version of the main ones and how they expose things:

- @use-gesture/react: hooks (`useDrag`, `useWheel`, `usePinch`, `useGesture`) that return a `bind()` you spread onto an element. Handlers get a rich state object (movement, velocity, pinch distance and angle), with per-gesture config for bounds, rubberbanding, modifier keys, `pointer: { touch }`, and passive/event options. The most React-native, and the closest match to the hooks layer above.

- Hammer.js: imperative, `new Hammer(el)` plus recognizers (Pan, Pinch, Tap, Press, Swipe) and `.on('pinch', ...)`. Framework-agnostic, older, no React story.

- interact.js: imperative, `interact(el).draggable({...}).gesturable({...})`, with inertia, restrictions and snapping.

- react-aria's `useMove` / `usePress` are also worth a look: lower-level and pointer-normalized, but strong on accessibility.


The pattern worth borrowing is use-gesture's: small composable gesture units, a consistent state object, and an easy opt-out, which is exactly what maps onto the hooks-and-components layering. I'll do a proper writeup with the exact signatures in the guide PR so we can decide how much to mirror.


Thanks for the research PDF too, I'll read through it (with the grain of salt about the 3.x branch).


So my immediate plan: the guide PR describing the composable API, then the state-injection and interaction layers underneath it. Brush, Minimap and follow-line / auto-scale all come after the foundation is settled.

Fefedu973:
@PavelVanecek The feature is now complete on my end, including the Brush rework, the Minimap, polar/standalone charts, docs and tests (~5-6k lines of source code, more in tests and docs). Here is how I set things up so the review stays manageable:

- **Integration preview: #7450 (draft, not meant to be merged).** It shows the finished feature as a whole. You can play with it and follow the commit trail to see how it was built step by step, which should be much easier than reviewing everything in one block.
- **The work is split into 12 stacked branches** on my fork (`zoom/01-state-foundation` ... `zoom/12-docs`), each one a small reviewable layer building on the previous. The full map with links is in the #7450 description. This PR (#7410) is layer 1.
- **Review process:** since Stacked PRs are not available to you (closed beta), I will open the PRs one at a time as each predecessor merges, so every PR is a small diff against main. As soon as this one merges I open layer 2 (applying the viewport to the axis range, still dormant). If the beta ever gets enabled for the repo, I can submit the whole stack at once instead.
- The guide in #7429 is up to date with the final implemented API, so the API discussion can keep happening there.

PavelVanecek:
Thanks @Fefedu973 Stacks are still not available. I can handle 12 small PRs or I can handle 1 PR with 12 commits, I can read code. I just need to find a bit of time for it. I'll start with the guide

Fefedu973:
Okay nice so you will be able to find the guide here: https://github.com/recharts/recharts/pull/7429 and if you want 1 PR with 12 commits wa can make this PR the main one then: https://github.com/recharts/recharts/pull/7450

PR #7428, opened 2026-06-07
fix(Rectangle): allow zero-dimension rendering for animation t=0 frames

### LINKED ISSUE #7377: Rectangle returning null at width/height === 0 strands Bar mount-animations on missed t=0 frames

In `src/shape/Rectangle.tsx`, the early-return on line 121 (`es6/shape/Rectangle.js`) bails out whenever `width === 0` or `height === 0`:

```js
if (x !== +x || y !== +y || width !== +width || height !== +height
    || width === 0 || height === 0) {
  return null;
}
```

This makes `<Bar>` mount-animations brittle. `Bar.renderRectanglesWithAnimation` interpolates from `width: 0` (or `height: 0` for horizontal layouts) at `t = 0` to the final value at `t = 1`. The `t = 0` frame passes a zero dimension into `<Rectangle>` → `<Rectangle>` returns `null` → no `<path>` is emitted that frame.

Normally this is fine: `react-smooth`'s `Animate` ticks `t` forward on the next animation frame, and the bar appears. But if anything stalls `Animate` (or causes the component to remount with `t = 0` and not re-tick — see ⚠ below), the `<g class="recharts-bar-rectangle">` placeholders **stay permanently empty** and the bars never render.

We hit this on production in a real app: BarChart inside ResponsiveContainer inside conditionally-rendered tab content. DOM evidence:

```
TopItemsBar  (layout="vertical"):  recharts-bar-rectangle <g> × 10  →  all empty
DaysOosHistogram (layout default): recharts-bar-rectangle <g> × 5   →  all empty
SupplierDonut (PieChart):          works fine (Sector renders even at endAngle === startAngle)
```

Axes, ticks, and grid all render correctly. React Fiber inspection confirms `Bar.curData` has the right `{ x, y, width, height }` for every entry — the geometry is computed, the rendering just never happens because the animation pipeline never advances past `t = 0`.

Setting `isAnimationActive={false}` from JSX **does not** help — the prop arrives at `Bar` as `true` (BarChart internally clones `<Bar>` and `Bar.defaultProps.isAnimationActive = !Global.isSsr` reapplies), so `renderRectangles()` still takes the animation branch.

⚠ I haven't fully traced *why* the `Animate` stalls in our specific setup. But that's a separate bug. The point of this report is that `<Rectangle>` returning `null` at width/height = 0 makes the surface behavior brittle — a single missed `Animate` frame leaves bars permanently invisible.

`<Rectangle>` should emit a `<path>` element even when `width === 0` or `height === 0`. A zero-size path is harmless visually, and it preserves the DOM structure so subsequent `Animate` frames replace it with the final shape. The existing `width !== +width || height !== +height` NaN/non-numeric guard is fine — it's the explicit `=== 0` checks that compound animation problems.

I don't have a minimal hosted repro for the upstream Animate-stall (haven't traced the trigger), but the relevant code is all in this repo:

- `src/shape/Rectangle.tsx` — the early-return at the start of the component body
- `src/cartesian/Bar.tsx` `renderRectanglesWithAnimation` — the interpolator that starts at `width: 0`

DOM-level confirmation from our production app: https://github.com/mlaia/predictor_frontend/pull/110 (see the PR description's "Root cause" section — full diagnosis with React Fiber inspection results).

## Suggested fix

Drop the `width === 0 || height === 0` clause from the early return:

```diff
- if (x !== +x || y !== +y || width !== +width || height !== +height
-     || width === 0 || height === 0) {
+ if (x !== +x || y !== +y || width !== +width || height !== +height) {
    return null;
  }
```

`getRectanglePath(x, y, 0, 0, radius)` returns a valid (zero-size) SVG path string `"M x,y h 0 v 0 h 0 Z"` — no crash. We've been running this patch via `patch-package` against `recharts@2.15.4` in production for a few hours, and:

- Bar charts that were rendering empty `<g>` shells now render correctly
- No visual regression on charts that were previously working
- No tests break in our suite or in the chart components that use `<Rectangle>`

Patch file: [predictor_frontend/patches/recharts+2.15.4.patch](https://github.com/mlaia/predictor_frontend/blob/main/patches/recharts+2.15.4.patch)

- **Recharts:** 2.15.4 (latest 2.x). The `|| width === 0 || height === 0` guard exists in 2.15.x; earlier versions had only the NaN/non-numeric check. Bisecting it isn't trivial but `git log -p src/shape/Rectangle.tsx` should locate when it was added.
- **Browsers:** Chrome 137 (Stable channel), confirmed on macOS Sequoia 15.x. Not browser-specific based on the React-Fiber-level diagnosis.
- **React:** 18.3.1, no `StrictMode` wrapping (so not the double-mount issue).
- **Worked in previous versions:** We haven't bisected — we were already on 2.15.4 when we hit the bug. But the guard hasn't always been there, so a `git blame` on the line would identify the regressing commit.

Thanks for maintaining Recharts! Happy to send a PR with the one-line fix if it'd be useful.

mixelburg:
Fixes #7377

The `Rectangle` component returns null when width or height is zero, which breaks Bar mount-animations. Bar.renderRectanglesWithAnimation interpolates from zero dimensions at t=0 to final values at t=1 — when Rectangle returns null at t=0, no path element is emitted and if the animation stalls, bars stay permanently invisible.

The zero-dimension guard in Rectangle is redundant — Bar already filters zero-dimension rectangles at the selector level in computeBarRectangles (for sparse stacked chart performance), so by the time a Rectangle renders with real data, the dimensions are non-zero. The only case where zero dimensions reach Rectangle is during animation interpolation, where they should produce a harmless zero-size path.

Removed the `|| width === 0 || height === 0` clause from the early return. The existing NaN/non-numeric guards remain. Updated the corresponding test and removed stale snapshot.

## Summary by CodeRabbit

* **New Features**
  * Rectangles with zero width or height are now rendered instead of being skipped.
* **Tests**
  * Updated rectangle tests to assert rendering for zero-dimension cases.

### DIFF
--- a/src/shape/Rectangle.tsx
+++ b/src/shape/Rectangle.tsx
@@ -220,7 +220,7 @@ export const Rectangle: React.FC<Props> = rectangleProps => {
   const animationIdInput = useMemo(() => ({ x, y, width, height, radius }), [x, y, width, height, radius]);
   const animationId = useAnimationId(animationIdInput, 'rectangle-');
 
-  if (x !== +x || y !== +y || width !== +width || height !== +height || width === 0 || height === 0) {
+  if (x !== +x || y !== +y || width !== +width || height !== +height) {
     return null;
   }
 
--- a/test/shape/Rectangle.spec.tsx
+++ b/test/shape/Rectangle.spec.tsx
@@ -34,7 +34,7 @@ describe('<Rectangle />', () => {
     expect(container).toMatchSnapshot();
   });
 
-  it("Shouldn't render anything when height === 0 || width === 0", () => {
+  it('Should render zero-size path when height === 0 || width === 0', () => {
     const { container } = render(
       <Surface width={400} height={400}>
         <Rectangle x={50} y={200} width={80} height={0} radius={5} fill="#ff7300" />
@@ -42,8 +42,7 @@ describe('<Rectangle />', () => {
       </Surface>,
     );
 
-    expect(container.querySelectorAll('.recharts-rectangle')).toHaveLength(0);
-    expect(container).toMatchSnapshot();
+    expect(container.querySelectorAll('.recharts-rectangle')).toHaveLength(2);
   });
 
   it("Shouldn't render any path when x, y, width or height is not a number", () => {
--- a/test/shape/__snapshots__/Rectangle.spec.tsx.snap
+++ b/test/shape/__snapshots__/Rectangle.spec.tsx.snap
@@ -101,17 +101,3 @@ exports[`<Rectangle /> > Shouldn't render any path when x, y, width or height is
   </svg>
 </div>
 `;
-
-exports[`<Rectangle /> > Shouldn't render anything when height === 0 || width === 0 1`] = `
-<div>
-  <svg
-    class="recharts-surface"
-    height="400"
-    viewBox="0 0 400 400"
-    width="400"
-  >
-    <title />
-    <desc />
-  </svg>
-</div>
-`;


mixelburg:
Closing due to inactivity (42 days with no reviews). Happy to reopen if there's interest.

PR #7429, opened 2026-06-08
docs(zoom): Zoom & Pan guide

Fefedu973:
Per the discussion on #7410, this ports the zoom/pan documentation from the Storybook story into a proper **website guide** (following `.github/skills/guide/SKILL.md`), as its own PR so the API can be discussed in parallel with the small state-change PRs.

It is written in the **final state** (the API as it will be once the whole feature lands), to work backwards from, as you suggested.

### What it covers
- The normalized per-axis **viewport** (`{ x?, y? }`, each a `[0, 1]` window of the axis), and why fractions rather than pixels or data values (resize-stable, and works the same for linear / time / band scales).
- A **composable, opt-in API**: interaction components you add as children, each tree-shakeable: `<MouseWheelZoom/>`, `<PanOnDrag/>`, `<DragToZoom/>`, `<ZoomPanKeyboard/>`, `<PinchZoom/>`, `<AxisZoom/>`, `<ZoomScrollbar/>`, `<DoubleClickReset/>`.
- **Controlled & uncontrolled** state (`zoomViewport` / `onZoomViewportChange` / `defaultZoomViewport`).
- **Custom UI / bring-your-own gestures** via a `useZoom()` hook and the low-level per-gesture hooks the components are built on.
- Touch, accessibility (keyboard navigation scoped to the visible window when zoomed), and the **off-by-default** guarantee (nothing zooms until you opt in; the existing Brush is unchanged).
- How it renders: stretch the axis **range** (not the domain) at one selector, so every renderer zooms automatically and nothing needs to learn about zoom.

This is a **proposal for the public API**; naming and granularity are open for discussion, which is exactly what this PR is for. The "Roadmap" section from the old story is intentionally dropped.

Refs #7410.

## Summary by CodeRabbit

* **New Features**
  * Added a full Zoom & Pan guide covering overview, quick-start, composing interactions, controlled vs uncontrolled modes, auto-scaling/LOD helpers, custom UI/gesture hooks, touch and keyboard behavior, and rendering semantics.
* **Documentation**
  * Guide page added to site navigation and guide listings; includes runnable Quick Start and Custom Controls examples demonstrating zoom/pan controls and presets.
* **Localization**
  * Added English and Chinese strings for the new guide title.

### DIFF
--- a/www/src/components/GuideView/ZoomAndPan/CustomControlsExample.tsx
+++ b/www/src/components/GuideView/ZoomAndPan/CustomControlsExample.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import { CartesianGrid, Line, LineChart, Tooltip, XAxis, YAxis, useZoom, useZoomState } from 'recharts';
+import { generateMockData } from '@recharts/devtools';
+
+const data = generateMockData(60, 11);
+
+function ZoomButtons() {
+  // The full helper API: zoomIn/zoomOut/pan/reset over the shared viewport.
+  const { zoomIn, zoomOut, reset, isZoomed } = useZoom();
+  // Or the useState-shaped tuple, for custom controls or your own gesture library.
+  const [zoom, setZoom] = useZoomState();
+
+  return (
+    <div style={{ display: 'flex', gap: 6, marginBottom: 8, alignItems: 'center' }}>
+      <button type="button" onClick={() => zoomIn()}>
+        Zoom in
+      </button>
+      <button type="button" onClick={() => zoomOut()}>
+        Zoom out
+      </button>
+      <button type="button" onClick={() => setZoom({ x: { start: 0.75, end: 1 } })}>
+        Last quarter
+      </button>
+      <button type="button" onClick={reset} disabled={!isZoomed}>
+        Reset
+      </button>
+      <span style={{ fontFamily: 'monospace', fontSize: 12 }}>
+        x: [{zoom.x.start.toFixed(2)}, {zoom.x.end.toFixed(2)}]
+      </span>
+    </div>
+  );
+}
+
+export default function CustomControlsExample() {
+  return (
+    <LineChart width={700} height={300} data={data} responsive>
+      <CartesianGrid strokeDasharray="3 3" />
+      <XAxis dataKey="label" />
+      <YAxis />
+      <Tooltip />
+      <Line type="monotone" dataKey="x" stroke="#8884d8" dot={false} isAnimationActive={false} />
+      {/* No interaction components mounted: the buttons are the only way to zoom. */}
+      <ZoomButtons />
+    </LineChart>
+  );
+}
--- a/www/src/components/GuideView/ZoomAndPan/CustomControlsExample.tsx
+++ b/www/src/components/GuideView/ZoomAndPan/CustomControlsExample.tsx
@@ -1,39 +1,38 @@
-import React from 'react';
 import { CartesianGrid, Line, LineChart, Tooltip, XAxis, YAxis, useZoom, useZoomState } from 'recharts';
 import { generateMockData } from '@recharts/devtools';
 
 const data = generateMockData(60, 11);
 
 function ZoomButtons() {
-  // The full helper API: zoomIn/zoomOut/pan/reset over the shared viewport.
   const { zoomIn, zoomOut, reset, isZoomed } = useZoom();
-  // Or the useState-shaped tuple, for custom controls or your own gesture library.
   const [zoom, setZoom] = useZoomState();
 
   return (
-    <div style={{ display: 'flex', gap: 6, marginBottom: 8, alignItems: 'center' }}>
-      <button type="button" onClick={() => zoomIn()}>
-        Zoom in
-      </button>
-      <button type="button" onClick={() => zoomOut()}>
-        Zoom out
-      </button>
-      <button type="button" onClick={() => setZoom({ x: { start: 0.75, end: 1 } })}>
-        Last quarter
-      </button>
-      <button type="button" onClick={reset} disabled={!isZoomed}>
-        Reset
-      </button>
-      <span style={{ fontFamily: 'monospace', fontSize: 12 }}>
-        x: [{zoom.x.start.toFixed(2)}, {zoom.x.end.toFixed(2)}]
-      </span>
-    </div>
+    <foreignObject x={8} y={308} width={684} height={46}>
+      <div style={{ display: 'flex', gap: 6, alignItems: 'center' }}>
+        <button type="button" onClick={() => zoomIn()}>
+          Zoom in
+        </button>
+        <button type="button" onClick={() => zoomOut()}>
+          Zoom out
+        </button>
+        <button type="button" onClick={() => setZoom({ x: { start: 0.75, end: 1 } })}>
+          Last quarter
+        </button>
+        <button type="button" onClick={reset} disabled={!isZoomed}>
+          Reset
+        </button>
+        <span style={{ fontFamily: 'monospace', fontSize: 12 }}>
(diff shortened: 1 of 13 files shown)


www/src/components/GuideView/ZoomAndPan/index.tsx
```diff
@@ -0,0 +1,402 @@
...
+        smooth sub-pixel panning even on a categorical axis. Note that <code>window</code> here is the visible slice of
+        an axis, it has nothing to do with the browser window, and resizing the browser never zooms the chart.
+      </p>
+
+      <h2>Quick start</h2>
+      <p>
+        For the usual experience in one line, add <code>&lt;ZoomAndPan /&gt;</code>. It bundles the sensible defaults:
+        wheel zoom, drag to pan, pinch, keyboard, double-click reset and scrollbars.
+      </p>
+      <pre>{`import { LineChart, XAxis, YAxis, Line, ZoomAndPan } from 'recharts';
```
PavelVanecek: These should be actual chart examples - see other guides for what `CodeEditorWithPreview` does

www/src/components/GuideView/ZoomAndPan/index.tsx
```diff
@@ -0,0 +1,402 @@
...
+
+// controlled: you hold the viewport and update it
+const [viewport, setViewport] = useState({ x: { start: 0.2, end: 0.6 } });
+
+<LineChart data={data}>
+  <ZoomAndPan viewport={viewport} onZoomChange={setViewport} />
+</LineChart>`}</pre>
+      <p>
+        <code>onZoomChange</code> fires on every change with the new viewport. Because it is the same shape you pass
+        back in, a controlled chart settles without a feedback loop, and you can sync several charts by sharing one
```
PavelVanecek, line 234: Good point about the synchronisation. Recharts has built-in chart sync. Should we have them sync zoom by default as well?

www/src/components/GuideView/ZoomAndPan/index.tsx
```diff
@@ -0,0 +1,402 @@
...
+      <p>
+        Leave it alone for an uncontrolled chart, set an initial value, or control it fully. These live on{' '}
+        <code>&lt;ZoomAndPan /&gt;</code> (and any individual interaction component).
+      </p>
+      <pre>{`// uncontrolled, with an initial zoom
+<LineChart data={data}>
+  <ZoomAndPan initialZoom={{ x: { start: 0.2, end: 0.6 } }} />
+</LineChart>
+
+// controlled: you hold the viewport and update it
```
PavelVanecek, line 230: Example with custom zoom controls would be great

www/src/components/GuideView/ZoomAndPan/index.tsx
```diff
@@ -0,0 +1,402 @@
...
+        <LinkToApi>RadarChart</LinkToApi>, <LinkToApi>RadialBarChart</LinkToApi>, <LinkToApi>FunnelChart</LinkToApi>,{' '}
+        <LinkToApi>Sankey</LinkToApi>, <LinkToApi>SunburstChart</LinkToApi>, <LinkToApi>Treemap</LinkToApi> — see{' '}
+        <a href="#special-charts">Polar &amp; standalone charts</a>). It is <strong>off by default</strong> and fully
+        composable: you opt in by adding the interaction components (or hooks) you want, and they all drive one shared,
+        normalized viewport.
+      </p>
+
+      <h2>The viewport</h2>
+      <p>
+        Zoom is described by a <strong>viewport</strong>: how much of each axis is currently visible, as a fraction in{' '}
```
PavelVanecek, line 45: I think this guide should also include Brush as a valid alternative. Recharts will now have two zoom modes: one driven by domain (Brush) another driven by range (ZoomAndPan)

www/src/components/GuideView/ZoomAndPan/index.tsx
PavelVanecek: So in my mind I had something like 

```tsx
const [zoom, setZoom] = useZoomState()
```

where `zoom` is the current state and `setZoom` is the writer and then people can implement their own controls, or subscribe to the current zoom state, with a single hook. And both Brush and the new ZoomAndPan components would be consumers of the same hook. Would that work?

Now this example is kinda sorta going in this direction but it uses ZoomAndPan as the source of events which I suppose works. Just thinking out loud

PavelVanecek: or Brush could have `const [{startIndex, endIndex }, setDataIndex] = useDataIndex(dataKey, axisType, axisId)` which would be the same thing but operating on data values instead of [0,1]

www/src/components/GuideView/ZoomAndPan/index.tsx
Fefedu973: Agreed, syncing zoom across `syncId` charts is where this should go. I would make it opt-in in v1 (say a `syncZoom` flag) rather than default: the viewport is a fraction of the axis, so charts with different data lengths under one syncId would need the same fraction-vs-value decision that tooltip sync solves with `syncMethod`, and defaulting on would change behavior for existing syncId users the moment they add zoom to one chart. It already works manually today by lifting one `useZoomState()` above both charts (there is a live example in the guide). I would ship sync as its own small PR in the series.

www/src/components/GuideView/ZoomAndPan/index.tsx
Fefedu973: Got it !

www/src/components/GuideView/ZoomAndPan/index.tsx
Fefedu973: Got it !

www/src/components/GuideView/ZoomAndPan/index.tsx
Fefedu973: Ok

www/src/components/GuideView/ZoomAndPan/index.tsx
PavelVanecek: Counter point: brush zoom synchronisation is turned on as soon as you provide a syncId. I would argue that we should stay consistent and have both zooms be included

www/src/components/GuideView/ZoomAndPan/CustomControlsExample.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { CartesianGrid, Line, LineChart, Tooltip, XAxis, YAxis, useZoom, useZoomState } from 'recharts';
+import { generateMockData } from '@recharts/devtools';
+
+const data = generateMockData(60, 11);
+
+function ZoomButtons() {
+  // The full helper API: zoomIn/zoomOut/pan/reset over the shared viewport.
+  const { zoomIn, zoomOut, reset, isZoomed } = useZoom();
```
PavelVanecek, line 7: Why do we have two different hooks for zooming? What's useZoom doing that useZoomState doesn't?

Fefedu973: `useZoomState()` and `useZoom()` are two layers over the same shared viewport, not two ways of doing the same thing:

- **`useZoomState()`** is the minimal primitive you suggested — `const [zoom, setZoom] = useZoomState()`, just the current state and the writer. It's what you reach for to subscribe to the viewport or wire fully custom controls, and it's the hook Brush, `ZoomAndPan` and the other components all consume.
- **`useZoom()`** is a convenience layer on top of that same state: it adds the ready-made actions, `zoomIn` / `zoomOut` / `pan` / `reset` / `isZoomed` — so common controls don't have to re-implement the zoom-factor maths or a `minZoom`-aware reset by hand.

So `useZoomState` is the foundation (state + setter), and `useZoom` is sugar for the typical buttons. The example just happened to show both at once, which makes them look parallel

PR #7450, opened 2026-06-11
feat(zoom): add composable zoom and pan

Fefedu973:
## Summary

This is the main review PR for composable zoom and pan in Recharts. The feature is opt-in and covers Cartesian, polar, and standalone charts through one viewport model and a set of composable controls.

Stacked PRs are not available for this repository, so the **Commits** tab is intentionally the review stack. The history has been rebuilt from the latest `main` into 15 ordered commits: each commit has one coherent responsibility, avoids temporary API churn, and keeps implementation, tests, exports, and build manifests together where its dependencies allow.

Rebased onto `main` at `1e855e9a469`.

## Recommended review order

| # | Commit | Review focus |
|---|---|---|
| 1 | [`622aadf`](https://github.com/recharts/recharts/pull/7450/commits/622aadf2604b44a7c09ebe5dbb38f470df6d067a) | Normalized viewport state, selectors, and pure viewport math |
| 2 | [`5c4f37a`](https://github.com/recharts/recharts/pull/7450/commits/5c4f37a210b469610cd1932ab2443d6961e7a0cd) | Apply viewport windows to Cartesian axes and rendering |
| 3 | [`c94072b`](https://github.com/recharts/recharts/pull/7450/commits/c94072b95ad151f56a235fadbfc363a004a9a9d4) | Controlled state plus modular mouse wheel, drag, and reset gestures |
| 4 | [`c0d25bf`](https://github.com/recharts/recharts/pull/7450/commits/c0d25bf8a2192e13407fb33074caea8845ce0903) | Keyboard and axis-band interactions |
| 5 | [`f175076`](https://github.com/recharts/recharts/pull/7450/commits/f175076b3ca4f785998ece748135ca06175646f7) | Touch gestures and stable double-tap targeting |
| 6 | [`931abb2`](https://github.com/recharts/recharts/pull/7450/commits/931abb25c0005c5acf7a675526064e35b304dbec) | Animation suppression and accessible on-canvas range controls |
| 7 | [`1c83ea4`](https://github.com/recharts/recharts/pull/7450/commits/1c83ea45fa0e2328a5919c4a17b6754815b1887f) | `ZoomAndPan`, `useZoom`, `useZoomState`, and controlled synchronization |
| 8 | [`550a4ea`](https://github.com/recharts/recharts/pull/7450/commits/550a4eae7ec4664b641a1b864a96db163f619323) | Granular interaction components and customization |
| 9 | [`3fb0dc3`](https://github.com/recharts/recharts/pull/7450/commits/3fb0dc30fe4a94fb51a3d63cbbab6a17184a9071) | Auto-scale, follow-series, and scatter LOD helpers |
| 10 | [`114a852`](https://github.com/recharts/recharts/pull/7450/commits/114a8523e42be2c593747cefbff1745558b772fc) | Vertical, reversed, and custom-axis layouts |
| 11 | [`97b8ba7`](https://github.com/recharts/recharts/pull/7450/commits/97b8ba7a604a7a2f2a78cdaf545fabfd1e81dfcd) | Brush zoom mode and composable Minimap controls |
| 12 | [`f74a2f6`](https://github.com/recharts/recharts/pull/7450/commits/f74a2f613f5e8d9c4c4816f3c05db8c0412ac93f) | Camera zoom for polar and standalone charts |
| 13 | [`4cea847`](https://github.com/recharts/recharts/pull/7450/commits/4cea8471b26c251c180bd8ac5e11f7aa728ef1bc) | Complete public type surface and shared limits |
| 14 | [`e3b1787`](https://github.com/recharts/recharts/pull/7450/commits/e3b1787b7cc93269a6c68ebeecf1872c44ad206d) | Storybook API pages and interactive examples |
| 15 | [`3e31950`](https://github.com/recharts/recharts/pull/7450/commits/3e31950b861d3aa174d5d8afd1f4824a9d1e4245) | Website guide, API navigation, and sitemap |

## What is included

- An opt-in `<ZoomAndPan />` bundle, granular interaction components, and programmatic hooks.
- Wheel, pointer, keyboard, touch, axis-band, scrollbar, Brush, and Minimap controls sharing the same limits.
- Controlled and uncontrolled viewports with stable callback semantics.
- Horizontal and vertical layouts, reversed/custom scales, series-level data, and nonlinear scales.
- Camera transforms and correctly projected HTML tooltips for polar and standalone charts.
- Interaction-island guards so embedded controls and Brush/Minimap handles do not trigger chart gestures.
- Accessibility support for keyboard navigation, range controls, Sunburst, Treemap, and interactive overlays.
- `AutoScaleAxis`, `FollowSeries`, and `useScatterLOD`.
- Storybook API documentation, examples, and the website guide.

The rebase preserves the latest upstream fixes for Brush controlled indices/data refresh, CartesianAxis tick reporting, themed axes/grid, Sankey algorithms and accessibility, and Treemap animation.

## Validation

- Full unit suite: 346 files; 17,270 passed, 6 expected-failure cases, 16 skipped, 1 todo.
- All TypeScript projects: library, tests, Storybook, visual tests, and website.
- ESLint: 0 errors (15 pre-existing warnings).
- Public export verification: 2/2.
- Build-output verification: 12/12.
- Full production build.
- Omnidoc checks: 1,067/1,067.
- Targeted zoom, Brush, Minimap, layout, tooltip, and special-chart suites were also run at their commit boundaries.

## Related

Continues the feature request in #710 and the earlier exploration in #5934. The initial stage/guide PRs #7410 and #7429 informed this work, but **#7450 is now the single review and merge target**.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive zoom and pan interactions, including wheel, drag, pinch, keyboard, axis, and double-click controls.
  * Added minimaps, zoom scrollbars, drag selection, controlled viewports, and public zoom hooks.
  * Added zoom support for Cartesian, polar, Sankey, Sunburst, Treemap, and other charts.
  * Added Brush zoom mode, vertical brushing, autoscaling, series following, and scatter-point level-of-detail handling.
  * Added accessibility improvements, tooltip alignment, clipping, and extensive interactive documentation examples.

* **Bug Fixes**
  * Improved rendering stability and animation behavior while charts are zoomed.


Fefedu973:
@coderabbitai review

src/cartesian/Brush.tsx
```diff
@@ -14,26 +13,49 @@ import { clsx } from 'clsx';
...
+import { isWellFormedNumberDomain } from '../util/isDomainSpecifiedByUser';
 
-type BrushTravellerType = ReactElement<SVGElement> | ((props: TravellerProps) => ReactElement<SVGElement>);
+type BrushTravellerType =
+  | ReactElement<SVGElement>
+  | ((props: Omit<TravellerProps, 'layout'>) => ReactElement<SVGElement>);
 
 // Why is this tickFormatter different from the other TickFormatters? This one allows to return numbers too for some reason.
 type BrushTickFormatter = (value: any, index: number) => number | string;
+type BrushLayout = 'horizontal' | 'vertical';
```
PavelVanecek: do we want to reuse `CartesianLayout` type?

src/cartesian/Brush.tsx
```diff
@@ -1011,41 +1558,139 @@ function BrushInternal(props: InternalProps) {
...
+        )
+      : undefined;
+  const zoomLimits: ZoomLimits = {
+    minZoom,
+    maxZoom,
+  };
 
-  const { startIndex, endIndex } = dataIndexes;
-  const { x, y, width } = brushDimensions;
+  const onZoomWindowChange = (start: number, end: number, railStart = primaryStart, activeRailLength = railLength) => {
```
PavelVanecek: What if this was a hook that we could export and have people implement their own controls? And Brush would be one of the OOTB optional implementations

src/cartesian/CartesianGrid.tsx
```diff
@@ -436,6 +439,52 @@ const defaultHorizontalCoordinatesGenerator: HorizontalCoordinatesGenerator = (
...
+  if (a === b) {
+    return true;
+  }
+  if (realScaleType == null) {
+    return false;
+  }
+  return areScalesApproximatelyEqual(a, b);
+}
+
+function areGridAxisPropsEqual(
```
PavelVanecek, line 473: We already have `src/util/propsAreEqual.ts`, can we reuse?

src/cartesian/CartesianGrid.tsx
```diff
@@ -555,6 +617,15 @@ export function CartesianGrid(props: Props) {
     }
   }
 
+  /*
+   * When zoomed, the generators place grid lines across the stretched range, so some land outside
+   * the plotting area. Keep only the lines that fall within the visible plot bounds.
+   */
+  if (isZoomed) {
+    horizontalPoints = horizontalPoints?.filter(p => p >= y - 0.5 && p <= y + height + 0.5);
```
PavelVanecek, line 654: This assumes a certain width of the lines, right? What if the lines are 2px wide? 4px wide? Isn't clipPath a better solution?

src/chart/zoom/touchGesture.ts
PavelVanecek, line 1: Is there an opportunity to reuse something like use-gesture or similar?

src/chart/zoom/useScatterLOD.ts
```diff
@@ -0,0 +1,127 @@
...
+ * appear - giving cheap rendering when zoomed out and full detail when zoomed in.
+ *
+ * Feed the result to a Scatter component with the data prop. It recomputes only when the data or the viewport
+ * changes, and returns the input unchanged until the scales are measured.
+ *
+ * @example
+ * const points = useScatterLOD(bigData, { x: 'x', y: 'y' });
+ * return <Scatter data={points} />;
+ */
+export function useScatterLOD<T>(data: ReadonlyArray<T>, options: ScatterLODOptions): ReadonlyArray<T> {
```
PavelVanecek: So there is LOD after all

PavelVanecek (commented):
Just a few observations

src/cartesian/Brush.tsx
Fefedu973: Yea, going to fix that !

src/chart/zoom/useScatterLOD.ts
Fefedu973: Yea I guess i have done it afterall.
I'm not 100% convinced by the result but you tell me !

src/cartesian/CartesianGrid.tsx
Fefedu973: The same function? Idk, I'm putting them at least in the same util file but merging the two functions i'm not 100% sure

src/cartesian/CartesianGrid.tsx
Fefedu973: Got it

src/cartesian/CartesianGrid.tsx
Fefedu973: the approximate-equality helpers now live in `util/propsAreEqual.ts` so all equality utilities share one home. Note they complement rather than replace the existing helpers - `propsAreEqual` does shallow React-props comparison, while these do epsilon-aware comparison of scales/ticks to keep store subscriptions idempotent under floating-point churn while zooming.

src/chart/zoom/touchGesture.ts
Fefedu973: I evaluated it. Most of our touch code is recharts-specific policy (one finger stays reserved for the tooltip, a pinch over an axis band zooms only that axis, double-tap-drag modes, coexistence with scrollbars and the brush panorama) rather than raw gesture recognition, which is the part use-gesture would replace. It would also be a new runtime dependency. The integration story for people who prefer use-gesture (or any gesture lib) is the controlled hook: turn the built-ins off and drive `useZoomState()` from their handlers - that is a ~10 line wiring. Happy to revisit if you would rather take the dependency.

src/cartesian/Brush.tsx
Fefedu973: Done !

Fefedu973:
Hey ! Any updates? 
I know this is an enormous PR but the more we wait the more difficult it will be to merge eventually. 
Good luck 🫡

PavelVanecek:
Hi @Fefedu973 I just released 3.9.0 and I'm going through the PR queue one by one. This is a significant change so it may take a while to get done. To give you an idea the animations took three months. Thank you for your patience. If you want to help you can try and find out why the dependency bumps are failing. I'm looking into the jsdom bump now.

Fefedu973:
The number of conflicts is starting to pile up ):
Github just broadly release stacked PR do you think it would help you if I converted my PR with a stacked PR or this is not the issue.

Fefedu973:
Well stacked PR seems to only be available for contributors and not forks anyway...

PavelVanecek:
Hi @Fefedu973 I'm now working on the theme and once that's done and rolled out I'll come back to zoom and pan. Conflicts we can handle. I want to be confident a feature is 100% ready before merging, especially for large one like this.

Fefedu973:
Okay thanks. No problem I absolutely agree this is not one small change. Good luck 🤞

PavelVanecek:
Okay themes done, time for zooming.

I'm playing around with the guide in the demo and I noticed a couple of problems:

On desktop
- as I drag using mouse on desktop, the text gets selected
- guide says "Shift pans horizontally, Ctrl/Cmd + Shift pans vertically." shift+scroll works but ctrl/cmd+scroll is zoom, not pan
- "<AutoScaleAxis /> re-fits the value axis to the data visible in the current window" this is great but the chart below has data such that they don't demonstrate this effect - the chart oscillates between 0 and 300 anyway so I can't see the effect
- there is a mention of keyboard controls, what are they?

on iPad
- the two-finger gesture does pan the chart but also it controls the Tooltip at the same time
- the "fog of war" area of Minimap is controlling Tooltip on the big chart - I would expect it should move the view instead (like the small "active" area does)
- when I hold one finger on the main chart and "drag" with the other one, the chart zooms in using the first finger as a pivot point. This is great - but when I pinch on the minimap it uses the second. I think the active area follows the second finger? They both should behave the same

on Android phone
- the charts overflow in X dimension and make the page scroll horizontally
- same touch control problems as the iPad

Other things that I want to test + clarify in the guide:

- Chart synchronization (using `syncId`) should also sync the zoom level
- "Auto-scaling, follow & level of detail" example doesn't have touch controls, we should add (and then test again)
- `Minimap` should use the same cartesian positioning system as Labels, and Legend
- Minimap in vertical chart
- does Minimap work with the old Brush?

Also it appears that stacked PRs are now GA so that should work.

All in all it looks promising so I will start merging things one by one if you allow.

Below is what Sonnet 5 found but AI being AI I would like to verify first

---

Code review
8 findings
touchGesture.ts

Confirmed
src/chart/zoom/touchGesture.ts:431.Touch selectZoom emits wrong post-zoom window

Reuse
src/chart/zoom/touchGesture.ts:18.Pinch-distance math duplicated across 3 files
Brush.tsx

Confirmed
src/cartesian/Brush.tsx:1845.Vertical Brush panorama renders at zero size

Simplification
src/cartesian/Brush.tsx:1036.Brush drag clamp math duplicated 4x
Sankey.tsx

Altitude
src/chart/Sankey.tsx:1425.Zoom wiring copy-pasted across 3 chart types
ZoomScrollbars.tsx

Efficiency
src/chart/ZoomScrollbars.tsx:89.matchMedia recomputed every scrollbar render
useScatterLOD.ts

Plausible
src/chart/zoom/useScatterLOD.ts:92.Scatter LOD rebuilds full grid every zoom tick
ZoomStateSync.tsx

Plausible
src/chart/ZoomStateSync.tsx:1.Zoom sync patches around two sources of truth
Reviewed [PR #7450](https://github.com/recharts/recharts/pull/7450) (zoom/pan feature, ~15k lines across 144 files) at medium effort. Key results:

Correctness (2 confirmed bugs):

touchGesture.ts:431 — the touch selectZoom drag mode zooms first, then computes the emitted selection region against the already-zoomed viewport using stale pre-zoom fractions, so onTouchSelectRegion reports a wrong, double-zoomed window.
Brush.tsx:1845 — BrushSettingsDispatcher never dispatches brush dimensions to the store for a vertical <Brush>, so a panorama chart nested inside a vertical Brush renders at x=0,y=0,width=0,height=0 (invisible).
Duplication/cleanup:

Zoom wiring (ClipPathProvider/ZoomTransformLayer/ZoomPropBridge) copy-pasted identically into Sankey, SunburstChart, and Treemap.
Brush's drag clamp/delta math duplicated 4x between live and frozen zoom-mode branches.
Pinch-distance math reimplemented separately in touchGesture.ts and Brush.tsx instead of reusing PolarUtils.distanceBetweenPoints.
Efficiency:

AxisScrollbar recomputes window.matchMedia(...) unmemoized on every render, including every scrollbar-drag frame.
Scatter LOD does a full O(n) grid rebuild on every scale change (plausible, partially mitigated already).
Architecture note: ZoomStateSync.tsx patches around two sources of truth (controlled prop vs. Redux zoom slice) with propose/rollback ref tracking — plausible fragility for future zoom features.

Fefedu973:
Thanks for the detailed review. I reproduced and checked each point, rebased the PR onto the latest main, and force-pushed a cleaned commit history.

Desktop

- Text selection while dragging: fixed. Text selection is now suppressed only while a plot or axis drag is active, and restored immediately afterwards.
- Wheel modifiers: the guide now explicitly says:
  - Shift + wheel pans horizontally
  - Ctrl/Cmd + Shift + wheel pans vertically
  - Ctrl/Cmd + wheel without Shift remains zoom
- AutoScaleAxis example: replaced the data with a series whose visible ranges have clearly different Y domains, so the autoscaling effect is now visible.

iPad / touch

- Two-finger gestures no longer update the Tooltip while they pan or pinch. The gesture also cancels any one-finger Tooltip update already queued for the next animation frame.
- Pressing the shaded/fog area of the Minimap now recenters the viewport there and immediately starts dragging it. Those events no longer leak to the main chart Tooltip.
- Main-chart and Minimap pinch gestures now use the same affine anchor behaviour. A stationary finger remains over the same datum in both.
- A Minimap one-finger touch is deferred until it is known not to be the beginning of a pinch, preventing the initial viewport jump.

Other requested cases

- Charts sharing a syncId now synchronize their semantic X/Y viewport, including programmatic and controlled updates. Receiving charts still enforce their own enabled axes and zoom limits.
- The auto-scale/follow/LOD example now includes touch pan and pinch controls.
- Minimap now uses the shared Cartesian positioning helper and the same inside*/center/object offset semantics as the other Cartesian overlays. Outside positions remain intentionally excluded because Minimap currently overlays the plot and does not reserve layout space. The legacy corner names remain supported.
- Vertical Minimap behaviour is covered, including an actual Y-axis drag while X remains unchanged.
- Minimap and the classic Brush can coexist without stealing each other’s events. The classic Brush still controls its independent data slice; Brush mode="zoom" should be used when Brush must control the same semantic viewport as Minimap.

AI findings

1. touchGesture selectZoom emitted region: confirmed and fixed. The selected region is now calculated before applying the new viewport, so it is not transformed twice.

2. Duplicated pinch-distance calculation: the Math.hypot calls are real, but I did not couple them to PolarUtils.distanceBetweenPoints. That helper is a private polar utility using chart {x, y} points, while these gestures use browser client coordinates. Exporting/coupling that utility for a single expression would make the dependency less clear.

3. Vertical Brush panorama at zero size: not reproduced. The nested panorama has its own chart store and receives explicit non-zero dimensions. Tests now assert the vertical preview’s x/y/width/height and exercise an actual vertical drag. Dispatching horizontal Brush dimensions here would incorrectly reserve bottom layout space.

4. Brush clamp duplication: confirmed and refactored into shared clamp/delta helpers.

5. Sankey/Sunburst/Treemap zoom wiring: the repeated chart-content boundary was extracted into a shared internal ZoomableChartContent component. ZoomPropBridge remains at the chart-specific boundaries where it is required.

6. matchMedia during scrollbar renders: fixed. The media-query result is memoized once per scrollbar mount instead of being recomputed on drag renders.

7. Scatter LOD rebuilding on every tick: confirmed. Rebuilds are now coalesced to one requestAnimationFrame and use the latest scale/data references.

8. ZoomStateSync sources of truth: I could not confirm a second render source. Redux remains the chart’s render source; the controlled bridge only tracks proposals and rollbacks to avoid loops. I added controlled synchronization and rollback coverage instead of replacing that architecture.

I also added the required @since 3.11 documentation for the new public exports and updated the build artifact manifests.

Regarding stacked PRs: I checked, but GitHub’s stacked-PR workflow still does not work for external PRs coming from forks. I therefore kept this as one PR and made the commit chain itself the reviewable logical stack.

PR #7457, opened 2026-06-13
Refactoring animation code

PavelVanecek:
So I wasn't happy with how the animation code is structured so I rewrote it from scratch. This didn't matter before but in 3.9 this will be a public API so let's get it nice and separated.

This is a refactor so the aim is to have 100% identical behavior to what it was before, just with new code.

Before:
- everything is kind of responsible for everything, multiple files are doing the same job
- test and production code are mixed together

After:
- clear separation of concerns: time management, animation state machine, connector, and React facade
- test code isolated for tests only, production code only ships what is required to run in browser

Not quite yet 100% settled, please share feedback:
- naming

Notable changes:
- ProgressAnimationManager and useAnimationManager are gone. I think these were the wrong abstraction, and the new AnimationController interface should replace them both. I will iterate on this in followup PR, I want to be able to override the AnimationController from the outside eventually
- This is not a breaking change because neither of those has been released to npm yet. These were new additions in 3.9 and the current version is 3.8.1 so they are both getting removed before they became a stable API
- easing code is greatly simplified
- in tests: I noticed that the old mock animation code was calling `onAnimationStart` asynchronously which is different from the production code which is calling it synchronously. I fixed the inconsistency which in turn made a few tests to fail so that's why you see changes in test code.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed `AnimationProgressProvider` and `useAnimationProgress` from the public API.
* **New Features**
  * Added a new animation state machine with JavaScript and CSS-transition implementations, including explicit `complete()` finalization and improved progress interpolation behavior.
* **Refactor**
  * Reworked animation driving around an `AnimationController`, updating animation lifecycle and easing wiring across relevant components.
  * Simplified animation manager usage to accept a controller instance.
* **Chores / Tests**
  * Added an `lgtm` script to run lint, build, tests, and type checks together.
  * Updated and expanded animation-related test coverage and documentation categories.


PavelVanecek:
@coderabbitai review

PavelVanecek:
@coderabbitai review

PR #7458, opened 2026-06-14
fix(Pie): pass computed innerRadius/outerRadius to Label children via PolarLabelContextProvider

### LINKED ISSUE #7445: viewBox property from LabelProps is not correctly populated
When using a custom `Label` component via the `content` prop inside a `<Pie>` component, the `viewBox` props passed to the label do not correctly position it in the center of the donut hole. Specifically, `viewBox.innerRadius` is `0` regardless of the `innerRadius` value set on `<Pie>`, making it impossible to calculate the correct bounding box for the label purely from `viewBox`.

A screenshot of what the Donut currently looks like:
Image

The `viewBox` props passed to a custom `Label` content component should include the correct resolved `innerRadius` (and `outerRadius`) values from the parent `<Pie>`, so that the label can be accurately positioned and sized within the donut hole. I expect that the custom label content will placed in the middle, just as if I were using the Label component with children like the [demo in the storybook](https://recharts.github.io/storybook-static/index.html?path=/story/api-chart-piechart--donut).

[Link to Demo](https://stackblitz.com/edit/react-vpmtfpvx-fznrewrx?file=src%2FExample.tsx)

Other Demos I found from GitHub Issues:
- I found this StackBlitz showing similar error: https://stackblitz.com/edit/react-vpmtfpvx?file=src%2FExample.tsx. On line 112 it has a comment saying `{/* centers on the whole thing not just the pie */}.`
- I also found this [PieCharts Labels receive an empty viewBox](https://github.com/recharts/recharts/issues/6030) with a similar problem that I am facing.

What I am currently using:
- **Recharts**: 3.8.1
- **React**: 19.2.6
- **Browser**: Chrome (latest)

0007aadil:
Fixes #7445

## Root Cause

When a `<Label content={MyLabel}>` is used as a child of `<Pie>`, the Label component reads its viewBox from `usePolarLabelContext()`. This hook falls back to `selectPolarViewBox` from the Redux store, which stores the **PieChart-level** `innerRadius: 0` default (set by `ReportPolarOptions` in `PolarChart`). The individual Pie's `innerRadius` (e.g. `80`) is used to compute sectors correctly but was never propagated to Label children — so `viewBox.innerRadius` was always `0`.

## Fix

Wrap `{props.children}` inside `SectorsWithAnimation` in a `PolarLabelContextProvider` that supplies the Pie's actual computed coordinates (`cx`, `cy`, `innerRadius`, `outerRadius`) taken from `sectors[0]`, which already holds the resolved values after percentage calculation.

## Before / After

**Before:** `viewBox.innerRadius` is always `0` in custom Label content, regardless of `<Pie innerRadius={80}>`

**After:** `viewBox.innerRadius` correctly reflects the Pie's computed `innerRadius`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pie chart label rendering to provide the correct geometric context for label `viewBox` when sectors are present, ensuring `innerRadius`/`outerRadius` values match the `Pie` settings.
  * Prevented the label content callback from running when the pie has no sectors.
* **Tests**
  * Added tests covering custom label content callbacks, verifying correct `viewBox.innerRadius`/`viewBox.outerRadius`, defaulting `viewBox.innerRadius` to `0` when `innerRadius` isn’t set, and ensuring callbacks are not invoked with empty data.


PavelVanecek:
Can we have a unit test for this please?

0007aadil:
Sure! Working on it.... will push a test shortly

0007aadil:
Done.... pushed two tests to test/chart/PieChart.spec.tsx:

Confirms viewBox.innerRadius correctly reflects the Pie's innerRadius prop (e.g. 80) — this was always 0 before the fix.
Checks the default case where no innerRadius is set (expects 0, which is correct).

Both pass locally with Vitest.

PavelVanecek (approved):

0007aadil:
Pushed a third test (6dd97fa) covering the props.children fallback branch when sectors is empty ,  this should resolve the Codecov gap on line 1024. The previous VR test failures were a Docker infrastructure error (permission denied on /root/.docker/config.json), unrelated to code changes. Could you approve the workflow run so the full CI suite can execute?

0007aadil:
Hi @PavelVanecek just a gentle ping! The PR is approved and the latest commit (a57b1ea) is green. The only remaining step is approving the workflow run so the VR tests can execute. Could you approve it when you get a chance? Thanks!

.gitignore
```diff
@@ -40,3 +40,6 @@ bundle-visualization.html
 
 # Website-generated bundle size data
 www/public/generated/bundleSizeData.generated.json
+
+# Local AI assistant instructions
+CLAUDE.md
```
PavelVanecek: This I believe should be checked in to the repository

0007aadil:
Good catch — that `.gitignore` line was incidental to this PR (leftover from local tooling setup), not something I meant to include. Pulled it out in 48777c1 to keep this PR scoped to the Label/viewBox fix.

I do like the idea of checking in `CLAUDE.md` for the project generally, but it needs a bit of cleanup first (the version I have locally has some contributor/session-specific notes that shouldn't ship in the shared repo, and it'll also want a `scripts/buildOutput.test.ts` snapshot update since root `*.md` files get published to npm via the `files` field in `package.json`). I'll follow up with a separate PR for that so it gets its own review.

PR #7471, opened 2026-06-18
fix: avoid Sankey nodes overlapping skipped-depth links

### LINKED ISSUE #5559: Sankey chart visual overlap between nodes
Bug

Sankey chart sometimes overlap visually between nodes.

Image

## If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via codesandbox or similar
https://codesandbox.io/p/sandbox/react-recharts-sankey-s0wgt

Node shouldn't be in other nodes "horizontal" space

v2.15.0 | Chrome

pupuking723:
## Description

Fixes a Sankey layout case where a high-value link can skip over an intermediate depth and visually cover smaller nodes in that depth.

This keeps the existing node sizing and `nodePadding` behavior unchanged. After the normal Sankey relaxation and collision passes complete, the layout now checks links that span across intermediate depth columns. If an intermediate node is fully contained inside one of those skipped-depth link bands, that link is treated as a fixed local obstacle and the affected depth is adjusted once.

## Related Issue

Fixes #5559.

## Motivation and Context

The workaround discussed in the issue is to increase `nodePadding`, but that does not scale well when a Sankey chart has many nodes. The root problem is that normal node collision handling only considers nodes in the same depth, while a link that skips a depth can still occupy visual space in that column.

The fix is intentionally scoped to the overlap case from the issue, so normal Sankey layouts are not rescaled and existing coordinate-sensitive behavior is preserved.

## How Has This Been Tested?

- `npm run test-lib -- test/chart/Sankey.spec.tsx -t "should keep intermediate nodes out of links that skip over their depth"`
- `npm run test-lib -- test/chart/Sankey.spec.tsx`
- `npm exec eslint src/chart/Sankey.tsx test/chart/Sankey.spec.tsx`
- `npm run check-types-lib`
- `npm run check-types-test`

Not included. The regression test checks that the intermediate nodes no longer sit inside the skipped-depth link band.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Sankey diagram layout to reduce and avoid visual overlaps between nodes and link paths, producing clearer, more readable flows.

* **Tests**
  * Added automated Sankey layout verification for complex/overlapping scenarios, including SVG path geometry checks and a screenshot regression test to confirm rendering correctness.


PavelVanecek:
Screenshots and/or a VR test would be welcome here

pupuking723:
Thanks for the note. This PR now includes both a VR case for the skipped-depth Sankey overlap and the generated Chromium/Firefox/WebKit snapshots, plus a unit-level geometry assertion for the same overlap case.

I also rechecked the targeted verification locally:

- `npm test -- --run test/chart/Sankey.spec.tsx`
- `npm run check-types-test -- --pretty false`

Could you please take another look when you have a chance?

PavelVanecek:
Just to compare here is what the chart looks like on main branch now:

image

Issue #7478, opened 2026-06-20
Sankey layout freezes on dense, heavily branching graphs


dm-gthb:
## What is the current behavior?

Rendering a <Sankey> chart can freeze the browser main thread for seconds or minutes during layout, before anything appears on screen.

Starting from relatively small graphs, this is already a problem — not only at “big data” scale. The graph is densely connected: nodes branch heavily and paths merge again, so the number of distinct paths from the first node to the last grows combinatorially even when node/link counts stay small.

With the reproducer below, the dataset has 85 nodes and 112 links, layout on current npm recharts freezes the UI for minutes.

## What is the expected behavior?

Layout should complete quickly and feel responsive. Cost should grow with the number of nodes and links, not combinatorially with the number of distinct paths through the graph.

## Please provide a demo of the problem in a sandbox

https://codesandbox.io/p/sandbox/recharts-sankey-performance-repro-wjwm4s 

Tune `SEGMENTS` to try different node/link counts. Recommended size: `SEGMENTS = 28` → **85 nodes**, **112 links**. On current `recharts` from npm, this typically freezes the UI for minutes.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

- **Recharts:** latest on npm (e.g. 2.x / 3.x)
- **Browser:** Chromium-based browsers (Chrome, Edge); likely all browsers (CPU-bound JS)
- **OS:** any
- **Previous versions:** not a regression from a specific release

---
**Note:** a related symptom was reported in #5531 (closed as not planned, larger dataset). This issue uses a smaller, more reliable reproduction that isolates the actual cause — layout cost scales with the number of distinct paths through the graph, not with node/link counts. I've opened a fix in [#7479](https://github.com/recharts/recharts/pull/7479)

PR #7479, opened 2026-06-20
fix(Sankey): avoid exponential depth traversal on dense graphs

dm-gthb:
## Description

Fix Sankey layout freezing on dense, heavily branching graphs by pruning depth assignment in `updateDepthOfTargets`.

When computing node `depth` (longest path from a source), the layout recurses into every target even when depth does not grow. On graphs where many paths share nodes, that re-walks every distinct path combinatorially. The freeze is not limited to very large datasets — it can happen with modest node/link counts when the graph branches heavily.

The fix only recurses when `curNode.depth + 1` exceeds the target's current depth. Once depth stops increasing, the subtree is already correct.

Also exports `computeData` from `Sankey.tsx` for unit tests (same pattern as Treemap's `computeNode`). It is not added to `src/index.ts`, so the public API is unchanged. New test coverage is described below.

## Related Issue

https://github.com/recharts/recharts/issues/7478

## Motivation and Context

Sankey layout can freeze the browser when depth assignment must follow the **longest path** to each node. Without pruning, the layout walks every distinct path combinatorially.

Starting from relatively small graphs, this is already a problem — not only at “big data” scale. For example, the issue reproducer uses only 85 nodes and 112 links (`makeBranchingChainData(28)`). 

The cost is driven by path count (how many distinct routes exist through shared nodes), not by raw node/link totals alone. Modest node/link counts can still hang when the graph branches heavily.

The fix is a small guard in `updateDepthOfTargets`. Existing Sankey tests that pin exact coordinates still pass, so layout geometry for typical graphs is unchanged.

## How Has This Been Tested?

Added unit tests in test/chart/Sankey.spec.tsx covering the Sankey layout for large, heavily branching graphs:

- **Depth correctness**: verifies a node's depth is the longest path from a source, even when links skip layers (e.g. a 0 -> 1 -> 2 chain plus a 0 -> 2 shortcut resolves node 2 to depth 2, not 1).
- **Performance regression guard**: builds a chain of 28 branching segments (small node/link count but 2 ** 28 distinct paths) and asserts computeData returns all nodes with the expected max depth, completing well within vitest's default timeout — the previous, unfixed implementation would time out on this input.
- **Rendering**: renders <Sankey /> with the same branching graph and checks all node and link elements are present.

N/A — performance fix with no visual change for existing graphs.

## Types of changes

## Checklist:

## Release Notes

* **Bug Fixes**
  * Improved Sankey layout depth propagation to avoid redundant recalculations and better handle complex branching graphs, resulting in more reliable node positioning.

* **Tests**
  * Added new Sankey tests using large, heavily branching datasets to verify depth assignment and expected rendered node/link counts.


dm-gthb:
@ckifer @PavelVanecek would appreciate a review when you have a moment - this fixes a freeze on dense branching graphs ([#7478](https://github.com/recharts/recharts/issues/7478)).

PavelVanecek:
@dm-gthb may I ask you to resolve the merge conflicts please? (Sorry if this is posted multiple times, GitHub is showing me errors when I send)

dm-gthb:
@PavelVanecek done, resolved the conflicts and merged latest main

Issue #7485, opened 2026-06-22
Bars don't animate to zero, but will animate from zero


6TELOIV:
## What is the current behavior?

When a bar chart value is set to zero, it disappears instantly.

## What is the expected behavior?

That it would animate to zero instead of disappearing

## Please provide a demo of the problem in a sandbox

https://stackblitz.com/edit/bcaemmfj?file=src%2FExample.tsx

Click the "Toggle Dataset (triggering animation)" button; observe how the bars animate from the first dataset to the second, with all non-zero values. Then notice how from the second to third, the bars all disappear instantly. Lastly, note how the bars animate from zero when switching back to the first dataset. That animation in reverse is what I'd expect to happen when a value is set to 0.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

I've tested in Firefox and Chrome and the issue is present in both; this is on the latest recharts version at the time of writing, 3.8.1

6TELOIV:
This might be related to #7377 but I'm not 100% sure, so I created this issue with a demo of the problem I'm observing.

6TELOIV:
In creating #7486 I have found that adding `shape={(props) => <Rectangle {...props}/>}` to the `<Bar />`s is a workaround for now, as it forces `hasCustomShape` to be true and render the rectangle as it shrinks to 0.

It still would be ideal for this to work by default, but that is a functional workaround for now.

pupuking723:
I’d like to investigate this if maintainers are open to a PR.

I’ll first reproduce the zero-width/zero-height transition path and compare the default `<Bar />` rectangle path with the `shape={(props) => <Rectangle {...props} />}` workaround mentioned above. If the root cause is the default rectangle skipping render at zero dimensions, I’ll keep the fix limited to preserving the animation path for bars that animate to zero, with a focused regression test so normal bar rendering is unchanged.

I’ll wait for confirmation before opening a PR.

PR #7486, opened 2026-06-22
fix(Bar): allow bar charts to animate from 0 values (WIP, help wanted)

6TELOIV:
## Description

This PR removes two behaviors:

1. Skipping data for zero width and/or zero height bar rectangles in `computeBarRectangles`
2. Skipping rendering of zero width and/or zero height rectangles in `Rectangle`

## Related Issue

#7485 

## Motivation and Context

Bar charts have janky behavior when animating bars to a zero value. Currently, they will instantly disappear instead of animating as expected. This is because they are entirely skipped in the render, meaning there is nothing to animate once the state changes.

These changes allow the zero-sized bars to still be rendered, so that bars visually collapse to zero instead of disappearing instantly.

## How Has This Been Tested?

Visually only with an added storybook story.

I have not fixed the tests yet, and would appreciate help in fixing tests, because these changes break at least 30 tests, and I'm not sure where my change needs more nuance vs where tests should be updated to match the new behavior.

Current behavior:
a bar disappearing and reappearing as it's value is updated

Corrected behavior:
animated bar moving between 0 and other values

## Types of changes

I believe this is a breaking change, but I'm not sure.

## Checklist:

I'm unsure if this requires a documentation change

## Summary by CodeRabbit

* **Updates**
  * Updated bar and rectangle rendering logic to support zero-dimension shapes in charts.

* **Tests**
  * Added Storybook example for testing bar chart behavior with dynamically changing data values.

### DIFF
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -1071,7 +1071,7 @@ function BarImpl(props: BarImplProps) {
 
 export function computeBarRectangles({
   layout,
-  barSettings: { dataKey, minPointSize: minPointSizeProp, hasCustomShape },
+  barSettings: { dataKey, minPointSize: minPointSizeProp },
   pos,
   bandSize,
   xAxis,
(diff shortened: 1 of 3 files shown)


PavelVanecek:
The "skip rendering zero sized bars" is a performance optimization so that has to stay in one form or another.

There has been several PRs on this before. Please first let's gather them all so we have the full story of what happened, then we can see what is the best fix.

6TELOIV:
@PavelVanecek it feels to me that the optimization is currently too greedy. Skipping rendering rectangles **animating torwards** 0 isn't a correct assumption, as not rendering them causes this animation bug.

Instead, rendering should be skipped once they reach 0.

As the code is written, I believe this would mean moving the optimization further down the component tree and I'm not sure how it would impact other areas of the code. And it would of course be slightly less efficient, because more react nodes are needed. But, if DOM rendering is skipped in the stable state (i.e, after animation) the preformance should be more/less the same.

Issue #3177, opened 2023-01-14
React does not recognize the `textBreakAll` prop on a DOM element


Nefcanto:
The issue which is not created via https://recharts.github.io/recharts-issue-helper will be closed immediately.

---

注意：不是用 https://recharts.github.io/recharts-issue-helper 创建的 issue 会被立即关闭。

Hello

As soon as I started using Recharts (2.3.2 at the moment of writing this issue) I see this error:

```
react_devtools_backend.js:4012 Warning: React does not recognize the `textBreakAll` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `textbreakall` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    at div
    at Label (https://admin.project.local/node_modules/.vite/deps/recharts.js?v=401e0d79:18058:23)
    at g
    at https://admin.project.local/node_modules/.vite/deps/recharts.js?v=401e0d79:5633:24
    at LabelList (https://admin.project.local/node_modules/.vite/deps/recharts.js?v=401e0d79:18336:20)
    at g
    at https://admin.project.local/node_modules/.vite/deps/recharts.js?v=401e0d79:5633:24
    at Line2 (https://admin.project.local/node_modules/.vite/deps/recharts.js?v=401e0d79:25051:5)
    at svg
    at Surface (https://admin.project.local/node_modules/.vite/deps/recharts.js?v=401e0d79:5565:24)
    at div
    at CategoricalChartWrapper (https://admin.project.local/node_modules/.vite/deps/recharts.js?v=401e0d79:27924:7)
    at div
    at ResizeDetector2 (https://admin.project.local/node_modules/.vite/deps/recharts.js?v=401e0d79:10597:24)
    at https://admin.project.local/node_modules/.vite/deps/recharts.js?v=401e0d79:10838:21
    at LineChart (https://admin.project.local/src/Components/Dashboard/LineChart.jsx?t=1673676230067:18:3)
    at Chart (https://admin.project.local/src/Aggregates/Chart.jsx?t=1673676230067:22:3)
    at div
    at Widget (https://admin.project.local/src/Components/Dashboard/Widget.jsx:19:3)
    at div
    at div
    at Section (https://admin.project.local/src/Components/Dashboard/Section.jsx:35:3)
    at div
    at Dashboard (https://admin.project.local/src/Components/Dashboard/Dashboard.jsx:20:3)
    at ProjectDashboard
[...]

PR #7524, opened 2026-07-02
docs: clarify custom labels and ticks need SVG elements

ishaanlabs-gg:
## Description

Clarifies that custom `label`, `Label`, `LabelList`, `XAxis tick`, and `YAxis tick` React elements need to be SVG elements/components, not HTML elements such as `<div>`.

## Related Issue

Refs #3177

## Motivation and Context

Issue #3177 includes a request to document React DOM warnings when an HTML element is used as chart label/tick content. These prop comments feed the generated API documentation, so this makes the supported custom-rendering surface explicit near the affected props.

## How Has This Been Tested?

- `git diff --check HEAD~1 HEAD`
- Verified the new guidance appears in the affected source comments with `grep`

N/A

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Documentation**
  * Clarified guidance across chart components for `label`, `tick`, and label `content` to use SVG-based React elements (e.g., `<text>`/`<g>`).
  * Added explicit warnings in prop docs that HTML elements like `<div>` are invalid inside SVG-rendered charts and may trigger React DOM warnings.
  * Updated label-related documentation consistently, including Treemap and axis label typings.

### DIFF
--- a/omnidoc/commentSimilarityExceptions.ts
+++ b/omnidoc/commentSimilarityExceptions.ts
@@ -124,6 +124,11 @@ export const commentSimilarityExceptions: ReadonlyArray<CommentSimilarityGroup> 
     reason: 'LabelList dataKey has special complications compared to other components',
   },
   {
+    components: ['LabelList'],
+    props: ['content'],
+    reason: 'LabelList content renders labels for data items, unlike Treemap content which renders tree nodes',
+  },
+  {
     components: ['Text'],
     props: ['width'],
     reason: 'has extra details on automated wrapping',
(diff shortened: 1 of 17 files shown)


PavelVanecek:
Can you double check the issue number? I don't think that's related.

ishaanlabs-gg:
Good catch. #3177 includes the docs request for HTML elements used as labels, but the issue also has broader follow-up discussion. I changed this from `Closes #3177` to `Refs #3177`, and pushed a consistency fix for the Omnidoc failures.

PR #7530, opened 2026-07-05
fix(Brush): keep controlled startIndex/endIndex on data updates

### LINKED ISSUE #7462: Cannot control Brush indices

The controlled indices are reset whenever the chart data changes.

The controlled indices should stay stable when the chart data changes.

[Click the “trigger bug” button](https://stackblitz.com/edit/x8aw3bdu?file=src%2Findex.tsx) and see the brush jump.

Recharts 3.8.1 – note this did not happen in 2.x

This may be related to #2672 and was mentioned in #7361

momomuchu:
## Description

Controlled `startIndex`/`endIndex` on `<Brush>` get silently overwritten when the chart `data` array changes. The sync effect that copies the props into Redux only depended on the index props themselves, so once `chartDataSlice` reset `dataStartIndex`/`dataEndIndex` on a data update, nothing re-applied the controlled values. Adding `chartData` to that effect's dependency array makes it re-assert the controlled indices any time the data changes, not just when the index props change.

This does not touch `chartDataSlice.ts` or the reducer's reset behavior at all, and does not affect uncontrolled Brush usage in any way (`setDataStartEndIndexes` is a no-op when both indices are undefined). It's unrelated to the referential-equality discussion in #7021 - that was about whether Brush should treat a same-length new array reference as "unchanged" for internally-managed selection state, which is a separate question from what a controlled component does with props it was explicitly given. This PR doesn't reopen that; it just makes controlled props actually win, the same way a controlled `<input value>` would.

## Related Issue

Fixes #7462

## Motivation and Context

Bug fix. Controlled Brush is a documented pattern and currently doesn't work reliably across data updates.

## How Has This Been Tested?

Added a unit test in test/cartesian/Brush.spec.tsx that renders a controlled Brush, changes the data array to a new reference of the same length, and asserts the controlled indices are preserved. Ran the full unit suite locally, all green.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Brush start/end range controls now remain consistent when the chart updates its underlying data, avoiding unintended resets of the selected brush positions.
  * Controlled brush indices are preserved across data re-renders, even when the chart receives a new data array instance.
* **Tests**
  * Added coverage to verify controlled Brush indices stay stable after parent chart re-renders with updated data identity.

### DIFF
--- a/src/cartesian/Brush.tsx
+++ b/src/cartesian/Brush.tsx
@@ -990,9 +990,14 @@ function BrushInternal(props: InternalProps) {
   const { startIndex: startIndexFromProps, endIndex: endIndexFromProps } = props;
 
   useEffect(() => {
-    // start and end index can be controlled from props, and we need them to stay up-to-date in the Redux state too
+    /*
+     * start and end index can be controlled from props, and we need them to stay up-to-date in the Redux state too.
+     * `chartData` is also a dependency here: whenever the chart data changes, `chartDataSlice` may reset
+     * `dataStartIndex` / `dataEndIndex` back to the full range. If the indices are controlled from props,
+     * this effect re-applies them so a controlled Brush selection is never silently clobbered by a data update.
+     */
     dispatch(setDataStartEndIndexes({ startIndex: startIndexFromProps, endIndex: endIndexFromProps }));
-  }, [dispatch, endIndexFromProps, startIndexFromProps]);
+  }, [dispatch, endIndexFromProps, startIndexFromProps, chartData]);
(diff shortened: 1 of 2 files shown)


PavelVanecek (approved):

PavelVanecek:
Sorry @momomuchu I merged https://github.com/recharts/recharts/pull/7542 and it appears that caused conflicts or your PR. May I please ask you to resolve

Edit: nevermind sorry it all looks good now

PR #7532, opened 2026-07-06
fix: keep active dot index on accessibility focus

### LINKED ISSUE #5589: The activeDot onClick function does not run on first click with accessibilityLayer enabled
Bug: the ``onClick`` function should run for the first point that you click on in a ``LineGraph`` when you have the accessibility layer on.

Currently, when you try to click on a point in a graph that has an accessibility layer, the point's``onClick`` function does not run and the active point is taken to the first point of the graph. It is only on the second click of this point that ``activeDot``'s ``onClick`` function runs.

CodeSandbox example: https://codesandbox.io/p/devbox/simple-line-chart-forked-z8zpqj

If a user clicks on a point in the graph, the ``onClick`` function should run and the point selected should not change.

From my understanding, this click is activating the accessibility layer instead - hence why the behaviour is not present when the accessibility layer is off. However; this leads to a degraded experience for most users - as the change in which point is selected and onClick not running is jarring. To contextualize, this functionality is important to display additional information for the point that is selected, like a dialog or alert.

Tested in Recharts 2.12.7 and 2.15.1 on Chrome with MacOS.

pupuking723:
Fixes #5589.

When `accessibilityLayer` is enabled, focusing the chart calls `AccessibilityManager.focus()`, which spoofs mouse movement from its internal index. Mouse hover/click already updates the chart's `activeTooltipIndex`, but the accessibility manager can still hold an older index, so the first pointer click can move the active dot back to the first point before the `activeDot` click handler runs.

This syncs the accessibility manager index from mouse-derived tooltip state, keeping focus from moving the active dot away from the hovered point.

Verification:

- `npm run test -- test/chart/LineChart.spec.tsx -t "focus should not move activeDot away from the hovered point with accessibilityLayer enabled"`
- `npm run test -- test/chart/LineChart.spec.tsx`

Note: I could not run local eslint/typecheck cleanly in this worktree because the 2.x checkout is using a temporary shared `node_modules` from the main branch; that resolves to newer ESLint/TypeScript versions than 2.x expects.

### DIFF
--- a/src/chart/generateCategoricalChart.tsx
+++ b/src/chart/generateCategoricalChart.tsx
@@ -1511,6 +1511,12 @@ export const generateCategoricalChart = ({
       eventCenter.removeListener(SYNC_EVENT, this.handleReceiveSyncEvent);
     }
 
+    syncAccessibilityIndex(state: CategoricalChartState) {
+      if (typeof state.activeTooltipIndex === 'number' && state.activeTooltipIndex >= 0) {
+        this.accessibilityManager.setIndex(state.activeTooltipIndex);
+      }
+    }
+
     handleLegendBBoxUpdate = (box: DOMRect | null) => {
       if (box) {
         const { dataStartIndex, dataEndIndex, updateId } = this.state;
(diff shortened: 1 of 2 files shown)


PavelVanecek:
I'm afraid I'm not comfortable merging any more changes to 2.x. Please encourage update to 3.x instead

pupuking723:
Thanks for the quick look. That makes sense; I’ll close this and leave 2.x as-is.

PR #7546, opened 2026-07-11
feat(Sankey): support accessibilityLayer, title and desc props

MahinAnowar:
## Description

Sankey currently rejects the `accessibilityLayer`, `title` and `desc` props at the type level, and even at runtime `title`/`desc` get stripped by `svgPropertiesNoEvents` before they reach `Surface`, so there's no way to give the chart an accessible name or description.

This adds the three props to `SankeyProps` and handles them locally in `SankeyImpl`: `title`/`desc` are passed explicitly to the existing `Surface` (which already knows how to render them), and `role`/`tabIndex` default to `application`/`0` when `accessibilityLayer` is on (the default, same as every other chart), while explicit `role`/`tabIndex` props still win. Same precedence logic as `MainChartSurface`.

I deliberately did not move Sankey onto `RootSurface` — that was the approach in #7153 and the concern there was that it drags in ZIndex portal support as a side effect. No changes to `Surface` either. Keyboard navigation wiring for Sankey nodes/links is out of scope here; this covers the props and the SVG attributes.

## Related Issue

Fixes #7151

## Motivation and Context

Users migrating from 2.x lost the ability to set `accessibilityLayer`/`title`/`desc` on Sankey (TS errors, and silently dropped at runtime). Every other chart supports them.

## How Has This Been Tested?

Added an `accessibility` describe block to `test/chart/Sankey.spec.tsx`: default role/tabindex, `accessibilityLayer={false}` omitting them, explicit `role`/`tabIndex` overriding the defaults, and `title`/`desc` rendering. The default-attributes and title/desc tests fail on current main and pass with this change; the full Sankey spec (33 tests), `AccessibilityLayer.spec.tsx` and `Sankey.typed.spec.tsx` all pass locally, plus `tsc` and eslint on the changed files.

n/a — attribute-only change, no visual difference.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added accessibility support to Sankey charts by default.
  * Charts now include appropriate keyboard navigation and application semantics.
  * Added support for custom titles and descriptions for screen readers.
  * Added options to disable the accessibility layer or override accessibility attributes.

### DIFF
--- a/src/chart/Sankey.tsx
+++ b/src/chart/Sankey.tsx
@@ -852,6 +852,14 @@ interface SankeyProps extends EventThrottlingProps {
   className?: string;
   children?: ReactNode;
   /**
+   * Turn on accessibility support for keyboard-only and screen reader users.
+   *
+   * @defaultValue true
+   */
+  accessibilityLayer?: boolean;
+  title?: string;
+  desc?: string;
+  /**
    * Empty space around the container.
    *
    * @defaultValue {"top":5,"right":5,"bottom":5,"left":5}
@@ -1203,6 +1211,7 @@ function AllNodeElements({
 }
 
 export const sankeyDefaultProps = {
+  accessibilityLayer: true,
   align: 'justify',
   dataKey: 'value',
   iterations: 32,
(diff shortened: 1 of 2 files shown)


PavelVanecek (approved):
Approve but I'm going to treat it as a new feature rather than a bugfix

PR #7565, opened 2026-07-20
fix(CartesianAxis): skip renderedTicks dispatch when tick values are unchanged

### LINKED ISSUE #7563: Maximum update depth exceeded via internal store dispatch in passive effects during rapid external re-renders (RenderedTicksReporter); hard crash under fast slider scrubbing
**Versions:** recharts 3.9.2, React 19.2.5, Next.js 16.2.10 (Turbopack), Chrome.

**Setup:** an `AreaChart`/`LineChart` inside `ResponsiveContainer` with `XAxis`, `YAxis`, `Tooltip`, `CartesianGrid`, and a few `ReferenceLine`s. The parent component re-renders in response to a range slider (an "as of" time scrubber): each input event calls `setState`, which recomputes the chart's `data` array and a `ReferenceLine` `x`.

**Problem 1 — hard crash under fast scrubbing.** With the app calling `setState` once per slider input event (no throttling), a fast mouse drag (input events arrive faster than frame rate) trips React's nested update limit:

```
Error: Maximum update depth exceeded. ...
The above error occurred in the <RenderedTicksReporter> component.
```

Captured stack: `RenderedTicksReporter.useEffect` → store `dispatch` → `notifyNestedSubs` → `forceStoreRerender`, firing during `commitPassiveUnmountOnFiber`, interleaved with the app's own `dispatchSetState` per slider event until the 50 nested update limit throws. In development this crashes the route to the nearest error boundary at genuine human drag speed; in a production build the same crash (minified React #185) reproduces under higher-rate event bursts.

**Problem 2 — the dev warning floor.** After fixing the app side — coalescing slider updates to one `setState` per animation frame, hoisting every static prop to module constants (`margin`, `axisLine`, `tickFormatter`, `label` objects), and memoizing the `ReferenceLine` elements so only genuinely data-driven props change identity — the crash is gone at any input rate. But during *continuous* scrubbing the dev-only "Maximum update depth exceeded" `console.error` still fires roughly once per 50 frames: each frame's legitimate data change (new `data` array + `ReferenceLine` `x`) is followed by at least one nested store dispatch from a recharts passive effect, so React's nested update counter never sees a quiet frame to reset.

**Question/ask:** could the axis/tick settings sync (and `RenderedTicksReporter`) skip dispatching when the incoming settings are value-equal to the stored ones, rather than reference-equal? That would let data-driven re-renders (animation/scrubbing use cases) complete without a nested store dispatch per frame, which both removes the dev warning floor and adds headroom against the hard crash for apps that haven't frame-coalesced their updates.

Related but distinct from #6613 (inline object props, fixed in 3.5.0) and #7463 (React 19 Suspense ref cleanup, open) — this one needs no Suspense and survives full prop identity stabilization. Happy to put together a minimal repro if useful.

kimlj:
## Description

`RenderedTicksReporter` dispatched `setRenderedTicks` from its passive effect on every render, because the `ticks` array it receives is rebuilt by a bare `getTicks()` call in render and so gets a new identity each time even when its contents are unchanged. On top of that, since `ticks` is in the effect's dependency array, every re-render also ran the effect cleanup, dispatching `removeRenderedTicks` — two store dispatches per axis per render.

This PR makes the reporter value-compare the outgoing tick items against the last dispatched ones (via a ref + `isEqual`) and skip the dispatch when nothing changed, and moves the `removeRenderedTicks` cleanup into its own effect so it only fires on actual unmount.

## Related Issue

Fixes #7563

## Motivation and Context

As described in #7563: when an app re-renders a chart rapidly (e.g. a range slider driving the chart's `data` each frame), every render triggered nested store dispatches from passive effects. Interleaved with the app's own `setState` calls, this trips React's maximum update depth limit (`RenderedTicksReporter.useEffect → dispatch → notifyNestedSubs → forceStoreRerender` during `commitPassiveUnmountOnFiber`) — a hard crash at real human drag speed in dev, and reproducible in production builds under higher-rate bursts. Even fully identity-stabilized apps still hit the dev-only warning floor because every legitimate data change was followed by a nested store dispatch.

With this change, data-driven re-renders with unchanged tick output complete without any nested store dispatch.

## How Has This Been Tested?

- New regression test in `test/cartesian/XAxis/XAxis.state.spec.tsx` asserting the rendered-ticks state stays referentially stable across a re-render with unchanged tick values. It fails against the previous implementation and passes with the fix.
- Full unit suite: 311 files, 16,228 tests passing (Windows, Node 18+).
- 29 existing render-count assertions across 10 test files needed their expected counts reduced by one: each chart now mounts with one fewer store update. The asserted values are unchanged — only the counts decreased, which is the measurable effect of removing the redundant dispatch.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced unnecessary state updates by only publishing rendered axis tick data when the tick contents actually change.
  * Improved cleanup behavior to remove rendered tick state when the axis changes or unmounts, without tying it to tick recalculation.

* **Tests**
  * Updated existing chart, axis, error bar, selector, and tooltip tests to match reduced redundant selector invocations.
  * Added an integration test asserting referentially stable rendered tick arrays when charts re-render without underlying tick value changes.

### DIFF
--- a/src/cartesian/CartesianAxis.tsx
+++ b/src/cartesian/CartesianAxis.tsx
@@ -5,6 +5,7 @@ import * as React from 'react';
 import { SVGProps, useState, useRef, useCallback, forwardRef, useImperativeHandle, useEffect } from 'react';
 
 import get from 'es-toolkit/compat/get';
+import isEqual from 'es-toolkit/compat/isEqual';
 import { clsx } from 'clsx';
 import { Layer } from '../container/Layer';
 import { Text, Props as TextProps, TextAnchor, TextVerticalAnchor, isValidTextAnchor } from '../component/Text';
@@ -311,9 +312,17 @@ function RenderedTicksReporter({
   axisId: AxisId | undefined;
 }) {
   const dispatch = useAppDispatch();
+  /*
+   * The ticks array gets a new identity on every render even when its contents are unchanged.
+   * If we dispatched on every identity change, then every parent re-render would trigger
+   * a store update from a passive effect. Under rapid external updates (e.g. a slider
+   * driving chart data), those nested dispatches can trip React's update depth limit.
+   * https://github.com/recharts/recharts/issues/7563
+   */
+  const lastDispatchedTicksRef = useRef<ReadonlyArray<TickItemType> | null>(null);
   useEffect(() => {
     if (axisId == null || axisType == null) {
-      return noop;
+      return;
     }
     // Filter out irrelevant internal properties before exposing externally
     const tickItems = ticks.map(tick => ({
@@ -322,7 +331,17 @@ function RenderedTicksReporter({
       offset: tick.offset,
       index: tick.index,
     }));
+    if (isEqual(lastDispatchedTicksRef.current, tickItems)) {
+      return;
+    }
+    lastDispatchedTicksRef.current = tickItems;
     dispatch(setRenderedTicks({ ticks: tickItems, axisId, axisType }));
+  }, [dispatch, ticks, axisId, axisType]);
+
+  useEffect(() => {
+    if (axisId == null || axisType == null) {
+      return noop;
+    }
     return () => {
       dispatch(
         removeRenderedTicks({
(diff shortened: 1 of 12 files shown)


kimlj:
Fixed in b7c0206e. The ref now stores axisId and axisType next to the tick contents and only skips the dispatch when all three are unchanged.

PavelVanecek (approved):

PR #7566, opened 2026-07-20
fix(Area): connectNulls connects across stacked points where every series is null

### LINKED ISSUE #6985: stackId and connectNulls don't work together for points with no data

kimlj:
Fixes #6985.

Stacked AreaChart with connectNulls, dataWithGaps dataset. Before: all-null point at Page C collapses to 0. After: Page C connects across, and the partial-null uv at Page F still renders as 0.

Top chart is master, bottom is this PR. The all-null point (Page C) connects across; the partial-null point (Page F, only `uv` missing) still renders as 0, so the #7073 regression does not return.

## Problem

In a stacked AreaChart with `connectNulls`, a data point where every series is null renders as a drop to 0 instead of connecting across the gap. d3-stack defaults nulls to 0, so `value1` is never null for stacked data, and the `!connectNulls` guard is the only thing controlling null behavior there.

## Change

`computeArea` in `src/cartesian/Area.tsx` marks a point as a break point when every dataKey in the stack group is null (`wholeStackIsNull`). The stack dataKeys are supplied by a new `selectStackDataKeys` selector in `src/state/selectors/areaSelectors.ts`.

The `!connectNulls && rawValue == null` guard is unchanged, so a point where only some series are null still renders as 0 in the stack.

- Every series null at a point: connects across (new behavior).
- Some series null at a point: unchanged, renders as 0 to preserve stack offsets.

## Prior art

This is the design proposed by @Harikrushn9118 in #7085: keep the existing guard, add a separate all-null check, and pass the stack dataKeys into `computeArea`. #7085 was closed by its author before completion. This PR implements that design and commits the VR snapshots #7085 was missing.

## History

#7041 removed the `!connectNulls` guard and was merged, then reverted in #7073 because it made stacked charts with a single null series render incorrectly. This PR keeps that guard and scopes the new behavior to the all-null case, so the #7073 regression does not return.

## Tests

- `test/chart/AreaChart.spec.tsx`: two regression tests, "connects across points where every stacked series is null if connectNulls is true" and "breaks at points where every stacked series is null if connectNulls is false". Both fail without the source change. The existing "Renders null points as 0 if stacked and connectNulls is true" test covers the partial-null case that must stay unchanged.
- `www` example `AreaChartConnectNulls.tsx`: adds a `dataWithGaps` dataset with an all-null point at Page C and a partial-null point at Page F.
- VR snapshots committed for chromium, firefox, and webkit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stacked area gap/break detection when an entire stacked point has all series values missing, so gaps are handled correctly.
  * `connectNulls` now reliably bridges fully missing stacked points, while keeping breaks when `connectNulls` is disabled.
  * Partial missing values still render with correct stacked contributions.
* **Documentation**
  * Added stacked area examples showing how gaps are rendered with and without `connectNulls`.
* **Tests**
  * Added coverage validating `connectNulls` behavior for all-missing stacked points in area charts.


PavelVanecek:
Why did we close https://github.com/recharts/recharts/pull/7085 ?

PavelVanecek (approved):

PR #7570, opened 2026-07-20
feat(XAxis): support height="auto" to size the axis to its tick labels

### LINKED ISSUE #7301: Allow XAxis to have auto-height similar to YAxis' auto-width

Currently, if we rotate the X-Axis tick labels, we need to manually fit the X-Axis label and tick labels via `margin.bottom` and `Label.offset`.

The ability to have auto-width was recently added to the Y-Axis as per this [PR](https://github.com/recharts/recharts/pull/5880). However, such a feature does not exist yet for the X-Axis. I might be wrong, but it should be possible to use a similar implementation for X-Axis.

I would like to be able to set a single string `"auto"` to `XAxis.height`.

Currently, I am using a hook that looks a little like the following to get things to auto-calculate the tick label height but this doesn't provide a good developer experience since this requires a ref and setting the height in a few places.

```ts
// Get the x-axis tick labels
const ticks = element.querySelectorAll<SVGTextElement>(
  '.recharts-xAxis-tick-labels .recharts-cartesian-axis-tick-value'
);

// Get the maximum height of the tick labels rounded up to the nearest integer
return Math.ceil(
  Array.from(ticks).reduce(
    (max, tick) => Math.max(max, tick.getBoundingClientRect().height),
    0
  )
);
```

**Additional context**

Here is a picture of what is desired:

Image

kimlj:
Fixes #7301.

Adds `height="auto"` to `XAxis`, mirroring the existing `YAxis` `width="auto"` (#5880). When set, the axis height is measured from the rendered tick labels and axis label after layout, then dispatched back into the axis state, so rotated or multi-line tick labels are no longer clipped and don't need manual `margin.bottom` tuning.

The implementation mirrors the Y-axis path end to end:

- `getCalculatedXAxisHeight` measures the tallest tick label plus tick size/margin and the axis label (analog of `getCalculatedYAxisWidth`).
- `CartesianAxis` exposes `getCalculatedHeight()` on its ref.
- `XAxisImpl` runs a `useLayoutEffect` that dispatches `updateXAxisHeight` when `height === 'auto'`, with the same oscillation guard, and the same skips when `label` is a function or a React element.
- `height` is typed `number | 'auto'`; the size and offset selectors fall back to `DEFAULT_X_AXIS_HEIGHT` (30) until measured, so the default behavior is unchanged.

## Tests

New `test/cartesian/XAxis/XAxis.autoHeight.spec.tsx` covers the fixed-height control case and the dynamic `"auto"` case (calculated height 88 = 80 tick + 6 tick size + 2 tick margin).

Local checks: `tsc --noEmit` clean, eslint clean, existing XAxis/YAxis/state suites (848 tests) still green.

## Summary by CodeRabbit

* **New Features**
  * Added automatic X-axis height calculation based on rendered tick labels and axis labels.
  * X-axis height now accepts `"auto"` in addition to numeric values.
  * Added a default X-axis height of 30 when no numeric height is available.

* **Bug Fixes**
  * Improved chart layout offsets when X-axis height is calculated dynamically.
  * Prevented minor height oscillations during automatic sizing.

* **Tests**
  * Added coverage for numeric and automatic X-axis height behavior.

### DIFF
--- a/omnidoc/commentSimilarityExceptions.ts
+++ b/omnidoc/commentSimilarityExceptions.ts
@@ -55,6 +55,11 @@ export const commentSimilarityExceptions: ReadonlyArray<CommentSimilarityGroup> 
     reason: 'YAxis width allows "auto" value, unlike other components',
   },
   {
+    components: ['XAxis'],
+    props: ['height'],
+    reason: 'XAxis height allows "auto" value, unlike other components',
+  },
+  {
     components: ['Brush'],
     props: ['width'],
     reason: 'Brush width defaults to chart width, unlike other components',
--- a/scripts/snapshots/es6Files.txt
+++ b/scripts/snapshots/es6Files.txt
@@ -239,6 +239,7 @@
   "es6/util/ReduceCSSCalc.js",
   "es6/util/ScatterUtils.js",
   "es6/util/TickUtils.js",
+  "es6/util/XAxisUtils.js",
   "es6/util/YAxisUtils.js",
   "es6/util/axisPropsAreEqual.js",
   "es6/util/createCartesianCharts.js",
--- a/scripts/snapshots/libFiles.txt
+++ b/scripts/snapshots/libFiles.txt
@@ -239,6 +239,7 @@
   "lib/util/ReduceCSSCalc.js",
   "lib/util/ScatterUtils.js",
   "lib/util/TickUtils.js",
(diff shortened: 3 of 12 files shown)


PavelVanecek (approved):

PavelVanecek:
Example on the website would be nice

PR #7582, opened 2026-07-24
fix(bar): keep barGap correct when maxBarSize clamps the bar width (#2774)

### LINKED ISSUE #2774: Gap between bars with `maxBarSize` in the same category is larger than the specified `barGap`

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/simple-bar-chart-forked-q3uop?file=/src/App.tsx)

### Steps to reproduce
* Create a wide bar chart with a single categories (or very few categories) and 2 bars in each category.
* Set the `barGap={1}` and `maxBarSize={80}` props.

### What is expected?
The gap between the 2 bars is that of the specific `barGap` prop.

Screenshot 2022-02-02 at 16 24 04

_This screenshot is achieved using `barSize={80}` prop. Using this prop isn't ideal in my case because the number of bars or width of the chart is completely arbitrary._

The gap between the 2 bars within the category doesn't respect the specified `barGap`. Instead the space seems evenly divided between the bars, resulting in a very large space between the 2 bars, which doesn't look very good in a very wide chart.

Screenshot 2022-02-02 at 16 23 59

| Environment | Info |
|---|---|
| Recharts | v2.0.10 |
| React | 17.0.2 |
| System | macOS Big Sur |
| Browser | Chrome 97 |

sobol-sudo:
## Description

In `getBarPositions` (`src/state/selectors/combiners/combineAllBarPositions.ts`), when `maxBarSize` clamps the computed bar width, each bar was still advanced by the *unclamped* stride and then centred inside its own original slot:

```ts
offset: offset + (originalSize + realBarGap) * i + (originalSize - size) / 2,
```

So the visible distance between two clamped bars became `realBarGap + (originalSize - size)` instead of `realBarGap`. This centres the clamped bars as a *group* and strides by the clamped width:

```ts
offset: offset + (len * (originalSize - size)) / 2 + (size + realBarGap) * i,
```

The clamped group occupies `len * size + (len - 1) * realBarGap`, the unclamped layout occupied `len * originalSize + (len - 1) * realBarGap`, so centring one inside the other is a right-shift of exactly `len * (originalSize - size) / 2`.

**When `size === originalSize` the shift term is `0` and the stride is `originalSize + realBarGap`, i.e. the new expression reduces algebraically to the old one.** Nothing moves for charts that don't hit the `maxBarSize` clamp, and no existing snapshot changes.

## Related Issue

Fixes #2774.

## Motivation and Context

Repro from the issue: `bandSize=400`, `barCategoryGap='10%'`, `barGap={1}`, `maxBarSize={80}`, two `<Bar>`s.

- `originalSize = Math.round((400 - 80 - 1) / 2) = 160`, clamped `size = 80`
- before: `[{ offset: 80, size: 80 }, { offset: 241, size: 80 }]` → an **81px** gap for a requested `barGap` of 1
- after: `[{ offset: 120, size: 80 }, { offset: 201, size: 80 }]` → gap is 1, group centred in the band

## How Has This Been Tested?

**Unit tests** — added two cases to `test/state/selectors/combiners/combineAllBarPositions.spec.ts`: the #2774 repro, and an explicit guard that a non-binding `maxBarSize` produces positions identical to `maxBarSize={undefined}`.

**VR tests** — added `test-vr/tests/Bar.2774.spec-vr.tsx` with three charts, following `Bar.6192.spec-vr.tsx`. It deliberately does not reuse the bar-alignment guide's playground as-is, because that example sets no `maxBarSize` and so cannot show this bug; it keeps the guide's three-`<Bar>` shape and adds a clamping `maxBarSize`.

1. `maxBarSize` that does **not** clamp — the control. Must look identical before and after this change.
2. `maxBarSize` on the chart that **does** clamp — the bug.
3. The same clamp set per-`<Bar>`, which resolves through the `childMaxBarSize` branch instead of `globalMaxBarSize`.

Running the new VR test against `main` (fix reverted, snapshots kept) gives exactly the intended split:

```
✘ BarChart with barGap and a maxBarSize that clamps
    26627 pixels (ratio 0.12 of all image pixels) are different.
✘ BarChart with barGap and a clamping maxBarSize set on each Bar
    26627 pixels (ratio 0.12 of all image pixels) are different.
✓ BarChart with barGap and a maxBarSize that does not clamp

  2 failed
  1 passed
```

The two clamped charts move; the unclamped control does not. Snapshots were generated through `docker compose run --rm test-vr playwright-test`, so they are the Linux baselines for all three browsers.

**Full suite** — `npm test` is green at 313 files / 16265 tests, plus `npm run check-types` clean and `npm run lint` at 0 errors.

Attached in the VR snapshots. The visible change with `barCategoryGap="10%"`, `barGap={4}`, `maxBarSize={12}` and three `<Bar>`s: before, the three thin bars sit spread across the category, each centred in the slot the wide bar would have occupied, and the requested `barGap` is nowhere to be seen; after, they sit `barGap` apart as a centred group.

## Types of changes

## Checklist:

---

One note on the choice of convention: the alternatives are left-aligning the clamped group against `barCategoryGap`, or scaling `barGap` proportionally with the clamp. I went with group-centring because it is the only one of the three that is byte-identical to today's output in the unclamped case, so it carries no risk for existing charts. Happy to switch if you'd rather have one of the others.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected horizontal bar positioning when maximum bar size clamps the computed bar width.
  - Bars are now centered as a group while preserving the requested gaps when clamping is applied.
  - When clamping does not affect the computed size, the bar layout remains the same.
- **Tests**
  - Added regression coverage for max bar size clamping behavior, including screenshot-based checks for chart-level and per-bar settings.


PavelVanecek:
We have this guide https://recharts.github.io/en-US/guide/barAlignment/ it would be interesting to see the changes visually there. At least as screenshots in the PR, if not a full VR test

sobol-sudo:
Here you go — `barCategoryGap="10%"`, `barGap={4}`, `maxBarSize={12}`, three `<Bar>`s, chromium.

<table>
<tr>
<th width="50%">Before</th>
<th width="50%">After</th>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

Before, the three thin bars are spread across the category — each one sits centred in the slot a full-width bar would have taken, so the requested `barGap={4}` is nowhere to be seen. After, they sit 4px apart as a group centred in the category.

I went with the VR test too: `test-vr/tests/Bar.2774.spec-vr.tsx`, in the shape of `Bar.6192.spec-vr.tsx`. I didn't reuse the guide's playground as-is because it sets no `maxBarSize`, so the bug can't show up there. The test keeps the guide's three-`<Bar>` layout and adds a clamping `maxBarSize`, plus a control chart where `maxBarSize` is too large to bind, and a third where the clamp is set per-`<Bar>` so the `childMaxBarSize` branch is covered as well.

Running it against `main` with the fix reverted:

```
✘ BarChart with barGap and a maxBarSize that clamps
    26627 pixels (ratio 0.12 of all image pixels) are different.
✘ BarChart with barGap and a clamping maxBarSize set on each Bar
    26627 pixels (ratio 0.12 of all image pixels) are different.
✓ BarChart with barGap and a maxBarSize that does not clamp

  2 failed
  1 passed
```

The two clamped charts move, the control doesn't — which is the property I care about most here, since it's what says existing charts don't shift. Snapshots are the Linux baselines for all three browsers, generated with `docker compose run --rm test-vr playwright-test`.

The "before" image above lives on a scratch branch of my fork, so it isn't part of this PR's diff.

Happy to add a `maxBarSize` lever to the guide playground as well, if you think that belongs there.

PavelVanecek:
That's quite a significant change in rendering logic if you ask me. I hope we don't make anyone too upset

sobol-sudo:
Thanks for merging!

On the blast radius — it's narrower than the diff makes it look. The old and new expressions are algebraically identical unless **both** of these hold: the category has 2+ bars, *and* `maxBarSize` actually binds (`size < originalSize`). With a single `<Bar>`, `len * (originalSize - size) / 2` reduces to the old `(originalSize - size) / 2`; with nothing clamped, that term is 0 and the stride is unchanged either way. I swept both formulas over ~7.5k combinations of `len` / `originalSize` / `size` / `barGap` / `offset` and every differing case had both properties — which is why no existing snapshot or test moved.

So the charts that shift are exactly the ones that were drawing a gap they never asked for. The people most likely to notice are those who'd been compensating with a hand-tuned `barGap`.

PR #7593, opened 2026-07-28
fix(Pie): interpolate paddingAngle during animation

### LINKED ISSUE #5307: Pie chart gap is sometimes missing
Bug: `PieChart` sometimes doesn't leave a gap between `Pie` segments when it should

The padding is missing between two pie slices under a specific scenario:
- Use a `PieChart` where `isAnimationActive` is true and `paddingAngle` > 0
- Render at least 3 Cells where a cell has a value of 0 that isn't the first or last cell
- First and last cells will have no gap or a reduced gap between them

Example: https://codesandbox.io/p/sandbox/recharts-padding-bug-m3835z

All cells should have the specified paddingAngle between them

This has been observed in Recharts 2.14.1 and 2.12.7

pratik-desgn:
## Description

Fixes #5307.

When a segment's value crosses to/from zero, its `paddingAngle` changes too (a zero-value segment gets no leading padding). The pie animation in `defaultPieAnimateItems` applied the new `paddingAngle` immediately on the very first animation frame instead of interpolating it over time like the arc angle itself. That threw off the position of every sector computed after it, including the closing gap between the last and first sector.

This only affects the animated code path (`isAnimationActive`). It's separate from #7113, which fixed the static `computePieSectors` cursor calculation for `isAnimationActive={false}`.

## Fix

`paddingAngle` is now interpolated between `item.prev.paddingAngle` and `item.next.paddingAngle` the same way the arc angle is, for `matched` items.

## Test plan

- Added `test/polar/Pie/PiePaddingAngleAnimation.spec.tsx`: renders a `PieChart` with a middle segment animating to/from a zero value and asserts the closing gap between the last and first sector stays stable throughout the animation.
- `npm run test` (Pie-related suites) passes.
- `npm run lint` passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pie chart animations so sector padding gaps transition smoothly when segment values change, preventing visible snapping during animation.
  - Maintained stable spacing between the first and last sectors throughout animated transitions.

- **Tests**
  - Added regression coverage for pie sector `paddingAngle` behavior during animated data updates.


PavelVanecek:
https://flakiness.io/recharts/recharts/pull/7593 this test failure is from another change that's already fixed. Please merge latest main

PR #7596, opened 2026-07-28
fix(Rectangle): keep the rectangle mounted while it animates to zero

### LINKED ISSUE #7485: Bars don't animate to zero, but will animate from zero

adarshsm:
## Description

`Rectangle` bails out early and returns `null` as soon as its `width` or `height` is `0`. That check runs before the `JavascriptAnimate` block, so a rectangle whose target height becomes `0` unmounts on the very first frame and never gets a chance to shrink. Growing *from* zero was unaffected, because in that direction the target dimension is non-zero — which is exactly the asymmetry described in the issue.

This splits the guard in two:

- the invalid-number check (`x !== +x` etc.) still returns `null` immediately, unchanged
- the zero-dimension check now keeps the rectangle mounted when `isUpdateAnimationActive` is set *and* the previous size was non-zero, i.e. only while it is animating out

Once the animation lands on zero, the tracked previous size is zero too, so the next render takes the normal early return and the rectangle unmounts as before.

## Related Issue

Closes #7485

## Motivation and Context

Bars, and anything else built on `Rectangle`, visibly pop out of existence instead of animating when their value drops to zero, while the reverse transition animates correctly. The inconsistency is most obvious on a bar chart whose data updates in place.

## How Has This Been Tested?

Added three cases to `test/shape/Rectangle.animation.spec.tsx` under `when the height changes to zero`:

- with `isUpdateAnimationActive`, the rectangle stays mounted after the height changes to `0` so it can animate out
- with `isUpdateAnimationActive={false}`, it is removed immediately — the existing behaviour
- a rectangle rendered with `height={0}` from the start still renders nothing

I checked the first test fails on `main` without the source change (rect count drops to `0` immediately) and passes with it, and that the other two pass either way, so the non-animated and initial-zero paths are untouched.

Also ran locally on Node 22:

- `npm test` — full unit suite
- `npm run lint`
- `npm run tsc` and `npm run check-types-test`

## Types of changes

## Checklist:

I did not add a VR test — the visual-regression suite needs Docker to generate matching snapshots and I could not produce reference images that would line up with CI. Happy to add one if you'd like, or if you can point me at the right story to extend.

---

Unrelated heads-up while I was setting up: `npm run check-types-test-vr` currently fails on a clean `main` (`8644f1a9`) with two errors in `test-vr/tests/www/fixtures.tsx`, where the `mount` fixture no longer matches Playwright's overloads — it looks like fallout from the Playwright bump in #7589. Since `check-types` is part of the `pre-push` hook, that blocks pushing from a fresh clone. Not touched by this PR, just flagging it in case it hasn't been noticed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rectangles now animate smoothly when their width or height changes to zero, completing the shrink animation before disappearing.
  * Zero-sized rectangles are hidden immediately when animation is disabled or when initially rendered at zero dimensions.
  * Invalid numeric dimensions continue to be excluded from rendering.

* **Tests**
  * Added coverage for shrinking rectangles and their rendering behavior during and after animation.

### DIFF
--- a/src/shape/Rectangle.tsx
+++ b/src/shape/Rectangle.tsx
@@ -220,7 +220,19 @@ export const Rectangle: React.FC<Props> = rectangleProps => {
   const animationIdInput = useMemo(() => ({ x, y, width, height, radius }), [x, y, width, height, radius]);
   const animationId = useAnimationId(animationIdInput, 'rectangle-');
 
-  if (x !== +x || y !== +y || width !== +width || height !== +height || width === 0 || height === 0) {
+  if (x !== +x || y !== +y || width !== +width || height !== +height) {
+    return null;
+  }
+
+  /*
+   * A rectangle with a zero width or height has nothing to paint, so usually we render nothing at all.
+   * The exception is a rectangle that is animating down to zero: bailing out here would unmount it
+   * immediately and the shrinking animation would never run - which is why rectangles used to animate
+   * away from zero but not towards it.
+   */
+  const hasZeroDimension = width === 0 || height === 0;
+  const wasVisibleBefore = prevWidthRef.current !== 0 && prevHeightRef.current !== 0;
+  if (hasZeroDimension && !(isUpdateAnimationActive && wasVisibleBefore)) {
     return null;
   }
 
--- a/src/shape/Rectangle.tsx
+++ b/src/shape/Rectangle.tsx
@@ -2,7 +2,7 @@
  * @fileOverview Rectangle
  */
 import * as React from 'react';
-import { SVGProps, useEffect, useMemo, useRef, useState } from 'react';
+import { SVGProps, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { clsx } from 'clsx';
 import { AnimationDuration } from '../util/types';
 import { resolveDefaultProps } from '../util/resolveDefaultProps';
@@ -220,6 +220,16 @@ export const Rectangle: React.FC<Props> = rectangleProps => {
   const animationIdInput = useMemo(() => ({ x, y, width, height, radius }), [x, y, width, height, radius]);
   const animationId = useAnimationId(animationIdInput, 'rectangle-');
 
+  /*
+   * Re-render once the shrink-to-zero animation is over. By then the tracked previous size is zero
+   * too, so the zero-dimension check below unmounts the rectangle instead of leaving an invisible
+   * zero-size path behind.
+   */
+  const [, setAnimationEndTick] = useState(0);
+  const handleAnimationEnd = useCallback(() => {
+    setAnimationEndTick(tick => tick + 1);
+  }, []);
+
   if (x !== +x || y !== +y || width !== +width || height !== +height) {
     return null;
   }
@@ -275,6 +285,7 @@ export const Rectangle: React.FC<Props> = rectangleProps => {
       easing={animationEasing}
       isActive={isUpdateAnimationActive}
       begin={animationBegin}
+      onAnimationEnd={hasZeroDimension ? handleAnimationEnd : undefined}
     >
       {(animationElapsedTime: number) => {
         const currWidth = interpolate(prevWidth, width, animationElapsedTime);
--- a/test/shape/Rectangle.animation.spec.tsx
+++ b/test/shape/Rectangle.animation.spec.tsx
@@ -381,7 +381,7 @@ describe('Rectangle animation', () => {
               y={50}
               width={80}
               height={height}
-              radius={4}
+              radius={0}
               fill="#ff7300"
               isAnimationActive
               isUpdateAnimationActive={isUpdateAnimationActive}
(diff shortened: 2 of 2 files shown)


test/shape/Rectangle.animation.spec.tsx
```diff
@@ -358,4 +358,80 @@ describe('Rectangle animation', () => {
...
+      return container.querySelectorAll('.recharts-rectangle').length;
+    }
+
+    it('should keep the rectangle rendered so that it can animate out', async () => {
+      const renderTestCase = heightTestCase({ startHeight: 100, isUpdateAnimationActive: true });
+      const { container } = renderTestCase();
+      expect(countRectangles(container)).toBe(1);
+
+      await prime(container);
+      expect(countRectangles(container)).toBe(1);
```
PavelVanecek, line 431: Can we extend this test so that it's progressing through the animation and checks for height and then checks it unmounted once the animation completed?

adarshsm:
Some context on how this interacts with existing work:

The zero-dimension early return was added in #6800 for performance. This PR keeps that filtering for the case it targeted — statically zero rectangles still return `null`, since they either have update animation off or were never visible, so `wasVisibleBefore` is false. Only rectangles that are mid-animation stay mounted, bounded by the animation duration.

To be clear on scope: this does **not** fix #7377, which is about mount animations in the from-zero direction. The third test added here asserts that an initially-zero rectangle still renders nothing, so that behaviour is unchanged. #7486 looks like the complement of this change and touches the same guard, so one of the two will probably need a rebase depending on merge order.

test/shape/Rectangle.animation.spec.tsx
adarshsm: Done in c6b261c6. The test now primes the animation, steps it through 0.25 / 0.5 / 0.75 and asserts the height is strictly decreasing at every step (and never negative), then calls `completeAnimation()` and asserts the rectangle is gone.

Writing that last assertion turned up a real bug, so the commit fixes the component too. `JavascriptAnimate` re-renders only its own children as the animation progresses, so `Rectangle` itself never re-ran after the animation ended — the zero-dimension early return was evaluated with the stale `prevHeightRef` and an invisible zero-size path stayed mounted forever. It now bumps a state counter from `onAnimationEnd` (only when the target dimension is zero), so the check runs once more after the tracked previous size has reached zero as well.

One incidental change: `HeightTestCase` now renders with `radius={0}` instead of `radius={4}`, which keeps the path as a plain `M 50,50 h 80 v {height} h -80 Z` so the height can be read straight out of the `v` command. With rounded corners the height is smeared across the arc commands and is not readable without reimplementing the corner maths in the test.

Verified locally: `test/shape` plus both Bar specs are green (130 passed, 1 expected fail), and `check-types-lib` / `check-types-test` are clean.

src/shape/Rectangle.tsx
```diff
@@ -220,7 +220,19 @@ export const Rectangle: React.FC<Props> = rectangleProps => {
...
+
+  /*
+   * A rectangle with a zero width or height has nothing to paint, so usually we render nothing at all.
+   * The exception is a rectangle that is animating down to zero: bailing out here would unmount it
+   * immediately and the shrinking animation would never run - which is why rectangles used to animate
+   * away from zero but not towards it.
+   */
+  const hasZeroDimension = width === 0 || height === 0;
+  const wasVisibleBefore = prevWidthRef.current !== 0 && prevHeightRef.current !== 0;
+  if (hasZeroDimension && !(isUpdateAnimationActive && wasVisibleBefore)) {
```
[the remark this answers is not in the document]
adarshsm: Good catch — this was real, and it is fixed in c6b261c6.

The root cause is slightly different from the diagnosis above. `JavascriptAnimate` re-renders only its own children as the animation progresses, so `Rectangle` itself never re-rendered after the animation finished; the early return was never re-evaluated at all, rather than being evaluated and returning `null`. The fix bumps a state counter from `onAnimationEnd` (wired up only when the target dimension is zero) so the zero-dimension check runs once more, by which point `prevWidthRef`/`prevHeightRef` have interpolated down to zero and the rectangle unmounts.

Post-animation cleanup is now covered: the test steps through the animation asserting the height strictly decreases, then calls `completeAnimation()` and asserts `countRectangles(container)` is 0.

PavelVanecek (approved):

PR #7600, opened 2026-07-29
fix(Bar): stop stacked minPointSize inflation from overlapping neighboring segments

### LINKED ISSUE #2460: minPointSize overlap issue with stacked bar chart

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/recharts-issue-template-forked-fxs9d?file=/src/index.js)

### Steps to reproduce
Set up the following:

- Set up a stacked bar chart
- Add minPointSize setting to bars
- Add data, with a a dataKey that contains a small value (smaller than minPointSize)

### What is expected?
Stacks after the bar clamped to minPointSize should shift

Stacks overlap

| Environment | Info |
|---|---|
| Recharts | v2.0.7 |
| React | react |
| System | linux |
| Browser | chrome |

Related issues

- https://github.com/recharts/recharts/issues/1001
- https://github.com/recharts/recharts/issues/2441

I can submit a PR if you are OK with the behaviour I'm proposing.

pratik-desgn:
## Summary

Fixes #2460. When a stacked bar segment is shorter than `minPointSize`, it gets inflated to that minimum size - but the inflation only extended into whatever segment was stacked next to it. Segments further from the base kept their *natural* (un-inflated) position, so they'd visually overlap the inflated segment below them.

@PavelVanecek said "feel free to send a PR" on the issue, so here it is.

## Root cause

In `computeBarRectangles` (`src/cartesian/Bar.tsx`), each `<Bar>` computes its own stacked segment's position independently from the axis scale, and its own `minPointSize` inflation always keeps its *base-ward* edge fixed and extends only the far edge. That's correct in isolation, but the next segment stacked on top starts from its own natural (un-shifted) position - which is exactly where the previous segment *would have* ended before inflation, not where it now actually ends.

## Fix

Added `selectStackedMinPointSizeShift` in `src/state/selectors/barSelectors.ts`. For a given stacked `<Bar>`, it walks the other bars earlier in the same stack (via the existing `StackGroup.graphicalItems`/`stackedData`, already available from the `BarStack` machinery) and sums up how far each of them was pushed outward by their own `minPointSize` inflation. Each segment's whole rectangle is then translated by that cumulative shift before its own inflation is applied, so segments stay attached to their stack neighbor without overlapping - stacks grow outward instead of overlapping, matching the behavior requested in the issue ("Stacks after the bar clamped to minPointSize should shift").

Handles both chart layouts (`horizontal`/`vertical`) and cascades correctly across more than two stacked segments.

## Test plan
- [x] Added 3 regression tests: horizontal layout, vertical layout, and a 3-segment cascade, all asserting zero gap/overlap between adjacent segments (`test/cartesian/Bar/Bar.spec.tsx`)
- [x] Full `test/cartesian`, `test/chart`, `test/state` suites pass with no regressions (4000+ tests)
- [x] `tsc --noEmit` clean
- [x] `eslint` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stacked bar behavior with `minPointSize`, preventing overlap and gaps by shifting undersized segments outward.
  * Applied cascading adjustments across multiple stacked segments and when the visible data window changes.
  * Ensured consistent behavior in horizontal and vertical layouts.
* **Documentation**
  * Clarified `minPointSize` behavior for stacked bar charts.
* **Tests**
  * Expanded coverage for stacked segment geometry, multi-segment stacks, and brush interactions.


pratik-desgn:
Addressed both CodeRabbit findings: fixed the `stackedMinPointSizeShift` indexing to account for `dataStartIndex` (added a regression test with a `Brush`-driven non-zero `dataStartIndex` that fails without the fix), and replaced the `as BarSettings` cast with a proper `isBarSettings` type guard.

PavelVanecek:
Can we have screenshots before & after please?

pratik-desgn:
Sure! Repro: `<BarChart>` with two stacked bars, `stackId="s"`, `minPointSize={20}`, values `1` and `100`.

**Before** — the tiny (red) segment gets inflated to `minPointSize`, but the big (blue) segment on top of it doesn't know that happened, so it still starts from the tiny segment's un-inflated position and completely covers it up:

before

**After** — the blue segment is pushed outward by exactly the amount the red one was inflated, so both are fully visible with no overlap and no gap:

after

src/state/selectors/barSelectors.ts
```diff
@@ -244,6 +245,78 @@ export const selectStackedDataOfItem: (
...
+      return undefined;
+    }
+    const shifts: Array<number> = [];
+    for (let seriesIndex = 0; seriesIndex < myIndex; seriesIndex++) {
+      const series = stackedData[seriesIndex];
+      const item = graphicalItems[seriesIndex];
+      if (series == null || item == null || !('type' in item) || (item as BarSettings).type !== 'bar') {
+        continue;
+      }
+      const minPointSizeProp = (item as BarSettings).minPointSize;
```
PavelVanecek: I still don't like the type cast. It looks safe but it's actually a hole, and bugs will spread from here. Why is this necessary?

src/state/selectors/barSelectors.ts
```diff
@@ -244,6 +245,78 @@ export const selectStackedDataOfItem: (
...
+    const { stackedData, graphicalItems } = stackGroup;
+    const myIndex = graphicalItems.findIndex(item => item.id === barSettings.id);
+    if (myIndex <= 0) {
+      return undefined;
+    }
+    const shifts: Array<number> = [];
+    for (let seriesIndex = 0; seriesIndex < myIndex; seriesIndex++) {
+      const series = stackedData[seriesIndex];
+      const item = graphicalItems[seriesIndex];
+      if (series == null || item == null || !('type' in item) || (item as BarSettings).type !== 'bar') {
```
PavelVanecek: Same here. Why is the `as` necessary? What happens when we remove it?

pratik-desgn:
yeah the cast was just hiding the real problem — `graphicalItems` was typed as the base interface with no `type` field, so there was nothing to narrow on. widened it to the actual union (Area/Bar/RadialBar settings, the only three that extend it) and shoved the one unavoidable assertion into `combineStackGroups` where it's actually true. barSelectors just checks `item.type !== 'bar'` now. pushed.

src/state/selectors/axisSelectors.ts
```diff
@@ -707,7 +710,13 @@ export const combineStackGroups = (
...
           // @ts-expect-error getStackedData requires that the input is array of objects, Recharts does not test for that
           stackedData: getStackedData(displayedData, dataKeys, stackOffsetType),
-          graphicalItems: orderedGraphicalItems,
+          // Area, Bar, and RadialBar are the only graphical items that can ever carry a stackId
+          // (see MaybeStackedGraphicalItem's implementors), so DefinitelyStackedGraphicalItem items
+          // reaching this point are guaranteed - by construction at every call site of this function -
+          // to be one of those three. This is the one place that invariant is asserted; downstream
+          // consumers (e.g. barSelectors.ts) get real discriminated-union narrowing on `item.type`
+          // from here on, with no further casts of their own.
+          graphicalItems: orderedGraphicalItems as ReadonlyArray<AreaSettings | BarSettings | RadialBarSettings>,
```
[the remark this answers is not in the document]
PavelVanecek: Yeah same comment as before. No `as` casts please.

PavelVanecek (changes requested):

src/state/types/StackedGraphicalItem.ts
```diff
@@ -24,8 +27,20 @@ export interface DefinitelyStackedGraphicalItem extends MaybeStackedGraphicalIte
...
 
+export type StackedGraphicalItem = (AreaSettings | BarSettings | RadialBarSettings) & DefinitelyStackedGraphicalItem;
+
 export function isStacked<T extends GraphicalItemSettings>(
   graphicalItem: T,
 ): graphicalItem is T & DefinitelyStackedGraphicalItem {
   return 'stackId' in graphicalItem && graphicalItem.stackId != null && graphicalItem.dataKey != null;
 }
+
+export function isStackedGraphicalItem<T extends GraphicalItemSettings>(
```
PavelVanecek, line 38: Why not use `isStacked` alone?

pratik-desgn: `isStacked` alone narrows only to `T & DefinitelyStackedGraphicalItem`. At these call sites, `T` can be the broader cartesian/polar/all-items union, so TypeScript still retains intersections such as `LineSettings & DefinitelyStackedGraphicalItem` instead of narrowing to the three settings types accepted by `StackGroup.graphicalItems`. The extra `type` check supplies that discriminated-union narrowing without bringing back an `as` cast. Runtime-wise, I agree that only Area/Bar/RadialBar should carry `stackId`; this predicate makes that invariant compiler-visible.

pratik-desgn:
Pavel, the earlier cast-related threads are now outdated and resolved, and the branch is green. The remaining thread explains why the extra discriminant check is needed for TypeScript narrowing. Could you take another look when you have a chance?

PavelVanecek (approved):

Issue #7609, opened 2026-08-03
Infinite loop in reduceCSSCalc when a value contains "$&"


jcross:
## What is the current behavior?

`reduceCSSCalc` does not terminate when the expression contains a `String.replace` substitution pattern such as `$&`.

`calculateParentheses` (`src/util/ReduceCSSCalc.ts`) passes the result of `calculateArithmetic` to `String.prototype.replace` as the replacement string:

```js
newExpr = newExpr.replace(PARENTHESES_REGEX, calculateArithmetic(parentheticalExpression));
```

If the text in the parentheses has no arithmetic operator, `calculateArithmetic` returns that text unchanged. In a replacement string, `$&` means "the matched substring". For the input `($&)` the replacement expands to `($&)`, which is the same value. `newExpr` does not change, the regular expression still matches, and the loop never advances.

The same applies to the other substitution patterns: `` $` ``, `$'`, `$1`, and `$$`.

This is reachable from the public `<Text>` component. `Text.tsx` builds the expression from the `capHeight` prop:

```js
startDy = reduceCSSCalc(`calc(${capHeight})`);
```

So this element makes the render hang:

```jsx
<Text capHeight="($&)">hello</Text>
```

The `try/catch` in `safeEvaluateExpression` does not help, because a loop that never ends throws nothing.

## What is the expected behavior?

`reduceCSSCalc` returns a value and the loop ends. The text `$&` has no arithmetic operator, so the correct result matches the current result for other non-arithmetic text:

```js
reduceCSSCalc('calc(abc)')    // 'abc'  (today)
reduceCSSCalc('calc(($&))')   // '$&'   (expected)
```

## Please provide a demo of the problem in a sandbox

A browser sandbox freezes the tab, so a Node command shows the problem better:

```
npm pack recharts@3.10.1 && tar -xzf recharts-3.10.1.tgz && cd package
npm install es-toolkit --ignore-scripts
timeout 5 node -e 'require("./lib/util/ReduceCSSCalc.js").reduceCSSCalc("calc(($&))")'
echo "exit=$?"     # 124 = the command timed out
```

A control call in the same build returns at once, which shows that the module loads correctly:

```
node -e 'console.log(require("./lib/util/ReduceCSSCalc.js").reduceCSSCalc("calc(100px + 20px)"))'   # 120px
```

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

[...]

PR #7620, opened 2026-08-06
fix: guard reduceCSSCalc against ReDoS from pathological calc() input

bunlongheng:
### Summary

`reduceCSSCalc` (used to resolve `calc()` expressions for text layout) is vulnerable to a **regular-expression denial of service (ReDoS)**. A long run of digits immediately followed by an arithmetic operator with no valid right-hand operand - e.g. `calc(999…999+)` or `calc(999…999*)` - makes the operand-matching regexes backtrack **quadratically**, blocking the rendering thread.

Measured on the current `main` (`reduceCSSCalc` called directly):

| input | time |
| --- | --- |
| `calc('9'×10000 + '+')` | ~470 ms |
| `calc('9'×20000 + '+')` | ~1.8 s |
| `calc('9'×40000 + '+')` | ~7.4 s |

Doubling the input roughly quadruples the time - classic O(n²) backtracking. `*`, and an invalid right operand such as `+@`, trigger the same blowup.

### Reachability

These `calc()` strings are built from the public `<Text>` component's `capHeight` / `lineHeight` props:

```
startDy = reduceCSSCalc(`calc(${capHeight})`);
startDy = reduceCSSCalc(`calc(${(wordsByLines.length - 1) / 2} * -${lineHeight} + (${capHeight} / 2))`);
```

so a pathological prop value reaches the evaluator unchecked and can freeze the UI. This is the same evaluator that was recently hardened for the `$&` infinite loop (#7610 / #7609).

### Fix

Add a length guard in `evaluateExpression`. Legitimate CSS length expressions for text layout are only a few dozen characters (the largest real expression here is ~23), so anything past a generous 1000-char bound is treated as invalid and returns an empty result. This bounds the worst case to sub-millisecond while leaving all valid expressions untouched.

### Tests

Adds regression tests asserting the pathological inputs resolve to `''` in well under a second. Full `ReduceCssCalcPrototype.spec.ts` suite and `Text.spec.tsx` pass; `tsc` and eslint are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added protection against excessively long CSS calculation expressions.
  * Prevented pathological or invalid expressions from causing slow processing.
  * Invalid oversized expressions now fail safely instead of being evaluated.

* **Tests**
  * Added regression coverage for very large and malformed expressions.

### DIFF
--- a/src/util/ReduceCSSCalc.ts
+++ b/src/util/ReduceCSSCalc.ts
@@ -158,7 +158,20 @@ function calculateParentheses(expr: string): string {
   return newExpr;
 }
 
+/*
+ * Guard against pathological input. A long run of digits followed by an
+ * arithmetic operator with no valid right-hand operand makes the operand
+ * regexes above backtrack quadratically (a ReDoS), which can freeze the
+ * rendering thread. Legitimate CSS length expressions used for text layout
+ * are only a few dozen characters, so anything longer is treated as invalid.
+ */
+const MAX_EXPRESSION_LENGTH = 1000;
+
 function evaluateExpression(expression: string): string {
+  if (expression.length > MAX_EXPRESSION_LENGTH) {
+    return STR_NAN;
+  }
+
   let newExpr = expression.replace(/\s+/g, '');
   newExpr = calculateParentheses(newExpr);
   newExpr = calculateArithmetic(newExpr);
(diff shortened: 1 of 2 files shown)


PavelVanecek:
This sounds to me like something the browser engine should be fixing instead. But fine.

PR #7644, opened 2026-08-14
fix(Bar): stop spurious re-animation on unrelated re-renders

### LINKED ISSUE #7643: Bar re-animates on unrelated re-renders, making labels flicker while hovering

When a `Bar` has animations enabled and the component *around* the chart re-renders, the Bar starts a new animation even though the `data` reference never changed. Labels are hidden while a Bar animates, so the numbers above the bars blank out and only reappear once `animationDuration` elapses.

Because this happens on every re-render, anything that re-renders while the pointer moves — a hover highlight, a cursor readout, a custom tooltip that lifts state — makes the labels flicker continuously for as long as the mouse is moving.

Two details that made this confusing to track down:

- **The bars themselves do not visibly move.** The restarted animation interpolates from the rectangles already on screen towards rectangles holding identical values, so the interpolation is a visual no-op. Only `isAnimating` flips, and only the labels react to it. The animation is very much running, it just has nothing to show.
- **`onAnimationStart` / `onAnimationEnd` fire again on each re-render**, and every rectangle re-renders at frame rate for the whole duration. So the cost is real even in charts with no labels at all.

It does not reproduce on every chart, which is what makes it look inconsistent. `Bar` is wrapped in `React.memo` with a custom comparator, and that comparator absorbs the parent re-render for simple charts. It stops absorbing when:

- `<Bar>` has children (`<LabelList>`, `<Cell>`, `<ErrorBar>`) — `children` is not on the `propsAreEqual` allowlist and JSX allocates a new element every render; or
- any prop compared by identity changes — an inline `onClick={() => {}}`, for example.

For contrast, the existing `ChangingDataKeyAndStacked` story renders its labels through the `label` prop and passes no children to `<Bar>`. There the memo holds, `BarImpl` never re-renders, and the numbers stay put.

### Root cause

`BarImpl` passes the `Cell` children into a memoized selector:

```tsx
// src/cartesian/Bar.tsx
const cells = findAllByType(props.children, Cell);

const rects = useAppSelector(state => selectBarRectangles(state, props.id, isPanorama, cells));
```

`findAllByType` allocates a new array on every call, and `selectBarRectangles` takes `cells` as an argument, so every render of `BarImpl` is a selector cache miss that returns a brand new `rects` array. `AnimatedItems` passes `rects` to `useAnimationId`, which compares by reference, so a new array means a new `animationId`, which re-keys `JavascriptAnimate` and starts a new animation.

`Pie`, `Scatter` and `RadialBar` wrap the same call in `useMemo(..., [props.children])`. That does not actually help, since JSX reallocates `props.children` on every parent render — but I have only verified the symptom on `Bar`, so I don't want to claim those reproduce without checking. `Funnel` has the unmemoized call like `Bar`, and additionally folds `svgPropertiesNoEvents(props)` into the same settings object, which has no stable memo key at all.

A re-render of the surrounding component should not start a new animation when the chart data has not changed. Labels should stay visible, and the animation callbacks should not fire.

Minimal reproduction — note `data` is at module scope, so its reference never changes. The only thing changing is state in the component around the chart:

```tsx
import React, { useState } from 'react';
import { Bar, BarChart, CartesianGrid, LabelList, XAxis, YAxis } from 'recharts';

const data = [
  { name: 'A', pv: 2400 },
  { name: 'B', pv: 4567 },
  { name: 'C', pv: 1398 },
  { name: 'D', pv: 9800 },
];

export default function Repro() {
  // Any state that updates while the pointer is over the chart will do.
  const [pointer, setPointer] = useState(null);

  return (
    <div onMouseMove={e => setPointer({ x: e.clientX, y: e.clientY })}>
      <p>Move the mouse. The numbers above the bars should stay visible.</p>
      <p>pointer: {pointer ? `${pointer.x}, ${pointer.y}` : 'outside'}</p>
      <BarChart width={500} height={300} data={data}>
        <XAxis dataKey="name" />
        <YAxis />
        <Bar dataKey="pv" fill="#8884d8" animationDuration={1500}>
          <LabelList dataKey="pv" position="top" />
        </Bar>
      </BarChart>
    </div>
  );
}
```

Moving the pointer anywhere over the div makes the numbers disappear and stay gone. The longer `animationDuration` just makes it easier to see; it happens at the default 400ms too.

Removing `<LabelList>` and adding `onClick={() => {}}` to the `<Bar>` shows the same spurious animation via `onAnimationStart`, with no children involved.

- Recharts `3.10.1`, reproduced against `main` (`55dfcc674a1`)
- React 18.3.1, macOS

Reproduced both in the repo's own Storybook and as a jsdom test: after the entrance animation completes the chart shows 4 labels, and a single `mousemove` over the wrapping div drops that to 0 while `onAnimationStart` fires a second time.

I have not bisected which version introduced it, so I can't say whether it worked previously.

---

I have a fix ready and am happy to open a PR: keep the `cells` reference stable across renders while the cells themselves are unchanged, so the selector keeps its cache. Happy to also cover `Funnel` / `Pie` / `Scatter` / `RadialBar` in the same PR or a follow-up, whichever the maintainers prefer.

StefanCodes17:
## Description

`BarImpl` passed the result of `findAllByType(props.children, Cell)` straight into `selectBarRectangles`:

```tsx
const cells = findAllByType(props.children, Cell);

const rects = useAppSelector(state => selectBarRectangles(state, props.id, isPanorama, cells));
```

`findAllByType` allocates a new array on every call, and `cells` is an argument to the memoized selector, so every render of `BarImpl` was a cache miss that returned a brand new `rects` array. `AnimatedItems` feeds `rects` to `useAnimationId`, which compares by reference, so a new array means a new `animationId`, which re-keys `JavascriptAnimate` and starts a new animation — even though the chart data never changed.

This adds `useCells`, which reads the `Cell` children and returns the previous reference while their types and props are unchanged, so the selector keeps its cache. When there are no `Cell` children at all it returns one shared empty array, making the common case a reference check rather than a loop. Props are compared with the existing `propsAreEqual`, so an inline `style={{ ... }}` on a `Cell` does not defeat it.

## Related Issue

Closes #7643

## Motivation and Context

The user-visible symptom is that labels flicker while hovering. Labels are hidden while a Bar animates, so a spurious animation blanks the numbers for the whole `animationDuration`, and anything that re-renders on pointer movement makes that repeat continuously.

Two things made this hard to recognise as an animation bug:

- **The bars do not visibly move.** The restarted animation interpolates from the rectangles already on screen towards rectangles holding identical values, so the interpolation is a visual no-op. Only `isAnimating` flips, and only the labels react to it.
- **It does not reproduce on every chart.** `Bar` is wrapped in `React.memo(BarFn, propsAreEqual)`, which absorbs the parent re-render for simple charts. It stops absorbing when `<Bar>` has children — `children` is not on the `propsAreEqual` allowlist and JSX allocates a new element every render — or when any prop compared by identity changes, such as an inline `onClick`.

The cost is not limited to labels. `onAnimationStart` / `onAnimationEnd` fire on every re-render, and every rectangle re-renders at frame rate for as long as the animation runs.

Measured on a chart re-render with `isAnimationActive={false}`, so these numbers exclude the animation loop that the fix also prevents:

| cells | before | after |
| --- | --- | --- |
| 100 | 10.73 ms | 7.84 ms |
| 1 000 | 99.05 ms | 68.47 ms |
| 10 000 | 943 ms | 623 ms |

It comes out ahead because one cheap comparison pass replaces a full `computeBarRectangles` recompute. In isolation `cellsAreEqual` costs about the same as the `findAllByType` call that already ran on every render (1.84 ms vs 1.65 ms at 10 000 cells).

### Scope

`Funnel` has the same unmemoized `findAllByType` call, and additionally folds `svgPropertiesNoEvents(props)` into the same settings object, which has no stable memo key at all — so `useCells` alone would not fix it.

`Pie`, `Scatter` and `RadialBar` wrap the call in `useMemo(..., [props.children])`, which does not actually help since JSX reallocates `props.children` on every parent render. `useCells` should drop into all three unchanged, but I have only verified the symptom on `Bar`, so I have kept this PR to the verified case rather than claiming fixes I have not demonstrated. Happy to extend it here or in a follow-up, whichever you prefer.

## How Has This Been Tested?

`test/util/useCells.spec.tsx` — 9 unit tests covering the hook directly: stable reference when cells are unchanged, when a prop is an equal-but-new object, and when there are no cells; new reference when a prop changes or a cell is added; and the shared empty array when the last cell is removed.

`test/cartesian/Bar/Bar.rerenderStability.spec.tsx` — 4 rendering tests, with and without `Cell` children, asserting that an unrelated re-render and a hover both leave the labels in place and do not fire the animation callbacks again, plus one test that changed `Cell` props still reach the DOM so the stability cache cannot go stale.

The three behavioural tests fail on `main`: after the entrance animation the chart shows its labels, and a single re-render drops the label count to 0 while `onAnimationStart` fires a second time.

Also run: `npm run lint` (0 errors), `npm run check-types`, `npm test`. The build-output snapshots are updated for the new source file. Two `unit:website` files (`color-mode.client.spec.tsx`, `navigation.spec.tsx`, 10 tests) fail identically on a clean checkout of `main` and are unrelated to this change.

Not included — the symptom is a label that disappears and reappears, which a still frame does not convey. The repro in #7643 shows it in a few lines.

## Types of changes

## Checklist:

On the last item: I had a story that demonstrated the flicker and removed it again, since the behaviour it showed requires moving the pointer over a live chart and a VR snapshot cannot capture it. The unit and rendering tests pin the behaviour instead. Happy to add a story back if you would like one for manual verification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chart rendering stability when surrounding components update or tooltips are hovered.
  * Prevented unnecessary bar animations and visual changes when cell styling remains unchanged.
  * Ensured updated cell colors are applied correctly.
* **Tests**
  * Added coverage for stable rerendering, labels, animations, tooltip interactions, and cell updates.

### DIFF
--- a/scripts/snapshots/es6Files.txt
+++ b/scripts/snapshots/es6Files.txt
@@ -287,6 +287,7 @@
   "es6/util/typedDataKey.js",
   "es6/util/types.js",
   "es6/util/useAnimationId.js",
+  "es6/util/useCells.js",
   "es6/util/useElementOffset.js",
   "es6/util/useId.js",
   "es6/util/usePrefersReducedMotion.js",
--- a/scripts/snapshots/libFiles.txt
+++ b/scripts/snapshots/libFiles.txt
@@ -287,6 +287,7 @@
   "lib/util/typedDataKey.js",
   "lib/util/types.js",
   "lib/util/useAnimationId.js",
+  "lib/util/useCells.js",
   "lib/util/useElementOffset.js",
   "lib/util/useId.js",
   "lib/util/usePrefersReducedMotion.js",
--- a/scripts/snapshots/typesFiles.txt
+++ b/scripts/snapshots/typesFiles.txt
@@ -287,6 +287,7 @@
   "types/util/typedDataKey.d.ts",
   "types/util/types.d.ts",
   "types/util/useAnimationId.d.ts",
+  "types/util/useCells.d.ts",
   "types/util/useElementOffset.d.ts",
   "types/util/useId.d.ts",
   "types/util/usePrefersReducedMotion.d.ts",
--- a/src/cartesian/Bar.tsx
+++ b/src/cartesian/Bar.tsx
@@ -15,7 +15,6 @@ import { Series } from 'victory-vendor/d3-shape';
 import { Props as RectangleProps, RectRadius } from '../shape/Rectangle';
 import { Layer } from '../container/Layer';
 import { ErrorBarDataItem, ErrorBarDataPointFormatter } from './ErrorBar';
-import { Cell } from '../component/Cell';
 import {
   CartesianLabelListContextProvider,
   CartesianLabelListEntry,
@@ -23,7 +22,6 @@ import {
   LabelListFromLabelProp,
 } from '../component/LabelList';
 import { interpolate, isNan, mathSign, noop } from '../util/DataUtils';
-import { findAllByType } from '../util/ReactUtils';
 import {
   BarPositionPosition,
   getBaseValueOfBar,
@@ -87,6 +85,7 @@ import { ZIndexable, ZIndexLayer } from '../zIndex/ZIndexLayer';
 import { DefaultZIndexes } from '../zIndex/DefaultZIndexes';
 import { getZIndexFromUnknown } from '../zIndex/getZIndexFromUnknown';
 import { propsAreEqual } from '../util/propsAreEqual';
+import { useCells } from '../util/useCells';
 import { AxisId } from '../state/cartesianAxisSlice';
 import { BarStackClipLayer, useStackId } from './BarStack';
 import { useRechartsTheme } from '../theme/RechartsThemeContext';
@@ -1043,7 +1042,12 @@ function BarImpl(props: BarImplProps) {
 
   const isPanorama = useIsPanorama();
 
-  const cells = findAllByType(props.children, Cell);
+  /*
+   * `cells` is an argument to the memoized selector below, so it has to be referentially stable,
+   * otherwise every render of this component is a cache miss that returns a brand new `rects` array,
+   * which changes the animationId, which starts a spurious animation.
+   */
+  const cells = useCells(props.children);
 
   const rects: ReadonlyArray<BarRectangleItem> | undefined = useAppSelector(state =>
     selectBarRectangles(state, props.id, isPanorama, cells),
--- a/src/util/useCells.ts
+++ b/src/util/useCells.ts
@@ -1,0 +1,54 @@
+import { ReactElement, ReactNode, useRef } from 'react';
+import { Cell } from '../component/Cell';
+import { findAllByType } from './ReactUtils';
+import { propsAreEqual } from './propsAreEqual';
+
+/**
+ * A single shared reference for "this graphical item has no Cell children",
+ * which is by far the most common case.
+ */
+const noCells: ReadonlyArray<ReactElement> = [];
+
+function cellsAreEqual(a: ReadonlyArray<ReactElement>, b: ReadonlyArray<ReactElement>): boolean {
+  if (a === b) {
+    return true;
+  }
+  if (a.length !== b.length) {
+    return false;
+  }
+  return a.every((cell, index) => {
+    const other = b[index];
+    return other != null && cell.type === other.type && propsAreEqual(cell.props, other.props);
+  });
+}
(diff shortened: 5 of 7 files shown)


PavelVanecek:
Why do you use Cell if I may ask?

Issue #7519, opened 2026-07-01
Nested Treemap back navigation not working


apcqa:
## What is the current behavior?
When using a nested Treemapthe navigate back by clicking on the index box does not work

## What is the expected behavior?
After select on a node and navigating to the children nodes click on the parent node name at the bottom should return.

## Please provide a demo of the problem in a sandbox
https://recharts.github.io/en-US/examples/NestedTreemap/

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of recharts@3.9.1
Windows, Firefox, Chrome

pupuking723:
I can take a look at this. I’ll first reproduce it against the NestedTreemap example and trace the parent/index-box click path, then keep any fix scoped to the back-navigation state rather than changing the Treemap layout calculation.

PR #7657, opened 2026-08-19
fix: preserve treemap navigation history

### LINKED ISSUE #7529: Allow tick formatter to access other rendered ticks
My goal is to render timestamps as time strings, except when there's a crossover of dates (e.g. 23:59 on Jun 6th displayed as time, and 00:01 on Jun 7th displayed as date and time). However, while the tick formatter has access to the index of the tick, it doesn't have access to the index of the tick within the set of ticks that are going to be rendered (since ticks can be dropped).

When formatting ticks, I'd like to know what (if any) ticks are before the current ticks--perhaps with an index into the  list of rendered ticks and a way to obtain the list of rendered ticks. I realize this may be tricky since, as I understand it, the tick labels need to be generated and measured in order to determine which ticks will be rendered, so there might be a chicken-or-the-egg situation going on here.

I tried using the `tickFormatter` prop of the `XAxis` component and indexed into the set of ticks. But I didn't have access to which ticks are going to be rendered.

I also tried using a custom `XAxis` component wrapper that uses `useXAxisTicks()` along with the `tickFormatter` prop. However, that led to an infinite recursion.

Dhritimann:
## Description

Fix Treemap nested navigation history by avoiding mutation of the existing `nestIndex` state array.

## Related Issue

Fixes #7529

## How Has This Been Tested?

Verified the change with `git diff` and confirmed that the existing `nestIndex` array is no longer mutated directly.

## Types of changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved treemap navigation when opening nested nodes, ensuring selections update reliably without affecting existing navigation state.

### DIFF
--- a/src/chart/Treemap.tsx
+++ b/src/chart/Treemap.tsx
@@ -973,12 +973,12 @@ class TreemapWithState extends PureComponent<InternalTreemapProps, State> {
       const formatRoot = squarify(root, aspectRatio, nodeInset, nodeGap);
 
       const { nestIndex } = this.state;
-      nestIndex.push(node);
+      const nextNestIndex = [...nestIndex, node];
 
       this.setState({
         formatRoot,
         currentRoot: root,
-        nestIndex,
+        nestIndex: nextNestIndex,
       });
     }
     if (onClick) {


PavelVanecek:
Is the related issue correct? It doesn't look related on first glance

PavelVanecek:
I will need a unit test for this change

PavelVanecek:
Perhaps related to https://github.com/recharts/recharts/issues/7519 instead

PR #6176, opened 2025-08-03
feat: implement LRU cache for string size measurements (#3955)

shreedharbhat98:
# String Size Caching Improvements

## Overview

This document outlines the implementation of LRU (Least Recently Used) cache for string size measurements in Recharts, addressing performance issues described in [issue #3955](https://github.com/recharts/recharts/issues/3955).

## Problem Statement

The original string size caching mechanism in Recharts had several limitations:

1. **Inefficient Cache Eviction**: Used a "clear all" strategy when cache exceeded 2000 items, causing poor cache hit ratios
2. **No Configurability**: Users couldn't adjust cache behavior for their specific use cases
3. **Performance Bottlenecks**: Excessive `getStringSize` calls were causing performance issues with large datasets
4. **Memory Management**: No intelligent eviction strategy led to suboptimal memory usage

### LRU Cache Implementation

Implemented a standalone `LRUCache` class with the following features:

- **O(1) Operations**: Get and set operations using JavaScript Map's insertion order
- **Configurable Size**: Default 2000 items, user configurable
- **Intelligent Eviction**: Removes least recently used items when capacity is reached
- **Memory Efficient**: Prevents unbounded memory growth

```typescript
// Example usage
const cache = new LRUCache<string, number>(1000);
cache.set('key', 42);
const value = cache.get('key'); // Returns 42, moves to most recent
```

### Enhanced DOMUtils

Improved the `getStringSize` function with:

- **Better Cache Keys**: Focus on text-relevant CSS properties only
- **Configurable Behavior**: Enable/disable caching, adjust cache size
- **Performance Optimization**: Reduced cache misses through better key generation

```typescript
// Configure caching behavior
configureTextMeasurement({
  cacheSize: 5000,     // Increase for large datasets
  enableCache: true    // Enable/disable caching
});

// Get cache statistics
const stats = getStringCacheStats();
console.log(`Cache size: ${stats.size}/${stats.maxSize}`);
```

#### `configureTextMeasurement(config)`

Configure text measurement behavior.

**Parameters:**
- `config.cacheSize` (number): Maximum cache size (default: 2000)
- `config.enableCache` (boolean): Enable/disable caching (default: true)

**Example:**
```typescript
configureTextMeasurement({
  cacheSize: 1000,
  enableCache: false
});
```

#### `getTextMeasurementConfig()`

Get current configuration.

**Returns:** Current `TextMeasurementConfig`

#### `clearStringCache()`

Clear the entire cache. Useful for testing or memory management.

[...]

PR #7659, opened 2026-08-20
fix(Text): include every style property in the text measurement cache key

MahinAnowar:
## Description

`getStringSize` caches measurements under a key built from six style properties, but `measureTextWithDOM` applies the *whole* style object to the measurement span. Any property outside those six is applied when measuring and then ignored when caching, so two different styles share one cache entry and the second caller gets back the first one's size.

```jsx
// same text, styles differ only in a property that isn't part of the key
getStringSize('Hello', { fontVariant: 'normal' })     // measured properly
getStringSize('Hello', { fontVariant: 'small-caps' }) // returns the width measured above
```

`wordSpacing`, `fontStretch`, `fontVariant` and the `font` shorthand all change how wide the text renders and none of them are in the key. The `font` shorthand is the worst of them, since it sets everything while leaving all six of the listed properties empty — every distinct `font` value collides.

This reaches users through `<Text>`, which passes the caller's `style` straight into `getStringSize` when it measures words for wrapping (`src/component/Text.tsx`). Two `<Text>` elements with the same content but, say, different `fontStretch` will wrap identically because the second one is laid out with the first one's measurements.

I found this by probing rather than from a report, so there's no issue to link. It's a regression from #6176, which swapped the old `JSON.stringify({ text, copyStyle })` key (whole style object) for hand-listed properties when it introduced the LRU cache.

## Motivation and Context

Wrong text measurements mean wrong word wrapping and wrong axis/label layout, and it's the kind of bug that only shows up once a second style is used on the same string, which makes it awkward to track down.

I kept the concatenation approach the current code chose for speed rather than going back to `JSON.stringify` — it just walks the style's own keys now instead of a fixed list, so nothing can silently fall out of the key again if someone applies a new property to the span later. Keys are sorted so `{ fontSize, fontStyle }` and `{ fontStyle, fontSize }` still land on one entry.

Values that are `null`/`undefined`/empty are skipped, which matches both the old `removeInvalidKeys` behaviour and the `|| ''` the current code uses.

## How Has This Been Tested?

Three tests added to `test/util/DomUtils.spec.tsx`, modelled on the existing "cache should handle different styles separately":

- two styles differing only in `wordSpacing` produce two cache entries
- the second of two `fontVariant` measurements returns its own size, not the cached one (this is the user-visible symptom, using `mockSequenceOfGetBoundingClientRect`)
- property order doesn't split one style across two entries

The first two fail on current `main` (the second returns `width: 25` where `40` is expected) and pass with the change. The third passes either way, it's there to pin the sorting.

`test/util` (47 files, 1057 tests), `test/component/Text.spec.tsx`, `test/cartesian/getTicks.spec.ts` and `test/chart/Treemap.spec.tsx` are green. `eslint`, `tsc --noEmit` and `tsc --project test/tsconfig.json` all clean.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved measurement accuracy when content uses different visual styles.
  * Prevented cached measurements from being incorrectly reused when styles change.
  * Ensured empty style values do not affect measurement results.
  * Cleared previously applied formatting between measurements to prevent stale styles from affecting layout.
  * Improved handling of style properties during measurement caching.

### DIFF
--- a/src/util/DOMUtils.ts
+++ b/src/util/DOMUtils.ts
@@ -33,15 +33,15 @@ function createCacheKey(text: string | number, style: CSSProperties): string {
   // Simple string concatenation for better performance than JSON.stringify.
   // Every property has to take part in the key because measureTextWithDOM applies the whole style
   // object to the measurement span - listing only a few of them makes two different styles share
-  // one cache entry. Keys are sorted so that the same style always produces the same key.
-  const names = Object.keys(style).sort();
+  // one cache entry.
   let key = `${text}`;
 
-  for (let i = 0; i < names.length; i++) {
-    const name = names[i];
-    const value = style[name as keyof CSSProperties];
-    if (value != null && value !== '') {
-      key += `|${name}:${value}`;
+  for (const name in style) {
+    if (Object.prototype.hasOwnProperty.call(style, name)) {
+      const value = style[name as keyof CSSProperties];
+      if (value != null && value !== '') {
+        key += `|${name}:${value}`;
+      }
     }
   }
 
--- a/src/util/DOMUtils.ts
+++ b/src/util/DOMUtils.ts
@@ -30,15 +30,22 @@ const SPAN_STYLE = {
 const MEASUREMENT_SPAN_ID = 'recharts_measurement_span';
 
 function createCacheKey(text: string | number, style: CSSProperties): string {
-  // Simple string concatenation for better performance than JSON.stringify
-  const fontSize = style.fontSize || '';
-  const fontFamily = style.fontFamily || '';
-  const fontWeight = style.fontWeight || '';
-  const fontStyle = style.fontStyle || '';
-  const letterSpacing = style.letterSpacing || '';
-  const textTransform = style.textTransform || '';
+  // Simple string concatenation for better performance than JSON.stringify.
+  // Every property has to take part in the key because measureTextWithDOM applies the whole style
+  // object to the measurement span - listing only a few of them makes two different styles share
+  // one cache entry. Keys are sorted so that the same style always produces the same key.
+  const names = Object.keys(style).sort();
+  let key = `${text}`;
 
-  return `${text}|${fontSize}|${fontFamily}|${fontWeight}|${fontStyle}|${letterSpacing}|${textTransform}`;
+  for (let i = 0; i < names.length; i++) {
+    const name = names[i];
+    const value = style[name as keyof CSSProperties];
+    if (value != null && value !== '') {
+      key += `|${name}:${value}`;
+    }
+  }
+
+  return key;
 }
 
 /**
--- a/src/util/DOMUtils.ts
+++ b/src/util/DOMUtils.ts
@@ -64,7 +64,10 @@ const measureTextWithDOM = (text: string | number, style: CSSProperties): Size =
       document.body.appendChild(measurementSpan);
     }
 
-    // Apply styles directly without unnecessary object creation
+    // The span is reused between calls, and Object.assign only adds properties - it never removes
+    // the ones a previous style set. Without the reset, a measurement inherits leftovers from
+    // whatever was measured before it.
+    measurementSpan.style.cssText = '';
     Object.assign(measurementSpan.style, SPAN_STYLE, style);
     measurementSpan.textContent = `${text}`;
 
--- a/test/util/DomUtils.spec.tsx
+++ b/test/util/DomUtils.spec.tsx
@@ -109,6 +109,21 @@ describe('DOMUtils', () => {
     expect(getStringSize('test', { fontVariant: 'small-caps' })).toEqual({ width: 40, height: 17 });
   });
 
+  test('measurement span should not keep styles from an earlier measurement', () => {
+    render(<span id="recharts_measurement_span">test</span>);
+    mockGetBoundingClientRect({
+      width: 25,
+      height: 17,
+    });
+    const measurementSpan = document.getElementById('recharts_measurement_span');
+
+    getStringSize('test', { fontStretch: 'expanded', wordSpacing: '9px' });
+    getStringSize('test', {});
+
+    expect(measurementSpan?.style.wordSpacing).toBe('');
+    expect(measurementSpan?.style.fontStretch).toBe('');
+  });
+
   test('configureTextMeasurement should update configuration', () => {
     const newConfig = {
       cacheSize: 1000,
--- a/test/util/DomUtils.spec.tsx
+++ b/test/util/DomUtils.spec.tsx
@@ -109,6 +109,19 @@ describe('DOMUtils', () => {
     expect(getStringSize('test', { fontVariant: 'small-caps' })).toEqual({ width: 40, height: 17 });
   });
 
+  test('cache should treat empty style values as absent', () => {
+    render(<span id="recharts_measurement_span">test</span>);
+    mockGetBoundingClientRect({
+      width: 25,
+      height: 17,
+    });
+
+    getStringSize('test', { fontSize: '14px' });
+    getStringSize('test', { fontSize: '14px', fontStretch: undefined, wordSpacing: '' });
+
+    expect(getStringCacheStats().size).toBe(1);
+  });
+
   test('measurement span should not keep styles from an earlier measurement', () => {
     render(<span id="recharts_measurement_span">test</span>);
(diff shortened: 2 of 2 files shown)


PavelVanecek:
https://github.com/recharts/recharts/pull/6176 was a performance fix. How is the new code performing compared to what is now in `main`?

MahinAnowar:
Fair question, and you were right to ask - my first version was noticeably slower, so I've pushed a faster one.

What I had was `Object.keys(style).sort()`, and the sort turned out to be the expensive part. It only kept `{fontSize, fill}` and `{fill, fontSize}` on a single cache entry, which costs a miss rather than a wrong answer, so I dropped it and walk the style with `for..in` instead (same idiom as `svgPropertiesNoEvents`). ns per `createCacheKey` call, 300k iterations, best of two runs:

| style | pre-#6176 | main | my first version | now |
|---|---|---|---|---|
| `{}` (Treemap) | 73.7 | 22.1 | 8.5 | 5.5 |
| `{fontSize, letterSpacing}` (axis ticks) | 129.5 | 32.0 | 81.0 | 33.6 |
| `{fontSize, fontFamily, fill}` | 164.2 | 32.8 | 113.2 | 50.7 |
| 8 properties | 307.7 | 39.1 | 303.6 | 120.6 |

The axis tick path is what `getTicks` actually calls and it's back within noise of main. Empty style is faster. It's still slower once a style carries properties main ignored, because those properties now have to go into the key - that's the fix itself, not overhead I can optimise away. Worst case there is about 80ns, against a `getBoundingClientRect` on the miss that forces layout and costs orders of magnitude more, and a Map lookup on the hit. Every column is well under what #6176 replaced.

One tradeoff worth naming: something like `fill` doesn't change text metrics but does now split the cache, so two labels differing only in colour get measured twice instead of once. I could skip properties known not to affect metrics, but that's a list that rots the same way the six-property list did, except quietly - if we ever miss one we're back to wrong sizes. Keying on everything applied to the span fails the safe way. Happy to go the other way if you'd rather.

MahinAnowar:
Went through the three bot findings. One was a real bug and is fixed, one is now moot, one I'd rather leave to you.

**Stale span - real, and it turned out to matter for this PR.** The span is reused and `Object.assign` only adds properties, so a style that set `wordSpacing` left it applied to whatever got measured next. I checked before believing it:

```
after getStringSize('test', {wordSpacing: '9px', fontStretch: 'expanded'})
  span.style.wordSpacing = "9px"  fontStretch = "expanded"
after getStringSize('test', {})
  span.style.wordSpacing = "9px"  fontStretch = "expanded"   <- leftovers
```

What makes it relevant here is that my change *exposes* it. On main those two calls share a cache entry, so the second never measures and the leak stays hidden. Once the keys are separate the second call really does measure, against a dirty span - so the key fix alone doesn't actually fix the reported behaviour. `measurementSpan.style.cssText = ''` before applying, plus a test asserting the span is clean afterwards. Good catch by the bot.

**Shorthand/longhand ordering - moot now.** That one was about the key sorting property names while the style is applied in its own order, so `{font: '16px serif', fontSize: '20px'}` and the reverse could share a key. I dropped the sort in the perf pass above, so the key follows insertion order and those two produce different keys.

**Delimiter collisions - real but I've left it alone, your call.** `getStringSize('a|fontSize:12px', {})` and `getStringSize('a', {fontSize: '12px'})` do land on the same key. It's not new: main collides the same way (`'a|12px|||||'` with no style vs `'a'` with `fontSize`), so it's pre-existing rather than something this PR introduces. Length-prefixing fixes it, and I measured what it costs on top of the numbers above:

| style | current | + text length prefix | + full length prefix |
|---|---|---|---|
| `{}` | 4.4 | 11.6 | 11.0 |
| `{fontSize, letterSpacing}` | 30.2 | 37.7 | 60.0 |
| `{fontSize, fontFamily, fill}` | 49.0 | 63.1 | 89.5 |
| 8 properties | 112.7 | 134.8 | 275.0 |

Prefixing just the text is the cheap 90% fix - it kills the realistic case (arbitrary label text) and leaves only a contrived one needing a property *value* that itself looks like `|name:value`. But it gives up the parity with main I just got back on the axis tick path, and you're the one weighing that, so say the word and I'll push it.

MahinAnowar:
The VR failure is this change doing its job, and I think the snapshot genuinely needs updating rather than the code needing a fix. `one-chart-four-themes` is off by 1587 pixels, ratio 0.01, and it's the only one.

That test renders five themed charts and then repeats all five inside a black div. `CustomThemeChart` sets `typography: { fontSize: 30, fontWeight: 'bold', fontFamily: 'Roboto' }`, while `lightTheme` and `darkTheme` only set `color` and the default sets no font at all. The measurement span is shared, and `Object.assign` never removes a property, so before this PR every chart measured after the custom-theme one inherited 30px bold Roboto. On that page the black-div copies of Default, Empty, Light and Dark all render right after `CustomThemeChart`, so their labels were being measured against a font they don't use, and laid out from those widths.

With the span reset they're measured with their own font, so the text widths shift slightly. Small diff, in the direction you'd want.

I can't regenerate the baselines here — `test-vr:update` needs the Docker image and the snapshots are `chromium-linux` / `firefox-linux`, so anything I produce locally would be from the wrong environment. Happy to run it and push the PNGs if you want them in this PR, or leave it to you if you'd rather regenerate them yourself. Just say which.

Worth flagging separately: the old cache key ignored `color`, so light and dark theme labels shared one cache entry. Same size either way, so nothing visible, but it's the same missing-property problem and it's fixed by the same change.

MahinAnowar:
Bumping this since it's been sitting red for a couple of weeks. The only thing failing is the VR snapshot I wrote about above — I'm still fairly confident the baseline needs regenerating rather than the code needing a change, since the shift is exactly the text measuring correctly now.

I can't regenerate it on my end. The baselines are `-linux.png` and the supported path is `npm run test-vr:update` through Docker, which on my machine would build arm64 and almost certainly rewrite every snapshot instead of the one that moved. So it'd make the diff worse, not better.

Happy to push the updated images if there's a way to pull them off a CI run, or to leave it with you if that's less hassle.

PavelVanecek:
`npm run test-vr:update` is the way to update snapshots, the whole pipeline runs in Docker. There is no CI for updating.

PavelVanecek:
The diff though shows an axis tick label disappeared even though it had plenty of space to render. Is that intentional?

MahinAnowar:
I reproduced the diff locally in chromium and went through it — no tick label is actually dropped. I counted the rendered `.recharts-cartesian-axis-tick-value` nodes with and without the fix and it's 27 either way. What changes is the position of the *last* x-axis label, and only that one:

```
                upstream      with fix
Iter: 5 (16px)  x=378.45  ->  x=380.67
Iter: 5 (30px)  x=359.60  ->  x=363.77
```

The reason it moves is the bug itself. I logged every `getStringSize` call while the ten themed charts mount together:

```
'Iter: 5' @16px   upstream 43.094   with fix 38.656
'Iter: 5' @30px   upstream 80.797   with fix 72.469
```

38.656 is the true width — you get exactly that from upstream too if you mount a single chart on its own. The inflated numbers only show up when several charts measure through the same span: `CustomThemeChart` sets `fontWeight: bold` and `fontFamily: Roboto`, `Object.assign` never removes them, and every later measurement of a label whose style doesn't mention those two inherits them and comes out bold-Roboto-wide. `getTicks` then clamps the final label inside the plot using that width, so an over-wide measurement was pushing it left. With the correct width it sits where it belongs, a couple of pixels right.

So the snapshot change is the fix working, and my read is the baseline wants updating rather than the code changing.

One thing I'd flag separately: on the 30px `CustomThemeChart` the last glyph of `Iter: 5` now sits right on the plot edge and gets clipped. That's the clamp in `getTicks` having less slack once the width is honest, not something the measuring introduced, but it's a real rough edge if you'd rather the last label got padded away from the boundary. Happy to look at that in its own PR.

On the snapshots — understood that it's Docker and there's no CI path. The thing I want to avoid is regenerating on this machine and handing you a diff where all three browsers changed for arm64-vs-amd64 font reasons rather than for this fix. I'll run it pinned to `linux/amd64` and, before writing anything, do a plain run first to confirm only `one-chart-four-themes` fails against your existing linux baselines. If anything else comes back red I'll stop and leave the snapshots to you rather than push noise.

PR #7710, opened 2026-08-30
fix: declare @types/react as an optional peer dependency

### LINKED ISSUE #7709: @types/react is not declared as a peer dependency, which breaks types under isolated installs
### Problem

`recharts` types import React, but `@types/react` is not declared as a peer dependency:

```ts
// types/index.d.ts and friends
import * as React from 'react';
```

Under a strict, non-hoisted `node_modules` the recharts type files cannot resolve React's types, so every component degrades and TypeScript reports:

```text
error TS2607: JSX element class does not support attributes because it does not have a 'props' property
error TS2786: 'Bar' cannot be used as a JSX component
error TS2786: 'XAxis' cannot be used as a JSX component
```

In one Electron app this produced 122 errors across 8 files, all from recharts.

### Why it happens

Bun's isolated linker (and pnpm, and Yarn PnP) symlinks each package into a store. Node and TypeScript resolve a symlink to its real path before walking parent directories, so a package can only see what it declares. `react` is declared and resolves; `@types/react` is not, so it does not.

`skipLibCheck` hides the errors inside `recharts`' own `.d.ts` files but not the broken types they produce downstream.

### Suggested fix

Declare `@types/react` as an optional peer dependency:

```json
"peerDependencies": {
  "@types/react": "*"
},
"peerDependenciesMeta": {
  "@types/react": { "optional": true }
}
```

Optional, so JavaScript consumers are not warned about a package they do not need.

This is what Radix UI and posthog-js already do:

- `@radix-ui/react-primitive` — `"@types/react": "*"` with `peerDependenciesMeta.optional`
- `posthog-js@1.422.5` — same, added to fix this exact class of bug

### Reproduce

```json
{ "devDependencies": { "recharts": "3.10.1", "react": "^19", "@types/react": "^19", "typescript": "^5" } }
```

```toml
# bunfig.toml
[install]
linker = "isolated"
```

`bun install`, then typecheck any file rendering `<BarChart><Bar /></BarChart>`.

Happy to open a PR.

tifandotme:
## Description

Declares `@types/react` as an optional peer dependency.

```json
"peerDependencies": {
  "@types/react": "*"
},
"peerDependenciesMeta": {
  "@types/react": { "optional": true }
}
```

Optional, so JavaScript consumers are not warned about a package they do not need.

## Related Issue

Fixes #7709.

## Motivation and Context

Recharts' type files do `import * as React from 'react'`, but `@types/react` is not declared. Under a strict, non-hoisted `node_modules` those types cannot resolve React, every component's types degrade, and consumers see:

```text
error TS2607: JSX element class does not support attributes because it does not have a 'props' property
error TS2786: 'Bar' cannot be used as a JSX component
```

Bun's isolated linker, pnpm and Yarn PnP all resolve a symlink to its real path before walking parent directories, so a package sees only what it declares. `react` is declared and resolves; `@types/react` is not.

Same fix that `@radix-ui/*` and `posthog-js@1.422.5` already ship.

## How Has This Been Tested?

In an Electron + React 19 app on Bun 1.4 with `linker = "isolated"`, recharts accounted for 122 TypeScript errors across 8 files. Applying this exact change as a local patch takes those to 0. Full app suite passes after: typecheck clean, lint clean, 168 unit tests.

`npm install --package-lock-only` regenerated the lockfile; the diff is the peer entry only.

## Types of changes

## Checklist:

Package metadata only, so there is no code path to unit test and no rendering behaviour to story. The regression it prevents is only observable from a real install under a strict linker.

## Summary by CodeRabbit

* **Chores**
  * Updated package compatibility requirements to support projects that use React type definitions.
  * Made the React type definitions dependency optional for improved installation flexibility.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -123,9 +123,15 @@
         "node": ">=18"
       },
       "peerDependencies": {
+        "@types/react": "*",
         "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0",
         "react-dom": "^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0",
         "react-is": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
+      },
+      "peerDependenciesMeta": {
+        "@types/react": {
+          "optional": true
+        }
       }
     },
     "node_modules/@actions/core": {
--- a/package.json
+++ b/package.json
@@ -90,7 +90,13 @@
   "peerDependencies": {
     "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0",
     "react-dom": "^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0",
-    "react-is": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
+    "react-is": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0",
+    "@types/react": "*"
+  },
+  "peerDependenciesMeta": {
+    "@types/react": {
+      "optional": true
+    }
   },
   "dependencies": {
     "@reduxjs/toolkit": "^1.9.0 || 2.x.x",


PavelVanecek:
Can we have a new recharts-integ test to prevent regressions?

tifandotme:
> Can we have a new recharts-integ test to prevent regressions?

@PavelVanecek I did at https://github.com/recharts/recharts-integ/pull/101

does this cover the regression you had in mind?

PR #7147, opened 2026-03-17
feat(RadialBar): add minAngle prop for minimum bar angle

mixelburg:
## Summary

Fixes #2986

The `RadialBar` component was missing the documented `minAngle` prop. The recharts API docs ([RadialBar#minAngle](https://recharts.org/en-US/api/RadialBar#minAngle)) list this prop, but it had never been implemented.

## Changes

- Added `minAngle?: number` prop to `RadialBarProps` (default: `0`)
- Added `minAngle: number` to `RadialBarSettings` interface
- In radial layout (the default for RadialBarChart), when a bar's `deltaAngle` is less than `minAngle`, the `endAngle` is extended to meet the minimum — ensuring small-value bars remain visible
- Wired `minAngle` through `RadialBarImpl`, `SetPolarGraphicalItem`, and `selectRadialBarSectors`
- Added unit tests verifying the behaviour
- Added a Storybook story demonstrating the prop with mixed-magnitude data

## How to test

```tsx
<RadialBarChart width={500} height={500} data={data}>
  <RadialBar dataKey="value" minAngle={15} />
</RadialBarChart>
```

Any bar whose natural angle falls below 15° will be extended to 15° so it stays visible.

## Release Notes

* **New Features**
  * Added `minAngle` prop to RadialBar component, allowing you to enforce a minimum angular width for radial bar segments. Bars smaller than the specified minimum angle will be automatically adjusted to meet the threshold, ensuring better visibility of small values in your radial charts.

PR #7717, opened 2026-09-01
fix(RadialBar): implement minAngle prop for minimum bar angle

### LINKED ISSUE #2986: RadialBar minAngle prop not working

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/simple-radial-bar-chart-forked-vmgkh8?file=/src/App.tsx)

### Steps to reproduce
See minAngle prop on RadialBar component, see the test (black) bar is less than 15 degrees

### What is expected?
Based on this https://recharts.org/en-US/api/RadialBar#minAngle I expect so see the smallest bar be 15 degrees

The bar is significantly less than 15 degrees

| Environment | Info |
|---|---|
| Recharts | v2.0.6 |
| React | 17.0.1 |
| System | CodeSandbox |
| Browser | Chrome |

Chandu-Putta:
## Description

`RadialBar` has a `minAngle` prop listed on the API docs, but it was never wired into the component — only `Pie` had it, so passing `minAngle` to `RadialBar` was silently ignored.

This adds a real `minAngle` prop to `RadialBar`: when a non-zero bar's natural angle falls below the threshold, its `endAngle` is extended to meet the minimum, so small values stay visible. Bars with a value of exactly zero are explicitly excluded, so they never get an artificial sliver.

## Related Issue

Fixes #2986

There was an earlier attempt at this in #7147, which implemented the same idea but never received a maintainer review and was closed from inactivity. It also had a bug where a zero-value bar could receive a fake sliver (`mathSign(deltaAngle || minAngle)` treats `deltaAngle === 0` as non-zero). This PR avoids that by only applying the expansion when `deltaAngle !== 0`.

## Motivation and Context

Bars representing small values in a `RadialBarChart` could render invisibly thin, or not visible at all, with no way for consumers to guarantee a minimum visible angle — despite the docs implying this was already supported.

## How Has This Been Tested?

- Added a new test verifying: a bar below the threshold is extended to `minAngle`, a bar already above the threshold is untouched, and a zero-value bar stays at 0° (no artificial sliver).
- Added a test confirming the `minAngle` prop on `<RadialBar>` is correctly forwarded into Redux state.
- Updated 11 existing test assertions that check the full settings object shape, since it now includes `minAngle`.
- Ran the full test suite (`npm test`), typecheck, and lint locally — all passing.

N/A

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added a `minAngle` option for radial bars to ensure non-zero segments meet the configured minimum angular span.
  * Zero-value segments remain at zero degrees, while segments already exceeding the threshold remain unchanged.
  * Negative minimum-angle values are handled consistently with positive values.
  * Improved the description of the pie chart minimum-angle option.

* **Tests**
  * Added coverage for minimum-angle behavior across radial bar configurations and selector scenarios.

### DIFF
--- a/src/polar/Pie.tsx
+++ b/src/polar/Pie.tsx
@@ -367,7 +367,7 @@ export interface PieProps<DataPointType = any, DataValueType = any>
   /** the max radius of pie */
   maxRadius?: number;
   /**
-   * The minimum angle of each unzero data.
+   * The minimum angle of each non-zero segment.
    * @defaultValue 0
    */
   minAngle?: number;
--- a/src/polar/RadialBar.tsx
+++ b/src/polar/RadialBar.tsx
@@ -798,8 +798,12 @@ export function computeRadialBarDataItems({
         }
 
         // Zero-value bars are excluded so a truly empty bar never gets an artificial sliver.
-        if (minAngle > 0 && deltaAngle !== 0 && Math.abs(deltaAngle) < minAngle) {
-          const delta = mathSign(deltaAngle) * (minAngle - Math.abs(deltaAngle));
+        // Negative minAngle is normalized the same way Pie normalizes pieSettings.minAngle.
+        // Note: like minPointSize above, this expands each bar independently, so a stacked
+        // bar's expansion can overlap into its neighbor's range. That is an existing
+        // limitation shared with minPointSize, not something new to minAngle.
+        if (Math.abs(minAngle) > 0 && deltaAngle !== 0 && Math.abs(deltaAngle) < Math.abs(minAngle)) {
+          const delta = mathSign(deltaAngle) * (Math.abs(minAngle) - Math.abs(deltaAngle));
           endAngle += delta;
         }
         backgroundSector = {
--- a/src/polar/RadialBar.tsx
+++ b/src/polar/RadialBar.tsx
@@ -416,6 +416,13 @@ export interface InternalRadialBarProps<DataPointType = any, DataValueType = any
   legendType?: LegendType;
   maxBarSize?: number;
   /**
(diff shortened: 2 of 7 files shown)


kdelay (commented):
`minAngle={-30}` is ignored here: the guard is `minAngle > 0`, while `Pie` normalizes with `Math.abs(pieSettings.minAngle)` (Pie.tsx:891). Measured on this branch with `[100, 5]`, the small bar stays at 18 degrees for `-30` and extends to 30 for `30`. One `Math.abs` keeps the two props in sync.

Stacked: one row `a=1, b=99`, both bars `minAngle={30}`. `a` becomes `[0, 30]` while `b` stays `[3.64, 360]`, so they overlap by 26.4 degrees. `minPointSize={30}` yields the same two ranges, so the shape is not new, but the added tests only cover the unstacked path.

Chandu-Putta:
Good catch on both, thanks for actually running the numbers.

Fixed the negative minAngle case — it now uses Math.abs() the same way Pie does, so -30 and 30 behave identically.

On the stacking overlap: you're right that it's real, and I confirmed minPointSize has the exact same issue already, so this isn't something new. Properly fixing it would mean redistributing angles across stacked bars, which felt like a bigger change than this PR should take on. I added a test that pins down the current behavior (including the overlap) so it's at least documented and won't silently get worse, and left a comment in the code explaining why.

Both are in the latest commit if you want to take another look.

Chandu-Putta:
Hi @ckifer / @PavelVanecek — this has been open a few days now. Would really appreciate a look when you get a chance. Happy to make any changes you'd suggest. Thanks!

PR #7719, opened 2026-09-02
fix(CartesianGrid): align grid lines with tick labels when custom fon…

### LINKED ISSUE #7713: Setting tick font size in the XAxis misaligns vertical tick lines

As you can see here the first tick lines are not aligned with the XAxis ticks:
Image

We expect them to be aligned like when using default font size:

Image

Screenshots or videos are welcome.

[Here is how: https://github.com/recharts/recharts/wiki/How-to-share-a-demo

The problem happens when you set tick.fontSize in the XAxis component. According to Claude the problem is the tick lines use a different method for spacing calculations than the x axis ticks themselves. It even proposed a patch for rechart lib that I am currently using in my project.

Here the calculation is based in computed styles (correct): https://github.com/recharts/recharts/blob/c161c10b0a3326578d8b71b59025b83596ca9757/src/cartesian/CartesianAxis.tsx#L566

Here it doesn't use style props so uses default fontSize for the body: https://github.com/recharts/recharts/blob/c161c10b0a3326578d8b71b59025b83596ca9757/src/cartesian/CartesianGrid.tsx#L432

https://stackblitz.com/edit/q5h6dqs6-z4hjmdxz?file=src%2Findex.tsx

It happens in latest version.

Medyooo:
## Description

Fixes vertical/horizontal grid lines misaligning with axis tick labels when a custom `tick.fontSize` is set. `CartesianGrid` computed tick positions without knowing the real font size, so it couldn't apply the same edge-of-chart adjustment `CartesianAxis` already does.

## Related Issue

Fixes #7713

## Motivation and Context

With a large `tick.fontSize`, the last tick's label shifts inward to stay inside the chart bounds  but the grid line stayed at the raw
position, causing a visible gap between the two.

## How Has This Been Tested?

Added a regression test comparing each vertical grid line's x1 to its corresponding tick label's x, with `tick={{ fontSize: 30 }}`. Verified manually via a StackBlitz repro (console script comparing DOM coordinates) before writing the fix. Ran the full CartesianGrid suite (416 tests) and the full unit:lib suite (6395 tests) all pass, no regressions.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected chart grid line positioning when axis tick labels use custom font sizes or letter spacing.
  * Grid lines now align with the actual rendered tick labels, including cases where tick coordinates are adjusted.
* **Tests**
  * Added coverage verifying grid lines remain aligned with x-axis labels when larger fonts affect tick placement.


Medyooo:
The code fix is solid  full unit:lib suite passes (6395 tests), no regressions. What's blocking merge is just 6 VR snapshot images
(OneChartFourThemes + dark-mode: examples list, across chromium/ firefox/webkit) that need regenerating after my fix legitimately shifts a few pixels on tick positioning even without a custom fontSize (getTicks() already computes a slightly-adjusted tickCoord by default, previously ignored by CartesianGrid).

Could someone confirm whether this permission error is a known/ pre-existing CI issue, and if so, help regenerate these 6 snapshots on a working environment?

PR #7724, opened 2026-09-03
Fix axis props with incomplete themes

### LINKED ISSUE #7723: Axis stroke props dropped when the theme has no axis slice

Render `<XAxis dataKey="label" stroke="gold" strokeWidth={2} />` in a `ComposedChart` inside `<RechartsThemeProvider value={{ graphicalItems: [] }}>`, i.e. a theme with no `axis` slice.

The axis line and the tick lines come out with `stroke=null`, and `strokeWidth`, `strokeOpacity` and `strokeDasharray` are absent too. Without the provider, or with an `axis` slice, `gold` is honored.

`useBackwardsCompatibleTheme` returns no styles for a missing slice, and `CartesianAxis` reads its stroke fields off that result, so they become `undefined`.

The explicit prop wins. Suppressing the legacy `#666` default is intended there; dropping the props is not.

## Versions

`main` at 3c671e38917 (no sandbox; reproduces in the repo's own RTL setup).

PavelVanecek:
## Description

Return explicit props instead of nothing when merging theme slice.

## Related Issue

Fixes https://github.com/recharts/recharts/issues/7723

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved user-provided styling when a theme exists but does not define matching styles.
  - Prevented legacy default styles from being applied in this situation.
  - Ensured axis and tick-line styling props are consistently applied.

- **Tests**
  - Added coverage for explicit axis and tick-line styling.
  - Updated theme compatibility tests to verify preservation of user-provided props.

### DIFF
--- a/src/theme/useBackwardsCompatibleTheme.ts
+++ b/src/theme/useBackwardsCompatibleTheme.ts
@@ -2,14 +2,12 @@ import { RechartsTheme } from './RechartsTheme';
 import { resolvePartialDefaultProps } from '../util/resolveDefaultProps';
 import { useRechartsTheme } from './RechartsThemeContext';
 
-const noStyles = {};
-
 /**
  * Backwards compatible hook for theming.
  *
  * - If the selector returns a theme slice, it is merged with `explicitProps` and acts as defaults.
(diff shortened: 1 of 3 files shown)


PavelVanecek:
@kdelay is this better?

PR #7727, opened 2026-09-05
fix(PolarGrid): drop the duplicate radial line where the angle axis wraps around

giaBaoJS:
## Problem

`PolarAngleAxis` gets its ticks from `selectPolarAngleAxisTicks`, which collapses ticks that land on the same angle once normalized to `[0, 360)`. That deduplication exists because the angle axis is circular, so a full circle domain produces a first and last tick at the same place.

`PolarGrid` did not use it. `selectPolarGridAngles` read the raw `selectPolarAxisTicks`, so with a numerical angle axis spanning a full circle the grid drew two radial lines exactly on top of each other, and `getPolygonPath` repeated the first vertex before the closing `Z`, adding a zero length segment to every concentric polygon.

The overlap is visible: the doubled line renders darker than its neighbours whenever `strokeOpacity` is below 1, and the radial line count disagrees with the tick count of the very axis it is supposed to follow.

Reproduction:

```jsx
<RadarChart width={300} height={300} data={data}>
  <Radar dataKey="uv" />
  <PolarAngleAxis type="number" />
  <PolarGrid />
</RadarChart>
```

13 radial grid lines are rendered for 12 angle axis ticks, and the first and last lines are the same line.

## Fix

`selectPolarGridAngles` now reads the same deduplicated ticks the axis uses. `selectPolarGridRadii` is untouched: the radius axis is not circular and has nothing to collapse.

## Tests

Added a regression test in `test/polar/PolarGrid.spec.tsx` asserting that no two radial grid lines share an endpoint and that the line count matches the rendered angle axis tick count. It fails before the fix with 12 unique endpoints for 13 lines.

Five existing expectations in the same file had the duplicate baked into them and are updated:

- `should select grid angles` lost the trailing `360`, which duplicated `0`
- `should select angles` lost the trailing `360` for the same reason
- two `should render lines` cases lost the repeated line at `x2="266" y2="150.00000000000003"`, identical to the first line at `x2="266" y2="150"`
- `should render polygons` lost the repeated first vertex in each path

`npm run test` and `npm run check-types` pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed polar angle grids so full-circle axes no longer render a duplicate grid line at the 360° wraparound point.
  - Corrected angle-axis tick selection for more accurate polar grid rendering.

- **Tests**
  - Added coverage to verify unique angle grid-line endpoints and correct alignment with numerical axis ticks.

### DIFF
--- a/src/state/selectors/polarGridSelectors.ts
+++ b/src/state/selectors/polarGridSelectors.ts
@@ -1,15 +1,16 @@
 import { createSelector } from 'reselect';
 import { RechartsRootState } from '../store';
 import { AxisId } from '../cartesianAxisSlice';
-import { selectPolarAxisTicks } from './polarScaleSelectors';
+import { selectPolarAngleAxisTicks, selectPolarAxisTicks } from './polarScaleSelectors';
 import { CartesianTickItem } from '../../util/types';
 
 export type PolarAngles = Array<number>;
 
 export type PolarRadius = Array<number>;
 
+// The angle axis is circular, so its first and last tick can land on the same angle.
 const selectAngleAxisTicks = (state: RechartsRootState, anglexisId: AxisId) =>
-  selectPolarAxisTicks(state, 'angleAxis', anglexisId, false);
+  selectPolarAngleAxisTicks(state, 'angleAxis', anglexisId, false);
 
 export const selectPolarGridAngles: (state: RechartsRootState, angleAxisId: AxisId) => PolarAngles | undefined =
   createSelector(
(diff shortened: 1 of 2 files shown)


kdelay (commented):
Verified at 128f297. Reverting only `polarGridSelectors.ts` to 70868ee keeps the new test red (12 unique endpoints for 13 lines) and the five edited expectations red too. Restored: 1044 pass across test/polar + test/state, lint clean.

Pre-existing, not a blocker: `startAngle={-90} endAngle={-450}` still draws 13 lines / 12 unique endpoints, and PolarAngleAxis shows 13 ticks there on main too. `(coordinate + 360) % 360` in `selectPolarAngleAxisTicks` returns -90 for -450. `((c % 360) + 360) % 360` collapses that row, leaves the partial 0->270 case at 13 distinct, 1044 still pass.

PavelVanecek (approved):

Issue #7696, opened 2026-08-26
interval="equidistantPreserveEnd" only renders a single tick when the end label overflows the axis


ChristianK43:
## What is the current behavior?

With `interval="equidistantPreserveEnd"` on a category XAxis, the axis often renders exactly one tick, the last one, no matter how wide the chart is. Wider charts are hit harder than narrow ones, which is the opposite of what you would expect.

The last tick is centered on the last band, so half of its label sticks out past the axis end. `isVisible` counts that as not fitting, so every `stepsize < len` gets rejected in `getEquidistantPreserveEndTicks`, and the only survivor is `stepsize === len`, whose single tick skips the check via `index === offset`. Hence one label.

Every other tick mode moves an overflowing end label inwards (`getTicksStart` does exactly this for `preserveStart` and `preserveStartEnd`). The equidistant variant is the one place where that is missing, so it drops all the labels instead.

## What is the expected behavior?

Evenly spaced labels with the end one kept, like the name says. Chart 3 in the demo, plus the end label.

## Please provide a demo of the problem in a sandbox

https://codesandbox.io/s/6332vv

Three identical 900px charts, same 100 data points, only `interval` differs:

| chart | interval | tick labels | last label |
| --- | --- | --- | --- |
| 1 | `equidistantPreserveEnd` | 1 | `2026-05-25 18:00` |
| 2 | `preserveStartEnd` | 6 | `2026-05-25 18:00` |
| 3 | `equidistantPreserveStart` | 6 | `2026-05-22 06:00` (end label missing) |

Pure function repro, no rendering needed:

```js
const ticks = Array.from({ length: 100 }, (_, i) => ({ value: i, coordinate: (i + 1) * 5 }));
getEquidistantPreserveEndTicks(1, { start: 0, end: 500 }, () => 100, ticks, 20);
// => [{ value: 99, coordinate: 500 }]  (one tick)
```

Raising the chart's `margin.right` to 60 in the demo makes all 6 labels appear, which confirms it is the end label overflow driving it.

## Which versions of Recharts, and which browser / OS are affected by this issue? Did this work in previous versions of Recharts?

3.10.1 and `main` (3.11.0-canary.0). Not a regression, it has behaved this way since the interval landed in #6661. Browser and OS independent, it is pure tick math.

PR #7729, opened 2026-09-06
fix(XAxis): move the first equidistantPreserveStart label inwards

giaBaoJS:
## Description

`getEquidistantTicks` always shows the first tick, and it shows it at its raw `coordinate`.
On a band axis that coordinate is the middle of the first band, so half of the first label
sits to the left of the axis start and gets clipped by the SVG. `getTicksStart` already
avoids this for `interval="preserveStart"` by moving the label through `tickCoord` while the
tick line stays on `coordinate`, and `getEquidistantPreserveEndTicks` does the same for the
end tick since #7697. `equidistantPreserveStart` was the only one of the four that did not.

This moves the first label inwards the same way, and measures the following ticks against the
moved label instead of the original coordinate, so the spacing search sees where the label
actually is.

To reproduce, an `XAxis` with `interval="equidistantPreserveStart"` on a category axis with
enough categories that the first band is narrower than its label: the first label is cut off
at the left edge of the chart, while the same axis with `interval="preserveStart"` renders it
fully.

## Related Issue

No open issue. Same class as #7696 / #7697, on the start side.

## Motivation and Context

The three other `preserve*` intervals keep their anchor tick inside the axis. This makes the
fourth one consistent, so a chart does not lose the label it was told to preserve.

## How Has This Been Tested?

`test/cartesian/getEquidistantTicks.spec.ts` gets a new case that mirrors the existing
`should move the end tick inwards instead of dropping every other tick`: ten ticks 50px apart
with the first one on the axis start and 100px wide labels. It asserts that the first tick
keeps `coordinate: 0` and gains `tickCoord: 50`, and that the returned ticks are `[0, 4, 8]`,
because the moved label now occupies the space from 0 to 100. Without the source change the
assertion on `tickCoord` reports `expected undefined to be 50`.

The existing parametrised case in the same file also asserts the new `tickCoord` on its first
tick; it uses ticks that start exactly on the axis boundary, so the first label overflows
there too. The set of ticks it returns is unchanged.

`npm run test-lib`, `npm run check-types` and `npm run lint` pass.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cartesian axis tick positioning when the first label would overflow the axis boundary.
  * Adjusted the first tick’s rendered position inward to improve label visibility and tick spacing.

### DIFF
--- a/src/cartesian/getEquidistantTicks.ts
+++ b/src/cartesian/getEquidistantTicks.ts
@@ -19,6 +19,7 @@ export function getEquidistantTicks(
   // For now, start from every tick
   let stepsize = 1;
   let start = initialStart;
+  let headCoord: number | undefined;
 
   while (stepsize <= result.length) {
     // Given stepsize, evaluate whether every stepsize-th tick can be shown.
@@ -28,7 +29,12 @@ export function getEquidistantTicks(
 
     // Break condition - If we have evaluated all the ticks, then we are done.
     if (entry === undefined) {
-      return getEveryNth(ticks, stepsize);
+      const shownTicks = getEveryNth(ticks, stepsize);
+      const head = shownTicks[0];
+      if (headCoord == null || head == null) {
+        return shownTicks;
+      }
+      return [{ ...head, tickCoord: headCoord }, ...shownTicks.slice(1)];
     }
 
     // Check if the element collides with the next element
@@ -42,7 +48,18 @@ export function getEquidistantTicks(
       return size;
     };
 
-    const tickCoord = entry.coordinate;
+    let tickCoord = entry.coordinate;
+
+    if (index === 0) {
+      // The first tick sits in the middle of the first band, so its label can hang over the start
+      // boundary; move it inwards, the same way getTicksStart does for `preserveStart`.
+      const headGap = sign * (tickCoord - (sign * getSize()) / 2 - start);
+      if (headGap < 0) {
+        tickCoord -= headGap * sign;
+        headCoord = tickCoord;
+      }
+    }
+
     // We will always show the first tick.
     const isShow = index === 0 || isVisible(sign, tickCoord, getSize, start, end);
 
--- a/test/cartesian/getEquidistantTicks.spec.ts
+++ b/test/cartesian/getEquidistantTicks.spec.ts
@@ -29,7 +29,30 @@ describe('getEquidistantTicks', () => {
       ticks.push({ value: ticksThatFit.includes(index) ? 10 : 1000, coordinate: index * 50, index, offset: 0 });
     }
     const result = getEquidistantTicks(1, { start: 0, end: 10 * 50 }, getTickSize, ticks, 0);
-    expect(result).toEqual(resultingTicks.map(index => ticks[index]));
+    const expectedTicks: Array<CartesianTickItem> = resultingTicks.map(index => ticks[index]);
+    // The first tick sits on the axis start, so half of its label hangs outside and its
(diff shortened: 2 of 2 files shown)


PavelVanecek:
We have VR tests for ticks alignment, none of them shows the difference? Can you add some screenshots of before&after pls?

PavelVanecek (approved):

giaBaoJS:
Sorry for the slow reply.

**Why none of the VR tests show a difference**

`equidistantPreserveStart` appears exactly once in the VR suite: `LineChartAxisInterval`, mounted at `test-vr/tests/www/LineChartApiExamples.spec-vr.tsx:26`, which renders `www/src/docs/exampleComponents/LineChart/LineChartAxisInterval.tsx` (the interval is set on line 30). The only other `interval` values anywhere under `test-vr/` are `preserveEnd`, at `test-vr/tests/App.spec-vr.tsx:10`, `:11` and `test-vr/tests/AreaChart/stacked.spec-vr.tsx:11`. The two specs I expected to cover tick alignment, `test-vr/tests/CartesianAxis/CartesianAxis.padding.spec-vr.tsx` and `test-vr/tests/www/CustomAxisTicks.spec-vr.tsx`, do not set `interval` at all.

`LineChartAxisInterval` does not move because the boundary the first label is measured against is the whole chart, not the plot area. `selectAxisViewBox` returns `{ x: 0, y: 0, width, height }` (`src/state/selectors/selectChartOffsetInternal.ts:138`) and `getTickBoundaries` turns that into `start = 0` (`src/util/TickUtils.ts:11`). That chart has a `YAxis` with `width="auto"`, so the plot area starts about 32px in from `x = 0` and the first tick sits there. `Iter: 0` is about 39px wide, so its left edge lands at +12.67, well inside the boundary, `headGap` comes out positive and the new branch never runs.

I ran that VR test in chromium on this branch and on `main` and measured the first XAxis label box relative to the SVG left edge. It is identical in both: `left 12.67, right 51.33`. So the snapshot genuinely is unchanged rather than the suite missing a diff.

To reach the new code path the first label has to cross the chart's own left edge, which needs a narrow first band and nothing (no `YAxis`, no left margin) pushing the plot area inwards.

**Before and after**

Both images were rendered through the VR harness in chromium, same viewport, same data, same axis config. The only thing swapped between them is `src/cartesian/getEquidistantTicks.ts`, taken from `main` for the first and from this branch for the second.

```jsx
const data = Array.from({ length: 24 }, (_, i) => ({
  name: `2026-01-01 ${String(i).padStart(2, '0')}:00`,
  visits: 40 + ((i * 37) % 60),
}));

<BarChart width={600} height={300} data={data} margin={{ top: 20, right: 20, bottom: 5, left: 0 }}>
  <XAxis dataKey="name" interval="equidistantPreserveStart" />
  <Bar dataKey="visits" fill="#8884d8" isAnimationActive={false} />
</BarChart>
```

Before, on `main`. The first label box is `left -45.47, right 69.64`, so 45px of a 115px label sits outside the SVG and is clipped:

equidistantPreserveStart before

After, on this branch. The box is `left 0, right 115.11`, the label is fully inside, and the axis goes from every 5th tick to every 7th because the spacing search now measures against where the label actually is:

equidistantPreserveStart after

**VR coverage**

I am happy to add a VR test with that chart so the case is covered from now on. Tell me where you would like it, a new file under `test-vr/tests/CartesianAxis/` or another case in an existing spec, and I will push it here.

PR #7733, opened 2026-09-06
fix(test-vr): make npm run test-vr:ui work reliably in Docker

visheshgubrani:
## Description

`npm run test-vr:ui` didn't work. Playwright UI mode re runs the config's global setup on every page connect, so each reload or new tab tore down the Vite webServer Playwright had started (with `--strictPort`) and launched a fresh one. Reloading the UI or opening a second tab killed the story gallery out from under in flight test runs, causing intermittent `ERR_CONNECTION_REFUSED`.

The docker entry wrapper (`test-vr/.bin/playwright-test`) now pre-starts a persistent Vite dev server when `--ui` is requested and waits until it is ready. Playwright adopts the running server (`reuseExistingServer` is enabled outside CI) and, since it never spawned it, never kills it.

The wrapper also detects Docker images built before the #7714 dependency change (`@playwright/experimental-ct-react` -> `@playwright/test` + `vite`) and prints an actionable `npm run test-vr:prepare` message instead of a bare `Cannot find module '@playwright/test'` crash.

Additional changes: publish port 3100 in the `test-vr:ui` script so the gallery can be opened from the host browser, and document both behaviors in `test-vr/README.md` and `DEVELOPING.md`.

## Related Issue

Follow-up to the review feedback on #7714: https://github.com/recharts/recharts/pull/7714#issuecomment-5560368113

## Motivation and Context

The story gallery migration replaced Playwright's internal CT dev server with an explicit `webServer`, which made the dev server lifecycle dependent on UI client connections. This restores a reliable UI mode workflow for reviewers and contributors.

## How Has This Been Tested?

Ran in Docker with a freshly prepared image:

- Headless run unchanged: `playwright-test --project=chromium App.spec-vr.tsx` passes
- UI mode: tests list, and single tests pass across page reloads and a second open tab while the Vite process PID stays constant (before this change it restarted on every connect)
- `curl -I http://localhost:3100/gallery/index.html` returns 200 from the host while the UI session runs
- Stale-image simulation prints the rebuild hint instead of a module-not-found stack trace
- `tsc --project test-vr`, eslint and prettier pass on the changed files

Note: run `npm run test-vr:prepare` once before testing, the Docker image must be rebuilt after the #7714 dependency change.

## Screenshots (if appropriate):

N/A

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

[...]

PR #7714, opened 2026-08-31
refactor(test-vr): migrate VR tests to the Playwright story gallery

visheshgubrani:
## Description

Migrates the visual regression suite from `@playwright/experimental-ct-react` to the official [Playwright story gallery model](https://playwright.dev/docs/test-components).

The new setup uses plain `@playwright/test` with a Vite-powered React gallery. Each VR spec mounts a named story through the new `mountStory` fixture instead of passing JSX through Playwright's experimental component test adapter.

This PR:
- Adds the Vite story gallery under `test-vr/gallery`
- Adds `mountStory` with the same root locator behavior as the previous `mount` fixture
- Converts all 87 VR spec files to use named stories
- Adds matching `*.story.tsx` modules for the migrated tests
- Preserves all 393 test cases and their existing titles
- Updates the Playwright configuration, Docker setup, CI workflow, dependency checks, and documentation
- Replaces `@playwright/experimental-ct-react` with `@playwright/test`
- Keeps `playwright` because it is still required by `@vitest/browser-playwright`
- Synchronizes the Playwright packages at version 1.62.1
- Isolates the VR Vite cache in `node_modules/.vite-test-vr`
- Limits Vite dependency scanning to the gallery and story files
- Keeps story-internal helper components private (unexported)
- Adds type-only imports and generic prop checking to parameterized `mountStory` calls

No production behavior or visual-difference tolerances were changed.

## Related Issue

Closes #7595

## Motivation and Context

Playwright now recommends moving away from experimental framework-specific component test packages to a framework-agnostic story-based setup.

The previous setup depended on Playwright's internal JSX transformation, component discovery, and Vite pipeline. The new setup makes the boundary simpler: stories are regular React modules served by Vite, while Playwright handles browser automation, screenshots, retries, and tracing.

This also removes the old `test-vr/playwright` scaffolding and makes the VR setup easier to understand and maintain.

## Implementation Notes

- The gallery exposes `window.mount` and `window.unmount`, resolves stories through `import.meta.glob`, and reuses a React root between mounts.
- The `mountStory` fixture preserves the locator behavior of the previous experimental fixture: when the gallery root contains one child, the returned locator targets that child; otherwise, it targets the gallery root.
- The existing themed and unthemed behavior is preserved:
  - The gallery does not add global CSS
  - Individual stories and helpers may opt into CSS
  - Website stories using `StoryTheme` continue to load `app.css`
  - Light, dark, and custom theme props continue to map through the existing theme helpers
[...]

Issue #7595, opened 2026-07-28
Migrate to new playwright tests


PavelVanecek:
https://playwright.dev/docs/test-components#migration-from-the-experimental-packages

https://github.com/recharts/recharts/pull/7589

PR #7734, opened 2026-09-06
Add interactive visual regression gallery

PavelVanecek:
Fix test-vr:ui command, plus add a gallery viewer for convenience

Followup from https://github.com/recharts/recharts/pull/7714

Related issue https://github.com/recharts/recharts/issues/7595

## Summary by CodeRabbit

* **New Features**
  * Added a browsable visual regression gallery with grouped stories, story selection, and default-prop previews.
  * Added support for opening the gallery in Playwright UI mode.
  * Made the gallery available through a dedicated local preview server.

* **Documentation**
  * Updated visual regression testing guidance for story-based workflows, targeted test runs, browser baselines, gallery usage, and preview URLs.
  * Clarified the distinction between the blank test mount and the human-facing story preview.

### DIFF
--- a/.agents/skills/vr-test/SKILL.md
+++ b/.agents/skills/vr-test/SKILL.md
@@ -5,93 +5,112 @@ description: Create visual regression tests
 
 # What is a visual regression test?
 
-VR test is a playwright test that renders a component and compares the rendered output to a baseline snapshot.
+VR tests are Playwright tests that render a story and compare the rendered output to a baseline
+snapshot. Recharts runs these tests in Docker using Playwright's component testing model and a Vite
+story gallery.
 
-# Writing a new test
+## Story and spec structure
 
-To write a new test, create a new file in the `test-vr/tests` directory. The file must end with `.spec-vr.tsx` to be picked up by the test runner.
+Each visual-regression spec has a matching story file in `test-vr/tests`:
 
-Here is an example of a simple test:
+```text
+test-vr/tests/App.story.tsx
+test-vr/tests/App.spec-vr.tsx
+```
+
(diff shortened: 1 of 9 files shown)


PavelVanecek:
Hi @visheshgubrani this is what my agent came up with and I was playing around with it and it appears to be stable. What do you think? It's quite similar to https://github.com/recharts/recharts/pull/7733 but I didn't see the restart problems.

visheshgubrani:
Looks good @PavelVanecek and yeah it's basically the same fix as #7733, we just chased the same root cause from different ends

The restart issue only shows up when playwright owns the dev server, UI mode reruns the global setup on every page connect and kills whatever webServer it spawned itself. Your version starts vite in the background before playwright boots, so playwright just adopts it and never owns it, which is why you didn't see the restarts. Same trick as my wrapper pre-start, so no mystery there :)

The preview.html viewer is a nice touch, clicking through stories beats staring at a blank mount page.

Also two things I noticed while poking at it, the `3100:3100` you added to docker-compose.yml might bite you. `docker compose run` doesn't apply service ports, and if the report container from `npm run test-vr:prepare` (`docker compose up -d`) is still running, it already holds host port 3100. So `npm run test-vr:ui` fails with "port is already allocated" before Playwright even starts. I reproduced that locally. And the inline `&` doesn't wait for vite to actually be ready, so on a cold start there's a small window where playwright could try to spawn its own server and hit the `--strictPort` conflict.

Since we're fixing the same thing, do you want me to close #7733 and help review this one instead? Whatever is least hassle for you

PavelVanecek:
Yes if you don't mind pls close #7733 and we will merge this one. I will undo the port mapping in the yaml file

PR #7756, opened 2026-09-08
fix: preserve repeated radar category positions

### LINKED ISSUE #4661: Issue with duplicated category in RadarCharts

Bug

Radar charts with duplicate category's fail to render values when there's duplicated categories. Tooltips still work perfectly though

When I write this code

``` js
import React, { PureComponent } from 'react';
import { Radar, RadarChart, PolarGrid, PolarAngleAxis, PolarRadiusAxis, ResponsiveContainer } from 'recharts';

const data =  [
    {
        "Key": "A",
        "Val": 12,
    },
    {
        "Key": "B",
        "Val": 3,
    },
    {
        "Key": "A", // this is the relevant line, change to C and it works
        "Val": 10,
    },
];

export default class Example extends PureComponent {
  render() {
    return (
        <RadarChart cx="50%" cy="50%" outerRadius="80%" data={data}>
          <PolarGrid />
          <PolarAngleAxis dataKey="Key" />
          <PolarRadiusAxis />
          <Radar  dataKey="Val" stroke="#8884d8" fill="#8884d8" fillOpacity={0.6} />
        </RadarChart>
      </ResponsiveContainer>
    );
  }
}
```

I don't get the radar chart to render correctly. 

Given allowDuplicatedCategory is not on I would expect it to render two A's, one with Val 10 and another with Val 12

Latest version from npm

vzer200:
## Description

Honor `PolarAngleAxis.allowDuplicatedCategory` and place repeated radar categories at separate axis positions. For names `A`, `B`, `A`, the default axis now renders three spokes instead of collapsing both `A` values onto one spoke. Explicit `allowDuplicatedCategory={false}` continues to deduplicate categories.

Radar uses occurrence positions while retaining each entry's name and payload. Axis, graphical-item, grid, and tooltip tick generation use those positions for custom band/point scales. Tooltip lookup retains the selected occurrence's value, with label lookup preserved when item data has a different ordering. Numeric-zero and empty-string data keys remain valid.

With a Brush, Radar already renders the full data range. Its duplicate angle metadata now uses that same range. Visible and graphical angle ticks in a RadarChart also use full-data metadata unless a RadialBar is registered on the same angle axis, including a hidden RadialBar. This preserves existing RadialBar sector/tick and categorical radius-axis behavior.

## Related Issue

Fixes #4661

## Motivation and Context

The public prop defaults to true, but the axis registered a hardcoded false. Respecting that prop creates an indexed duplicate domain, so mapping category names into it cannot identify individual occurrences. The same occurrence must remain aligned across points, ticks, grids, and tooltip values.

The Brush correction follows Radar's existing full-data contract. Shared Tooltip selectors still use the Brush slice, so some pre-existing tooltip label/value mismatches remain. A Radar and RadialBar sharing one angle axis also retain their existing different data-range requirements; this change preserves the RadialBar contract rather than redesigning that mixed-chart behavior.

## How Has This Been Tested?

- 137 added regression/compatibility cases in total. The initial 48 cover repeated categories, scales, data keys, tooltips, and Cartesian alignment; 37 fail on the original source. The additional 89 cover Brush ranges, multiple/hidden series, mixed Radar/RadialBar consumers, panorama ticks, and categorical radius identity.
- The final 197 affected tests pass; the same tests against the initially published PR source produce 25 failures and 172 passes.
- Full unit suite: 16,907 passed; 6 expected failures, 16 skipped, 1 todo, across 353 files.
- Full build, including 13 build-output tests; all five TypeScript projects; unmodified ESLint command with zero errors and 17 existing warnings.
- A separate automated reviewer independently ran 942 semantic probes and 1,264 related repository tests (9 skipped), including direct RadialBar sector/label and Radar radius/value comparisons against the published revision.
- Chromium rendering and pointer checks across original, published, and final code: 48 charts and 192 pointer moves, with no observed browser errors or previously correct in-slice tooltip check becoming incorrect. Seven existing in-slice tooltip mismatches remain, as described above.
- Added the `RepeatedCategories` Storybook example.

Validation ran on Windows with Node 24.16.0. Dependencies used the documented Windows `npm install --force` workaround. Docker visual-regression snapshots were not run or changed.

## Types of changes

## Checklist:

Prepared with Codex and independently reviewed by a separate automated agent. Review findings were corrected and rechecked against the final file hashes before publication. No human review is claimed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed duplicated categorical values in Radar and polar charts so points, axis ticks, grid lines, and tooltips remain correctly aligned.
  - Improved tooltip selection for repeated categories, including numeric categories and filtered chart ranges.
  - Preserved full category domains and correct Brush behavior across Radar and RadialBar charts.

- **Examples**
  - Added a RadarChart example demonstrating repeated categories.

- **Tests**
  - Added comprehensive coverage for duplicated categories, tooltips, scales, axes, and Brush interactions.

### DIFF
--- a/src/polar/Radar.tsx
+++ b/src/polar/Radar.tsx
@@ -200,6 +200,7 @@ export type AngleAxisForRadar = {
   scale: RechartsScale;
   type: 'number' | 'category';
   dataKey: DataKey<any> | undefined;
+  duplicateDomain: ReadonlyArray<unknown> | undefined;
   cx: number;
   cy: number;
   range: AxisRange;
@@ -318,7 +319,7 @@ export function computeRadarPoints({
   displayedData.forEach((entry, i) => {
     const name = getValueByDataKey(entry, angleAxis.dataKey, i);
     const value = getValueByDataKey(entry, dataKey);
-    const angle: number = (angleAxis.scale.map(name) ?? 0) + angleBandSize;
+    const angle: number = (angleAxis.scale.map(angleAxis.duplicateDomain ? i : name) ?? 0) + angleBandSize;
     const pointValue = Array.isArray(value) ? last(value) : value;
     const radius: number = isNullish(pointValue) ? 0 : (radiusAxis.scale.map(pointValue) ?? 0);
 
--- a/src/state/selectors/polarScaleSelectors.ts
+++ b/src/state/selectors/polarScaleSelectors.ts
@@ -17,11 +17,34 @@ import {
 } from './polarAxisSelectors';
 import { CartesianTickItem } from '../../util/types';
 import { selectChartLayout } from '../../context/chartLayoutContext';
-import { selectPolarAppliedValues, selectPolarAxisCheckedDomain, selectPolarNiceTicks } from './polarSelectors';
+import {
+  selectPolarAppliedValues,
+  selectPolarAxisCheckedDomain,
+  selectPolarNiceTicks,
+  selectUnfilteredPolarItems,
+} from './polarSelectors';
 import { pickAxisType } from './pickAxisType';
 import { RechartsScale, rechartsScaleFactory } from '../../util/scale/RechartsScale';
 import { CustomScaleDefinition } from '../../util/scale/CustomScaleDefinition';
 import { combineConfiguredScale } from './combiners/combineConfiguredScale';
+import { selectChartName } from './rootPropsSelectors';
+
+const selectPolarDuplicateDomain = (
+  state: RechartsRootState,
+  axisType: 'angleAxis' | 'radiusAxis',
+  axisId: AxisId,
+  isPanorama: boolean,
+): ReadonlyArray<unknown> | undefined => {
+  // Radar angles use the full data range even with a Brush. A RadialBar sharing
+  // the angle axis still uses the Brush slice, so retain its existing tick domain.
+  const axisIdKey = axisType === 'angleAxis' ? 'angleAxisId' : 'radiusAxisId';
+  const hasRadialBar = selectUnfilteredPolarItems(state).some(
+    item => item.type === 'radialBar' && item[axisIdKey] === axisId,
+  );
+  const ignoreIndexes =
+    isPanorama || (axisType === 'angleAxis' && selectChartName(state) === 'RadarChart' && !hasRadialBar);
+  return selectDuplicateDomain(state, axisType, axisId, ignoreIndexes);
+};
 
 export const selectPolarAxis = (state: RechartsRootState, axisType: 'angleAxis' | 'radiusAxis', axisId: AxisId) => {
   switch (axisType) {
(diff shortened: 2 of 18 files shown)


src/polar/Radar.tsx
```diff
@@ -318,7 +319,7 @@ export function computeRadarPoints({
   displayedData.forEach((entry, i) => {
     const name = getValueByDataKey(entry, angleAxis.dataKey, i);
     const value = getValueByDataKey(entry, dataKey);
-    const angle: number = (angleAxis.scale.map(name) ?? 0) + angleBandSize;
+    const angle: number = (angleAxis.scale.map(angleAxis.duplicateDomain ? i : name) ?? 0) + angleBandSize;
```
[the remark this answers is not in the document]
vzer200: Addressed in 08bad09. Radar already consumes the full chart data through `selectChartDataAndAlwaysIgnoreIndexes`, so applying a Brush start offset to its point indexes would use the wrong range. The correction makes duplicate angle metadata use that existing full-data contract instead. Visible/graphical angle ticks follow it only in RadarChart without a same-axis RadialBar, preserving RadialBar sector labels and categorical radius behavior.

Added 89 Brush/mixed-consumer/radius regression cases; the final 197 affected tests pass, with 25 failures on the initially published source. Independent review also compared 942 boundary cases. The older shared Tooltip slice mismatch and mixed same-axis range conflict remain outside this fix and are documented in the PR description.

src/state/selectors/polarScaleSelectors.ts
```diff
@@ -17,11 +17,34 @@ import {
...
+  axisId: AxisId,
+  isPanorama: boolean,
+): ReadonlyArray<unknown> | undefined => {
+  // Radar angles use the full data range even with a Brush. A RadialBar sharing
+  // the angle axis still uses the Brush slice, so retain its existing tick domain.
+  const axisIdKey = axisType === 'angleAxis' ? 'angleAxisId' : 'radiusAxisId';
+  const hasRadialBar = selectUnfilteredPolarItems(state).some(
+    item => item.type === 'radialBar' && item[axisIdKey] === axisId,
+  );
+  const ignoreIndexes =
```
PavelVanecek, line 44: What is `ignoreIndexes` and why do we use it instead of `isPanorama`? This looks like extra complex logic, with special exception for RadarChart? Can this be simpler?

vzer200: `ignoreIndexes` means “use the full chart-data range.” The boolean passed down as `isPanorama` currently selects that range. Radar already uses the full range outside the Brush panorama, so with a Brush-selected subset, passing the actual `isPanorama` value alone can pair full-data points with sliced tick labels.

On a shared categorical angle axis, with `[A, B, A, C, B]` and Brush `1..4`, RadialBar needs `[B, A, C, B]` labels while Radar retains all five points. Removing the RadialBar guard changes labels without moving its sectors. Extending the override to radius axes also breaks the category-radius coordinate tests.

The panorama argument is confusing here. A dedicated full-data duplicate-domain selector could express that choice more clearly, but it would still need the consumer distinction. The three substitutions above failed existing regressions, so I have kept the code unchanged. I have not implemented the dedicated selector.

Issue #7796, opened 2026-09-11
Translate chart example names for the zh-CN locale


CupOfBread:
**Is your feature request related to a problem? Please describe.**
The website supports a `zh-CN` locale (locale switcher, translated nav and guides), but chart example names are hardcoded English strings (`name` fields in `www/src/docs/exampleComponents/*/index.ts(x)`). Both the examples sidebar navigation and the example page title render these strings directly, bypassing the locale system, so the `/zh-CN/` site still shows English names like "Simple Line Chart" while everything else is in Chinese.

**Describe the solution you'd like**
Add Chinese translations for all chart example names to the `examples` group in `www/src/locale/zh-CN.ts`, keyed by their registry keys. Make the examples sidebar (`www/src/navigation.ts`) and the example page title (`www/src/views/ExamplesView.tsx`) look up `localeGet(locale, 'examples', key)` first and fall back to the hardcoded name when no translation exists — the same pattern already used for guide and API navigation items. This keeps `en-US.ts` untouched and lets untranslated examples fall back to English automatically.

**Describe alternatives you've considered**
- Hardcoding Chinese names directly into the example registries: rejected because it would change the English site and violate the project rule against hardcoding localized strings.
- Translating inside `en-US.ts`: rejected because English is the fallback source of truth; duplicating names there would create two sources of truth.

**Additional context**
I have a working change ready (124 example name entries + two render-path lookups) and will submit a PR referencing this issue. Scope is the website only; the library is unaffected.

PR #7797, opened 2026-09-11
feat(www): add Chinese translations for all chart example names

CupOfBread:
## Description

Chart example names are hardcoded English strings (`name` in `www/src/docs/exampleComponents/*/index.ts(x)`), rendered directly by the examples sidebar (`www/src/navigation.ts`) and example page title (`www/src/views/ExamplesView.tsx`), bypassing the locale system — so the `/zh-CN/` site shows English names. This PR adds 124 example name entries (keyed by registry keys) to the `examples` group in `www/src/locale/zh-CN.ts`, and makes the sidebar and page title look up `localeGet(locale, 'examples', key)` first with fallback to the hardcoded name. `en-US.ts` is untouched: English behavior is unchanged, and untranslated examples fall back to English instead of rendering blank.

## Related Issue

https://github.com/recharts/recharts/issues/7796

## Motivation and Context

The website ships a `zh-CN` locale (locale switcher, translated nav and guides), but example names stayed hardcoded English, making the Chinese site inconsistent. This change reuses the existing `localeGet` pattern already used for guide and API navigation items; it only touches the website, not the library.

## How Has This Been Tested?

- Script-verified that the 124 keys added to `zh-CN.ts` exactly match the 124 registry keys in `www/src/docs/exampleComponents/*/index.ts(x)` (no missing or extra keys)
- Reviewed the fallback path: for `en-US`, `localeGet` returns `undefined` for these keys, so English rendering is identical to before
- Type checking and build will be exercised by the pre-push hook (`check-types-website` included)

No new unit/VR tests: website-only locale data plus two render-path lookups following the existing `localeGet` pattern; no library, Storybook, or VR changes.

## Types of changes

## Checklist:

- [x I have added tests to cover my changes.

## Summary by CodeRabbit

- **New Features**
  - Added Simplified Chinese translations for chart examples and component demonstrations.
  - Example navigation now displays localized names when translations are available.
  - Example page titles now adapt to the selected locale, with fallback names shown when needed.


PavelVanecek:
Hi @CupOfBread I'm considering unshipping the chinese translation completely, and have users rely on their own translation services. Do you think it's valuable to keep it? I don't understand Chinese myself.

CupOfBread:
@PavelVanecek I think this is necessary. Ideally, the website should have built-in multi-language support — if users rely on their own browser translation services, the translations are often inaccurate.

PavelVanecek:
Yes I agree that would be ideal. There was however no interest from translators - and we don't have budget for actual translation services. I was considering migrating the website to something like crowdin, I even started a project here https://crowdin.com/project/recharts but without people to translate it, we might just as well rely on google translate.

CupOfBread:
Maintaining translations on Crowdin is actually a good option too. If no one is available to translate, using Google Translate as a starting point is fine — users can then refine the translations in Crowdin when they spot issues. I think that would be a viable approach.

www/src/navigation.ts
```diff
@@ -93,7 +93,7 @@ function exampleNavItems(locale: SupportedLocale): ReadonlyArray<NavCategory> {
     displayName: categoryName,
     items: Object.entries(category.examples).map(([name, example]) => ({
       key: name,
-      displayName: example.name,
+      displayName: (localeGet(locale, 'examples', name) as string | undefined) ?? example.name,
```
PavelVanecek: Can we do this without the "as" cast?

PR #7798, opened 2026-09-12
docs(XAxis): add a website example for height="auto"

kimlj:
## Description

Adds a website example for `XAxis height="auto"`, which shipped in #7570 without one.

The example renders the same bar chart twice with rotated tick labels. The first leaves `height` at its default of 30, where the labels are clipped. The second sets `height="auto"`, where the axis measures its ticks and grows to fit them. The prose notes the `YAxis width="auto"` parallel and says when to keep a fixed numeric height (to keep several charts aligned).

Two files, +87 lines, no source changes:

- `www/src/docs/exampleComponents/XAxis/XAxisAutoHeightExample.tsx` (new)
- `www/src/docs/exampleComponents/XAxis/index.ts` (registers it)

No API doc edit is needed. Omnidoc reads `www/src/docs/exampleComponents/` and links an example from the props it actually uses, so after `npm run omnidoc` the example appears under the `XAxis` `height` prop on its own.

## Related Issue

Follow-up to the review on #7570: https://github.com/recharts/recharts/pull/7570#issuecomment-5025807484

> Example on the website would be nice

The underlying feature request was #7301, closed by #7570.

## Motivation and Context

`height="auto"` was undocumented by example. Before this, `height="auto"` appeared nowhere in `www/src` or `storybook/`, so the only description of it was the prop table text. The rotated-label case is the one where a fixed height cannot be known ahead of time, so it is the case worth showing.

## How Has This Been Tested?

- `npx vitest run www/test`: 15 files, 9252 tests passed
- `npm run test-omnidoc`: 14 files, 1098 tests passed, including the comment-similarity guard that needed an exception in #7570
- `npx tsc --noEmit -p www/tsconfig.json`: 77 errors on this branch and 77 on stock `main`, byte-identical diff, so nothing introduced. Those are pre-existing and unrelated to this change.
- `npm run omnidoc`: regenerated, and the example appears under the `height` prop as expected
- Verified the example's claim rather than assuming it: rendering both charts with `mockGetBoundingClientRect({ width: 30, height: 80 })` gives a rendered axis `height` of 30 for the default and greater than 30 for `height="auto"`. That check was a scratch file and is not included, since #7570 already covers the behavior in `test/cartesian/XAxis/XAxis.autoHeight.spec.tsx`.

Environment: Windows 11, Node 22.

Not attached. The example is the screenshot: the first chart clips its rotated labels, the second does not.

## Types of changes

Documentation only. None of the three boxes fit, so none are ticked.

## Checklist:

On the last two: the behavior is already covered by `XAxis.autoHeight.spec.tsx` from #7570, and a duplicate test on a docs change seemed like noise. This adds a website example rather than a storybook story, which is what the review asked for. I can add a story as well if you would prefer both.

## Summary by CodeRabbit

- **New Features**
  - Added an X-axis example demonstrating automatic height adjustment alongside a fixed-height chart.
  - Added sample visitor data and responsive bar charts to compare automatic and fixed sizing.
- **Documentation**
  - Documented automatic X-axis sizing, fixed-height behavior, and the limitation that labels provided as functions or React elements do not expand the automatic height.

### DIFF
--- a/www/src/docs/exampleComponents/XAxis/XAxisAutoHeightExample.tsx
+++ b/www/src/docs/exampleComponents/XAxis/XAxisAutoHeightExample.tsx
@@ -1,0 +1,79 @@
+import { Bar, BarChart, CartesianGrid, Tooltip, XAxis, YAxis } from 'recharts';
+
+// #region Sample data
+const data = [
+  {
+    name: 'Organic search',
+    visitors: 4000,
+  },
+  {
+    name: 'Paid search and display',
+    visitors: 3000,
+  },
+  {
+    name: 'Email newsletter',
+    visitors: 2000,
+  },
+  {
+    name: 'Referral from partners',
+    visitors: 2780,
+  },
+  {
+    name: 'Social media campaigns',
+    visitors: 1890,
+  },
+  {
+    name: 'Direct traffic',
+    visitors: 2390,
+  },
+];
+
+// #endregion
+export default function XAxisAutoHeightExample() {
+  return (
+    <div style={{ width: '100%' }}>
+      <BarChart
+        style={{ width: '100%', maxWidth: '700px', maxHeight: '30vh', aspectRatio: 1.618 }}
+        responsive
+        data={data}
+      >
+        <CartesianGrid />
+        <XAxis dataKey="name" angle={-35} textAnchor="end" />
+        <YAxis width="auto" />
+        <Tooltip />
+        <Bar dataKey="visitors" />
+      </BarChart>
+
+      <BarChart
+        style={{ width: '100%', maxWidth: '700px', maxHeight: '30vh', aspectRatio: 1.618 }}
+        responsive
+        data={data}
+      >
+        <CartesianGrid />
+        <XAxis dataKey="name" angle={-35} textAnchor="end" height="auto" />
+        <YAxis width="auto" />
+        <Tooltip />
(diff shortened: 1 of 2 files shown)


PavelVanecek:
The example should also include `RechartsDevtools` in the JSX so that the hook inspector and other tools can work.

VR test would be nice too

kimlj:
Added RechartsDevtools to both charts and a VR test in test-vr/tests/www/XAxisAutoHeightExample.spec-vr.tsx, snapshots for all 9 browser/theme projects.

www/src/docs/exampleComponents/XAxis/XAxisAutoHeightExample.tsx
```diff
@@ -0,0 +1,83 @@
...
+        style={{ width: '100%', maxWidth: '700px', maxHeight: '30vh', aspectRatio: 1.618 }}
+        responsive
+        data={data}
+      >
+        <CartesianGrid />
+        <XAxis dataKey="name" angle={-35} textAnchor="end" />
+        <YAxis width="auto" />
+        <Tooltip />
+        <Bar dataKey="visitors" />
+        <RechartsDevtools />
```
PavelVanecek, line 46: The website is only prepared for inspecting a single chart so adding `RechartsDevtools` twice may show weird data.

PavelVanecek (approved):

PR #7805, opened 2026-09-12
fix(z-index): drop tabIndex from z-index portal groups to stop pointer focus rings

### LINKED ISSUE #7799: `tabindex="-1"` on the z-index layer groups makes every chart layer mouse-focusable
### What happens

Every chart renders thirteen `<g tabindex="-1" class="recharts-zIndex-layer_N">` groups inside the `<svg>`, one per default z-index, whether the layer holds anything or not.

`tabindex="-1"` removes an element from the tab order **and makes it focusable by pointer**. So clicking anywhere inside a chart focuses the nearest focusable ancestor of whatever is under the cursor, which is one of those layer groups rather than the chart surface. Any app with a global `:focus-visible` rule then paints a focus ring around it, and in WebKit an outline on an SVG element traces the geometry inside rather than a bounding box, so the ring comes out as the silhouette of the bars, or of a `ReferenceArea`, depending on which layer was clicked. It reads to a user as though they had selected a region of the chart.

### Where

`src/zIndex/ZIndexPortal.js`:

```js
// these g elements should not be tabbable
return <g tabIndex={-1} ref={ref} className={`recharts-zIndex-layer_${zIndex}`} />
```

The comment states the intent, and a bare SVG `<g>` already satisfies it: `<g>` is not tabbable without `tabindex`. Adding `tabindex="-1"` does not remove tabbability it never had; it adds pointer focusability it should not have.

`src/zIndex/zIndexSelectors.js` concatenates `Object.values(DefaultZIndexes)` unconditionally, so the full set of layer groups is rendered by every chart, used or not.

### Reproduction

Any `<BarChart>` in Safari, in a document with:

```css
:focus-visible { outline: 2px solid blue; outline-offset: 2px; }
```

Click a bar. The outline follows the union of the bar paths.

### Suggested fix

Drop the `tabIndex` prop from the `<g>` in `ZIndexPortal`. The comment's stated goal holds without it.

### Version

recharts 3.10.1.

devtechedge:
Fixes #7799.

`tabindex="-1"` on the z-index layer `<g>` elements makes them pointer-focusable. Clicking anywhere in a chart focuses the nearest layer group, and any app with a global `:focus-visible` rule then paints a focus ring on it. In WebKit an outline on an SVG element traces its geometry, which looks like a selected bar or ReferenceArea.

This removes the `tabIndex={-1}` prop from `ZIndexSvgPortal`. A bare SVG `<g>` is neither tabbable nor pointer-focusable, so the intent documented in PR #6687 (preventing extraneous focusable surfaces) still holds without the attribute.

Tests: the AllZIndexPortals spec now selects layer groups by their `recharts-zIndex-layer_` class instead of by tabindex, and adds an assertion that no rendered layer group carries a `tabindex` attribute. That assertion fails on main and passes with this change.

Note: the issue cites `ZIndexPortal.js`; the actual file is `src/zIndex/ZIndexPortal.tsx`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated SVG layer focus behavior so portal layers are not assigned a `tabindex` attribute, while remaining non-tabbable and non-pointer-focusable.
* **Tests**
  * Updated coverage to verify all portal layers render correctly without a `tabindex` attribute.

### DIFF
--- a/src/zIndex/ZIndexPortal.tsx
+++ b/src/zIndex/ZIndexPortal.tsx
@@ -15,8 +15,8 @@ function ZIndexSvgPortal({ zIndex, isPanorama }: { zIndex: number; isPanorama: b
       dispatch(unregisterZIndexPortalElement({ zIndex, isPanorama }));
     };
   }, [dispatch, zIndex, isPanorama]);
-  // these g elements should not be tabbable
-  return <g tabIndex={-1} ref={ref} className={`recharts-zIndex-layer_${zIndex}`} />;
+  // a bare <g> is not tabbable or pointer-focusable; do not add tabIndex
+  return <g ref={ref} className={`recharts-zIndex-layer_${zIndex}`} />;
 }
 
 export function AllZIndexPortals({ children, isPanorama }: { children?: React.ReactNode; isPanorama: boolean }) {
--- a/test/zIndex/AllZIndexPortals.spec.tsx
+++ b/test/zIndex/AllZIndexPortals.spec.tsx
@@ -36,8 +36,10 @@ describe('AllZIndexPortals', () => {
     const newZIndexDefinitelyNotOneOfTheDefaults =
       Object.values(DefaultZIndexes).reduce((acc, val) => Math.max(acc, val), Number.NEGATIVE_INFINITY) + 1;
 
-    // all z-index portals should be rendered (they are <g> elements with tabIndex={-1})
-    expect(container.querySelectorAll('g[tabindex="-1"]')).toHaveLength(allZIndexes.length);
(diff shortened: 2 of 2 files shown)


PavelVanecek (approved):

devtechedge:
Thanks for merging, @PavelVanecek.
I learned that `tabIndex={-1}` on a portal SVG `<g>` still leaves an empty group reachable, and that WebKit will draw a focus ring around traced geometry because of it.

Removing the attribute and locking the behaviour with a test was a small change with a very specific cause.
Thanks for taking it.

